PR #11, opened 2022-09-01
Initial implementation of `LdkLite`

tnull:
Based on #8, #9, #10.

This PR wraps the previously mentioned PRs and provides initial implementations of `LdkLite`, `LdkLiteBuilder`, as well as basic API methods.


tnull:
Rebased on #9/#10.

src/io_utils.rs
```diff
@@ -0,0 +1,66 @@
...
+
+use std::fs;
+use std::io::{BufReader, Write};
+use std::path::Path;
+use std::sync::Arc;
+
+pub(crate) fn read_or_generate_seed_file(config: Arc<Config>) -> [u8; 32] {
+	let keys_seed_path = format!("{}/keys_seed", config.storage_dir_path);
+	let keys_seed = if Path::new(&keys_seed_path).exists() {
+		let seed = fs::read(keys_seed_path.clone()).expect("Failed to read keys seed file");
```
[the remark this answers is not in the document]
tnull: Right, the idea is to give the user a range of options how to supply the entropy, e.g., as a raw file path, raw bytes, or Bip39Mnemonic (cf. #14 / https://github.com/tnull/ldk-lite/blob/4a184f03c16e90f721d8feed1a3a85af69cabd21/src/lib.rs#L163-L180). 

We may then also give him the option to use a seed provided via any persistence backend, same goes with the network graph. I however still think that this kind of 'general use' of the persister might need some updates to the trait.

src/lib.rs
```diff
@@ -0,0 +1,1012 @@
...
+/// Represents the configuration of an [`LdkLite`] instance.
+pub struct Config {
+	/// The path where the underlying LDK and BDK persist their data.
+	pub storage_dir_path: String,
+	/// The URL of the utilized Esplora server.
+	pub esplora_server_url: String,
+	/// The used Bitcoin network.
+	pub network: bitcoin::Network,
+	/// The TCP port the network node will listen on.
+	pub listening_port: u16,
```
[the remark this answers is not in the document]
tnull: Done in https://github.com/lightningdevkit/ldk-lite/pull/11/commits/245cf9ec1967a96eecb68b1c84ecdc9317e3cf6a.

src/lib.rs
```diff
@@ -0,0 +1,1012 @@
...
+
+		// TODO: Check that we can be sure that the Esplora client re-connects in case of failure
+		// and and exits cleanly on drop. Otherwise we need to handle this/move it to the runtime?
+		let blockchain = EsploraBlockchain::new(&config.esplora_server_url, BDK_CLIENT_STOP_GAP)
+			.with_concurrency(BDK_CLIENT_CONCURRENCY);
+
+		let chain_access = Arc::new(ChainAccess::new(blockchain, bdk_wallet, Arc::clone(&logger)));
+
+		// Step 3: Initialize Persist
+		let persister = Arc::new(FilesystemPersister::new(ldk_data_dir.clone()));
```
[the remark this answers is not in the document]
tnull: Right, simply because I haven't come around to wrapping the `sled` database in a reasonable abstraction and implementing the necessary trait on it. I'm kind of holding off on this a bit since a) BDK seems to suggest using SQLite now and b) BDK's current use of the database makes `bdk::Wallet` not `Send`, which is messing with `async`. This is however due to change in BDK core / 1.0, which we therefore should adopt ASAP. 

So, I so far was holding off touching the database things too much to avoid finding a workaround (and possibly introducing some trade-offs), just to have to change it again anyways. But, will definitely change this in an upcoming PR soon.

src/lib.rs
```diff
@@ -0,0 +1,1012 @@
...
+					lightning_net_tokio::setup_inbound(
+						Arc::clone(&peer_mgr),
+						tcp_stream.into_std().unwrap(),
+					)
+					.await;
+				});
+			}
+		});
+
+		// Regularly reconnect to channel peers.
```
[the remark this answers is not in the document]
tnull: Hm, not quite sure what you're referring to? Looking them up to connect to them? But if we have a channel, they should be in the peer store?

src/lib.rs
```diff
@@ -0,0 +1,1012 @@
...
+			tokio::runtime::Builder::new_multi_thread().enable_all().build().unwrap();
+
+		// Setup wallet sync
+		let chain_access = Arc::clone(&self.chain_access);
+		let sync_cman = Arc::clone(&self.channel_manager);
+		let sync_cmon = Arc::clone(&self.chain_monitor);
+		let sync_logger = Arc::clone(&self.logger);
+		let stop_wallet_sync = Arc::new(AtomicBool::new(false));
+		let stop_sync = Arc::clone(&stop_wallet_sync);
+
```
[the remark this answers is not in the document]
tnull: Noted, opened https://github.com/lightningdevkit/ldk-lite/issues/29 for this. My approach will likely be to check that all confirmables are on the same page in `chain_access.sync` and otherwise enter a recovery process there.

tnull:
Had to squash some fixups after I closed #9 and reopened #33 with wallet functionality.

tnull:
Alright, after ripping out the transaction sync I had to switch to using BDK's `async-interface` feature here already in order to be able to use async Esplora calls in `Wallet`, this now also includes the changes from #26, which I just closed.
Moreover, I now include the first full end-to-end test against `ElectrsD`/`BitcoinD` backends in `regtest`. In order to clean everything up I updated the PRs this depends on and had to squash fixups everywhere.

Will get back to the open feedback ASAP.

src/lib.rs
```diff
@@ -0,0 +1,1012 @@
...
+		});
+
+		// Regularly reconnect to channel peers.
+		let connect_cm = Arc::clone(&self.channel_manager);
+		let connect_pm = Arc::clone(&self.peer_manager);
+		let connect_logger = Arc::clone(&self.logger);
+		let connect_peer_store = Arc::clone(&self.peer_store);
+		let stop_connect = Arc::clone(&stop_networking);
+		tokio_runtime.spawn(async move {
+			let mut interval = tokio::time::interval(Duration::from_secs(1));
```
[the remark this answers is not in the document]
tnull: I'm not sure I understand the issue fully. Wouldn't that simply mean that the first iteration might fail, but as soon as the BP disconnected the sockets properly the next one should be able to reconnect the sockets?

src/lib.rs
```diff
@@ -0,0 +1,1012 @@
...
+					}
+				}
+				rounds = (rounds + 1) % 5;
+
+				let confirmables = vec![
+					&*sync_cman as &(dyn Confirm + Sync),
+					&*sync_cmon as &(dyn Confirm + Sync),
+				];
+				let now = Instant::now();
+				match chain_access.sync(confirmables).await {
```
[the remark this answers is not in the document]
tnull: Hm, we spawn the corresponding tokio task/thread before anything else, in particular, before we call `background_processor::start()`. Is this enough or is it mandatory to block on the wallet syncs once before proceeding?

src/lib.rs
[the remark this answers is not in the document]
tnull: Mh, but wouldn't that mean that we need to keep track when we added the peer and see if any update since then changed the address? If so, we probably should default to not store addresses for public nodes separately and always try to look them up from gossip?

Cargo.toml
```diff
@@ -1,30 +1,57 @@
...
+#lightning-background-processor = { git = "https://github.com/lightningdevkit/rust-lightning", branch="main" }
+#lightning-rapid-gossip-sync = { git = "https://github.com/lightningdevkit/rust-lightning", branch="main" }
+
+lightning = { git = "https://github.com/tnull/rust-lightning", branch="2022-11-add-transaction-sync-crate", features = ["max_level_trace", "std"] }
+lightning-invoice = { git = "https://github.com/tnull/rust-lightning", branch="2022-11-add-transaction-sync-crate" }
+lightning-net-tokio = { git = "https://github.com/tnull/rust-lightning", branch="2022-11-add-transaction-sync-crate" }
+lightning-persister = { git = "https://github.com/tnull/rust-lightning", branch="2022-11-add-transaction-sync-crate" }
+lightning-background-processor = { git = "https://github.com/tnull/rust-lightning", branch="2022-11-add-transaction-sync-crate" }
+lightning-rapid-gossip-sync = { git = "https://github.com/tnull/rust-lightning", branch="2022-11-add-transaction-sync-crate" }
+lightning-transaction-sync = { git = "https://github.com/tnull/rust-lightning", branch="2022-11-add-transaction-sync-crate", features = ["esplora-async"] }
```
MaxFangX, line 32: Either (1) the declaration for `lightning-transaction-sync` should specify `default-features = false` to avoid including the blocking dependency graph, or (2) `esplora-blocking` should not be a default feature of `lightning-trannsaction-sync`. I think (2) is the best solution. See https://github.com/lightningdevkit/rust-lightning/pull/1870/files#r1068653127 for details.

tnull: Ah, good catch. Now fixed by removing the default feature of the upstream create.

src/wallet.rs
```diff
@@ -0,0 +1,345 @@
...
+		};
+
+		// We double-check that no inputs try to spend non-witness outputs. As we use a SegWit
+		// wallet descriptor this technically shouldn't ever happen, but better safe than sorry.
+		for input in &psbt.inputs {
+			if input.witness_utxo.is_none() {
+				log_error!(self.logger, "Tried to spend a non-witness funding output. This must not ever happen. Panicking!");
+				panic!("Tried to spend a non-witness funding output. This must not ever happen.");
+			}
+		}
```
MaxFangX, line 148: FYI this check is redundant because LDK already checks for this internally:

https://github.com/lightningdevkit/rust-lightning/blob/01fe9badc034fdc16a4450a2ce9ba5552560329f/lightning/src/ln/channelmanager.rs#L2644-L2650

It could still be kept around though, up to you

tnull: Ah, right, good point. Now removed it as this would be triply redundant (as we use Bip84 wallet descriptor, this shouldn't ever happen anyways).

src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+					user_config,
+					chain_params,
+				)
+			}
+		};
+
+		let channel_manager = Arc::new(channel_manager);
+
+		// Step 8: Give ChannelMonitors to ChainMonitor
+		for (_blockhash, channel_monitor) in channel_monitors.drain(..) {
```
jkczyz, line 345: Can drop `drain(..)` to use an implicit `into_iter()`, I believe.

src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+		));
+
+		//// Step 11: Initialize the PeerManager
+		let onion_messenger: Arc<OnionMessenger> = Arc::new(OnionMessenger::new(
+			Arc::clone(&keys_manager),
+			Arc::clone(&keys_manager),
+			Arc::clone(&logger),
+			IgnoringMessageHandler {},
+		));
+		let ephemeral_bytes: [u8; 32] = rand::thread_rng().gen();
```
jkczyz, line 364: Not sure if it matters, but should we use `KeysManager` for this?

src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+
+		let runtime = runtime_lock.as_ref().unwrap();
+
+		// TODO: ensure we never tried paying the given payment hash before
+		let status = match runtime.invoice_payer.pay_invoice(&invoice) {
+			Ok(_payment_id) => {
+				let payee_pubkey = invoice.recover_payee_pub_key();
+				// TODO: is this unwrap safe? Would a payment to an invoice with None amount ever
+				// succeed? Should we allow to set the amount in the interface or via a dedicated
+				// method?
```
jkczyz, line 848: You would have gotten an `payment::PaymentError::Invoice` error. So you'd call `pay_zero_value_invoice` instead.

src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+		Arc<InvoicePayer<Arc<EventHandler<Arc<FilesystemPersister>, Arc<FilesystemLogger>>>>>,
+	stop_networking: Arc<AtomicBool>,
+	stop_wallet_sync: Arc<AtomicBool>,
+}
+
+/// The main interface object of the simplified API, wrapping the necessary LDK and BDK functionalities.
+///
+/// Needs to be initialized and instantiated through [`Builder::build`].
+pub struct Node {
+	running: RwLock<Option<Runtime>>,
```
jkczyz, line 451: Thinking aloud here... Regarding runtime management, you could have both a `Node` and a `RunningNode`, where the latter owns the former and a `Runtime`. `Node::start()` could take a `Runtime` and return a `RunningNode`, which would have any methods needing a `Runtime`. `RunningNode::stop()` could give back the `Node`.

Haven't really thought it all the way through, but the general idea would be to lean on the type system as much as possible.

jkczyz (commented):
Only partway through.

src/io_utils.rs
```diff
@@ -0,0 +1,63 @@
...
+use lightning::util::ser::ReadableArgs;
+
+use rand::{thread_rng, RngCore};
+
+use std::fs;
+use std::io::{BufReader, Write};
+use std::path::Path;
+use std::sync::Arc;
+
+pub(crate) fn read_or_generate_seed_file(config: Arc<Config>) -> [u8; 32] {
```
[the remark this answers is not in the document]
jkczyz: Can't you deref the `Arc` and then borrow?

src/lib.rs
jkczyz: I don't have a strong opinion in this case, especially if we want an option to pay more than the invoice amount. That said, more generally, I typically would prefer to add a method rather than see a method with a growing list of `Option` parameters where users need to pass `None` for all of them to get the default behavior.

src/lib.rs
jkczyz: > Mh, generally a nice idea to rethink the approach taken for the `Runtime`. However, I see two issues with the outlined alternative approach:
> 
> 1. Giving a `Runtime` to `start()` significantly complicates the interface, especially since `Runtime` on its own would have no meaning or utility to the user. So it would be just an additional step they have to look up, i.e., an inconvenience to the user.

True, though it need not be a parameter if `start` could just create it. I hadn't looked closely enough to realize it wasn't `pub`.

> 2. While Rust's move semantics would allow us to make sure only one of the objects can be used at a time, I don't think this would work in bindings, where, as mentioned above, we'd never return just a `Node`/`RunningNode`, but always an `Arc<Node>`/`Arc<RunningNode>`. Result would be that the users might hold on to the former and hence they'd have two objects lying around that semantically should be the same thing.

I see...

I was thinking a bit about bindings for the builders used in offers. Currently each method takes `self` and returns either `Self` or `Result<Self, SemanticError>`. This won't work for bindings as you mentioned. I'm considering making a wrapper for bindings with a similar bindings-compatible interface, which would delegate to the idiomatic Rust builder and manage that internally.

Not sure if it makes sense to try something similar here where the bindings-compatible struct could internally manage `Node` / `RunningNode`. Drawback I suppose would be that the bindings interface would be different from the Rust interface.

src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+	}
+
+	/// Sets the IP address and TCP port on which [`Node`] will listen for incoming network connections.
+	///
+	/// Format: `ADDR:PORT`
+	/// Default: `0.0.0.0:9735`
+	pub fn set_listening_address(&mut self, listening_address: String) -> &mut Self {
+		self.config.listening_address = Some(listening_address);
+		self
+	}
```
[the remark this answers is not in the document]
jkczyz: Hmm... so `Config` seem ephemeral and could be replaced with setting fields on the builder, I suppose. Then again, builders sometimes have a struct like this used to setup the object it's creating, possibly using it directly to create the object. See `OfferBuilder`, `OfferContents`, and `Offer`. So I guess I'd just choose whatever variation fits this use case the best given there's different ways to implement the builder pattern.

src/lib.rs
jkczyz: To expand upon the idea, the bindings version would hold an internal enum with `Running(RunningNode)` and `Stopped(Node)` variants and have the full set of methods, erroring on the ones requiring a `RunningNode`.

src/io_utils.rs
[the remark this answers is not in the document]
jkczyz: Can you do something similar for the other parameters? Calling `deref()` on the `Arc` rather than cloning it.

src/lib.rs
jkczyz: Yeah, the rust version wouldn't let you call any methods needing a runtime if the node is not running. Maybe not worth it if the interfaces will differ and if it won't help with ridding an `Option` from the runtime in the wallet.

src/peer_store.rs
```diff
@@ -0,0 +1,209 @@
...
+		if peer_addr.is_err() || peer_addr.as_ref().unwrap().is_none() {
+			return Err(Error::PeerInfoParseFailed);
+		}
+
+		let pubkey = hex_utils::to_compressed_pubkey(pubkey.unwrap());
+		if pubkey.is_none() {
+			return Err(Error::PeerInfoParseFailed);
+		}
+
+		Ok(PeerInfo { pubkey: pubkey.unwrap(), address: peer_addr.unwrap().unwrap() })
```
jkczyz, line 175: nit: could combine `let` and `match` above to avoid a lot of unwrapping

src/lib.rs
```diff
@@ -0,0 +1,1060 @@
...
+			.duration_since(SystemTime::UNIX_EPOCH)
+			.expect("System time error: Clock may have gone backwards");
+		let keys_manager = Arc::new(KeysManager::new(
+			&seed,
+			cur_time.as_secs(),
+			cur_time.subsec_nanos(),
+			Arc::clone(&wallet),
+		));
+
+		// Step 12: Initialize the network graph, scorer, and router
```
jkczyz, line 269: Re-number?

src/lib.rs
```diff
@@ -0,0 +1,1060 @@
...
+
+		// Step 7: Initialize the ChannelManager
+		let mut user_config = UserConfig::default();
+		user_config.channel_handshake_limits.force_announced_channel_preference = false;
+		let channel_manager = {
+			if let Ok(mut f) = fs::File::open(format!("{}/manager", ldk_data_dir)) {
+				let mut channel_monitor_mut_references = Vec::new();
+				for (_, channel_monitor) in channel_monitors.iter_mut() {
+					channel_monitor_mut_references.push(channel_monitor);
+				}
```
jkczyz, line 300: Use `map` / `collect`?

src/lib.rs
jkczyz: What's the plan for zero-amount invoices?

src/tests/functional_tests.rs
```diff
@@ -0,0 +1,232 @@
...
+	assert_eq!(node_a.on_chain_balance().unwrap().get_spendable(), 100000);
+	assert_eq!(node_b.on_chain_balance().unwrap().get_spendable(), 100000);
+
+	println!("\nA -- connect_open_channel -> B");
+	let node_b_addr =
+		format!("{}@{}", node_b.node_id().unwrap(), node_b.listening_address().unwrap());
+	node_a.connect_open_channel(&node_b_addr, 50000, true).unwrap();
+
+	// Wait a sec so the funding tx can 'propagate' via EsploraD to BitcoinD.
+	std::thread::sleep(Duration::from_secs(1));
```
jkczyz, line 151: Should we query this somehow instead? Could be a source of flakiness in CI.

src/lib.rs
```diff
@@ -0,0 +1,1159 @@
...
+					},
+				);
+				Err(Error::PaymentFailed)
+			}
+		}
+	}
+
+	/// Send a payement given a so-called "zero amount" invoice, i.e., an invoice that leaves the
+	/// amount paid to be determined by the user.
+	pub fn send_adjustable_value_payment(
```
jkczyz, line 927: Maybe `send_payment_using_amount`? Seems like "adjustable value" is adding terminology that may not be necessary.

src/lib.rs
```diff
@@ -0,0 +1,1159 @@
...
+		let payment_hash = PaymentHash((*invoice.payment_hash()).into_inner());
+		let payment_secret = Some(*invoice.payment_secret());
+
+		if invoice.amount_milli_satoshis().is_some() {
+			log_error!(
+				self.logger,
+				"Failed to pay the given invoice: expected \"zero-amount\" invoice, please use send_payment."
+			);
+			return Err(Error::InvalidInvoice);
+		}
```
jkczyz, line 945: Would we want to allow paying an invoice with amount over the stated amount? (e.g. tip) Or maybe we don't support that in LDK?

src/lib.rs
```diff
@@ -0,0 +1,1159 @@
...
+			Err(payment::PaymentError::Sending(e)) => {
+				log_error!(self.logger, "Failed to send payment: {:?}", e);
+
+				outbound_payments_lock.insert(
+					payment_hash,
+					PaymentInfo {
+						preimage: None,
+						secret: payment_secret,
+						status: PaymentStatus::Failed,
+						amount_msat: invoice.amount_milli_satoshis(),
```
jkczyz, line 983: Should this be the given amount?

src/lib.rs
```diff
@@ -0,0 +1,1159 @@
...
+					},
+				);
+				Err(Error::PaymentFailed)
+			}
+		}
+	}
+
+	/// Returns a payable invoice that can be used to request and receive a payment.
+	pub fn receive_payment(
+		&self, amount_msat: Option<u64>, description: &str, expiry_secs: u32,
```
jkczyz, line 1051: Could you document `amount_msat`? Do we care that we don't have a separate method for zero-amount invoices?

jkczyz (commented):
Feel fee to squash fixups.

src/io_utils.rs
```diff
@@ -0,0 +1,61 @@
...
+	}
+}
+
+pub(crate) fn read_network_graph(
+	config: &Config, logger: Arc<FilesystemLogger>,
+) -> Result<NetworkGraph, Error> {
+	let ldk_data_dir = format!("{}/ldk", config.storage_dir_path);
+	let network_graph_path = format!("{}/network_graph", ldk_data_dir);
+
+	if let Ok(file) = fs::File::open(network_graph_path) {
```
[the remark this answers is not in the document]
jkczyz: Ah, I missed that. Why do we need to use a `Result` then?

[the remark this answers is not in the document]
jkczyz: Is this logic robust enough? For instance, if we get `PermissionDenied`, we'll create a new `NetworkGraph` and then presumably fail to write it later.

src/lib.rs
```diff
@@ -0,0 +1,1211 @@
...
+					user_config,
+					channel_monitor_references,
+				);
+				let (_hash, channel_manager) =
+					<(BlockHash, ChannelManager)>::read(&mut f, read_args)
+						.expect("Failed to read channel manager from disk");
+				channel_manager
+			} else {
+				// We're starting a fresh node.
+				let dummy_block_hash = bitcoin::blockdata::constants::genesis_block(config.network)
```
jkczyz, line 319: nit: genesis_block_hash

src/lib.rs
```diff
@@ -0,0 +1,1211 @@
...
+			keys_manager,
+			network_graph,
+			gossip_sync,
+			persister,
+			logger,
+			scorer,
+			inbound_payments,
+			outbound_payments,
+			peer_store,
+		}
```
jkczyz, line 431: Consider packing these on a few lines if rustfmt supports that. Probably worth doing in a follow-up later if it will result in a lot of churn.

src/lib.rs
```diff
@@ -0,0 +1,1211 @@
...
+		let _background_processor = BackgroundProcessor::start(
+			Arc::clone(&self.persister),
+			Arc::clone(&event_handler),
+			Arc::clone(&self.chain_monitor),
+			Arc::clone(&self.channel_manager),
+			BPGossipSync::p2p(Arc::clone(&self.gossip_sync)),
+			Arc::clone(&self.peer_manager),
+			Arc::clone(&self.logger),
+			Some(Arc::clone(&self.scorer)),
+		);
```
jkczyz, line 670: Should we use the async interface now?

src/lib.rs
```diff
@@ -0,0 +1,1211 @@
...
+		let con_success = Arc::new(AtomicBool::new(false));
+		let con_success_cloned = Arc::clone(&con_success);
+		let con_logger = Arc::clone(&self.logger);
+		let con_pm = Arc::clone(&self.peer_manager);
+
+		tokio::task::block_in_place(move || {
+			runtime.tokio_runtime.block_on(async move {
+				let res = connect_peer_if_necessary(
+					con_peer_info.pubkey,
+					con_peer_info.address,
```
jkczyz, line 746: Could consider copy/cloning these fields here and below instead of the whole struct. That way you aren't cloning the whole struct twice if it happens to grown in number of fields.

src/peer_store.rs
```diff
@@ -0,0 +1,243 @@
...
+impl<K: KVStorePersister> PeerInfoStorage<K> {
+	pub(crate) fn new(persister: Arc<K>) -> Self {
+		let peers = RwLock::new(HashMap::new());
+		Self { peers, persister }
+	}
+
+	pub(crate) fn add_peer(&self, peer_info: PeerInfo) -> Result<(), Error> {
+		let mut locked_peers = self.peers.write().unwrap();
+
+		locked_peers.insert(peer_info.pubkey, peer_info);
```
jkczyz, line 31: Is it intended to override an existing peer if present?

jkczyz (commented):
Feel free to punt on anything that requires a large change.

src/lib.rs
```diff
@@ -949,14 +949,7 @@ impl Node {
...
 		let payment_secret = Some(*invoice.payment_secret());
-		let expiry_time = invoice
-			.clone()
-			.into_signed_raw()
-			.raw_invoice()
-			.data
-			.timestamp
-			.as_duration_since_epoch()
-			+ invoice.expiry_time();
+		let expiry_time = invoice.duration_since_epoch() + invoice.expiry_time();
```
jkczyz, line 992: Probably should use `saturating_add` to prevent overflows from inputs.

src/tests/functional_tests.rs
```diff
@@ -83,6 +83,22 @@ fn wait_for_tx(txid: Txid) {
...
+	let mut is_spent = !get_electrsd().client.script_get_history(&txout_script).unwrap().is_empty();
+	loop {
+		if is_spent {
+			break;
+		}
+		is_spent = exponential_backoff_poll(|| {
+			get_electrsd().trigger().unwrap();
+			get_electrsd().client.ping().unwrap();
+			Some(!get_electrsd().client.script_get_history(&txout_script).unwrap().is_empty())
+		});
```
jkczyz, line 98: Will we ever give up? Shouldn't have the test hang if there is a problem either, I suppose.

jkczyz (approved):

PR #13, opened 2022-09-06
Initial payment store implementation

tnull:
Based on #8, #9, #10, #11.

This PR implements a `PaymentInfoStorage` that persists individual `PaymentInfo` objects after every update, which is done via a `KVStorePersister::persist()` under the key `payments/{payment_hash}`. 

(So essentially same approach as taken for events and peer storage)

As the data is currently only read from disk upon restart, it still assumes a `FilesystemPersister`, but may be adapted in the future to support the interface more generically, if `KVStorePersister` would expose some more functionality.


src/event.rs
```diff
@@ -347,49 +362,50 @@ where
...
 							amount_msat: Some(amount_msat),
-						});
-					}
-				}
+							direction: PaymentDirection::Inbound,
+							status: PaymentStatus::Succeeded,
+						}
+					};
+
+				self.payment_store.insert(payment_info).expect("Failed to access payment store");
```
jkczyz, line 385: May want to expose something like `HashMap`'s `entry().and_modify().or_insert_with()` otherwise you'll lock twice and could race if events are processed in parallel. Perhaps locking shouldn't be internal to the store?

src/payment_store.rs
```diff
@@ -0,0 +1,138 @@
...
+	persister: K,
+}
+
+impl<K: Deref> PaymentInfoStorage<K>
+where
+	K::Target: KVStorePersister,
+{
+	pub(crate) fn from_payments(mut payments: Vec<PaymentInfo>, persister: K) -> Self {
+		let payments = Mutex::new(HashMap::from_iter(
+			payments.drain(..).map(|payment_info| (payment_info.payment_hash, payment_info)),
```
jkczyz, line 88: Can use `into_iter()` instead of `drain(..)`, I believe, and remove `mut` from `payments`.

src/io_utils.rs
```diff
@@ -83,3 +86,37 @@ pub(crate) fn read_payment_info(config: &Config) -> Vec<PaymentInfo> {
 
 	payments
 }
+
+/// Provides an interface that allows a previously persisted key to be unpersisted.
+pub trait KVStoreUnpersister {
```
jkczyz, line 91: I wonder if a persister is the right abstraction. Seems we want some abstract data storage where persistence is more of an implementation detail?

jkczyz: @G8XSU How does VSS deal with removed data?

src/payment_store.rs
```diff
@@ -0,0 +1,232 @@
...
+				format!("{}/{}", PAYMENT_INFO_PERSISTENCE_PREFIX, hex_utils::to_string(&key.0));
+
+			match self.inner.entry(*key) {
+				hash_map::Entry::Vacant(_) => {
+					self.persister.unpersist(&store_key).expect("Persistence failed");
+				}
+				hash_map::Entry::Occupied(e) => {
+					self.persister.persist(&store_key, e.get()).expect("Persistence failed");
+				}
+			};
```
jkczyz, line 200: So the drawback here is that we must crash if persistence fails. I suppose we already do that elsewhere when called from the event handler. More generally, is this what we should expect to do when persistence fails? We may need approach that gives a better UX.

src/io_utils.rs
G8XSU: currently vss doesn't have a delete/remove data api. 
from my understanding, ldk-node uses del/remove for events, i am not sure if it will be using vss for that. 
Also in later commit i see that unpersist was removed.

src/io/fs_store.rs
```diff
@@ -0,0 +1,310 @@
...
+		}
+	};
+}
+
+#[cfg(target_os = "windows")]
+fn path_to_windows_str<T: AsRef<OsStr>>(path: T) -> Vec<winapi::shared::ntdef::WCHAR> {
+	path.as_ref().encode_wide().chain(Some(0)).collect()
+}
+
+pub struct FilesystemStore {
```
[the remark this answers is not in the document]
jkczyz: Does this hold for `FilesystemStore ::remove` though? At the end it checks if the file exists, which could happen if a write happens after `fs::remove_file`.

src/io/fs_store.rs
```diff
@@ -0,0 +1,302 @@
...
+		let msg = format!("Could not persist file for key {}.", prefixed_key);
+		let dest_file = PathBuf::from_str(prefixed_key).map_err(|_| {
+			lightning::io::Error::new(lightning::io::ErrorKind::InvalidInput, msg.clone())
+		})?;
+
+		let parent_directory = dest_file.parent().ok_or(lightning::io::Error::new(
+			lightning::io::ErrorKind::InvalidInput,
+			msg.clone(),
+		))?;
+		let namespace = parent_directory.display().to_string();
```
jkczyz, line 247: Why is `display()` needed here?

src/io/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+/// Keys and namespaces are required to be valid ASCII strings and the empty namespace (`""`) is
+/// assumed to be valid namespace.
+pub trait KVStore {
+	/// Returns a [`Read`] for the given `namespace` and `key` from which [`Readable`]s may be
+	/// read.
+	///
+	/// Returns an `Err` if the given `key` could not be found in the given `namespace`.
+	///
+	/// [`Readable`]: lightning::util::ser::Readable
+	fn read(&self, namespace: &str, key: &str) -> std::io::Result<Box<dyn Read>>;
```
jkczyz, line 45: Why not use an associated type instead of `Box<dyn Read>`?

src/io/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+	/// Returns a [`TransactionalWrite`] for the given `key` to which [`Writeable`]s may be written.
+	///
+	/// Will create the given `namespace` if not already present in the store.
+	///
+	/// Note that [`TransactionalWrite::commit`] MUST be called to commit the written data, otherwise
+	/// the changes won't be persisted.
+	///
+	/// [`Writeable`]: lightning::util::ser::Writeable
+	fn write(&self, namespace: &str, key: &str) -> std::io::Result<Box<dyn TransactionalWrite>>;
```
jkczyz, line 54: Likewise.

src/io/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+
+/// A [`Write`] asserting data consistency.
+///
+/// Note that any changes need to be `commit`ed for them to take effect, and are lost otherwise.
+pub trait TransactionalWrite: Write {
+	/// Persist the previously made changes.
+	fn commit(&mut self) -> std::io::Result<()>;
+}
+
+// TODO: Use this generic impl once we have Node<K>
```
jkczyz, line 73: What's `Node<K>`? Is this meant to commented out?

src/payment_store.rs
```diff
@@ -0,0 +1,249 @@
...
+			log_error!(
+				self.logger,
+				"Committing payment data for key {}/{} failed due to: {}",
+				PAYMENT_INFO_PERSISTENCE_NAMESPACE,
+				store_key,
+				e
+			);
+			Error::PersistenceFailed
+		})?;
+		Ok(())
```
jkczyz, line 205: nit: can simply return the result of `writer.commit().map_err()`

src/peer_store.rs
```diff
@@ -56,23 +54,67 @@ impl<K: KVStorePersister> PeerInfoStorage<K> {
...
+			log_error!(
+				self.logger,
+				"Committing peer data to key {}/{} failed due to: {}",
+				PEER_INFO_PERSISTENCE_NAMESPACE,
+				PEER_INFO_PERSISTENCE_KEY,
+				e
+			);
+			Error::PersistenceFailed
+		})?;
+		Ok(())
```
jkczyz, line 94: Likewise

src/lib.rs
```diff
@@ -464,31 +502,40 @@ impl Builder {
...
-		} else {
-			Arc::new(EventQueue::new(Arc::clone(&persister)))
+		let payment_store = match io::utils::read_payments(Arc::clone(&kv_store)) {
+			Ok(payments) => {
+				Arc::new(PaymentStore::new(payments, Arc::clone(&kv_store), Arc::clone(&logger)))
+			}
+			Err(e) => {
+				log_error!(logger, "Failed to read payment information: {}", e.to_string());
+				panic!("Failed to read payment information: {}", e.to_string());
+			}
```
jkczyz, line 512: Should we check against `NotFound`?

src/lib.rs
```diff
@@ -1086,15 +1135,13 @@ impl Node {
...
-				outbound_payments_lock.insert(
-					payment_hash,
-					PaymentInfo {
-						preimage: None,
-						secret: payment_secret,
-						status: PaymentStatus::Failed,
-						amount_msat: Some(amount_msat),
-					},
-				);
+				self.payment_store.update(
```
jkczyz, line 1138: Why not `insert`?

src/payment_store.rs
```diff
@@ -133,16 +128,36 @@ where
 		self.payments.lock().unwrap().contains_key(payment_hash)
 	}
 
-	pub(crate) fn set_status(
-		&self, payment_hash: &PaymentHash, payment_status: PaymentStatus,
-	) -> Result<(), Error> {
+	pub(crate) fn update(
+		&self, payment_hash: &PaymentHash, update_preimage: Option<Option<PaymentPreimage>>,
+		update_secret: Option<Option<PaymentSecret>>, update_amount_msat: Option<Option<u64>>,
+		update_status: Option<PaymentStatus>,
```
jkczyz: May be more readable at the call sites if instead of having all these parameters, have some struct with all the the fields that can be modified defaulting to `None`, and only setting the ones that need to be modified.

jkczyz (commented):
Feel free to squash the fixups.

src/event.rs
```diff
@@ -338,9 +339,10 @@ where
 							amount_msat,
 						);
 						self.channel_manager.fail_htlc_backwards(&payment_hash);
-						self.payment_store
-							.update(&payment_hash, None, None, None, Some(PaymentStatus::Failed))
-							.expect("Failed to access payment store");
+
+						let mut update = PaymentDetailsUpdate::new(payment_hash);
+						update.status = Some(PaymentStatus::Failed);
```
jkczyz, line 344: These may be cleaner using the following notation, especially as the number of fields to set grows. Your call.

```rust
let updata = PaymentDetailsUpdate {
    status: Some(PaymentStatus::Failed),
    ..PaymentDetailsUpdate::new(payment_hash)
};

**Review by jkczyz — COMMENTED** [review]
LGTM. Feel free pass on the nit. Good to squash now.

**Review by jkczyz — APPROVED** [review]

File: src/io/mod.rs [hunk 16]
```diff
@@ -0,0 +1,73 @@
...
+	fn read(&self, namespace: &str, key: &str) -> std::io::Result<Self::Reader>;
+	/// Returns a [`TransactionalWrite`] for the given `key` to which [`Writeable`]s may be written.
+	///
+	/// Will create the given `namespace` if not already present in the store.
+	///
+	/// Note that [`TransactionalWrite::commit`] MUST be called to commit the written data, otherwise
+	/// the changes won't be persisted.
+	///
+	/// [`Writeable`]: lightning::util::ser::Writeable
+	fn write(&self, namespace: &str, key: &str) -> std::io::Result<Self::Writer>;
```
G8XSU, line 56: I am still not sure if KVStore should be a stateful object, keeping track of written buffer till now. 
Can we define behavior in case of key overwrite ? 

Also if client just uses a write couple of times and never commits, are we left with dangling in-memory footprint of that buffer? (because depending on kvstore impl, one might have to keep in-memory writer object for each key not commited yet)
Moreover, what will be the expectation in case of lost in-memory buffers when commit hasn't been called yet? (due to restart or crash).

Imo, it would make more sense to have Writer buffer out of KVStore and write once we are ready with full buffer for key. It makes for a much cleaner interface and we can avoid some of the missteps.

PR #25, opened 2022-10-20
UniFFI bindings

tnull:
Based on #11.

This PR provides the starting point for UniFFI bindings. Depending on how much code changes are required for the binding generation,  we may want to move this to a separate repository.

PR #39, opened 2023-01-11
Add JVM bindings library

thunderbiscuit:
This PR builds a Kotlin JVM library which can be published to a local Maven repository, to then be used as a dependency in local projects.


thunderbiscuit:
I am attempting to build the android library PR, but I got bogged down in first trying to fix this PR and simply use the new code in the `event.rs` file (basically accepting all changes). Once I do that the glue code doesn't build anymore (a ton of errors). Wondering if I'm better off waiting for #25 to get rebased and pick up those changes before moving on?

PR #52, opened 2023-03-28
Unify interface types and expose `connect` / `disconnect` methods

tnull:
~~Based on #13.~~

So far the interface sometimes took `String`s, `&str`s, or actual Rust
types. Moreover we sometimes took the arguments by reference and
sometimes not. In order to make the interface more uniform, we here take a step towards
the Rust types and taking arguments by reference.

Note that some of the affected parts are due to change in the future,
e.g., once we transition to using upstream's `NetAddress` for peer
addresses.

Also, by user request, we expose the capability to *just* connect a peer, which is useful for example when awaiting an incoming channel from an LSP.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -103,9 +103,9 @@ use logger::{log_error, log_info, FilesystemLogger, Logger};
 
 use lightning::chain::keysinterface::EntropySource;
 use lightning::chain::{chainmonitor, BestBlock, Confirm, Watch};
-use lightning::ln::channelmanager::{self, RecipientOnionFields};
 use lightning::ln::channelmanager::{
-	ChainParameters, ChannelDetails, ChannelManagerReadArgs, PaymentId, Retry,
+	self, ChainParameters, ChannelDetails, ChannelManagerReadArgs, PaymentId, RecipientOnionFields,
+	Retry,
 };
 use lightning::ln::peer_handler::{IgnoringMessageHandler, MessageHandler};
 use lightning::ln::{PaymentHash, PaymentPreimage};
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -819,8 +821,8 @@ impl Node {
 		self.channel_manager.get_our_node_id()
 	}
 
-	/// Returns our own listening address and port.
-	pub fn listening_address(&self) -> Option<String> {
+	/// Returns our own listening address.
+	pub fn listening_address(&self) -> Option<SocketAddr> {
 		self.config.listening_address.clone()
```
jkczyz, line 825: Should we return an `Option<&SocketAddr>` instead of cloning (i.e., call `as_ref` instead)?

src/lib.rs
```diff
@@ -845,7 +847,8 @@ impl Node {
 	///
 	/// Returns a temporary channel id
 	pub fn connect_open_channel(
-		&self, node_pubkey_and_address: &str, channel_amount_sats: u64, announce_channel: bool,
+		&self, node_id: &PublicKey, address: &SocketAddr, channel_amount_sats: u64,
```
jkczyz, line 916: Hmm... I'd rather take by value instead of cloning if the method needs to own the data. I believe that is more idiomatic. The caller can clone if need be.

src/lib.rs
```diff
@@ -843,7 +843,74 @@ impl Node {
...
+				con_success_cloned.store(res.is_ok(), Ordering::Release);
+			})
+		});
+
+		if !con_success.load(Ordering::Acquire) {
+			return Err(Error::ConnectionFailed);
+		}
+
+		if permanently {
+			self.peer_store.add_peer(peer_info.clone())?;
```
jkczyz, line 882: Probably can avoid this clone by logging prior.

src/lib.rs
```diff
@@ -843,7 +843,74 @@ impl Node {
...
+		let runtime_lock = self.running.read().unwrap();
+		if runtime_lock.is_none() {
+			return Err(Error::NotRunning);
+		}
+
+		let runtime = runtime_lock.as_ref().unwrap();
+
+		let peer_info = PeerInfo { pubkey: node_id, address };
+
+		let con_peer_pubkey = peer_info.pubkey.clone();
```
jkczyz, line 861: I think the clone isn't needed here because `PublicKey` implements `Copy`.

jkczyz (commented):
LGTM modulo a couple minor comments. Feel free to squash.

PR #61, opened 2023-04-04
Allow for withdrawing on-chain funds

### LINKED ISSUE #58: Missing onchain operations
Going over the interface I didn't see a way to interact with the underline onchain wallet so currently for example I don't see a way to spend utxos that are the result of closed channels.
Are there any plans to add such API to the interface? Or is it deliberately designed like this and the intention it to access the wallet in some other way?

tnull:
~~Based on #13~~, fixes #58.

Previously we only exposed methods to generate new addresses, retrieve
the balance, and fund channels. Here, we fix the oversight and allow
users to actually withdraw their funds again.

We also rename some API methods for consistent use of the term `onchain`.

### DIFF
--- a/src/wallet.rs
+++ b/src/wallet.rs
@@ -22,7 +22,7 @@ use bitcoin::bech32::u5;
 use bitcoin::secp256k1::ecdh::SharedSecret;
 use bitcoin::secp256k1::ecdsa::{RecoverableSignature, Signature};
 use bitcoin::secp256k1::{PublicKey, Scalar, Secp256k1, Signing};
-use bitcoin::{Script, Transaction, TxOut};
+use bitcoin::{Script, Transaction, TxOut, Txid};
 
 use std::collections::HashMap;
 use std::sync::{Arc, Condvar, Mutex, RwLock};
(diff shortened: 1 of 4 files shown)


src/wallet.rs
```diff
@@ -208,6 +208,60 @@ where
 		Ok(self.inner.lock().unwrap().get_balance()?)
 	}
 
+	pub(crate) fn send_to_address(
```
roeierez, line 211: Perhaps it is better to allow pass an optional fee rate?
Also a lightning node user would probably want to send all the onchain funds (empty the wallet) so perhaps this use case should be taken into account when exposing this ability at the Node level.

src/wallet.rs
```diff
@@ -208,6 +208,112 @@ where
...
+			"Created new transaction {} sending {}sats on-chain to address {}",
+			txid,
+			amount_sats,
+			address
+		);
+
+		Ok(txid)
+	}
+
+	pub(crate) fn drain_to_address(&self, address: &bitcoin::Address) -> Result<Txid, Error> {
```
jkczyz, line 265: Could we dry this up by delegating to `send_to_address` or I suppose some common helper taking `Option<u64>` for the amount?

jkczyz (commented):
LGTM. Please squash.

jkczyz (approved):

PR #63, opened 2023-04-11
Allow pushing some msats on channel open

tnull:
We had so far omitted the `push_msat` field, but it's likely useful, so we add it here.

### DIFF
--- a/src/test/functional_tests.rs
+++ b/src/test/functional_tests.rs
@@ -20,20 +20,24 @@ fn channel_full_cycle() {
 	node_b.start().unwrap();
 	let addr_b = node_b.new_funding_address().unwrap();
 
+	let premine_amount_sat = 100_000;
+
 	premine_and_distribute_funds(
 		&bitcoind,
 		&electrsd,
 		vec![addr_a, addr_b],
-		Amount::from_sat(100000),
+		Amount::from_sat(premine_amount_sat),
 	);
 	node_a.sync_wallets().unwrap();
 	node_b.sync_wallets().unwrap();
-	assert_eq!(node_a.on_chain_balance().unwrap().get_spendable(), 100000);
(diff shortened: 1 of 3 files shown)


src/test/functional_tests.rs
```diff
@@ -34,14 +34,15 @@ fn channel_full_cycle() {
...
-	node_a.connect_open_channel(&node_b_addr, 50000, true).unwrap();
+	node_a.connect_open_channel(&node_b_addr, 80_000, Some(40_000 * 1000), true).unwrap();
 
 	let funding_txo = loop {
 		let details = node_a.list_channels();
 
 		if details.is_empty() || details[0].funding_txo.is_none() {
 			std::thread::sleep(Duration::from_secs(1));
 		} else {
+			assert_eq!(details[0].inbound_capacity_msat, 39000000);
```
jkczyz, line 45: Just to refresh my own understanding, why is this not `40000000`? A reserve?

src/test/functional_tests.rs
```diff
@@ -54,8 +55,8 @@ fn channel_full_cycle() {
 	node_b.sync_wallets().unwrap();
 
 	let node_a_balance = node_a.on_chain_balance().unwrap();
-	assert!(node_a_balance.get_spendable() < 50000);
-	assert!(node_a_balance.get_spendable() > 40000);
+	assert!(node_a_balance.get_spendable() < 20000);
+	assert!(node_a_balance.get_spendable() > 15000);
```
jkczyz, line 59: Why not `10000`? Any reason why we don't assert exact values?

jkczyz: Yeah, I'd like to at least make the code clearer that this happening by writing the asserts as the arithmetic of variables with meaningful names.

jkczyz (commented):
LGTM. Please squash.

jkczyz (approved):

PR #66, opened 2023-04-20
Allow to configure seed via a BIP 39 mnemonic

### LINKED ISSUE #47: Allow to read Bip39 mnemonic
While specifying the seed bytes or a seed file location has been covered in #11 already, we want to enable reading in a mnemonic string. 

Currently still blocked on availability of a version-compatible `bip39` crate exposed via BDK.

tnull:
Fixes #47, Based on #64 

Since it's unblocked now, we allow users to give the seed via a BIP 39 mnemonic.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,10 +41,11 @@ lightning-transaction-sync = { git = "https://github.com/lightningdevkit/rust-li
 #lightning-rapid-gossip-sync = { path = "../rust-lightning/lightning-rapid-gossip-sync" }
 #lightning-transaction-sync = { path = "../rust-lightning/lightning-transaction-sync", features = ["esplora-async"] }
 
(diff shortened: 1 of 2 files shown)


danielnordh:
Does this also expose the ability to create a mnemonic?
I.e via BDK's Mnemonic() function.

As I currently use BDK to generate the mnemonic for new users I would like to be able to do the same with ldk-node

danielnordh:
> > Does this also expose the ability to create a mnemonic? I.e via BDK's Mnemonic() function.
> > As I currently use BDK to generate the mnemonic for new users I would like to be able to do the same with ldk-node
> 
> In Rust we to this end re-export the `bip39` crate for users to have it handy. In bindings, I imagine we'll expose this as a function that just takes the Mnemonic as a string and builds the object for the user.

I would also want a function that GETs a new mnemonic, for first time users.

jkczyz (approved):

PR #68, opened 2023-04-21
Switch to async background processing and cleanup runtime handling.

### LINKED ISSUE #37: Use async event handler and background processor
An async variant of the background processor shipped with 0.0.112 (see https://github.com/lightningdevkit/rust-lightning/pull/1657).

We should switch to this, also requiring us to make the event handler async.

This is kind of straightforward, *except* for the current implementation of `process_pending_events_async` holding a `RwLockGuard` over `await` boundaries (https://github.com/lightningdevkit/rust-lightning/blob/b79ff71fe7587b03ab0b7fe8b6229eb4bc2d7f08/lightning/src/ln/channelmanager.rs#L5262-L5279), which is not `Send` and therefore doesn't play with tokio nicely. 

We therefore first need to address this upstream, e.g., by introducing a `Send` lock variant.

tnull:
Fixes #37~~, currently based on https://github.com/lightningdevkit/rust-lightning/pull/2199.~~
We switch to an async BP as upstream now allows us to.

We furthermore cleanup our runtime handling: 
So far we instantiated a dedicated `Runtime` object which also held the BP. This was clunky and somewhat error-prone under different circumstances. As we now got rid of the BP object, we here remove the dedicated `Runtime` object and just pass around an `Arc<RwLock<Option<Runtime>>>` upon init, which allows us to get rid of the `set`/`drop` methods on `Wallet` as an added benefit.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -347,7 +347,13 @@ impl Builder {
 			EsploraBlockchain::from_client(tx_sync.client().clone(), BDK_CLIENT_STOP_GAP)
 				.with_concurrency(BDK_CLIENT_CONCURRENCY);
 
-		let wallet = Arc::new(Wallet::new(blockchain, bdk_wallet, Arc::clone(&logger)));
+		let runtime = Arc::new(RwLock::new(None));
+		let wallet = Arc::new(Wallet::new(
+			blockchain,
+			bdk_wallet,
+			Arc::clone(&runtime),
+			Arc::clone(&logger),
+		));
 
 		let kv_store = Arc::new(FilesystemStore::new(ldk_data_dir.clone().into()));
 
@@ -556,10 +562,11 @@ impl Builder {
 			}
 		};
 
-		let running = RwLock::new(None);
+		let stop_running = Arc::new(AtomicBool::new(false));
 
 		Node {
-			running,
+			runtime,
+			stop_running,
 			config,
 			wallet,
 			tx_sync,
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -25,22 +25,30 @@ description = "A ready-to-go node implementation built using LDK."
...
 #lightning-transaction-sync = { path = "../rust-lightning/lightning-transaction-sync", features = ["esplora-async"] }
 
+lightning = { git = "https://github.com/tnull/rust-lightning", branch="2023-04-fix-async-event-processing", features = ["max_level_trace", "std"] }
+lightning-invoice = { git = "https://github.com/tnull/rust-lightning", branch="2023-04-fix-async-event-processing" }
+lightning-net-tokio = { git = "https://github.com/tnull/rust-lightning", branch="2023-04-fix-async-event-processing" }
+lightning-persister = { git = "https://github.com/tnull/rust-lightning", branch="2023-04-fix-async-event-processing" }
+lightning-background-processor = { git = "https://github.com/tnull/rust-lightning", branch="2023-04-fix-async-event-processing", features = ["futures"] }
+lightning-rapid-gossip-sync = { git = "https://github.com/tnull/rust-lightning", branch="2023-04-fix-async-event-processing" }
+lightning-transaction-sync = { git = "https://github.com/tnull/rust-lightning", branch="2023-04-fix-async-event-processing", features = ["esplora-async"] }
+
```
jkczyz, line 51: Should we use 115 now?

src/lib.rs
```diff
@@ -785,19 +782,45 @@ impl Node {
...
+		let background_event_handler = Arc::clone(&event_handler);
+		let background_chain_mon = Arc::clone(&self.chain_monitor);
+		let background_chan_man = Arc::clone(&self.channel_manager);
+		let background_gossip_sync = BPGossipSync::p2p(Arc::clone(&self.gossip_sync));
+		let background_peer_man = Arc::clone(&self.peer_manager);
+		let background_logger = Arc::clone(&self.logger);
+		let background_scorer = Arc::clone(&self.scorer);
+		let stop_background_processing = Arc::clone(&stop_runtime);
+		let sleeper = move |d| {
+			let stop = Arc::clone(&stop_background_processing);
```
jkczyz, line 762: Is a clone needed here?

Cargo.toml
```diff
@@ -29,7 +29,7 @@ lightning-transaction-sync = { version = "0.0.115", features = ["esplora-async-h
 #lightning-invoice = { path = "../rust-lightning/lightning-invoice" }
 #lightning-net-tokio = { path = "../rust-lightning/lightning-net-tokio" }
 #lightning-persister = { path = "../rust-lightning/lightning-persister" }
-#lightning-background-processor = { path = "../rust-lightning/lightning-background-processor" }
+#lightning-background-processor = { path = "../rust-lightning/lightning-background-processor", features = ["futures"] }
```
jkczyz, line 32: Should this be added to the block used for `branch="main"`

jkczyz (approved):

PR #69, opened 2023-04-21
Allow for trusted inbound 0conf channels

tnull:
~~Based on #25.~~

We set the peer to manually handle incoming channel request and allow inbound 0conf channels for trusted peers.

Setting this to draft as a) I'm not sure we'd always want to go the manually accept route and b) we want to always expose this  as an additional bool on connect methods and c) this needs additional tests.

### DIFF
--- a/src/peer_store.rs
+++ b/src/peer_store.rs
@@ -32,6 +32,10 @@ where
 	pub(crate) fn add_peer(&self, peer_info: PeerInfo) -> Result<(), Error> {
 		let mut locked_peers = self.peers.write().unwrap();
 
+		if locked_peers.contains_key(&peer_info.node_id) {
+			return Ok(());
+		}
+
 		locked_peers.insert(peer_info.node_id, peer_info);
 		self.persist_peers(&*locked_peers)
 	}


src/lib.rs
```diff
@@ -247,6 +248,12 @@ pub struct Config {
 	///
 	/// **Note:** A minimum of 10 seconds is always enforced.
 	pub fee_rate_cache_update_interval_secs: u64,
+	/// A list of peers which we allow to establish zero confirmation channels to us.
```
jkczyz, line 251: s/which/that

src/lib.rs
```diff
@@ -247,6 +248,12 @@ pub struct Config {
...
+	///
+	/// **Note:** Allowing payments via zero-confirmation channels is potentially insecure if the
+	/// funding transaction ends up never being confirmed on-chain. Zero-confirmation channels
+	/// should therefore only be accepted from trusted peers.
+	pub peers_trusted_0conf: Vec<PublicKey>,
```
jkczyz, line 256: How about `trusted_peers_0conf` for consistency with `accept_inbound_channel_from_trusted_peer_0conf`

src/peer_store.rs
```diff
@@ -32,6 +32,10 @@ where
 	pub(crate) fn add_peer(&self, peer_info: PeerInfo) -> Result<(), Error> {
 		let mut locked_peers = self.peers.write().unwrap();
 
+		if locked_peers.contains_key(&peer_info.node_id) {
+			return Ok(());
+		}
```
jkczyz, line 37: Could you make this a separate commit stating why this is needed? Assuming this is needed regardless of the larger change.

src/lib.rs
```diff
@@ -569,6 +577,11 @@ impl Builder {
 		// Initialize the ChannelManager
 		let mut user_config = UserConfig::default();
 		user_config.channel_handshake_limits.force_announced_channel_preference = false;
+		if !config.peers_trusted_0conf.is_empty() {
+			// Manually accept inbound channels if we expect 0conf channel requests, avoid
+			// generating the events otherwise.
+			user_config.manually_accept_inbound_channels = true;
+		}
```
jkczyz, line 584: Should users be able to add trusted peers while the node is running?

jkczyz (commented):
LGTM. Feel free to squash fixups.

jkczyz (approved):

PR #70, opened 2023-04-25
Implement configurable `GossipSource` to allow RGS gossip updates

### LINKED ISSUE #2: Support `RapidGossipSync`
Currently we only sync the gossip state via the P2P network. 
We should however allow to configure a RGS server and use it for the initial sync.

We may also consider using RGS when we haven't synced in a long time and it may be faster to forget our old state and use RGS to start from scratch.

tnull:
Fixes #2, ~~Based on #68.~~

Here we implement a `GossipSource` object that allows us to configure P2P or RGS gossip sources **without** the need to propagate and leak LDK type parameters upwards. To this end, `GossipSource` wraps the corresponding variants and implements a `RoutingMessageHandler` that is delegating or ignoring the incoming messages.

### DIFF
--- a/src/gossip.rs
+++ b/src/gossip.rs
@@ -30,10 +30,11 @@ impl GossipSource {
 	}
 
 	pub fn new_rgs(
-		server_url: String, network_graph: Arc<NetworkGraph>, logger: Arc<FilesystemLogger>,
+		server_url: String, latest_sync_timestamp: u32, network_graph: Arc<NetworkGraph>,
+		logger: Arc<FilesystemLogger>,
 	) -> Self {
 		let gossip_sync = Arc::new(RapidGossipSync::new(network_graph, Arc::clone(&logger)));
-		let latest_sync_timestamp = AtomicU32::new(0);
+		let latest_sync_timestamp = AtomicU32::new(latest_sync_timestamp);
 		Self::RapidGossipSync { gossip_sync, server_url, latest_sync_timestamp, logger }
 	}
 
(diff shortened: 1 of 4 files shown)


jkczyz:
> Fixes #2, ~Based on #68.~
> 
> Here we implement a `GossipSource` object that allows us to configure P2P or RGS gossip sources **without** the need to propagate and leak LDK type parameters upwards. To this end, `GossipSource` wraps the corresponding variants and implements a `RoutingMessageHandler` that is delegating or ignoring the incoming messages.

Would `dyn RoutingMessageHandler` work? Then you wouldn't need the boilerplate of implementing `RoutingMessageHandler` for `GossipSource`.

src/lib.rs
```diff
@@ -546,21 +548,44 @@ impl Builder {
...
+					onion_message_handler: onion_messenger,
+				};
+				Arc::new(PeerManager::new(
+					msg_handler,
+					cur_time.as_secs().try_into().expect("System time error"),
+					&ephemeral_bytes,
+					Arc::clone(&logger),
+					IgnoringMessageHandler {},
+					Arc::clone(&keys_manager),
+				))
```
jkczyz, line 566: Can DRY this up by moving it outside the `match`.

src/gossip.rs
```diff
@@ -0,0 +1,90 @@
...
+	pub fn is_p2p(&self) -> bool {
+		if let Self::P2PNetwork { .. } = self {
+			true
+		} else {
+			false
+		}
+	}
+
+	pub fn is_rgs(&self) -> bool {
+		!self.is_p2p()
```
jkczyz, line 49: May want to explicitly check for `RapidGossipSync` in case the enum changes in the future to have more variants.

src/lib.rs
```diff
@@ -546,21 +548,44 @@ impl Builder {
...
+			}
+			GossipSync::None => {
+				panic!();
```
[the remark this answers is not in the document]
jkczyz: `unreachable!()`?

jkczyz (commented):
Feel free to sqush.

PR #84, opened 2023-05-02
Call `stop()` on `Node` drop

### LINKED ISSUE #79: Call `Node::stop()` on `drop()`

tnull:
Fixes #79 

We simply call `stop()` on dropping the `Node` object to allow for a cleaner shutdown by default.

We also include a minor commit which changes the API to take `SocketAddrs` and `PublicKey`s by value as they'll be copied/dereferenced in most cases anyways and this makes the API a tiny bit more uniform.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -953,7 +953,7 @@ impl Node {
 	///
 	/// Will also remove the peer from the peer store, i.e., after this has been called we won't
 	/// try to reconnect on restart.
-	pub fn disconnect(&self, counterparty_node_id: &PublicKey) -> Result<(), Error> {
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1440,6 +1440,12 @@ impl Node {
 	}
 }
 
+impl Drop for Node {
+	fn drop(&mut self) {
+		let _ = self.stop();
```
jkczyz, line 1445: If the node is already stopped, this will return `Error::NotRunning`. I guess that is ok though.

jkczyz (approved):

PR #88, opened 2023-05-09
Break builder pattern to expose all setters via bindings 

### LINKED ISSUE #65: Use enums for Network types instead of strings
To simplify switch statements depending on the Network type it would be helpful if `Network` was enums instead of strings.
(I assume this is similar in all languages but specifically wanted this in Swift)

See BDK:

```
public enum Network {
    
    case `bitcoin`
    case `testnet`
    case `signet`
    case `regtest`
}
```

tnull:
Closes #65 

Unfortunately there is no good way of exposing the builder pattern via Uniffi bindings currently. We therefore resort to internal mutability and using setters on the `Builder` object.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -86,6 +86,8 @@ mod peer_store;
 #[cfg(test)]
 mod test;
 mod types;
+#[cfg(feature = "uniffi")]
+mod uniffi_types;
 mod wallet;
 
 pub use bip39;
@@ -99,6 +101,9 @@ use error::Error;
 
 pub use event::Event;
 pub use types::NetAddress;
+
+#[cfg(feature = "uniffi")]
+use {bitcoin::OutPoint, lightning::ln::PaymentSecret, uniffi_types::*};
 
 use event::{EventHandler, EventQueue};
 use gossip::GossipSource;
@@ -122,7 +127,7 @@ use lightning::ln::channelmanager::{
 	self, ChainParameters, ChannelManagerReadArgs, PaymentId, RecipientOnionFields, Retry,
 };
 use lightning::ln::peer_handler::{IgnoringMessageHandler, MessageHandler};
-use lightning::ln::{PaymentHash, PaymentPreimage, PaymentSecret};
+use lightning::ln::{PaymentHash, PaymentPreimage};
 use lightning::routing::scoring::{ProbabilisticScorer, ProbabilisticScoringParameters};
 
 use lightning::util::config::{ChannelHandshakeConfig, ChannelHandshakeLimits, UserConfig};
@@ -147,7 +152,7 @@ use bitcoin::Network;
 
 use bip39::Mnemonic;
 
-use bitcoin::{Address, BlockHash, OutPoint, Txid};
+use bitcoin::{Address, BlockHash, Txid};
 
 use rand::Rng;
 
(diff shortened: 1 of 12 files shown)


src/lib.rs
```diff
@@ -223,107 +225,107 @@ enum GossipSourceConfig {
...
 	/// Configures the [`Node`] instance to source its wallet entropy from a seed file on disk.
 	///
 	/// If the given file does not exist a new random seed file will be generated and
 	/// stored at the given location.
-	pub fn set_entropy_seed_path(&mut self, seed_path: String) -> &mut Self {
-		self.entropy_source_config = Some(EntropySourceConfig::SeedFile(seed_path));
-		self
+	pub fn set_entropy_seed_path(&self, seed_path: String) {
+		*self.entropy_source_config.write().unwrap() =
+			Some(EntropySourceConfig::SeedFile(seed_path));
```
jkczyz, line 285: Does Uniffi not even support `&mut self` methods?

bindings/kotlin/ldk-node-jvm/lib/src/test/kotlin/org/lightningdevkit/ldknode/LibraryTest.kt
```diff
@@ -61,8 +61,7 @@ class LibraryTest {
     @Test fun fullCycle() {
         setup()
 
-        val network: Network = "regtest"
-        assertEquals(network, "regtest")
+        val network = Network.REGTEST
```
jkczyz, line 64: Does Uniffi map `Regtest` to `REGTEST`?

src/lib.rs
```diff
@@ -161,6 +161,13 @@ use std::time::{Duration, Instant, SystemTime};
 
 uniffi::include_scaffolding!("ldk_node");
 
+// Config defaults
+const DEFAULT_STORAGE_DIR_PATH: &str = "/tmp/ldk_node/";
+const DEFAULT_NETWORK: Network = Network::Bitcoin;
```
jkczyz, line 166: Should leave this as `Regtest` and update to `Bitcoin` in a separate commit.

src/lib.rs
```diff
@@ -389,14 +410,30 @@ impl Builder {
...
-			Arc::clone(&logger),
-		));
-
-		let blockchain =
-			EsploraBlockchain::from_client(tx_sync.client().clone(), BDK_CLIENT_STOP_GAP)
-				.with_concurrency(BDK_CLIENT_CONCURRENCY);
+		let (blockchain, tx_sync) = if let Some(chain_data_source_config) =
+			&*self.chain_data_source_config.read().unwrap()
+		{
+			match chain_data_source_config {
```
jkczyz, line 422: This may be more readable if you combine the `if let` and `match` into a single match:

```rust
let (blockchain, tx_sync) = match &*self.chain_data_source_config.read().unwrap() {
	Some(ChainDataSourceConfig::Esplora(server_url)) => {
	},
	None => {
	},
};

File: src/lib.rs [hunk 5]
```diff
@@ -163,7 +163,7 @@ uniffi::include_scaffolding!("ldk_node");
 
 // Config defaults
 const DEFAULT_STORAGE_DIR_PATH: &str = "/tmp/ldk_node/";
-const DEFAULT_NETWORK: Network = Network::Regtest;
+const DEFAULT_NETWORK: Network = Network::Bitcoin;
```
jkczyz: Message on the previous commit needs to be updated to remove the part about updating `DEFAULT_NETWORK`.

jkczyz (commented):
Looks good minus a couple easy comments.

jkczyz (approved):

PR #95, opened 2023-05-12
Include non-permanently connected peers in `list_peers()`

### LINKED ISSUE #62: Include non-permanently connected peers in `list_peers`
As of #56, we allow to retrieve all permanently stored peers and add a `is_connected: bool` field based on their connection status.

It would however be nice to eventually return the superset of stored *and* connected (i.e., also peer not marked as `permanently`) peers. 

This is much easier after https://github.com/lightningdevkit/rust-lightning/issues/2056, which is why we postpone until then.

tnull:
Closes #62, ~~based on #85~~

This is a follow-up to #56, as we now are able to include non-permanently connected peers in `list_peers()`. 
We also include a number of minor doc fixes as a second commit.


src/types.rs
```diff
@@ -130,26 +130,26 @@ impl Readable for UserChannelId {
...
 /// Details of a channel as returned by [`Node::list_channels`].
 ///
-/// [`Node::list_channels`]: [`crate::Node::list_channels`]
+/// [`Node::list_channels`]: crate::Node::list_channels
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ChannelDetails {
-	/// The channel's ID (prior to funding transaction generation, this is a random 32 bytes,
-	/// thereafter this is the transaction ID of the funding transaction XOR the funding transaction
-	/// output).
+	/// The channel ID (prior to funding transaction generation, this is a random 32 byte
```
jkczyz, line 136: s/32 byte/32-byte

src/types.rs
```diff
@@ -130,26 +130,26 @@ impl Readable for UserChannelId {
...
 	/// The channel's funding transaction output, if we've negotiated the funding transaction with
 	/// our counterparty already.
 	pub funding_txo: Option<OutPoint>,
-	/// The value, in satoshis, of this channel as appears in the funding output.
-	pub channel_value_satoshis: u64,
-	/// The value, in satoshis, that must always be held in the channel for us. This value ensures
-	/// that if we broadcast a revoked state, our counterparty can punish us by claiming at least
-	/// this value on chain.
+	/// The value, in satoshis, of this channel as it appears in the funding output.
+	pub channel_value_sats: u64,
```
jkczyz, line 149: We use the plural sats but the singular msat below, it seems.

src/lib.rs
```diff
@@ -1833,17 +1833,43 @@ impl<K: KVStore + Sync + Send + 'static> Node<K> {
...
+			let is_connected = true;
+			let details = PeerDetails { node_id, address, is_persisted, is_connected };
+			peers.push(details);
+		}
+
+		// Now add all known-but-offline peers, too.
+		for p in self.peer_store.list_peers() {
+			if peers.iter().find(|d| d.node_id == p.node_id).is_some() {
+				continue;
+			}
```
jkczyz, line 1860: Seems we can be more efficient here by only iterating through the initial set of connected peers instead of the growing list.

src/lib.rs
```diff
@@ -1837,24 +1837,25 @@ impl<K: KVStore + Sync + Send + 'static> Node<K> {
...
 		// First add all connected peers, preferring to list the connected address if available.
 		let connected_peers = self.peer_manager.get_peer_node_ids();
-		for (node_id, con_addr_opt) in connected_peers {
+		for (node_id, con_addr_opt) in &connected_peers {
 			let stored_peer = self.peer_store.get_peer(&node_id);
 			let stored_addr_opt = stored_peer.as_ref().map(|p| p.address.clone());
+
 			let address = match (con_addr_opt, stored_addr_opt) {
-				(Some(con_addr), _) => NetAddress(con_addr),
+				(Some(con_addr), _) => NetAddress(con_addr.clone()),
```
jkczyz, line 1845: If you store the number of connected peers, you can avoid this clone by using `peers.iter().take(len)` below and keeping the code as it originally was.

jkczyz (commented):
LGTM. Please squash fixups.

jkczyz (approved):

PR #96, opened 2023-05-12
Only avoid duplicate payments if we didn't fail sending (and expose `list_payments`)

tnull:
Previously, we went from not tracking payments when they immediately failed to send to always tracking them but denying any retries for payments whose payment hash was known to us. However, this is of course much too restrictive as payments might fail due to local connections issues or unavailability of usable channels, in which case the invoices/payment hashes would be 'burnt' and couldn't get retried.

Here, we introduce a new `PaymentStatus::SendingFailed` to differentiate send failures and allow retrying tracked payments if they failed to send immediately. We also rename some error types accordingly for further clarification.

Additionally, we expose a simple `Node::list_payments` variant, mainly to be used in bindings where we can't make use of the `_with_filter` variant.

### DIFF
--- a/src/payment_store.rs
+++ b/src/payment_store.rs
@@ -209,7 +209,7 @@ mod tests {
 	use std::sync::Arc;
 
 	#[test]
-	fn persistence_guard_persists_on_drop() {
+	fn payment_info_is_persisted() {
 		let store = Arc::new(TestStore::new());
 		let logger = Arc::new(TestLogger::new());
 		let payment_store = PaymentStore::new(Vec::new(), Arc::clone(&store), logger);


src/error.rs
```diff
@@ -86,7 +86,9 @@ impl fmt::Display for Error {
 			Self::InvalidInvoice => write!(f, "The given invoice is invalid."),
 			Self::InvalidChannelId => write!(f, "The given channel ID is invalid."),
 			Self::InvalidNetwork => write!(f, "The given network is invalid."),
-			Self::DuplicatePayment => write!(f, "A payment with the given hash has already been initiated."),
+			Self::DuplicatePayment => {
+				write!(f, "A payment with the given hash has already been initiated.")
+			}
```
jkczyz: Looks like this belongs in the previous commit.

src/lib.rs
```diff
@@ -1339,11 +1341,16 @@ impl Node {
...
 				};
 				self.payment_store.insert(payment)?;
 
-				Err(Error::PaymentFailed)
+				match e {
+					channelmanager::RetryableSendFailure::DuplicatePayment => {
+						Err(Error::DuplicatePayment)
+					}
+					_ => Err(Error::PaymentSendingFailed),
+				}
```
jkczyz, line 1353: Wouldn't we want to not update the `payment_store` if it is a duplicate?

src/lib.rs
```diff
@@ -1334,22 +1334,23 @@ impl Node {
...
+						let payment = PaymentDetails {
+							preimage: None,
+							hash: payment_hash,
+							secret: payment_secret,
+							amount_msat: invoice.amount_milli_satoshis(),
+							direction: PaymentDirection::Outbound,
+							status: PaymentStatus::SendingFailed,
+						};
+
+						self.payment_store.insert(payment)?;
```
jkczyz, line 1351: So we have any tests that check the payment store? Might be worth adding a test case for duplicate payments.

jkczyz: But we don't check that the status remains the same as before in case of a duplicate error, IIUC.

jkczyz (commented):
LGTM. Please squash.

jkczyz (approved):

PR #99, opened 2023-05-18
Feature: sign arbitrary msg and verify msg' sig

### LINKED ISSUE #97: Add message signing support
We should allow users to have LDK node sign arbitrary data

telelvis:
Closes #97 

This adds new methods to Node & KeysManager to sign any message and verify signature.
Both rely on ldk::util to do it and use node' secret and derived public key.

Plus it adds a test for above methods.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1556,6 +1556,14 @@ impl Node {
 			})
 			.collect()
 	}
+
+	pub fn sign_message(&self, msg: &[u8]) -> Result<String, Error> {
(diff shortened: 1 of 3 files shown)


telelvis:
Appreciate the review and comments @tnull ! I've made required changes.
I just realized that verify method is not extremely useful, probably user would want to check signature produced by some other node and not self. But we can do it later.

telelvis:
thank you for patience! I've made adjustments

PR #100, opened 2023-05-18
Add `SqliteStore` backend

tnull:
~~Based on #101.~~

Finally, we add an SQLite-based `KVStore` implementation which becomes the new default
default for `Node::build`. The `FilesystemStore` is still configurable via `Node::build_with_fs_store`

### DIFF
--- a/src/io/sqlite_store.rs
+++ b/src/io/sqlite_store.rs
@@ -1,0 +1,192 @@
+use super::*;
+
+use lightning::util::persist::KVStorePersister;
+use lightning::util::ser::Writeable;
+
+use rusqlite::{named_params, Connection};
+
+use std::fs;
+use std::io::Cursor;
+use std::path::PathBuf;
+use std::sync::{Arc, Mutex};
+
+// The database file name.
+const SQLITE_DB_FILE: &str = "ldk_node.sqlite";
+
+// The table in which we store all data.
+const KV_TABLE_NAME: &str = "ldk_node_data";
+
+// The current SQLite `user_version`, which we can use if we'd ever need to do a schema migration.
+const SCHEMA_USER_VERSION: u16 = 1;
+
+/// A [`KVStore`] implementation that writes to and reads from an [SQLite] database.
+///
+/// [SQLite]: https://sqlite.org
+pub struct SqliteStore {
+	connection: Arc<Mutex<Connection>>,
+}
+
+impl SqliteStore {
+	pub(crate) fn new(dest_dir: PathBuf) -> Self {
+		fs::create_dir_all(dest_dir.clone()).unwrap_or_else(|_| {
+			panic!("Failed to create database destination directory: {}", dest_dir.display())
+		});
+		let mut db_file_path = dest_dir.clone();
+		db_file_path.push(SQLITE_DB_FILE);
+
+		let connection = Connection::open(db_file_path.clone()).unwrap_or_else(|_| {
+			panic!("Failed to open/create database file: {}", db_file_path.display())
+		});
+
+		connection
+			.pragma(Some(rusqlite::DatabaseName::Main), "user_version", SCHEMA_USER_VERSION, |_| {
+				Ok(())
+			})
+			.unwrap_or_else(|_| panic!("Failed to set PRAGMA user_version"));
+
+		let sql = format!(
+			"CREATE TABLE IF NOT EXISTS {} (
+			namespace TEXT NOT NULL,
+			key TEXT NOT NULL CHECK (key <> ''),
+			value BLOB, PRIMARY KEY ( namespace, key )
+		);",
+			KV_TABLE_NAME
+		);
+		connection
+			.execute(&sql, [])
+			.unwrap_or_else(|_| panic!("Failed to create table: {}", KV_TABLE_NAME));
+
+		let connection = Arc::new(Mutex::new(connection));
+		Self { connection }
+	}
+}
+
+impl KVStore for SqliteStore {
+	type Reader = Cursor<Vec<u8>>;
+
+	fn read(&self, namespace: &str, key: &str) -> std::io::Result<Self::Reader> {
+		let locked_conn = self.connection.lock().unwrap();
+		let sql =
+			format!("SELECT value FROM {} WHERE namespace=:namespace AND key=:key;", KV_TABLE_NAME);
+
+		let res = locked_conn
+			.query_row(
+				&sql,
+				named_params! {
+					":namespace": namespace,
+					":key": key,
+				},
+				|row| row.get(0),
+			)
+			.map_err(|e| match e {
+				rusqlite::Error::QueryReturnedNoRows => {
+					let msg =
+						format!("Failed to read as key could not be found: {}/{}", namespace, key);
+					std::io::Error::new(std::io::ErrorKind::NotFound, msg)
+				}
+				e => {
+					let msg = format!("Failed to read from key {}/{}: {}", namespace, key, e);
+					std::io::Error::new(std::io::ErrorKind::Other, msg)
+				}
+			})?;
+		Ok(Cursor::new(res))
+	}
+
+	fn write(&self, namespace: &str, key: &str, buf: &[u8]) -> std::io::Result<()> {
+		let locked_conn = self.connection.lock().unwrap();
+
+		let sql = format!(
+			"INSERT OR REPLACE INTO {} (namespace, key, value) VALUES (:namespace, :key, :value);",
+			KV_TABLE_NAME
+		);
+
+		locked_conn
+			.execute(
+				&sql,
+				named_params! {
+					":namespace": namespace,
+					":key": key,
+					":value": buf,
+				},
+			)
+			.map(|_| ())
+			.map_err(|e| {
+				let msg = format!("Failed to write to key {}/{}: {}", namespace, key, e);
+				std::io::Error::new(std::io::ErrorKind::Other, msg)
+			})
+	}
+
+	fn remove(&self, namespace: &str, key: &str) -> std::io::Result<bool> {
+		let locked_conn = self.connection.lock().unwrap();
+
+		let sql = format!("DELETE FROM {} WHERE namespace=:namespace AND key=:key;", KV_TABLE_NAME);
+		let changes = locked_conn
+			.execute(
+				&sql,
+				named_params! {
+					":namespace": namespace,
+					":key": key,
+				},
+			)
+			.map_err(|e| {
+				let msg = format!("Failed to delete key {}/{}: {}", namespace, key, e);
+				std::io::Error::new(std::io::ErrorKind::Other, msg)
+			})?;
+
+		let was_present = changes != 0;
+
+		Ok(was_present)
+	}
+
+	fn list(&self, namespace: &str) -> std::io::Result<Vec<String>> {
+		let locked_conn = self.connection.lock().unwrap();
+
+		let sql = format!("SELECT key FROM {} WHERE namespace=:namespace", KV_TABLE_NAME);
+		let mut stmt = locked_conn.prepare(&sql).map_err(|e| {
+			let msg = format!("Failed to prepare statement: {}", e);
+			std::io::Error::new(std::io::ErrorKind::Other, msg)
+		})?;
+
+		let mut keys = Vec::new();
+
+		let rows_iter = stmt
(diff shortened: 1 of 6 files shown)


G8XSU:
Can we move the last commit of SQLite impl is separate PR, would like to review it separately. :)

src/io/sqlite_store.rs
```diff
@@ -0,0 +1,228 @@
...
+/// [SQLite]: https://sqlite.org
+pub struct SqliteStore {
+	connection: Arc<Mutex<Connection>>,
+}
+
+impl SqliteStore {
+	pub(crate) fn new(dest_dir: PathBuf) -> Self {
+		let msg =
+			format!("Failed to create database destination directory: {}", dest_dir.display());
+		fs::create_dir_all(dest_dir.clone()).expect(&msg);
```
jkczyz, line 34: Consider using a variant where you don't need to allocate the string for the successful case. Likewise throughout.

src/io/sqlite_store.rs
```diff
@@ -0,0 +1,228 @@
...
+
+/// A [`KVStore`] implementation that writes to and reads from an [SQLite] database.
+///
+/// [SQLite]: https://sqlite.org
+pub struct SqliteStore {
+	connection: Arc<Mutex<Connection>>,
+}
+
+impl SqliteStore {
+	pub(crate) fn new(dest_dir: PathBuf) -> Self {
```
jkczyz, line 31: Should this return an error so that any failure doesn't cause a mobile app to crash? I'm assuming that's what `expect` will do, though I'm not sure if there is any way to catch this in other languages.

src/io/sqlite_store.rs
```diff
@@ -0,0 +1,228 @@
...
+		))?;
+		let namespace = parent_directory.display().to_string();
+
+		let dest_without_namespace = dest_file_path
+			.strip_prefix(&namespace)
+			.map_err(|_| lightning::io::Error::new(lightning::io::ErrorKind::InvalidInput, msg))?;
+		let key = dest_without_namespace.display().to_string();
+
+		self.write(&namespace, &key, &object.encode())?;
+		Ok(())
```
jkczyz: Can make this one expression?

src/io/sqlite_store.rs
```diff
@@ -0,0 +1,228 @@
...
+			let sqlite_store = SqliteStore::new(rand_dir.into());
+			let namespace = "testspace";
+			let key = "testkey";
+
+			// Test the basic KVStore operations.
+			sqlite_store.write(namespace, key, &data).unwrap();
+
+			let listed_keys = sqlite_store.list(namespace).unwrap();
+			assert_eq!(listed_keys.len(), 1);
+			assert_eq!(listed_keys[0], "testkey");
```
jkczyz, line 219: Re-use `key`.

src/io/sqlite_store.rs
```diff
@@ -0,0 +1,228 @@
...
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use crate::test::utils::random_storage_path;
+	use lightning::util::persist::KVStorePersister;
+	use lightning::util::ser::Readable;
+
+	use proptest::prelude::*;
+	proptest! {
```
jkczyz, line 205: Perhaps I'm misunderstanding, but what is the value of using `proptest` here. The data written doesn't seem relevant to exercising the code paths.

src/lib.rs
```diff
@@ -320,11 +321,21 @@ impl Builder {
 		self
 	}
 
+	/// Builds a [`Node`] instance with a [`SqliteStore`] backend and according to the options
+	/// previously configured.
+	pub fn build(&self) -> Arc<Node<SqliteStore>> {
+		let storage_dir = self.config.storage_dir_path.clone();
+		fs::create_dir_all(storage_dir.clone()).expect("Failed to create LDK data directory");
+		let kv_store = Arc::new(SqliteStore::new(storage_dir.into()).into());
+		self.build_with_store(kv_store)
```
jkczyz, line 330: Does `build_with_store` need to be `pub`?

src/lib.rs
```diff
@@ -473,8 +479,12 @@ impl Builder {
...
 			Ok(monitors) => monitors,
 			Err(e) => {
-				log_error!(logger, "Failed to read channel monitors: {}", e.to_string());
-				panic!("Failed to read channel monitors: {}", e.to_string());
+				if e.kind() == std::io::ErrorKind::NotFound {
+					Vec::new()
+				} else {
+					log_error!(logger, "Failed to read channel monitors: {}", e.to_string());
+					panic!("Failed to read channel monitors: {}", e.to_string());
+				}
```
jkczyz, line 487: Looks like this should be a separate commit?

src/lib.rs
```diff
@@ -320,11 +321,21 @@ impl Builder {
 		self
 	}
 
+	/// Builds a [`Node`] instance with a [`SqliteStore`] backend and according to the options
+	/// previously configured.
+	pub fn build(&self) -> Arc<Node<SqliteStore>> {
+		let storage_dir = self.config.storage_dir_path.clone();
+		fs::create_dir_all(storage_dir.clone()).expect("Failed to create LDK data directory");
```
jkczyz, line 328: Could you DRY this up and make a separate function to use in both methods?

src/test/functional_tests.rs
```diff
@@ -321,6 +321,55 @@ fn start_stop_reinit() {
 	reinitialized_node.stop().unwrap();
 }
 
+#[test]
+fn start_stop_reinit_fs_store() {
```
jkczyz, line 325: Should we also do this for `SqliteStore`?

src/io/sqlite_store.rs
```diff
@@ -0,0 +1,238 @@
...
+		#[test]
+		fn read_write_remove_list_persist(data in any::<[u8; 32]>()) {
```
G8XSU, line 207: can we refactor this testcase so that all impl use same set of tests ?

src/lib.rs
```diff
@@ -320,15 +321,29 @@ impl Builder {
...
+		let kv_store = Arc::new(SqliteStore::new(storage_dir.into()).into());
+		self.build_with_store(kv_store)
+	}
 
-		let ldk_data_dir = format!("{}/ldk", config.storage_dir_path);
-		fs::create_dir_all(ldk_data_dir.clone()).expect("Failed to create LDK data directory");
+	/// Builds a [`Node`] instance with a [`FilesystemStore`] backend and according to the options
+	/// previously configured.
+	pub fn build_with_fs_store(&self) -> Arc<Node<FilesystemStore>> {
```
G8XSU, line 335: can't we do similar to set_esplora_server_url ? 
and use same builder pattern for this.

G8XSU: moved to another PR

src/io/sqlite_store.rs
```diff
@@ -0,0 +1,228 @@
...
+		connection
+			.pragma(Some(rusqlite::DatabaseName::Main), "user_version", SCHEMA_USER_VERSION, |_| {
+				Ok(())
+			})
+			.expect(&msg);
+
+		let sql = format!(
+			"CREATE TABLE IF NOT EXISTS {} (
+			namespace TEXT NOT NULL,
+			key TEXT NOT NULL,
```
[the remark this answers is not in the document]
G8XSU: i think we should add the non-empty constraint on namespace as well, it is unexpected to have empty namespace, specially when we allow for clients to use custom schema with pluggable backend kvStore. 

And instead use a static namespace in place of an empty one where needed.

src/io/sqlite_store.rs
```diff
@@ -0,0 +1,228 @@
...
+			keys.push(k.map_err(|_| std::io::Error::new(std::io::ErrorKind::Other, msg))?);
+		}
+
+		Ok(keys)
+	}
+}
+
+impl KVStorePersister for SqliteStore {
+	fn persist<W: Writeable>(&self, prefixed_key: &str, object: &W) -> lightning::io::Result<()> {
+		let msg = format!("Could not persist file for key {}.", prefixed_key);
```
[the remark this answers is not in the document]
G8XSU: "having the different backends decide to rename namespaces is a bad idea". yes exactly
that is why we should not be using anything to do with parent directories and key parsing from ldk-key as namespace keys. (for all kvstores including filesystem)

afaiu, ldk doesn't enforce anything by having keys like "monitors/[txid][index]", ldk is just providing a prefix.
 we could have something like "ldk_core" as namespace and "monitors/[txid][index]" as full key. 
 
 otherwise we are letting ldk-core create namespaces dynamically instead of ldk-node having fixed set of namespaces.(even though we own ldk-core, this is not ideal from ldk-node user perspective)
 
otherwise, in future, if ldk decides to store something new as "xyzMeta/" we would end up creating a namespace out of it and migration code will need to change.

src/io/sqlite_store.rs
```diff
@@ -0,0 +1,193 @@
...
+
+		connection
+			.pragma(Some(rusqlite::DatabaseName::Main), "user_version", SCHEMA_USER_VERSION, |_| {
+				Ok(())
+			})
+			.unwrap_or_else(|_| panic!("Failed to set PRAGMA user_version"));
+
+		let sql = format!(
+			"CREATE TABLE IF NOT EXISTS {} (
+			namespace TEXT NOT NULL,
```
G8XSU, line 49: i think we should add the non-empty constraint on namespace as well, it is unexpected to have empty namespace, specially when we allow for clients to use custom schema with pluggable backend kvStore.

And instead use a static namespace in place of an empty one where needed.

src/io/sqlite_store.rs
```diff
@@ -0,0 +1,193 @@
...
+			key TEXT NOT NULL CHECK (length(key) > 0),
```
G8XSU, line 50: not empty check should look like 
CHECK(key <> '') instead of checking length

src/io/sqlite_store.rs
```diff
@@ -0,0 +1,193 @@
...
+					":namespace": namespace,
+					":key": key,
+					":value": buf,
+				},
+			)
+			.map_err(|e| {
+				let msg = format!("Failed to write to key {}/{}: {}", namespace, key, e);
+				std::io::Error::new(std::io::ErrorKind::Other, msg)
+			})?;
+		Ok(())
```
jkczyz, line 116: Can also remove this explicit `Ok(())`.

src/io/sqlite_store.rs
G8XSU: its a nit but depending on database constraint optimization, length might need to compute length before comparison.

src/io/sqlite_store.rs
G8XSU: why do we want to have this inconsistency/workaround instead of opting for a simpler thing ?

src/io/sqlite_store.rs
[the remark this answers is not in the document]
G8XSU: >No, it's clearly not just a prefix but an implied namespace, actually even an implied path as all assumptions are made around FilesystemPersister

there is no enforced implied namespace, cash-lit doesn't care about it, ldk doesn't care about it. We are making that assumption right now in this PR, as we split the key into a file-path for a non-filesystem-directory datastore. We shouldn't be thinking of this having any relation to file-path or filesystem-directory/subdirectory.

src/io/sqlite_store.rs
```diff
@@ -0,0 +1,193 @@
...
+			})?);
+		}
+
+		Ok(keys)
+	}
+}
+
+impl KVStorePersister for SqliteStore {
+	fn persist<W: Writeable>(&self, prefixed_key: &str, object: &W) -> lightning::io::Result<()> {
+		let (namespace, key) = get_namespace_and_key_from_prefixed(prefixed_key)?;
```
G8XSU, line 173: We shouldn't split the key into a file-path for a non-filesystem-directory datastore. We shouldn't be thinking of this having any relation to file-path or filesystem-directory/subdirectory. This is a flawed assumption reminiscent of filesystem-datastore.

as being discussed in ([link](https://github.com/lightningdevkit/ldk-node/pull/100#discussion_r1207362176))

jkczyz:
> @G8XSU @jkczyz Can I squash the fixups here?

SGTM

G8XSU (approved):
Overall LGTM

jkczyz (approved):

imaginator:
What's the reasoning behind having KV storage instead of a full SQL schema?

PR #101, opened 2023-05-19
Make `KVStore` configurable

tnull:
Firstly, we simplify the `KVStore` interface: in the first commit we refactor `KVStore::write` to take a `buf: &[u8]` directly and serialize the data into corresponding buffers before calling `KVStore::write`, thereby reducing the complexity for any
implementation of the `KVStore` interface.

Furthermore, we previously had `Node` take a concrete `FilesystemStore` struct. In the second commit we switch to a generic `KVStore` type parameter. To this end we switched from `K: Deref` to concrete `Arc<K>` in all of the modules to avoid confusion of the type parameters or requirements to track `Arc`-ed and non-`Arc`ed version of the `K: KVStore` paramter. Moreover, as UniFFI doesn't support exposing generics we now expose a concretized `LDKNode`
type alias in bindings, which will use `SqliteStore` in the future.

### DIFF
--- a/src/io/fs_store.rs
+++ b/src/io/fs_store.rs
@@ -75,10 +75,13 @@ impl KVStore for FilesystemStore {
 		dest_file_path.push(namespace);
 		dest_file_path.push(key);
 
-		let msg = format!("Could not retrieve parent directory of {}.", dest_file_path.display());
 		let parent_directory = dest_file_path
 			.parent()
-			.ok_or(std::io::Error::new(std::io::ErrorKind::InvalidInput, msg))?
+			.ok_or_else(|| {
+				let msg =
+					format!("Could not retrieve parent directory of {}.", dest_file_path.display());
+				std::io::Error::new(std::io::ErrorKind::InvalidInput, msg)
+			})?
 			.to_path_buf();
 		fs::create_dir_all(parent_directory.clone())?;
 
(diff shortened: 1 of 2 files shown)


src/io/mod.rs
```diff
@@ -67,11 +68,3 @@ pub trait KVStore {
 	/// Will return an empty list if the `namespace` is unknown.
 	fn list(&self, namespace: &str) -> std::io::Result<Vec<String>>;
```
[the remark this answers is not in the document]
G8XSU: i think this should be done out of bounds, 
an application needs to know its namespaces for e.g "channel_store", "peer_store" etc. instead of list_namespaces.

G8XSU (commented):
Overall LGTM!

src/io/utils.rs
```diff
@@ -195,38 +195,20 @@ where
...
-		log_error!(
-			logger,
-			"Committing data to key {}/{} failed due to: {}",
-			LATEST_RGS_SYNC_TIMESTAMP_NAMESPACE,
-			LATEST_RGS_SYNC_TIMESTAMP_KEY,
-			e
-		);
-		Error::PersistenceFailed
-	})
+	Ok(())
```
jkczyz, line 204: Drop `Ok(())` and return the result of the previous expression without using `?`. Likewise elsewhere.

src/io/fs_store.rs
```diff
@@ -52,28 +52,84 @@ impl FilesystemStore {
...
+		let _guard = inner_lock_ref.write().unwrap();
+
+		let mut dest_file_path = self.dest_dir.clone();
+		dest_file_path.push(namespace);
+		dest_file_path.push(key);
+
+		let msg = format!("Could not retrieve parent directory of {}.", dest_file_path.display());
+		let parent_directory = dest_file_path
+			.parent()
+			.ok_or(std::io::Error::new(std::io::ErrorKind::InvalidInput, msg))?
```
jkczyz, line 81: Unrelated to this PR, but may want to use `ok_or_else` here and elsewhere to avoid allocating a string unnecessarily.

jkczyz (commented):
Feel free to squash

src/lib.rs
```diff
@@ -320,8 +320,18 @@ impl Builder {
...
+	/// previously configured.
+	pub fn build(&self) -> Arc<Node<FilesystemStore>> {
+		let ldk_data_dir = format!("{}/ldk", self.config.storage_dir_path);
+		let kv_store = Arc::new(FilesystemStore::new(ldk_data_dir.clone().into()));
+		self.build_with_store(kv_store)
+	}
+
 	/// Builds a [`Node`] instance according to the options previously configured.
-	pub fn build(&self) -> Arc<Node> {
+	pub fn build_with_store<K: KVStore + Sync + Send + 'static>(
```
G8XSU, line 332: can't we do similar to set_esplora_server_url ?
and use similar builder pattern for this.

G8XSU: there's one more related or unrelated item. 
some KVStore's like Vss might need a runtime, 

Current process of build_with_store assumes kvstore will be provided externally, but we will have no way to inject common runtime without something like set_runtime.
any thoughts or solutions to that?

PR #102, opened 2023-05-19
Update fee rate cache on startup and make sync interval configurable

tnull:
~~Based on #93.~~
~~Based on #108.~~

So far we'd immediately start background syncing upon `start()` which would work fine, except when these initial syncs failed, in which case we'd use our fallback fee. This might lead to being disconnected from peers as our fallback values are likely out of current fee ranges.

To mitigate this we here simply enforce that an initial update of our fee rate cache in `start()` is successful and hence we'd at least once got a recent fee estimation update.

Moreover, we make the background syncing intervals configurable, however enforcing a floor of 10 seconds.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -174,6 +174,7 @@ const DEFAULT_LISTENING_ADDR: &str = "0.0.0.0:9735";
 const DEFAULT_CLTV_EXPIRY_DELTA: u32 = 144;
 const DEFAULT_BDK_WALLET_SYNC_INTERVAL_SECS: u64 = 60;
 const DEFAULT_LDK_WALLET_SYNC_INTERVAL_SECS: u64 = 20;
+const DEFAULT_FEE_RATE_CACHE_UPDATE_INTERVAL_SECS: u64 = 60 * 10;
 const DEFAULT_LOG_LEVEL: LogLevel = LogLevel::Debug;
 
 // The 'stop gap' parameter used by BDK's wallet sync. This seems to configure the threshold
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -175,6 +175,7 @@ const DEFAULT_LISTENING_ADDR: &str = "0.0.0.0:9735";
 const DEFAULT_CLTV_EXPIRY_DELTA: u32 = 144;
 const DEFAULT_BDK_WALLET_SYNC_INTERVAL_SECS: u64 = 60;
 const DEFAULT_LDK_WALLET_SYNC_INTERVAL_SECS: u64 = 20;
+const DEFAULT_FEE_RATE_CACHE_UPDATE_INTERVAL_SECS: u64 = 60 * 60;
```
jkczyz, line 178: Should we consider a smaller default?

src/lib.rs
```diff
@@ -1373,39 +1375,31 @@ impl<K: KVStore + Sync + Send + 'static> Node<K> {
...
+							log_error!(sync_logger, "Sync of Lightning wallet failed: {}", e);
+							Err(e.into())
 						}
 					}
 				},
 			)
-		})?;
-
-		let sync_logger = Arc::clone(&self.logger);
-		tokio::task::block_in_place(move || {
```
jkczyz, line 1388: Could you expand upon the choices of runtime in the commit message and comments?

jkczyz (commented):
LGTM. Please squash.

jkczyz (approved):

PR #108, opened 2023-05-23
Improve logging

### LINKED ISSUE #109: Expose easy way to control log level
Adjusting the log level, e.g., via a Builder method will be very useful going forward for debugging purposes. 

We should probably default to `Debug` for now and allow enabling `Trace` and `Gossip` when needed.

tnull:
~~Based on #93~~, Fixes #109.

We improve our logging output by logging more detailed errors and introduce logging to `start`/`stop` methods.

Additionally, we expose the log level via `Config::log_level` and `Builder::set_log_level`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1123,7 +1123,9 @@ impl<K: KVStore + Sync + Send + 'static> Node<K> {
 	///
 	/// **Note:** This **MUST** be called after each event has been handled.
 	pub fn event_handled(&self) {
-		self.event_queue.event_handled().unwrap();
+		self.event_queue
+			.event_handled()
+			.expect("Couldn't mark event handled due to persistence failure");
 	}
 
 	/// Returns our own node id
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -17,7 +17,7 @@ pub(crate) struct FilesystemLogger {
 impl FilesystemLogger {
 	pub(crate) fn new(file_path: String, level: Level) -> Self {
 		if let Some(parent_dir) = Path::new(&file_path).parent() {
-			fs::create_dir_all(parent_dir).unwrap();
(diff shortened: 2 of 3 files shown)


src/logger.rs
```diff
@@ -1,27 +1,32 @@
...
 		}
-		Self { file_path }
+		Self { file_path, level }
 	}
 }
 impl Logger for FilesystemLogger {
 	fn log(&self, record: &Record) {
+		if record.level < self.level {
+			return;
+		}
```
jkczyz, line 29: Hmm... shouldn't this be enforced at compile time by the log macros?

jkczyz: Gotcha... for bindings is there a similar issue in that we construct the log message even if the we ultimately filter?

https://github.com/lightningdevkit/rust-lightning/blob/32eb89474cdc1e400ef89eab76d590d19da4746a/lightning/src/util/logger.rs#L101-L103

jkczyz: Ah, right. We aren't exposing the `Logger` interface here.

jkczyz (commented):
LGTM. Please squash.

src/lib.rs
```diff
@@ -207,6 +209,7 @@ const WALLET_KEYS_SEED_LEN: usize = 64;
 /// | `network`                   | Network::Bitcoin |
 /// | `listening_address`         | 0.0.0.0:9735     |
 /// | `default_cltv_expiry_delta` | 144              |
+/// | `log_level`                 | `Debug`          |
```
jkczyz, line 212: nit: inconsistency with `Network::Bitcoin` and here. The former doesn't use ticks and the latter doesn't use the enum name prefix.

jkczyz (approved):

PR #115, opened 2023-05-29
Move `Builder` to own module, allow for shared runtime reference.

tnull:
Firstly, as `lib.rs` got quite big and `Builder` made up a good chunk of it, we now move it to a dedicated submodule.

We then split the actual logic of `build_with_store` from the API method, which is needed as VSS in the
future will need to share the Runtime, i.e., will need to hand in a `Arc<Rwlock<Runtime>>>` parameter that we'll set `Some` in `Node::start`.

Finally, we do exactly that: we allow `build_with_store_internal` to take a runtime parameter.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -74,29 +74,29 @@ enum GossipSourceConfig {
 /// - Chain data is sourced from the Esplora endpoint `https://blockstream.info/api`
 /// - Gossip data is sourced via the peer-to-peer network
 #[derive(Debug)]
-pub struct Builder {
-	config: RwLock<Config>,
-	entropy_source_config: RwLock<Option<EntropySourceConfig>>,
-	chain_data_source_config: RwLock<Option<ChainDataSourceConfig>>,
-	gossip_source_config: RwLock<Option<GossipSourceConfig>>,
+pub struct NodeBuilder {
+	config: Config,
+	entropy_source_config: Option<EntropySourceConfig>,
+	chain_data_source_config: Option<ChainDataSourceConfig>,
+	gossip_source_config: Option<GossipSourceConfig>,
 }
 
-impl Builder {
+impl NodeBuilder {
 	/// Creates a new builder instance with the default configuration.
 	pub fn new() -> Self {
-		let config = RwLock::new(Config::default());
-		let entropy_source_config = RwLock::new(None);
-		let chain_data_source_config = RwLock::new(None);
-		let gossip_source_config = RwLock::new(None);
+		let config = Config::default();
+		let entropy_source_config = None;
+		let chain_data_source_config = None;
+		let gossip_source_config = None;
 		Self { config, entropy_source_config, chain_data_source_config, gossip_source_config }
 	}
 
 	/// Creates a new builder instance from an [`Config`].
 	pub fn from_config(config: Config) -> Self {
-		let config = RwLock::new(config);
-		let entropy_source_config = RwLock::new(None);
-		let chain_data_source_config = RwLock::new(None);
-		let gossip_source_config = RwLock::new(None);
+		let config = config;
+		let entropy_source_config = None;
+		let chain_data_source_config = None;
+		let gossip_source_config = None;
 		Self { config, entropy_source_config, chain_data_source_config, gossip_source_config }
 	}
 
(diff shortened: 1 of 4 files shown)


src/builder.rs
```diff
@@ -196,358 +196,368 @@ impl Builder {
...
-			Arc::clone(&kv_store),
-			Arc::clone(&network_graph),
-			Arc::clone(&logger),
-		) {
-			Ok(scorer) => Arc::new(Mutex::new(scorer)),
+/// Builds a [`Node`] instance according to the options previously configured.
+fn build_with_store_internal<'a, K: KVStore + Sync + Send + 'static>(
+	config: Arc<Config>, entropy_source_config: Option<&'a EntropySourceConfig>,
+	chain_data_source_config: Option<&'a ChainDataSourceConfig>,
+	gossip_source_config: Option<&'a GossipSourceConfig>, kv_store: Arc<K>,
```
jkczyz, line 314: `'a` lifetime can be inferred.

src/builder.rs
```diff
@@ -74,117 +74,110 @@ enum GossipSourceConfig {
...
 	}
 
 	/// Configures the [`Node`] instance to source its wallet entropy from a seed file on disk.
 	///
 	/// If the given file does not exist a new random seed file will be generated and
 	/// stored at the given location.
-	pub fn set_entropy_seed_path(&self, seed_path: String) {
-		*self.entropy_source_config.write().unwrap() =
-			Some(EntropySourceConfig::SeedFile(seed_path));
+	pub fn set_entropy_seed_path(&mut self, seed_path: String) {
```
jkczyz, line 107: Any reason we can't use move semantics now (i.e., take `self` and return `Self`)? That way you can chain mutators.

src/builder.rs
```diff
@@ -74,117 +74,110 @@ enum GossipSourceConfig {
...
-		config.log_level = level;
+	pub fn set_log_level(&mut self, level: LogLevel) {
+		self.config.log_level = level;
 	}
 
 	/// Builds a [`Node`] instance with a [`SqliteStore`] backend and according to the options
 	/// previously configured.
-	pub fn build(&self) -> Arc<Node<SqliteStore>> {
-		let storage_dir_path = self.config.read().unwrap().storage_dir_path.clone();
+	pub fn build(&self) -> Node<SqliteStore> {
```
jkczyz, line 170: Similarly, should the build methods consume `self`?

src/builder.rs
jkczyz: > I think move semantics still wouldn't work seamlessly as the `ArcedNodeBuilder` wrapper needs to own the inner `NoderBuilder` and we couldn't move/modify the `RwLock`'ed value in place? I guess we could work around that by `ArcedNodeBuilder` holding an `inner: RwLock<Option<NodeBuilder>>` and taking and re-setting the value upon each call. Not sure this is preferable to taking `&mut` and returning `&mut Self`?

Right, it would require using an `Option`, I believe.

> 
> Generally, the lack of mutator chaining is a good point: I simply had forgotten to add the `&mut Self` return values, now added.

One reason I'd prefer returning `Self` over `&mut Self` is the former makes for a cleaner way of chaining when you need to do intermediary calculations before using the builder again. The latter requires initializing a variable with the builder before doing any chaining. Otherwise, you'd return a reference to a temporary, which the compiler will complain about:

```
creates a temporary which is freed while still in use
```

jkczyz: > 
> Mh, I now experimented with it and have it working in an experimental branch, see: [tnull@3efa72e](https://github.com/tnull/ldk-node/commit/3efa72e1354bda556518873358f452ae29b3af9b)
> 
> Few thoughts:
> 
> 1. Move semantics indeed 'feel' a bit 'cleaner' too me as well.
> 2. While moving indeed allow to 'quick safe' the builder state, it also requires the user to do so, i.e., if they don't want to do it all in a single line, they'd have to do
> 
> ```rust
> let builder = Builder::new().set_X()
> ```
> 
> rather than just
> 
> ```rust
> let mut builder = Builder.new();
> builder.set_X();
> ```

Actually, this will cause a compilation error when the builder is later used since `set_X` consumed it but never reassigned it to `builder`. But maybe I'm misunderstanding what version each of the above is demonstrating...

> 
> This is a bit unfortunate also as the latter corresponds to the syntax they'd have to use in the bindings, i.e., our Rust docs would be a bit less transferable/relevant to the binding languages.
> 

That's a fair point.

> 3. Consuming restricts the use to a 1:1 relationship between `Builder` and `Node`. However, AFAIK there is no reason we shouldn't allow creation of multiple nodes with the same builder. Having the `Builder` be non-reusable also means that in bindings we would panic on a second `build()` command for now (https://github.com/tnull/ldk-node/blob/3efa72e1354bda556518873358f452ae29b3af9b/src/builder.rs#L314) and would need to introduce a dedicated `BuildError::AlreadyBuilt` type that's only returned for that case in [Prefer returning errors over panicking where possible #119](https://github.com/lightningdevkit/ldk-node/pull/119).
> 

Alternatively, `build` functions don't necessarily need to consume if we want reuse. But if it consumes, then we could implement `Clone` to support building additional nodes. If not, then note that we are already internally cloning config fields to pass to the node. The consuming variation is a bit more efficient as those fields could be moved instead and only cloned when the user explicitly clones the builder.

The bindings version could internally clone the config to avoid needing a `BuildError::AlreadyBuilt`. 

> Not entirely convinced one way or the other yet tbh. Moving/consuming 'feels' more paradigmatic, but `&mut self` would allow us to keep the APIs closer together.

I don't feel very strongly if you prefer to keep the APIs similar. The common case of building a node in one statement looks the same either way. So I suppose it's a matter of whether API similarity is more important than how uncommon cases are formulated.

jkczyz (commented):
LGTM. Please squash fix-ups.

jkczyz (approved):

PR #119, opened 2023-06-06
Prefer returning errors over panicking where possible

### LINKED ISSUE #50: Avoid panics where possible
While panicking at a point of no return might be considered reasonable error handling in Rust, it turns out that the messages given to `except` don't translate in any way to some of the binding languages. We therefore should consider to avoid any panics in favor of using `Result`s.

Maybe should go hand-in-hand with #31.

tnull:
~~Based on #115.~~
Fixes #50.

### DIFF
--- a/src/test/functional_tests.rs
+++ b/src/test/functional_tests.rs
@@ -324,7 +324,6 @@ fn connect_to_public_testnet_esplora() {
 	builder.set_esplora_server("https://blockstream.info/testnet/api".to_string());
 	let node = builder.build().unwrap();
 	node.start().unwrap();
-	node.sync_wallets().unwrap();
 	node.stop().unwrap();
 }
 


src/wallet.rs
```diff
@@ -450,20 +451,35 @@ where
...
+						panic!("Invalid shutdown script.");
+					},
+				);
+			}
+			_ => {
+				log_error!(
+					self.logger,
+					"Tried to use a non-witness address. This must not ever happen."
+				);
+				panic!("Tried to use a non-witness address. This must not ever happen.");
```
jkczyz, line 481: nit: s/not ever/never

jkczyz (commented):
LGTM. Please squash fixups.

jkczyz (approved):

PR #122, opened 2023-06-09
Allow to specify and update `ChannelConfig`

tnull:
This is useful/necessary to set forwarding fee rates, as well as allowing for overriding the dust HTLC exposure limit.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,7 +48,7 @@
 //!
 //! 	let node_id = PublicKey::from_str("NODE_ID").unwrap();
 //! 	let node_addr = NetAddress::from_str("IP_ADDR:PORT").unwrap();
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -1504,6 +1504,16 @@ impl<K: KVStore + Sync + Send + 'static> Node<K> {
 		}
 	}
 
+	/// Update the config for a previously opened channel.
+	pub fn update_channel_config(
+		&self, channel_id: &ChannelId, counterparty_node_id: PublicKey,
+		channel_config: ChannelConfig,
```
jkczyz, line 1512: Maybe take a reference since the method doesn't need ownership? Or is that an issue for bindings?

jkczyz (commented):
LGTM. Please squash fixups.

jkczyz (approved):

PR #125, opened 2023-06-15
Remove unnecessary `PaymentStatus::SendingFailed`

tnull:
Previously, I had a misunderstood in what event retrying payments is safe and erred on the conservative side. It turns out that it's indeed safe to retry a payment as soon as we receive `LdkEvent::PaymentFailed` and the misleading docs have since been removed (https://github.com/lightningdevkit/rust-lightning/pull/2340).

Therefore, we can here now only refuse to send duplicate payments if our previous status is `Succeeded` or `Pending`, which renders the additional distinction between `Failed` and `SendingFailed` superfluous. Accordingly, we remove this `PaymentStatus` variant.

### DIFF
--- a/src/payment_store.rs
+++ b/src/payment_store.rs
@@ -57,19 +57,16 @@ impl_writeable_tlv_based_enum!(PaymentDirection,
 pub enum PaymentStatus {
 	/// The payment is still pending.
 	Pending,
-	/// The sending of the payment failed and is safe to be retried.
-	SendingFailed,
 	/// The payment suceeded.
 	Succeeded,
-	/// The payment failed and is not retryable.
+	/// The payment failed.
 	Failed,
 }
 
 impl_writeable_tlv_based_enum!(PaymentStatus,
 	(0, Pending) => {},
-	(2, SendingFailed) => {},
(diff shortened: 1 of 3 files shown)


src/payment_store.rs
```diff
@@ -57,19 +57,16 @@ impl_writeable_tlv_based_enum!(PaymentDirection,
...
 	Failed,
 }
 
 impl_writeable_tlv_based_enum!(PaymentStatus,
 	(0, Pending) => {},
-	(2, SendingFailed) => {},
-	(4, Succeeded) => {},
-	(6, Failed) => {};
+	(2, Succeeded) => {},
+	(4, Failed) => {};
```
jkczyz, line 69: Can't recall if `SendingFailed` made it into the alpha release. But if so, are we ok breaking serialization here? This could blindly result in a `PaymentStatus` that was previously one value become a different one, IIUC.

Side note: should we add release tags to github?

jkczyz: Gotcha. We may want to suggest wiping data when upgrading then. Otherwise, successful payments will show up as failed.

jkczyz (approved):

PR #137, opened 2023-07-05
bugfix: Return early if no non-static outputs

MaxFangX:
If there is a cooperative closure, there will only be one output to spend, and it will be a `StaticOutput`. Since `WalletKeysManager`'s `spend_spendable_outputs` filters out static outputs, `KeysManager`'s `spend_spendable_outputs` gets passed an empty list of descriptors to spend, which results in an error (with no further information provided).


MaxFangX:
> I wonder if a similar change should land upstream

It would be better if `KeysManager::spend_spendable_outputs`' `Err` contained more information than just `()`. I ran into this problem when integration testing cooperative closes, and guessed (correctly) it was because of all of the static outputs being filtered out. With explicit errors the problem would have been more obvious.

PR #138, opened 2023-07-05
VSS Store Prefactors

tnull:
The VSS store will need access to the seed bytes to derive de-/encryption keys. As we'll need to set up the store before entering `build_with_store_internal`, we now move reading the `EntropySourceConfig` outside as a prefactor.

Additionally, we drop the return value from `KVStore::remove`.


jkczyz (commented):
LGTM. Please squash.

PR #141, opened 2023-07-24
Add anchor support

tnull:
Based on ~~#243~~, ~~#250, #253.~~

### DIFF
--- a/src/wallet.rs
+++ b/src/wallet.rs
@@ -166,6 +166,12 @@ where
 		Ok(address_info.address)
 	}
 
+	fn get_new_internal_address(&self) -> Result<bitcoin::Address, Error> {
+		let address_info =
+			self.inner.lock().unwrap().get_internal_address(AddressIndex::LastUnused)?;
+		Ok(address_info.address)
+	}
+
 	pub(crate) fn get_balances(
 		&self, total_anchor_channels_reserve_sats: u64,
 	) -> Result<(u64, u64), Error> {
@@ -349,9 +355,10 @@ where
 
 	fn get_change_script(&self) -> Result<ScriptBuf, ()> {
 		let locked_wallet = self.inner.lock().unwrap();
-		let address_info = locked_wallet.get_address(AddressIndex::LastUnused).map_err(|e| {
-			log_error!(self.logger, "Failed to retrieve new address from wallet: {}", e);
-		})?;
+		let address_info =
+			locked_wallet.get_internal_address(AddressIndex::LastUnused).map_err(|e| {
+				log_error!(self.logger, "Failed to retrieve new address from wallet: {}", e);
+			})?;
 
 		Ok(address_info.address.script_pubkey())
 	}
@@ -569,7 +576,7 @@ where
 	L::Target: Logger,
 {
 	fn get_change_destination_script(&self) -> Result<ScriptBuf, ()> {
-		let address = self.wallet.get_new_address().map_err(|e| {
+		let address = self.wallet.get_new_internal_address().map_err(|e| {
 			log_error!(self.logger, "Failed to retrieve new address from wallet: {}", e);
 		})?;
 		Ok(address.script_pubkey())


src/lib.rs
```diff
@@ -1176,27 +1176,66 @@ impl Node {
...
+	/// channel funds time-locked, i.e., they will only be available after the counterparty had
+	/// time to contest our claim. Force-closing channels also more costly in terms of on-chain
+	/// fees. So cooperative closure should always be preferred (and tried first).
+	///
+	/// Broadcasting the closing transactions will be omitted for Anchor channels if we trust the
+	/// counterparty to broadcast for us (see [`AnchorChannelsConfig::trusted_peers_no_reserve`]
+	/// for more information).
 	pub fn close_channel(
-		&self, user_channel_id: &UserChannelId, counterparty_node_id: PublicKey,
+		&self, user_channel_id: &UserChannelId, counterparty_node_id: PublicKey, force: bool,
```
jkczyz, line 1190: Should we add a separate method for this? `bool` params are meh

src/config.rs
```diff
@@ -120,6 +139,62 @@ impl Default for Config {
...
+	/// insecure as the channel may not be closed if they refuse to do so, potentially leaving the
+	/// user funds stuck.
+	pub trusted_peers_no_reserve: Vec<PublicKey>,
+	/// The amount of satoshis we keep as an emergency reserve in our on-chain wallet in order to
+	/// be able to spend the Anchor output on channel close.
+	///
+	/// **Note:** Depending on the fee market at the time of closure, this reserve amount might or
+	/// might not suffice to successfully spend the Anchor output, i.e., you may want to adjust
+	/// this value accordingly.
+	pub per_channel_reserve_sats: u64,
```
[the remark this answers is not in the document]
jkczyz: Yeah, I can see why some technical details are necessary to expose, but at least the _implementation_ details (i.e., anchors) can be hidden it seems. Other than in the config, it is only exposed in `ChannelType::Anchors`. And the alternative, `ChannelType::StaticRemoteKey` is not providing any useful information because anchor channels imply that.

Seems instead of `ChannelType` we want something like `ReserveType`. And that can be more fine-grained than two variants as the trusted peer case could be an additional variant. That would provide more useful information as otherwise the user needs to cross reference `ChannelDetails` with the config to see if the counterparty is in `trusted_peers_no_reserve`.

src/config.rs
```diff
@@ -120,6 +139,68 @@ impl Default for Config {
...
+	/// Channels with these peers won't count towards the retained on-chain reserve and we won't
+	/// take any action to get the required transactions confirmed ourselves.
+	///
+	/// **Note:** Trusting the channel counterparty to take the necessary actions to get the
+	/// required Anchor spending and HTLC transactions confirmed on-chain is potentially insecure
+	/// as the channel may not be closed if they refuse to do so, potentially leaving the user
+	/// funds stuck *or* even allow the counterparty to steal any in-flight funds after the
+	/// corresponding HTLCs time out.
+	pub trusted_peers_no_reserve: Vec<PublicKey>,
+	/// The amount of satoshis we keep as an emergency reserve in our on-chain wallet in order to
```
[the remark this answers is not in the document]
jkczyz: Hmm... "per channel" is a bit misleading then. It's more of a multiplier? Not sure if an alternative name is any better, but I think at least the docs need to be clear on: (a) how the reserve can be used, (b) how it is checked, and (c) that it can be exceeded (which, IIUC could prevent further channels from being opened).

src/config.rs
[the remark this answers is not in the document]
jkczyz: Was thinking more along the lines `Predetermined`, `Dynamic`, `NoReserve`. Probably could iterate on the exact names, but basically something that indicates how the reserve is determined.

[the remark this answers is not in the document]
jkczyz: Maybe not necessarily `Predetermined` but something indicating that it is part of the channel balance. Whereas the other is from on-chain funds.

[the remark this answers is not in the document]
jkczyz: > Tbh. I'm not entirely convinced we should go this way. The main reason why I introduced `ChannelType` is for users to check whether any opened channel is or isn't using Anchors, as this has several other consequences other than reserve. E.g., LND requires to use Anchor channels for zero-conf. So I'm afraid entirely dropping `Anchors` from the type obfuscates things even further.

Not sure I understand why the user needs to check this and what actions they would take. Could this be documented?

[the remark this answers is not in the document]
jkczyz: Yeah, either just drop it or use a bool. No need to define a type, IMO. My earlier thought was really on how to tie the reserve configuration to each channel, which might be valuable? Don't really have a strong opinion though.

src/lib.rs
```diff
@@ -1176,30 +1176,84 @@ impl Node {
...
+	/// Will force-close the channel, potentially broadcasting our latest state. Note that in
+	/// contrast to cooperative closure, force-closing will have the channel funds time-locked,
+	/// i.e., they will only be available after the counterparty had time to contest our claim.
+	/// Force-closing channels also more costly in terms of on-chain fees. So cooperative closure
+	/// should always be preferred (and tried first).
+	///
+	/// Broadcasting the closing transactions will be omitted for Anchor channels if we trust the
+	/// counterparty to broadcast for us (see [`AnchorChannelsConfig::trusted_peers_no_reserve`]
+	/// for more information).
+	pub fn force_close_channel(
```
jkczyz, line 1199: Update the commit message in terms of new API method rather than `force` flag detail.

src/wallet.rs
```diff
@@ -343,9 +349,10 @@ where
 
 	fn get_change_script(&self) -> Result<ScriptBuf, ()> {
 		let locked_wallet = self.inner.lock().unwrap();
-		let address_info = locked_wallet.get_address(AddressIndex::LastUnused).map_err(|e| {
-			log_error!(self.logger, "Failed to retrieve new address from wallet: {}", e);
-		})?;
+		let address_info =
+			locked_wallet.get_internal_address(AddressIndex::LastUnused).map_err(|e| {
+				log_error!(self.logger, "Failed to retrieve new address from wallet: {}", e);
+			})?;
```
jkczyz, line 355: Should we use `get_new_internal_address` instead?

src/lib.rs
```diff
@@ -1013,18 +1021,26 @@ impl Node {
...
 		let peer_info = PeerInfo { node_id, address };
 
 		let con_node_id = peer_info.node_id;
 		let con_addr = peer_info.address.clone();
 		let con_cm = Arc::clone(&self.connection_manager);
 
+		let cur_anchor_reserve_sats =
+			total_anchor_channels_reserve_sats(&self.channel_manager, &self.config);
+		let spendable_amount_sats =
+			self.wallet.get_balances(cur_anchor_reserve_sats).map(|(_, s)| s).unwrap_or(0);
```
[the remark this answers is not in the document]
jkczyz: Yeah, ideally `Wallet` would just be aware of the anchors reserve (instead of passing it) and return an adjusted `Balance`, though that involves introducing dependencies to the `Wallet`.  Even if it were aware, the reserve concept can't be applied to one of the `Balance` components without affecting the total, unfortunately.

What about defining our own internal `Balance` as:

```rust
struct Balance {
	pub anchor_reserve_sats: u64,
	pub spendable_amount_sats: u64,
	pub total_onchain_balance_sats: u64,
}
```

Where `Wallet::get_balances` takes `ChannelManager` and `Config` to compute the anchor reserve internally instead of computing it externally at each call site and passing it in? That would keep the calculations in one place, including the one that adjusts the reserve based on the total balance.

src/config.rs
```diff
@@ -120,6 +141,77 @@ impl Default for Config {
...
+/// opening. This required to estimate what fee rate would be sufficient to still have the
+/// closing transactions be spendable on-chain (i.e., not be considered dust). This legacy
+/// design of pre-anchor channels proved inadequate in the unpredictable, often turbulent, fee
+/// markets we experience today.
+///
+/// In contrast, Anchor channels allow to determine an adequate fee
+/// rate *at the time of channel closure*, making them much more robust in the face of fee spikes.
+/// In turn, they require to maintain a reserve of on-chain funds to have the channel closure
+/// transactions confirmed on-chain, at least if the channel counterparty can't be trusted to do
+/// this for us.
```
jkczyz, line 163: nit: re-wrap

src/config.rs
```diff
@@ -120,6 +141,77 @@ impl Default for Config {
...
+	/// Channels with these peers won't count towards the retained on-chain reserve and we won't
+	/// take any action to get the required transactions confirmed ourselves.
+	///
+	/// **Note:** Trusting the channel counterparty to take the necessary actions to get the
+	/// required Anchor spending and HTLC transactions confirmed on-chain is potentially insecure
+	/// as the channel may not be closed if they refuse to do so, potentially leaving the user
+	/// funds stuck *or* even allow the counterparty to steal any in-flight funds after the
+	/// corresponding HTLCs time out.
+	pub trusted_peers_no_reserve: Vec<PublicKey>,
+	/// The amount of satoshis per anchors-negotiated channel with an untrusted peer we keep as an
```
jkczyz, line 191: Reads a bit better as "that we keep ..."

src/config.rs
```diff
@@ -120,6 +141,77 @@ impl Default for Config {
...
+	/// emergency reserve in our on-chain wallet in order to have the required Anchor output
+	/// spending and HTLC transactions confirmed when the channel is closed.
```
jkczyz, line 193: Sentence is a bit lengthy now. I'd recommend splitting it. Would recommend something like `s/wallet in order to have/wallet. Allows for having`.

tests/common/mod.rs
```diff
@@ -355,11 +373,12 @@ pub fn open_channel(
 
 pub(crate) fn do_channel_full_cycle<E: ElectrumApi>(
 	node_a: TestNode, node_b: TestNode, bitcoind: &BitcoindClient, electrsd: &E, allow_0conf: bool,
+	expect_anchor_channel: bool, force_close: bool,
```
jkczyz, line 376: May want to consider enums for these at some point.

jkczyz (commented):
Other than the earlier comment on balance computation, just some minor comments and I think we're good.

jkczyz:
> > Other than the earlier comment on balance computation, just some minor comments and I think we're good.
> 
> Addressed the remaining doc comments. See my response to the balance/reserve above. Let me know if I can squash the fixups.

Yup, go ahead and squash.

src/event.rs
```diff
@@ -815,9 +820,67 @@ where
...
+						} else {
+							anchor_channels_config.per_channel_reserve_sats
+						};
+
+						if spendable_amount_sats < required_amount_sats {
+							log_error!(
+							self.logger,
+							"Rejecting inbound Anchor channel from peer {} due to insufficient available on-chain reserves.",
+							counterparty_node_id,
+						);
```
jkczyz, line 855: Indentation off.

src/event.rs
```diff
@@ -1062,6 +1126,35 @@ where
...
+
+				if let Some(anchor_channels_config) = self.config.anchor_channels_config.as_ref() {
+					if anchor_channels_config
+						.trusted_peers_no_reserve
+						.contains(counterparty_node_id)
+					{
+						log_debug!(self.logger,
+								"Ignoring BumpTransactionEvent for channel {} due to trusted counterparty {}",
+								channel_id, counterparty_node_id
+							);
```
jkczyz, line 1151: Weird indentation?

jkczyz (approved):

jkczyz (approved):

PR #143, opened 2023-07-26
Balance values are Long rather than Int

galderz:
I run the LDK node Kotlin test standalone and it was failing with:

```
org.opentest4j.AssertionFailedError: expected: kotlin.UInt@602e0143<100000> but was: kotlin.ULong@2c07545f<100000>
Expected :100000
Actual   :100000
<Click to see difference>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1142)
	at org.mendrugo.korko.LdkNodeTest.testFullLifecycle(LdkNodeTest.kt:85)
```

According to the decompiled class:

```
public open fun spendableOnchainBalanceSats(): kotlin.ULong { /* compiled code */ }
```

### DIFF
--- a/bindings/kotlin/ldk-node-jvm/lib/src/test/kotlin/org/lightningdevkit/ldknode/LibraryTest.kt
+++ b/bindings/kotlin/ldk-node-jvm/lib/src/test/kotlin/org/lightningdevkit/ldknode/LibraryTest.kt
@@ -155,10 +155,10 @@ class LibraryTest {
         println("Spendable balance 2: $spendableBalance1")
         println("Total balance 1: $totalBalance1")
         println("Total balance 2: $totalBalance1")
-        assertEquals(100000u, spendableBalance1)
-        assertEquals(100000u, spendableBalance2)
-        assertEquals(100000u, totalBalance1)
-        assertEquals(100000u, totalBalance2)
+        assertEquals(100000uL, spendableBalance1)
+        assertEquals(100000uL, spendableBalance2)
+        assertEquals(100000uL, totalBalance1)
+        assertEquals(100000uL, totalBalance2)
 
         node1.connectOpenChannel(nodeId2, listenAddress2, 50000u, null, null, true)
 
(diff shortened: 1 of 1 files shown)


galderz:
Tried this with Java 17 and it works:

```
./scripts/uniffi_bindgen_generate_kotlin.sh
cd bindings/kotlin/ldk-node-jvm/
./gradlew test
```

Though I see things like this:

```
'compileJava' task (current target is 17) and 'compileKotlin' task (current target is 1.8) jvm target compatibility should be set to the same Java version.
By default will become an error since Gradle 8.0+! Read more: https://kotl.in/gradle/jvm/target-validation
```

I thought maybe the difference is in the Kotlin version since you're using `1.7.10` and my project is using `1.8.22`, but I tried your project with `1.8.22` and seems to work fine again.

So, I don't really know why the difference right now. I'm neither a Kotlin nor Gradle expert. Feel free to close the PR.

galderz:
I do think it would be good practice to have types aligned when comparing them. IOW, if the API returns a Long, compare it with a Long, and if it returns an Int, compare with an Int.

galderz:
> Still a bit dubious why this would be required.

Hmmm, I think the `assertEquals` being invoked here is the one that takes `Object` rather than Java primitives, so you can replicate the same issue with plain Java by doing:

```
assertEquals(Integer.valueOf(100000), Long.valueOf(100000L));
```

Which fails with:

```
org.opentest4j.AssertionFailedError: expected: java.lang.Integer@29774679<100000> but was: java.lang.Long@3ffc5af1<100000>
Expected :100000
Actual   :100000
```

I don't understand how this can work when `./gradlew test` is executed though in the `ldk-node-jvm/lib` module.

PR #156, opened 2023-08-16
Skip channels used for probing based on available liquidity

### LINKED ISSUE #155: Add heuristic to LDK Node code
Possibly using LDK's implementation if we decide to put it in LDK code. See here: https://github.com/lightningdevkit/rust-lightning/issues/2499

tnull:
Fixes #155.

As pre-flight probes might take up some of the available liquidity, we here introduce that channels whose available liquidity is less than the required amount times `Config::probing_liquidity_limit_multiplier` won't be used to send pre-flight probes.

(cc @TonyGiorgio)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -149,6 +149,7 @@ use bitcoin::{Address, Txid};
 
 use rand::Rng;
 
+use std::collections::HashMap;
 use std::default::Default;
 use std::net::ToSocketAddrs;
 use std::sync::{Arc, Mutex, RwLock};
@@ -164,6 +165,7 @@ const DEFAULT_CLTV_EXPIRY_DELTA: u32 = 144;
 const DEFAULT_BDK_WALLET_SYNC_INTERVAL_SECS: u64 = 80;
 const DEFAULT_LDK_WALLET_SYNC_INTERVAL_SECS: u64 = 30;
 const DEFAULT_FEE_RATE_CACHE_UPDATE_INTERVAL_SECS: u64 = 60 * 10;
+const DEFAULT_PROBING_LIQUIDITY_LIMIT_MULTIPLIER: u64 = 3;
 const DEFAULT_LOG_LEVEL: LogLevel = LogLevel::Debug;
 
 // The 'stop gap' parameter used by BDK's wallet sync. This seems to configure the threshold
@@ -210,6 +212,7 @@ const WALLET_KEYS_SEED_LEN: usize = 64;
 /// | `wallet_sync_interval_secs`            | 30                 |
 /// | `fee_rate_cache_update_interval_secs`  | 600                |
 /// | `trusted_peers_0conf`                  | []                 |
+/// | `probing_liquidity_limit_multiplier`   | 3                  |
 /// | `log_level`                            | Debug              |
 ///
 pub struct Config {
@@ -243,6 +246,11 @@ pub struct Config {
 	/// funding transaction ends up never being confirmed on-chain. Zero-confirmation channels
 	/// should therefore only be accepted from trusted peers.
 	pub trusted_peers_0conf: Vec<PublicKey>,
+	/// The liquidity factor by which we filter the outgoing channels used for sending probes.
+	///
+	/// Channels with available liquidity less than the required amount times this value won't be
+	/// used to send pre-flight probes.
+	pub probing_liquidity_limit_multiplier: u64,
 	/// The level at which we log messages.
 	///
 	/// Any messages below this level will be excluded from the logs.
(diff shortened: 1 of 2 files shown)


moneyball:
ACK

moneyball (dismissed):

src/lib.rs
```diff
@@ -1390,6 +1399,21 @@ impl<K: KVStore + Sync + Send + 'static> Node<K> {
...
 			}
 
+			if let Some(first_route_hop) = path.hops.first() {
+				if let Some(first_hop) = first_hops
+					.iter()
+					.find(|h| h.short_channel_id == Some(first_route_hop.short_channel_id))
+				{
+					let path_value = path.final_value_msat() + path.fee_msat();
+					if first_hop.next_outbound_htlc_limit_msat
+						< path_value * self.config.probing_liquidity_limit_multiplier
```
jkczyz, line 1409: If a user has one channel and is trying to make a payment that spends greater than half of its value, won't this still be a problem if the payment is split equally along three paths?

Say that `next_outbound_htlc_limit_msat` is initially `100` and the payment is split equally along paths each with a `path_value` of `19` for a total payment value of `57`. All three paths would be probed as their value is under 1/3 `next_outbound_htlc_limit_msat` even when adjusting for the previously probed path, but we are taking over half the channel liquidity. Worse case scenario is all three probes don't come back and we can't make the payment because our only channel's liquidity isn't enough for the full payment.

Should we instead keep a running sum of `path_value` for paths using the same first hop, and probe until we exceeded the limit with a fixed `next_outbound_htlc_limit_msat`?

jkczyz: It would depend on the scorer used, it's state, which channels are probed, and the values for each path. Yes, for `ProbabilisticScorer`, as it's trying to learn channel liquidity balances and how often they fall within a given bucket.

As for the follow-up payment, if a probe fails for one of the paths, then we'll know to avoid a channel or trying splitting differently.

src/lib.rs
```diff
@@ -1390,6 +1391,17 @@ impl<K: KVStore + Sync + Send + 'static> Node<K> {
 				Error::ProbeSendingFailed
 			})?;
 
+		let mut liquidity_map: HashMap<u64, u64> = first_hops
+			.iter()
+			.filter_map(|h| {
+				if let Some(scid) = h.short_channel_id {
```
jkczyz, line 1400: Should we use `get_outbound_payment_scid()`?

src/lib.rs
```diff
@@ -1400,17 +1412,17 @@ impl<K: KVStore + Sync + Send + 'static> Node<K> {
...
-					.iter()
-					.find(|h| h.short_channel_id == Some(first_route_hop.short_channel_id))
+				let path_value = path.final_value_msat() + path.fee_msat();
+				if let Some(remaining_liquidity) =
+					liquidity_map.get_mut(&first_route_hop.short_channel_id)
 				{
-					let path_value = path.final_value_msat() + path.fee_msat();
-					if first_hop.next_outbound_htlc_limit_msat
+					if *remaining_liquidity
 						< path_value * self.config.probing_liquidity_limit_multiplier
```
jkczyz: You'll have the same problem as before because the way the division is moved to the other side of the inequality as a multiplication. From the previous example:

`remaining_liquidity`: 100, 81, 62
`path_value * self.config.probing_liquidity_limit_multiplier`: 19 * 3 = 57

Time to write a test? 😅

src/lib.rs
```diff
@@ -1402,17 +1404,22 @@ impl<K: KVStore + Sync + Send + 'static> Node<K> {
...
+					h.get_outbound_payment_scid() == Some(first_path_hop.short_channel_id)
+				}) {
 					let path_value = path.final_value_msat() + path.fee_msat();
+					let used_liquidity =
+						used_liquidity_map.entry(first_path_hop.short_channel_id).or_insert(0);
+
 					if first_hop.next_outbound_htlc_limit_msat
-						< path_value * self.config.probing_liquidity_limit_multiplier
+						< (*used_liquidity + path_value)
+							* self.config.probing_liquidity_limit_multiplier
```
jkczyz, line 1417: Isn't `next_outbound_htlc_limit_msat` reduced on each probe? So you would potentially stop sending probes even though `1 / probing_liquidity_limit_multiplier` has not been exhausted for the given *payment*? (i.e., a smaller limit is being compared against each path rather than being fixed).

jkczyz: Maybe we don't care, but just wanted to mention that in case it was not intentional.

jkczyz: Ah, right. I forgot that those are copies of `Channel` data.

src/lib.rs
jkczyz: I'm fine with testing upstream. Indifferent on whether you want to land this first.

jkczyz (commented):
Feel free to squash.

jkczyz (approved):

PR #158, opened 2023-08-18
Add VSS Support in ldk-node

G8XSU:
Note: 
This PR does not add full support of VSS, there will be additional PR's for some of the changes. 
Currently it introduces only VssKVStore which is major chunk of code.

We will call it full support only once we make changes to public api for ldk-node.

Note on testing:
* Running a Java WAR file within a Rust test is not a straightforward process, as the execution of a Java WAR typically requires a Java runtime, which can not be natively managed by Rust.
* Currently it is tested against a local instance of VSS. But we will be testing this as part of our CI pipeline as discussed in sync, currently working on it.
* Alternative to env-var was feature-gate, but I didn't see much reason to introduce new feature for this test and it could be confusing. Setting env-var to any value will result in test-failure if vss doesn't work properly or it is not a correct endpoint. So we can easily verify this.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -37,6 +37,8 @@ use lightning_persister::fs_store::FilesystemStore;
 
 use lightning_transaction_sync::EsploraSyncClient;
 
+#[cfg(any(vss, vss_test))]
+use crate::io::vss_store::VssStore;
 use bdk::bitcoin::secp256k1::Secp256k1;
 use bdk::blockchain::esplora::EsploraBlockchain;
 use bdk::database::SqliteDatabase;
@@ -267,6 +269,16 @@ impl NodeBuilder {
 			.map_err(|_| BuildError::StoragePathAccessFailed)?;
 		let kv_store = Arc::new(FilesystemStore::new(storage_dir_path));
 		self.build_with_store(kv_store)
+	}
+
+	/// Builds a [`Node`] instance with a [`VssStore`] backend and according to the options
+	/// previously configured.
+	#[cfg(any(vss, vss_test))]
+	pub fn build_with_vss_store(
+		&self, url: &str, store_id: String,
+	) -> Result<Node<VssStore>, BuildError> {
+		let vss = Arc::new(VssStore::new(url, store_id));
+		self.build_with_store(vss)
 	}
 
 	/// Builds a [`Node`] instance according to the options previously configured.
--- a/src/io/vss_store.rs
+++ b/src/io/vss_store.rs
@@ -1,0 +1,194 @@
+use io::Error;
+use std::io;
+use std::io::ErrorKind;
+#[cfg(test)]
+use std::panic::RefUnwindSafe;
+
+use crate::io::utils::check_namespace_key_validity;
+use lightning::util::persist::KVStore;
+use tokio::runtime::Runtime;
+use vss_client::client::VssClient;
+use vss_client::error::VssError;
+use vss_client::types::{
+	DeleteObjectRequest, GetObjectRequest, KeyValue, ListKeyVersionsRequest, PutObjectRequest,
+};
+
+/// A [`KVStore`] implementation that writes to and reads from a [VSS](https://github.com/lightningdevkit/vss-server/blob/main/README.md) backend.
+pub struct VssStore {
+	client: VssClient,
+	store_id: String,
+	runtime: Runtime,
+}
+
+impl VssStore {
+	pub(crate) fn new(base_url: &str, store_id: String) -> Self {
+		let client = VssClient::new(base_url);
+		let runtime = tokio::runtime::Builder::new_multi_thread().enable_all().build().unwrap();
+		Self { client, store_id, runtime }
+	}
+
+	fn build_key(
+		&self, primary_namespace: &str, secondary_namespace: &str, key: &str,
+	) -> io::Result<String> {
+		if primary_namespace.is_empty() {
+			Ok(key.to_string())
+		} else {
+			Ok(format!("{}#{}#{}", primary_namespace, secondary_namespace, key))
+		}
+	}
+
+	fn extract_key(&self, unified_key: &str) -> io::Result<String> {
+		let mut parts = unified_key.splitn(3, '#');
+		let (_primary_namespace, _secondary_namespace) = (parts.next(), parts.next());
+		match parts.next() {
+			Some(actual_key) => Ok(actual_key.to_string()),
+			None => Err(Error::new(ErrorKind::InvalidData, "Invalid key format")),
+		}
+	}
+
+	async fn list_all_keys(
+		&self, primary_namespace: &str, secondary_namespace: &str,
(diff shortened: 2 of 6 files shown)


src/io/vss_store.rs
```diff
@@ -0,0 +1,192 @@
...
+			store_id: self.store_id.to_string(),
+			key: self.build_key(namespace, sub_namespace, key)?,
+		};
+
+		let resp =
+			tokio::task::block_in_place(|| self.runtime.block_on(self.client.get_object(&request)))
+				.map_err(|e| match e {
+					VssError::NoSuchKeyError(..) => {
+						let msg = format!(
+							"Failed to read as key could not be found: {}/{}/{}. Details: {}",
```
jkczyz, line 91: Let's word this as `Failed to read from key {}/{}/{}: {}` for consistency with the other errors and since "could not be found" is redundant with both `VssError::NoSuchKeyError` used in the message as `e` and `ErrorKind::NotFound` when wrapped below.

src/io/vss_store.rs
```diff
@@ -0,0 +1,192 @@
...
+	#[cfg(feature = "vss")]
+	pub(crate) fn new(base_url: &str, store_id: String) -> Self {
+		let client = VssClient::new(base_url);
+		let runtime = tokio::runtime::Builder::new_multi_thread().enable_all().build().unwrap();
+		Self { client, store_id, runtime }
+	}
+
+	fn build_key(&self, namespace: &str, sub_namespace: &str, key: &str) -> io::Result<String> {
+		if key.is_empty() {
+			return Err(Error::new(ErrorKind::Other, "Empty key is not allowed"));
```
jkczyz, line 33: Should we use `InvalidInput`?

src/io/vss_store.rs
```diff
@@ -0,0 +1,192 @@
...
+impl VssStore {
+	#[cfg(feature = "vss")]
+	pub(crate) fn new(base_url: &str, store_id: String) -> Self {
+		let client = VssClient::new(base_url);
+		let runtime = tokio::runtime::Builder::new_multi_thread().enable_all().build().unwrap();
+		Self { client, store_id, runtime }
+	}
+
+	fn build_key(&self, namespace: &str, sub_namespace: &str, key: &str) -> io::Result<String> {
+		if key.is_empty() {
```
jkczyz, line 32: Seems like this check is covered by `check_namespace_key_validity`, FWIW.

src/io/vss_store.rs
jkczyz: Hmm... we don't make such distinctions for other VSS error types. And it _would_ be given first since `ErrorKind::NotFound` comes before the message, no? In practice, I think you'll find that this error message will be extremely verbose and redundant even without the explicit "could not be found" here, IMHO. With our server implementation, the key will even be given again.

src/io/vss_store.rs
```diff
@@ -0,0 +1,201 @@
...
+		if primary_namespace.is_empty() {
+			Ok(key.to_string())
+		} else {
+			Ok(format!("{}#{}#{}", primary_namespace, secondary_namespace, key))
+		}
+	}
+
+	fn extract_key(&self, unified_key: &str) -> io::Result<String> {
+		let mut parts = unified_key.splitn(3, '#');
+		let (_, _) = (parts.next(), parts.next()); // Discard primary_namespace & secondary_namespace
```
jkczyz, line 43: nit: You can fold the comment into the variable names instead:

```rust
(_primary_namespace, _secondary_namespace) = (parts.next(), parts.next());
```

src/io/vss_store.rs
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	async fn list_all_keys(&self, namespace: &str, sub_namespace: &str) -> io::Result<Vec<String>> {
+		let mut page_token = None;
+		let mut keys = vec![];
+		let key_prefix = format!("{}#{}", namespace, sub_namespace);
+		while page_token != Some("".to_string()) {
+			let request = ListKeyVersionsRequest {
+				store_id: self.store_id.to_string(),
+				key_prefix: Some(key_prefix.to_string()),
```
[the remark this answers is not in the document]
jkczyz: Yeah, `clone` would be more idiomatic. There are a few more places where this can be done for `self.store_id`.

src/io/vss_store.rs
```diff
@@ -0,0 +1,201 @@
...
+		Self { client, store_id, runtime }
+	}
+
+	fn build_key(
+		&self, primary_namespace: &str, secondary_namespace: &str, key: &str,
+	) -> io::Result<String> {
+		if primary_namespace.is_empty() {
+			Ok(key.to_string())
+		} else {
+			Ok(format!("{}#{}#{}", primary_namespace, secondary_namespace, key))
```
jkczyz, line 37: Should we use `/` for the separator so that error messages don't have two different keys when the server returns one in its error message?

src/io/vss_store.rs
```diff
@@ -0,0 +1,201 @@
...
+							primary_namespace, secondary_namespace, key, e
+						);
+						Error::new(ErrorKind::NotFound, msg)
+					}
+					_ => {
+						let msg = format!(
+							"Failed to read from key {}/{}/{}: {}",
+							primary_namespace, secondary_namespace, key, e
+						);
+						Error::new(ErrorKind::Other, msg)
```
jkczyz, line 106: nit: Could DRY up the construction of `msg` and `Error` by only conditioning for the `ErrorKind`.

src/builder.rs
```diff
@@ -48,6 +48,8 @@ use bip39::Mnemonic;
 
 use bitcoin::BlockHash;
 
+#[cfg(any(vss, vss_test))]
```
jkczyz, line 51: Do we need both? Would just `vss` suffice?

jkczyz (commented):
Largely looks good. Don't have any other comments besides these.

src/io/vss_store.rs
```diff
@@ -0,0 +1,195 @@
...
+					let msg = format!(
+						"Failed to read from key {}/{}/{}: {}",
+						primary_namespace, secondary_namespace, key, e
+					);
+					match e {
+						VssError::NoSuchKeyError(..) => Error::new(ErrorKind::NotFound, msg),
+						_ => Error::new(ErrorKind::Other, msg),
+					}
+				})?;
+		Ok(resp.value.unwrap().value)
```
jkczyz, line 103: Oh, do we ever check if the server returns malformed data? (i.e., if the response's value is `None`)

Worse case scenario is we crash because our server is misbehaving. How would a caller normally handle a read failure?

src/io/vss_store.rs
jkczyz: FWIW, those are front-slashes, which aren't use for escaping. 😄  But makes sense if the backend may be using a filesystem instead of database, I suppose.

jkczyz (approved):

jkczyz (approved):

PR #162, opened 2023-09-19
Add script and config for C# bindings

dennisreimann:
Sharing the code for the C# integration, but this isn't fully complete yet, as the `uniffi-bindgen-cs` templates use different casings for the generated types. I tried adapting the config vars `module_name` and `type_name`, but the generator spits out different casings in those two templates:

- `LdkNodeSafeHandle` https://github.com/NordSecurity/uniffi-bindgen-cs/blob/main/bindgen/templates/ObjectTemplate.cs#L6
- `LDKNodeSafeHandle` https://github.com/NordSecurity/uniffi-bindgen-cs/blob/main/bindgen/templates/NamespaceLibraryTemplate.cs#L16

### DIFF
--- a/scripts/uniffi_bindgen_generate_csharp.sh
+++ b/scripts/uniffi_bindgen_generate_csharp.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+# cargo install uniffi-bindgen-cs --git https://github.com/NordSecurity/uniffi-bindgen-cs
+BINDINGS_DIR="./bindings/csharp"
+
+mkdir -p $BINDINGS_DIR
+
+cargo build --release --features uniffi || exit 1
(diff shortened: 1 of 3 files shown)


scripts/uniffi_bindgen_generate_csharp.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# cargo install uniffi-bindgen-cs --git https://github.com/NordSecurity/uniffi-bindgen-cs
```
dennisreimann, line 2:
```suggestion
# cargo install uniffi-bindgen-cs --git https://github.com/NordSecurity/uniffi-bindgen-cs --tag v0.2.3+v0.23.
```

PR #169, opened 2023-09-27
Add `is_running` function to `Node`

### LINKED ISSUE #164: Check if node is running
It would be useful to expose a function to check if the node is running (ie `node.start()` ) was invoked. 

A possible solution:
```rust
/// Check if node runtime is up
pub fn is_node_running(&self) -> bool {
    if self.runtime.read().unwrap().is_some() {
          true
     }
    else { false }
}
```

jbesraa:
resolves #164


src/lib.rs
```diff
@@ -708,6 +708,11 @@ impl<K: KVStore + Sync + Send + 'static> Node<K> {
 		Ok(())
 	}
 
+    /// Checks if [`Node`] runtime is available.
```
jbesraa, line 711: indentation is a bit off, will fix it.

anything else I should mentioned in the comment? 
because this check is more "codey" check and not really has anything to do with the network, should we mention this?

jbesraa:
looks like we need to pin some crate(s) here?

PR #224, opened 2024-01-03
Introduce `next_event_async` allowing to poll event queue

tnull:
We implement a way to asynchronously poll the queue for new events, providing an async alternative to `wait_next_event`.

~~Bindings exposure is still blocked on the next UniFFI release.~~
~~Now also exposed in bindings as UniFFI 0.26 has been released, now based on #230.~~

### DIFF
--- a/bindings/kotlin/ldk-node-android/lib/build.gradle.kts
+++ b/bindings/kotlin/ldk-node-android/lib/build.gradle.kts
@@ -43,6 +43,7 @@ android {
 dependencies {
     implementation("net.java.dev.jna:jna:5.12.0@aar")
     implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk7")
+    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.4")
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -800,6 +800,15 @@ impl<K: KVStore + Sync + Send + 'static> Node<K> {
 		self.event_queue.next_event()
 	}
 
+	/// Returns the next event in the event queue.
+	///
+	/// Will asynchronously poll the event queue until the next event is ready.
+	///
+	/// **Note:** this will always return the same event until handling is confirmed via [`Node::event_handled`].
+	pub async fn next_event_async(&self) -> Event {
```
jbesraa, line 808: its worth following the naming convention with `wait_next_event` and call this `async_next_event`
also, Its a bit confusing to have `wait` and `async` as usually they both mean async something..
maybe wait_next_event should be sync_next_event?

jbesraa: Thanks for the explanation. 

I would also go with `poll_` , would look better than `async fn name_async()`. 
`future_next_event` could be another option..

PR #230, opened 2024-01-24
Upgrade UniFFI to 0.26.0

tnull:
.. which notably features support for `async` methods and `non_exhaustive` enums.


thunderbiscuit:
Hey thanks for pointing out to us that you were doing the upgrade!

Question for ya: do you know if the `uniffi-android.toml` file now called by the `--config` flag simply _adds_ to the configurations already defined in `uniffi.toml`, or if it becomes the only configurations read by the build process?

thunderbiscuit:
Oh thanks! They keep adding pages and I try to make sure I see all the new ones but totally missed that. 👍

PR #243, opened 2024-02-07
Upgrade to LDK v0.0.123-beta

tnull:
We're upgrading to LDK v0.0.122.

This is a WIP in progress PR and we'll incorporate more changes from `rust-lightning`'s `main` branch over time.


src/io/utils.rs
```diff
@@ -200,34 +204,107 @@ where
...
+		kv_store,
+		logger.clone(),
+	);
+	OutputSweeper::read(&mut reader, args).map_err(|e| {
+		log_error!(logger, "Failed to deserialize OutputSweeper: {}", e);
+		std::io::Error::new(std::io::ErrorKind::InvalidData, "Failed to deserialize OutputSweeper")
+	})
+}
+
 /// Read previously persisted spendable output information from the store.
```
jkczyz, line 233: Could you update the docs?

src/io/utils.rs
```diff
@@ -200,34 +204,107 @@ where
...
 		})?;
-		res.push(output);
+		let descriptors = vec![output.descriptor.clone()];
+		let spend_delay = Some(best_block.height + 2);
+		sweeper.track_spendable_outputs(descriptors, output.channel_id, true, spend_delay);
+		if let Some(tracked_spendable_output) =
+			sweeper.tracked_spendable_outputs().iter().find(|o| o.descriptor == output.descriptor)
+		{
+			match tracked_spendable_output.status {
+				OutputSpendStatus::PendingInitialBroadcast { delayed_until_height } => {
```
jkczyz, line 265: It isn't clear to me why this is the only state we would expect. Could you add a comment somewhere describing the migration?

src/io/utils.rs
```diff
@@ -200,34 +204,118 @@ where
...
+					if delayed_until_height == spend_delay {
+						kv_store.remove(
+							DEPRECATED_SPENDABLE_OUTPUT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
+							DEPRECATED_SPENDABLE_OUTPUT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
+							&stored_key,
+							false,
+						)?;
```
jkczyz, line 283: Should we persist the `Sweeper` before removing these?

jkczyz: Oh, never mind. It looks like `track_spendable_outputs` does just that.

jkczyz: Though it will only log but otherwise fail silently. Is this a concern?

jkczyz: Feel free to squash. We can update this in a follow-up as soon as a new RC is available.

PR #244, opened 2024-02-07
Drop `KVStore` type parameter from `Node`

tnull:
Based on #243, depends on https://github.com/lightningdevkit/rust-lightning/pull/2883

Previously, LDK Node's main object has a generic trait bound on `KVStore`, i.e., `Node<K: KVStore + Send + Sync + 'static>`. However, this has always been a painful issue as our bindings generator UniFFI doesn't have generic support, so we were required to type-def and expose only a concrete type, i.e., `Node<SqliteStore>`, in bindings. This will now become even more painful and borderline infeasible with the upcoming introduction of the `VssStore` alternative.

Here we therefore finally drop the generic trait bound, allowing us expose a cleaner `Node` object everywhere.

Draft as blocked on upstream for now.

### DIFF
--- a/bindings/ldk_node.udl
+++ b/bindings/ldk_node.udl
@@ -35,6 +35,8 @@ interface Builder {
 	void set_listening_addresses(sequence<SocketAddress> listening_addresses);
 	[Throws=BuildError]
 	Node build();
+	[Throws=BuildError]
+	Node build_with_fs_store();
 };
 
 interface Node {


tests/integration_tests_rust.rs
```diff
@@ -157,7 +159,8 @@ fn start_stop_reinit() {
 
 	let esplora_url = format!("http://{}", electrsd.esplora_url.as_ref().unwrap());
 
-	let test_sync_store = Arc::new(TestSyncStore::new(config.storage_dir_path.clone().into()));
+	let test_sync_store: Arc<dyn KVStore + Sync + Send> =
```
jkczyz, line 162: nit: `DynStore`

jkczyz (commented):
LGTM

jkczyz (approved):

PR #256, opened 2024-02-21
Add BOLT12 support

tnull:
~~Based on https://github.com/lightningdevkit/ldk-node/pull/243.~~
~~Based on https://github.com/lightningdevkit/ldk-node/pull/244.~~
~~Based on https://github.com/lightningdevkit/ldk-node/pull/266.~~
~~Based on https://github.com/lightningdevkit/ldk-node/pull/270.~~

~~WIP, hence draft.~~

Ready for review.

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -2,7 +2,7 @@
 #![allow(dead_code)]
 
 use ldk_node::io::sqlite_store::SqliteStore;
-use ldk_node::payment::{PaymentDirection, PaymentStatus};
+use ldk_node::payment::{PaymentDirection, PaymentKind, PaymentStatus};
 use ldk_node::{Builder, Config, Event, LogLevel, Node, NodeError};
 
 use lightning::ln::msgs::SocketAddress;
@@ -447,9 +447,11 @@ pub(crate) fn do_channel_full_cycle<E: ElectrumApi>(
 	assert_eq!(node_a.payment(&payment_id).unwrap().status, PaymentStatus::Succeeded);
 	assert_eq!(node_a.payment(&payment_id).unwrap().direction, PaymentDirection::Outbound);
 	assert_eq!(node_a.payment(&payment_id).unwrap().amount_msat, Some(invoice_amount_1_msat));
+	assert!(matches!(node_a.payment(&payment_id).unwrap().kind, PaymentKind::Bolt11 { .. }));
 	assert_eq!(node_b.payment(&payment_id).unwrap().status, PaymentStatus::Succeeded);
 	assert_eq!(node_b.payment(&payment_id).unwrap().direction, PaymentDirection::Inbound);
 	assert_eq!(node_b.payment(&payment_id).unwrap().amount_msat, Some(invoice_amount_1_msat));
+	assert!(matches!(node_b.payment(&payment_id).unwrap().kind, PaymentKind::Bolt11 { .. }));
 
 	// Assert we fail duplicate outbound payments and check the status hasn't changed.
 	assert_eq!(Err(NodeError::DuplicatePayment), node_a.bolt11_payment().send(&invoice));
@@ -492,9 +494,11 @@ pub(crate) fn do_channel_full_cycle<E: ElectrumApi>(
 	assert_eq!(node_a.payment(&payment_id).unwrap().status, PaymentStatus::Succeeded);
 	assert_eq!(node_a.payment(&payment_id).unwrap().direction, PaymentDirection::Outbound);
 	assert_eq!(node_a.payment(&payment_id).unwrap().amount_msat, Some(overpaid_amount_msat));
+	assert!(matches!(node_a.payment(&payment_id).unwrap().kind, PaymentKind::Bolt11 { .. }));
 	assert_eq!(node_b.payment(&payment_id).unwrap().status, PaymentStatus::Succeeded);
 	assert_eq!(node_b.payment(&payment_id).unwrap().direction, PaymentDirection::Inbound);
 	assert_eq!(node_b.payment(&payment_id).unwrap().amount_msat, Some(overpaid_amount_msat));
+	assert!(matches!(node_b.payment(&payment_id).unwrap().kind, PaymentKind::Bolt11 { .. }));
 
 	// Test "zero-amount" invoice payment
 	println!("\nB receive_variable_amount_payment");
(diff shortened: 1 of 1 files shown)


src/payment/bolt12.rs
```diff
@@ -0,0 +1,272 @@
...
+						let kind = PaymentKind::Bolt12 { hash: None, preimage: None, secret: None };
+						let payment = PaymentDetails {
+							id: payment_id,
+							kind,
+							amount_msat: Some(*offer_amount_msat),
+							direction: PaymentDirection::Outbound,
+							status: PaymentStatus::Failed,
+						};
+						self.payment_store.insert(payment)?;
+						Err(Error::PaymentSendingFailed)
```
jkczyz, line 117: Should we use something like `InvoiceRequestCreationFailed`?

src/payment/bolt12.rs
```diff
@@ -0,0 +1,272 @@
...
+			},
+			None => amount_msat,
+		};
+
+		if amount_msat < offer_amount_msat {
+			log_error!(
+				self.logger,
+				"Failed to pay as the given amount needs to be at least the offer amount: required {}msat, gave {}msat.", offer_amount_msat, amount_msat);
+			return Err(Error::InvalidAmount);
+		}
```
jkczyz, line 163: Maybe do this inside the above match arm so that you don't later shadow `offer_amount_msat`.

src/payment/bolt12.rs
```diff
@@ -0,0 +1,272 @@
...
+		) {
+			Ok(()) => {
+				let payee_pubkey = offer.signing_pubkey();
+				let offer_amount_msat = offer.amount();
+				log_info!(
+					self.logger,
+					"Initiated sending {:?} to {:?}",
+					offer_amount_msat,
+					payee_pubkey
+				);
```
jkczyz, line 182: Hmm... shouldn't this actually be `amount_msats`?

src/payment/bolt12.rs
```diff
@@ -0,0 +1,272 @@
...
+			payer_note,
+			payment_id,
+			retry_strategy,
+			max_total_routing_fee_msat,
+		) {
+			Ok(()) => {
+				let payee_pubkey = offer.signing_pubkey();
+				log_info!(
+					self.logger,
+					"Initiated sending {:?} to {:?}",
```
jkczyz, line 86: Could we give the units? Likewise below.

src/payment/bolt12.rs
```diff
@@ -0,0 +1,272 @@
...
+		let offer = offer_builder
+			.amount_msats(amount_msat)
+			.description(description.to_string())
+			.build()
+			.map_err(|e| {
+				log_error!(self.logger, "Failed to create offer: {:?}", e);
+				Error::OfferCreationFailed
+			})?;
+
+		let id = PaymentId(offer.id().0);
```
jkczyz, line 233: Not quite sure if I followed the transition plan. IIIUC, someone paying the offer multiple times would result in duplicate payment ids, right?

src/payment/bolt12.rs
```diff
@@ -0,0 +1,272 @@
...
+						Err(Error::PaymentSendingFailed)
+					},
+				}
+			},
+		}
+	}
+
+	/// Returns a payable offer that can be used to request and receive a payment of the amount
+	/// given.
+	pub fn receive(&self, amount_msat: u64, description: &str) -> Result<Offer, Error> {
```
jkczyz, line 219: Any reason why we don't take a `String` if we need to own it?

src/event.rs
```diff
@@ -445,6 +455,7 @@ where
 						self.channel_manager.fail_htlc_backwards(&payment_hash);
 
 						let update = PaymentDetailsUpdate {
+							hash: Some(Some(payment_hash)),
```
jkczyz, line 458: Is this only needed for BOLT12?

src/event.rs
```diff
@@ -502,48 +514,75 @@ where
...
-							payment_preimage
-						} else {
-							self.channel_manager
-								.get_payment_preimage(payment_hash, payment_secret)
-								.ok()
-						}
+						payment_preimage.or(self
+							.channel_manager
+							.get_payment_preimage(payment_hash, payment_secret)
+							.ok())
```
jkczyz, line 520: Need to consider this further, but I thought a `None` preimage meant that the user needs to supply it, right? How could we get it from `ChannelManager`?

src/lib.rs
```diff
@@ -846,6 +846,32 @@ impl Node {
 		))
 	}
 
+	/// Returns a payment handler allowing to create and pay [BOLT 12] offers and refunds.
```
jkczyz, line 852: Will refund support come in a follow-up?

src/payment/store.rs
```diff
@@ -176,6 +176,18 @@ pub enum PaymentKind {
 		/// [`LdkChannelConfig::accept_underpaying_htlcs`]: lightning::util::config::ChannelConfig::accept_underpaying_htlcs
 		lsp_fee_limits: LSPFeeLimits,
 	},
+	/// A [BOLT 12] payment.
+	///
+	/// [BOLT 12]: https://github.com/lightning/bolts/blob/master/12-offer-encoding.md
+	// TODO: Bolt12 { offer: Option<Offer>, refund: Option<Refund> },
```
jkczyz, line 182: Hmm... seems we may want a separate kind for offer and refund?

src/payment/bolt12.rs
```diff
@@ -264,4 +267,62 @@ impl Bolt12Payment {
...
 	}
+
+	/// Requests a refund payment for the given [`Refund`].
+	///
+	/// The returned [`Bolt12Invoice`] is for informational purposes only (i.e., isn't needed to
+	/// retrieve the refund).
+	pub fn request_refund(&self, refund: &Refund) -> Result<Bolt12Invoice, Error> {
+		self.channel_manager.request_refund_payment(refund).map_err(|e| {
+			log_error!(self.logger, "Failed to request refund payment: {:?}", e);
+			Error::InvalidRefund
```
jkczyz, line 278: Similarly here I'm not sure if we want to differentiate when we fail to create the invoice / reply path.

src/payment/bolt12.rs
```diff
@@ -264,4 +267,62 @@ impl Bolt12Payment {
...
+				max_total_routing_fee_msat,
+			)
+			.map_err(|e| {
+				log_error!(self.logger, "Failed to create refund builder: {:?}", e);
+				Error::RefundCreationFailed
+			})?;
+		let refund = refund_builder.build().map_err(|e| {
+			log_error!(self.logger, "Failed to create refund: {:?}", e);
+			Error::RefundCreationFailed
+		})?;
```
jkczyz, line 310: Probably can chain these two.

src/payment/bolt12.rs
```diff
@@ -264,4 +267,62 @@ impl Bolt12Payment {
 
 		Ok(offer)
 	}
+
+	/// Requests a refund payment for the given [`Refund`].
+	///
+	/// The returned [`Bolt12Invoice`] is for informational purposes only (i.e., isn't needed to
+	/// retrieve the refund).
+	pub fn request_refund(&self, refund: &Refund) -> Result<Bolt12Invoice, Error> {
+		self.channel_manager.request_refund_payment(refund).map_err(|e| {
```
jkczyz, line 276: Should we also insert this into the payment store as an inbound payment given we have the payment hash?

jkczyz:
> Addressed the open feedback, @jkczyz let me know if I can squash.

Yes, please squash

src/event.rs
```diff
@@ -549,42 +549,7 @@ where
...
-							Err(e) => {
-								log_error!(
-									self.logger,
-									"Failed to insert payment with ID {}: {}",
-									payment_id,
-									e
-								);
-								debug_assert!(false);
-							},
-						}
```
jkczyz: If possible, could you remove this from the earlier commit since it was not needed.

src/payment/store.rs
```diff
@@ -176,6 +175,30 @@ pub enum PaymentKind {
...
+		preimage: Option<PaymentPreimage>,
+		/// The secret used by the payment.
+		secret: Option<PaymentSecret>,
+		/// The ID of the offer this payment is for.
+		offer_id: OfferId,
+	},
+	/// A [BOLT 12] 'refund' payment, i.e., a payment without a prior offer.
+	///
+	/// [BOLT 12]: https://github.com/lightning/bolts/blob/master/12-offer-encoding.md
+	Bolt12Refund {
```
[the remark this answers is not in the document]
jkczyz: How about `initiate_refund` for creation and possibly `request_payment` for requesting? Using `request_refund` makes it seem like it comes before `create_refund`, even though the order is reversed.

PR #257, opened 2024-02-21
Payjoin integration

jbesraa:
**April 18 Update:**

 This pull request adds two new features:
 1) Allowing the node to accept payjoin transactions to its bitcoin wallet.
 2) Allowing the node to open a channel from funds received in a  payjoin transaction, saving the user an additional on-chain transaction.

* The first 3 commits are about adding the new `lightning_payjoin` crate and will be moved to a different repo once the code is approved. 

* Added new crate `lightning_payjoin` responsible for gluing together  `ldk-node` and `rust-payjoin` . The library exports a `Receiver` trait and a `LightningPayjoin` struct which gives the ability to handle payjoin transactions.
The main code is basically living in this crate including the code for accepting normal payjoin transactions, handling lightning channel opening from payjoin transaction, validating transactions and other http stuff.

* The current flow for opening a channel with payjoin transaction:
  1. Node runner calls `open_payjoin_channel` function
     `open_payjoin_channel` will:
     1.1 Add new channel to scheduler
     1.2 The channel manager will open a new channel and stop after `ChannelAccepted` msg because its part of the    `ChannelScheduler`
     1.3 Function will return BIP21 for the user to pay
   2. Node runner  scan BIP21 with a wallet supporting payjoin
   3. Node will accept the payment on the `payjoin_server_port` its listening on
   4. Look for scheduled channel from `ChannelScheduler` and change the `output_script` to the funding tx multisig
   5. Call `channel_manager.funding_generation_ready`
   6. Return response to payjoin sender after `FundingSigned` received from counterparty node

  We could change the order so when the user calls `open_payjoin_channel` its only adding a new channel to the `ChannelScheduler` and when the payments comes in we could start the channel opening and finalize it before we return a response. It would require exposing a bit more functionality to the `PayjoinHandler` like `PeerManager` and probably the `Logger`.  there will be some duplicate code if we change it as the channel opening defined in `lib` would also be in `payjoin_handler`. I dont have a strong feeling about either.

* The current implementation is following Payjoin V1 which requires the receiver to run an http server where in Payjoin V2 its possible to only have an http client. IMO its worth first supporting v1 as thats whats currently widely used by wallets and using v2 now could limit the usage of this feature. as all of the http code is setting in `lightning-payjoin` I think its a good choice for now to go with v1.

-----
Original

The main goal of this pull request is to allow users to create channel with funds outside the node.

Flow that should be covered:
1) Alice run Node A enabling "payjoin" and acting as a "payjoin receiver"
2) Alice generates payjoin uri
3) Alice receives a request for payjoin payment from Charlie
4) Alice want to use the funds from the payjoin payment to open a channel with a single transaction.
5) Alice negotiates channel opening with Bob(using the psbt it received in step 3)
6) Alice get `FundingSigned` from Bob
7) Alice send the finalized psbt to Charlie
8) Charlie broadcast the tx

** This PR should cover Alice part **

There are few tricky points, mainly around 6 and 7. Currently transactions are immediately  broadcasted after `FundingSigned`, that wont work with payjoin because the transaction wont be valid.

The channel opened is Outbound channel from Alice -> Bob but it will be broadcasted by Charlie, so we need to add some functionality to let Alice node know that its an Outbound channel. 

some reference for payjoin integration with lnd 
https://github.com/payjoin/nolooking


DanGould:
This sequence diagram might help those getting familiar with the how lightning channels can be opened from payjoin transactions

```
    The `merchant` hosts the payjoin receiver.

     ┌──────────────┐                ┌─────────────────┐                  ┌──────┐
     │Lightning Peer│                │    `merchant`   │                  │Sender│
     └──────┬───────┘                └───────┬─────────┘                  └───┬──┘
            │                                │                                │
            │            BOLT 2              ├─────── Bip21 with ?pj= ───────►│
            │     Channel Establishment      │                                │
            │                                │◄────── Original PSBT ──────────┤
            │                                │                                │
            │                                │                                │
            │◄──────── open_channel ─────────┤                                │
            │                                │                                │
            ├──────── accept_channel ───────►│                                │
            │                                │             BIP 78             │
            │                                │                                │
            │◄─────── funding_created ───────┤                                │
            │                                │                                │
            ├──────── funding_signed ───────►│                                │
            │                                │                                │
            │                                │        PayJoin Proposal        │
            │                                ├──────       PSBT       ───────►│
            │                                │                                │
            │                                │                                │
            │                                │    ┌─ PayJoin + Funding ───────┤
            │                                │    │     Transaction           │
            │                                │    │                           │
           x│xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx│xxx ▼ xxxxxxxxxxxxxxxxxxxxxxxxxx│x
           xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
           xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx BITCOIN NETWORK xxxxxxxxxxxxxxxxxxxxx
           xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
           x│xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx│xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx│x
            │                                │                                │
            │◄────────channel_ready ─────────┤                                │
            │                                │                                │
            ├──────── channel_ready ────────►│                                │
            │                                │                                │

```

jbesraa:
A few points to discuss:
1) In the PR `hyper` is used to start and deal with the `http_server`. while all hyper functionality could be implemented using tokio, I choose it first as its faster to setup and some functionality like shared state is more intuitive to do.
2) The node runner will need to set whom they are interested to open channel with, I added two end points to http server to do that. this need concept ack.
3) `payjoin_handler` will probably need to access `channel_manager`(see the todos in the function) but it doesnt implement `clone` and I am not feeling very comfortable adding it). I would appreciate input about this.
@DanGould

DanGould:
> 1. In the PR `hyper` is used to start and deal with the `http_server`. while all hyper functionality could be implemented using tokio, I choose it first as its faster to setup and some functionality like shared state is more intuitive to do.

Hyper serves https and makes use of the async tokio runtime being used in LDK node already. This makes more sense to me than rewriting an http stack.

Indeed, a payjoin receiver needs to share concurrent state. I suggest writing down exactly what state needs to be shared and what each component it's shared with needs to do with it, and whether or not it needs its own thread.

> 2. The node runner will need to set whom they are interested to open channel with, I added two end points to http server to do that. this need concept ack.

Indeed, this is the first documentation of who needs what state.

`payjoin_handler` in my opinion should only handle payjoin POST requests. Seems like node management is done elsewhere by exposing methods on the `Node` which could then be called by an downstream program. HTTP should probably only be used to handle the payjoin protocol itself. `Node` control methods should be used to control it. I believe my suggestion is consistent with the current LDK-Node design.

> 3. `payjoin_handler` will probably need to access `channel_manager`(see the todos in the function) but it doesnt implement `clone` and I am not feeling very comfortable adding it). I would appreciate input about this.

shared state can be passed as you figured out `let clone_ccr = create_channel_request.clone();` but I'm not totally sure that a "CreateChannelRequest" should itself be that shared state. More likely, you want a shared scheduler service that listens for create channel request events. [Nolooking has this scheduler architecture](https://github.com/payjoin/nolooking/blob/master/src/scheduler.rs) you can follow. `payjoin_handler` would then have shared access to the scheduler. Does LDK-Node offers batching already? This scheduler could be combined with that service if so. By sharing a listening service instead of a single request that gets mutated in place, the node maintain a queue of user intentions with a clear, abstract, purpose-built API.

Clone would be a bit of a hack since channel_manager is a singleton service that needs to have its shared state altered while the program is running. I would guess that the channel_manager would be a dependency of whatever batched scheduler service you come up with. To share the state, choose an appropriate memory container and access pattern. This flow chart has helped me a ton. Rather than wrap each field of shared state in its own `Arc<Mutex>`, perhaps it would be appropriate to wrap the whole shared state in a single Mutex or other appropriate container.

image

DanGould:
> If a reachable HTTP endpoint is a hard requirement for the proposed PJ scheme the handler probably should live in a different crate and we need to find a good interface in order to configure LDK node with it.

The big issue with Version 1 payjoin is exactly this dependency on an HTTP server stack. Fortunately [the Payjoin Version 2 BIP](https://github.com/bitcoin/bips/pull/1483) forgoes this dependency for a mere HTTP client. This draft uses V1 but I imagine the merged PR will use V2. While the `payjoin` crate itself does no I/O, a crate to do basic payjoin networking for most implementations is a solid idea.

src/lib.rs
```diff
@@ -462,6 +471,42 @@ impl<K: KVStore + Sync + Send + 'static> Node<K> {
 			});
 		}
 
+		use hyper::server::conn::http1;
+		use hyper_util::rt::TokioIo;
+		use std::net::SocketAddr;
+		use tokio::net::TcpListener;
+
+		// Start the HTTP server for payjoin
+		let (payjoin_queue_sender, mut payjoin_queue_receiver) = mpsc::channel::<PayjoinRequest>(1);
```
jbesraa, line 480: @DanGould starting a channel here that is accessed from the payjoin endpoint. the payjoin endpoint will add `PayjoinRequest` to the queue, and wait for response(funding psbt) via different channel from `create_channel_from_pj`.

src/lib.rs
```diff
@@ -675,6 +720,41 @@ impl<K: KVStore + Sync + Send + 'static> Node<K> {
 		self.runtime.read().unwrap().is_some()
 	}
 
+	async fn create_channel_from_pj(
+		pj_req: PayjoinRequest, _payjoin_executer: Arc<PayjoinExecuter<K>>,
+	) {
+		let psbt = pj_req.clone().original_psbt();
+		let pj_response = PayjoinResponse::new(psbt);
+		// Send OpenChannel message
```
jbesraa, line 728: This is where the channel negotiations should happen and the `queue` the response back to the http handler

src/lib.rs
```diff
@@ -675,6 +720,41 @@ impl<K: KVStore + Sync + Send + 'static> Node<K> {
...
+		// Wait for AcceptChannel message
+		// Send FundingCreated message
+		// Wait for FundingSigned message
+		// Build PayjoinResponse
+		// Send PayjoinResponse to queue
+		pj_req.clone().queue(pj_response);
+	}
+
+	/// Request a new channel to be opened with a remote peer.
+	pub fn payjoin_channel(
```
jbesraa, line 738: this function is currently doing two things which is not great.
it should be broke into two functions and in the "payjoin_uri" docs we should mention that the `pending_channels` array should be filled first

jbesraa:
@DanGould In the last commit I addressed the scheduler we discussed in the last discussion.

Added a new struct `PayjoinScheduler` that is utilised on an incoming payjoin request, that only pushes the request to the queue and wait for response.

You can see now other new structs like `PayjoinRequest`, `PayjoinResponse`, `PendingChannels` and others added as well. those are not mature enough yet and they will probably change, this is just the start. I am also not a big fan of the naming yet.

`mpsc` channels are used so its consistent with other structs with similar functionality(like tx broadcaster).

src/pj.rs
```diff
@@ -0,0 +1,175 @@
...
+impl Service<Request<IncomingBody>> for PayjoinScheduler {
+	type Response = hyper::Response<Full<bytes::Bytes>>;
+	type Error = hyper::Error;
+	type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
+
+	fn call(&self, http_request: Request<IncomingBody>) -> Self::Future {
+		let state = Arc::new(Mutex::new(self.clone()));
+		let res = match (http_request.method(), http_request.uri().path()) {
+			_ => payjoin_handler(http_request, state),
+		};
```
DanGould, line 171: I think methods are being matched twice here for no reason. Unless you're going to transform a custom error type into an http response, you should probably call payjoin_handler directly without the match

DanGould (commented):
My biggest confusion with the design is spawning loops to listen for mpsc channel updates. Rather than spawn a loop listening on an mpsc channel, you can use an async function like PayjoinScheduler::queue_channel and avoid that complexity. This can encapsulate `create_channel_from_pj` within a payjoin module.

I have the same concern with using mpsc loop to respond to payjoin requests. POST requests to `/payjoin` shouldn't schedule anything, they should check against the queue from the scheduler to attempt to pop from it and open a channel, otherwise accept the funds into a single-sig address. HTTP response must be synchronous anyhow, it does you no good to wait for a channel update if you can't respond to the HTTP request with it. There is no need to manually parse the PSBT from the request body, the payjoin crate has all of the methods to do that for you. Just marshal HTTP I/O data [into and out of the payjoin crate methods](https://docs.rs/payjoin/latest/payjoin/receive/struct.UncheckedProposal.html#method.from_request). Define a function that takes the request and returns a response.

I think what you are trying to do is separate the scheduler logic from the http server, which does make sense. However, the server could reference the scheduler directly as a field instead of an opaque mpsc::Sender. This lets it interface with rich function signatures instead of channels in loops. Both the receiver server and the Node should share reference to the Scheduler and interface with its methods rather than mpsc channels.

jbesraa:
> > If a reachable HTTP endpoint is a hard requirement for the proposed PJ scheme the handler probably should live in a different crate and we need to find a good interface in order to configure LDK node with it.
> 
> The big issue with Version 1 payjoin is exactly this dependency on an HTTP server stack. Fortunately [the Payjoin Version 2 BIP](https://github.com/bitcoin/bips/pull/1483) forgoes this dependency for a mere HTTP client. This draft uses V1 but I imagine the merged PR will use V2. While the `payjoin` crate itself does no I/O, a crate to do basic payjoin networking for most implementations is a solid idea.

I would like not to rely on v2 for this integration for the following reasons:
1) we link this PR to the approval of the bip which can delay it
2) the single goal of this pr is allowing the node runner to open a channel with funds from a separate wallet, so we will end up with the user communicating with their self over a relay.. 

The idea of having a separate crate for a payjoin-server is very reasonable. We can start with v1 and once v2 is more mature we can upgrade the payjoin-server and do minimal changes in ldk-node.

jbesraa:
@DanGould @tnull 
I think this is ready for review. 

There are two main files to look into:
first `pj_new_crate.rs`, the code in this file will move to a separate repo+crate and will be added as a dependency to ldk-node. once you guys approve the content of the file, im happy to do the above and break the commits accordingly.
in `pjoin.rs` I added the implementation `ldk-node` needs to write to integrate with `pj_new_crate`. notice that this is aiming to complete the initial full setup. and the bulk of the work is yet to be implemented. I added `request_to_psbt` to the `PayjoinExecuter` trait and thats where the channel opening functionality should go. the trait is have async function and that breaks the msrv, I wonder if there is a way to still use async trait ?

Integration tests added in `integration_tests_payjoin.rs` and they cover the currently functionality, but one problem I encountered was the inability to call async functions implemented in `Node` like `.schedule_payjoin_channel`  from the test context. I tried to run under `tokio::test` but that failed and I think because it conflicts with the node runtime. but iam not sure tbh. would appreciate some guidance on that.

DanGould:
I'm attempting a comprehensive design review of the abstraction intended to enable payjoin receipt for lightning channel opens funded directly by outside funds.

## Async Design

The past review raised concerns with the async design, suggesting the use of async functions and scoped memory containers instead of mpsc channels. I am satisfied with the direction of this abstraction 🎊🚀 it's much easier to read, reason about, and should be easier to change.

## New crate

@tnull voiced skepticism with regard to the inclusion of an HTTP server to receive payjoin in LDK-Node.

> I'm generally very skeptical to go down the outlined route in respect to adding an HTTP server stack to LDK Node itself. If a reachable HTTP endpoint is a hard requirement for the proposed PJ scheme the handler probably should live in a different crate and we need to find a good interface in order to configure LDK node with it.

*Assuming* the stack is a hard requirement, I agree, an HTTP stack would benefit from its own crate. This could be shared between `payjoin-cli` and `ldk-node` for the purpose of accepting v1 payjoins. It would make sense to live in the `rust-payjoin` repository (`payjoin-io` ?), and could abstract both send and receive io functionality. For this reason, I'm skeptical of the drafted new crate including scheduling logic. It seems more natural for such a crate to handle only the http stack and leave lightning specifics to LDK-Node. Should we come across multiple implementations of lightning payjoin functionality, another `payjoin-lightning` crate could be born.

However, a payjoin v2 receiver is viable for this PR exclude the requirement for an http stack. @jbesraa raised 2 points I would like to respond to.

> 1. we link this PR to the approval of the bip which can delay it

BIPs don't have an "approval" process per se, but an [open workflow](https://github.com/bitcoin/bips/blob/master/bip-0002.mediawiki#bip-workflow). Sure, the actual file can get approved to be included in the repository. But even that process includes implementing the BIP in the wild by multiple implementations before being merged.

> 2. the single goal of this pr is allowing the node runner to open a channel with funds from a separate wallet, so we will end up with the user communicating with their self over a relay.

It's a good idea to narrow the scope of this PR to do one thing. For that reason, it makes sense to continue to document how a new crate might help, but in my opinion also suggests that the development of a new crate is outside of scope.

Regarding "communicating to their self over a relay," doing so forgoes the http server requirement and the TLS/onion hidden service setup that would otherwise be required. What exactly are the downsides you predict in using a relay? I have listened to concerns about IP and privacy leaks, but BIP 77 explicitly addresses these with OHTTP and authenticated encryption. Is latency a concern? I believe that's addressed by the asynchronous nature of BIP 77 as well.

Even with payjoin v2, a `payjoin-io` crate could still be useful for default payjoin http i/o implementations and ought to make its way into our payjoin roadmap.

## Payjoin Executer

I'm not certain the `PayjoinExecuter` needs to exist as a named abstraction. While I admire the effort to design contained simple abstractions, it seems like the executer is an extra layer of complexity inside `PayjoinScheduler` since each of ts methods call into scheduler but it's not clear to me what exactly it's abstracting.  My understanding of the suggested new crate was to implement the HTTP I/O to be maintained outside of LDK-Node, not lightning-specific payjoin operations.

The `PayjoinExecuter` is also defined as a trait in the new pseudo-crate module, but I'm not sure what it is generic over? If it's not generic over a particular service or data, What's keeping it from being a mere struct impl, or even just fields and methods inside `PayjoinScheduler`. I'm not sure of the purpose of `request_to_psbt`, since the `payjoin` state machine transformas a request entirely into a response. The necessary interface will probably not appear until the rest of the implementation is drafted, so I wouldn't let this design block attempting the rest of the integration.

## `serve` in `PayjoinScheduler`

Assuming the ldk `Node` only hangs on to one piece of payjoin data, and it's OK to couple the payjoin server to the ldk-node implementation, I think including serve in the scheduler is a great idea. It simplifies the design a ton. I am a bit concerned that it constructs a new scheduler rather than cloning self, however. Doesn't that break the singleton pattern?

## `make_http_response`

`make_http_response("404".into())` will return a `Response` struct having Status code 200 with body "404". Make sure you're setting the status code and body appropriately. I prefer to build new mutable `Response` structs and setting their fields rather than the builder outside of tests so you don't have to handle errors with `unwrap()` too.

## Testing

[grug brained developer take is to avoid mocking](https://grugbrain.dev/#grug-on-testing), and I agree. An integration with a real payjoiner has examples and would prevent incorrect mocking from causing bugs that don't get caught down the road. It looks like despite the "mock_payjoin_sender" name you do in fact implement a regtest payjoin sender, and don't mock one. Funny, a `payjoin-io` crate could get rid of some of the complext http handling in these tests too. I'm seeing a pattern.

>  [I'm unable] to call async functions implemented in Node like .schedule_payjoin_channel from the test context. I tried to run under tokio::test but that failed and I think because it conflicts with the node runtime. 

True, `setup_node` is blocking and tokio::test is a non-blocking context. You should be able to use tokio spawn_blocking to call such methods in a scope of that context. You may need to manage similar blocking vs non-blocking tasks with your reqwest clients. reqwest fortunatly makes both types available with/without the `blocking` feature.

### Random port

Why use randomness when you can have a deterministic answer? Binding to port 0 makes the OS pick an available port for you. [Even in python](https://stackoverflow.com/a/1365284). Random port selection could lead to random test failure.

e.g. in rust

```rs
fn find_free_port() -> u16 {
    let listener = std::net::TcpListener::bind("0.0.0.0:0").unwrap();
    listener.local_addr().unwrap().port()
}
```

## On Review submission

It's much easier to have some certainty about whether or not a design is possible if you can show passing CI, even if that means changing (and documenting) the CI itself. This concern relates to the new crate idea, too. *If* you want to prove out a new crate (IMO better as a separate PR/discussion at this stage), make a new crate in the same workspace rather than mock it with a module. That would help design its dependencies and interface concretely vs a mocked crate in a new module.

## Final thoughts

This PR has come a long way to addressing the issue of getting channels open from a payjoin, especially with regard to managing concurrent tasks and I'm stoked to see that. Answering the questions about generic business PayjoinExecutor is doing (if any) will help address the async trait issue. From that answer you should be able to begin implementing the channel logic and get a better idea of what the implementation looks like inside the scheduler. From my view it appears like it can exist entirely inside the PayjoinScheduler.

Until some end-to-end implementation is committed bits and bytes. we're reviewing a design with one eye blind. It'll almost certainly change, and that's OK. Solid progress.

lightning-payjoin/src/error.rs
```diff
@@ -0,0 +1,123 @@
+use lightning::util::errors::APIError;
+
+#[derive(Debug)]
+pub enum Error {
+	Internal(InternalError),
```
jbesraa, line 5: Internal error referring to `lightining_payjoin` crate error

lightning-payjoin/src/lib.rs
```diff
@@ -1 +1,316 @@
...
+use std::sync::Arc;
+use tokio::net::TcpStream;
+use tokio::sync::Mutex;
+
+/// Payjoin `Receiver` trait defines the behavior of the receiver side in a payjoin transaction
+/// withing a lightning network context.
+///
+/// Utilising payjoin transactions, a lightning node setup as a receiver can fund a channel from an
+/// incoming payjoin transaction saving additional on-chain transaction.
+pub trait Receiver: Send + Sync + 'static + Clone {
```
jbesraa, line 25: This trait is implemented by `ldk-node` in the `payjoin_handler.rs` file. the following struct `LightningPayjoin` will use those functions to either accept normal pj transaction or open a channel

lightning-payjoin/src/lib.rs
```diff
@@ -1 +1,316 @@
...
+			Ok(body) => body.to_bytes().to_vec(),
+			Err(_e) => panic!("panicesh"),
+		};
+		let query = parts.uri.query().unwrap_or("").to_string();
+		let headers = parts.headers;
+		let headers = RequestHeaders::from(headers);
+		Ok((body, query, headers))
+	}
+
+	fn validate_payjoin_request(
```
jbesraa, line 134: These validations are based on bip78 https://github.com/bitcoin/bips/blob/master/bip-0078.mediawiki#user-content-Receivers_original_PSBT_checklist

lightning-payjoin/src/lib.rs
```diff
@@ -1 +1,316 @@
...
+	) -> Result<String, Error> {
+		let (body, query, headers) = self.extract_request_parts(request).await?;
+		let provisional_proposal =
+			self.validate_payjoin_request(body.as_slice(), &query, headers.clone(), None)?;
+		let payjoin_amount = provisional_proposal.payjoin_amount();
+		if let Some(channel) = self.get_next_channel(payjoin_amount).await {
+			let mut proposal = provisional_proposal.clone();
+			let output_script = channel.output_script().expect("Output script should exist");
+			// substitute the output script with the channel output script
+			proposal.substitute_output_script(output_script);
```
jbesraa, line 176: https://github.com/payjoin/rust-payjoin/pull/233/commits/28533dabe712d07bc1454739a832efcb5f2c9ff4#diff-5fba6fc39c59b07e5980d87da92537138068364d192a2b63eee5e57af01a07bbR476

lightning-payjoin/src/lib.rs
```diff
@@ -1 +1,316 @@
...
+			let temporary_channel_id =
+				channel.temporary_channel_id().expect("Temporary channel id should exist");
+			// tell the counterparty node that the funding tx has been created
+			let _ = self.receiver_handler.funding_transaction_generated(
+				&temporary_channel_id,
+				channel.node_id(),
+				funding_tx.clone(),
+			)?;
+			// wait for the counterparty node to return FundingSigned message
+			match tokio::time::timeout(tokio::time::Duration::from_secs(3), async move {
```
jbesraa, line 195: The only way for me to know if the `FundingSigned` was received is by catching the tx in the `tx_broadcaster` https://github.com/lightningdevkit/ldk-node/pull/257/commits/d682093b62bbbc623c20b8fe210bec85dafcabd9#diff-b707fa4c317cfc8afdfeb972e8ae1b6fd4dc36361289cc53703af17d08dd6630R53 so I wait here a bit until the `tx_broadcaster` updates the `ChannelScheduler` state.. need a better way to handle this

lightning-payjoin/src/lib.rs
jbesraa: this is where we change output of the psbt to point to the funding tx multisig

lightning-payjoin/src/lib.rs
```diff
@@ -1 +1,316 @@
...
+			},
+			_ => Ok(hyper::Response::builder().body(Full::new(bytes::Bytes::from("404")))?),
+		}
+	}
+}
+
+#[derive(Clone)]
+struct RequestHeaders(HashMap<String, String>);
+
+impl payjoin::receive::Headers for RequestHeaders {
```
jbesraa, line 233: This is required by the `payjoin` crate

src/payjoin_handler.rs
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	fn sign_tx(&self, psbt: &Psbt) -> Result<Psbt, PayjoinError> {
+		let mut sign_options = SignOptions::default();
+		sign_options.try_finalize = false;
+		Ok(self.wallet.sign_tx(psbt, Some(sign_options))?)
+	}
+
+	fn can_broadcast(&self, _transaction: &bitcoin::Transaction) -> Result<bool, PayjoinError> {
+		Ok(true)
```
jbesraa, line 83: oops, this should be fixed. I dont think the current blockchain or wallet implementation in ldk-node have this functionality. basically need equivalent to https://developer.bitcoin.org/reference/rpc/testmempoolaccept.html.
I think its achievable with `consensus::verify` module in `bitcoin`

src/payjoin_handler.rs
```diff
@@ -0,0 +1,85 @@
...
+struct PayjoinReceiver<K: KVStore + Sync + Send + 'static> {
+	channel_manager: Arc<ChannelManager<K>>,
+	wallet: Arc<Wallet>,
+}
+
+impl<K> Clone for PayjoinReceiver<K>
+where
+	K: KVStore + Sync + Send + 'static,
+{
+	fn clone(&self) -> Self {
```
jbesraa, line 45: I needed to implement this manually

lightning-payjoin/src/error.rs
```diff
@@ -0,0 +1,123 @@
...
+impl std::fmt::Display for Error {
+	fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+		match *self {
+			Self::Internal(ref e) => write!(f, "Internal lightning_payjoin error: {}", e),
+			Self::Payjoin(ref e) => write!(f, "Payjoin error: {}", e),
+		}
+	}
+}
+
+// Maybe this should live in LDK-Node instead?
```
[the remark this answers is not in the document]
jbesraa: Oh sorry what I meant here is implement it to the ldk node `Error` rather than for `lightning-payjoin::Error` so we dont add here stuff that are specific for ldk node handling

lightning-payjoin/src/scheduler.rs
```diff
@@ -0,0 +1,378 @@
...
+		Self { channels: vec![] }
+	}
+	/// Schedule a new channel.
+	///
+	/// The channel will be created with `ScheduledChannelState::ChannelCreated` state.
+	pub fn schedule(
+		&mut self, channel_value_satoshi: u64, push_msat: Option<u64>, announce_channel: bool,
+		node_id: PublicKey, user_channel_id: u128, address: SocketAddress,
+	) {
+		let channel = ScheduledChannel::new(
```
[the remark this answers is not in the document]
jbesraa: It would, I thought in case of failure and a retry, we would want to have the data in hand but iam not sure its something we want to implement for now..

src/tx_broadcaster.rs
```diff
@@ -23,72 +25,90 @@ where
...
-						},
-						_ => {
-							log_error!(
-								self.logger,
-								"Failed to broadcast transaction {}: {}",
-								tx.txid(),
-								e
-							);
+				let mut channel_scheduler = self.channel_scheduler.lock().await;
+				if channel_scheduler.is_funding_tx_created(tx.txid()) {
```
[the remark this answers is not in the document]
jbesraa: so when the user calls `payjoin_open_channel` the channel is opened and stopped after `FundingGenerationReady` (https://github.com/lightningdevkit/ldk-node/pull/257/files#diff-14da7c350de5742eaa248de110fefeae9b4b639c0358e26024d6040ccf174a41R359) and the user gets back the bip21 they need to pay, nothing with payjoin until now. the channel now is already in `ChannelAccepted` phase and will be continued when the user pays the bip21, here we make sure we finish the communication with the counterparty node(ie get FundingSigned) before we return response(https://github.com/lightningdevkit/ldk-node/pull/257/files#diff-a5e41e49b8b7e8dd8ec88ed6b1c5ba0b530c2a186e37db14617223e34b11ca31R174) this is so we dont lock the funds in case we return a response the the payjoin sender without getting FundingSigned from the node we want to open channel with.

lightning-payjoin/Cargo.toml
```diff
@@ -11,4 +11,11 @@ version = "0.0.1"
 
 [dependencies]
 bitcoin = "0.30.2"
+bytes = "1.5.0"
+http-body-util = "0.1.0"
+hyper = {version = "1.2.0", features = ["http1", "server"]}
```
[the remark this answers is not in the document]
DanGould: v2 receivers are backwards compatible with v1 senders. I don't see a good reason to deploy a v1 receiver here.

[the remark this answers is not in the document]
jbesraa: Started new commit with initial implementation for payjoin v2..still need to replace the current v1. I want to first update the tests to work with v2 and that was a bit challenging with some sync/async in the test.   This change will give us more flexibility on where and how to execute the payjoin interaction.

jbesraa:
I am breaking this PR into smaller PRs. First PR regarding the payjoin sender is opened here https://github.com/lightningdevkit/ldk-node/pull/295, and ill create another two prs to integrate the receiver, one for regular payjoin receiving and another for receiving with opening a lightning channel. 

Ill leave it open for a bit longer as i am using it to check the e2e flow and how the CI is responding to changes, until I open the first payjoin receiver PR.

jbesraa:
closing in favor of #301

PR #260, opened 2024-02-22
Add `reason` to `Event::PaymentFailed` and `Event::ChannelClosed`

### LINKED ISSUE #247: Channel Closed "message"
for the `Event` of `channelClosed(channelId, userChannelId, counterpartyNodeId)` any thoughts on adding an additional `message` item in there? 

For instance i tried to open up a channel for 250k sats but saw in the logs that i needed a min 400k which had the message `invalid funding_amount=200000 sat (min=400000 sat)` which would be nice to bubble up to possibly display to a user as a reason they just saw a channel closed event in the UI when they tried opening the channel (so then they might retry opening the channel with the appropriate channel size above the min).

Just was wondering your thoughts on whether that was even a good idea to start or not-

### LINKED ISSUE #259: Expose `PaymentFailureReason` in `PaymentFailed`
.. we should do this.

tnull:
Fixes #259.
Fixes #247.

To provide some more insight why a payment failed, we expose the `PaymentFailureReason` in our event. We also expose the `reason` in `ChannelClosed`.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -855,7 +855,7 @@ where
 				counterparty_node_id,
 				..
 			} => {
-				log_info!(self.logger, "Channel {} closed due to: {:?}", channel_id, reason);
+				log_info!(self.logger, "Channel {} closed due to: {}", channel_id, reason);
 				self.event_queue
 					.add_event(Event::ChannelClosed {
 						channel_id,


src/event.rs
```diff
@@ -858,6 +861,7 @@ where
 						channel_id,
 						user_channel_id: UserChannelId(user_channel_id),
 						counterparty_node_id,
+						reason: Some(reason),
```
jkczyz, line 864: A few lines up we log `reason` with `Debug`. Should we use `Display` given it is implemented for `ClosureReason`?

jkczyz (commented):
LGTM. Please squash.

jkczyz (approved):
Any thoughts on including the reason in `PaymentStatus`?

PR #262, opened 2024-02-26
Bump max inflight HTLC value to 100% if we're an LSPS2 client

tnull:
When we're an LSPS2 client, the LSP might open the channel to us with a capacity only barely more than the initial payment size. We therefore need to be able to receive close to 100% of the channel size in a single HTLC.

Here we therefore bump the default max HTLC value inflight config to 100%. We should eventually be able to set this on a per-channel basis, but for now we just bump or default for all channels.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -685,7 +685,11 @@ fn build_with_store_internal<K: KVStore + Sync + Send + 'static>(
 		},
 	};
 
-	// Initialize the ChannelManager
+	// Initialize the default config values.
+	//
+	// Note that methods such as Node::connect_open_channel might override some of the values set
+	// here, e.g. the ChannelHandshakeConfig, meaning these default values will mostly be relevant
+	// for inbound channels.
 	let mut user_config = UserConfig::default();
 	user_config.channel_handshake_limits.force_announced_channel_preference = false;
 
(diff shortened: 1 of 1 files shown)


src/builder.rs
```diff
@@ -695,11 +699,19 @@ fn build_with_store_internal<K: KVStore + Sync + Send + 'static>(
...
 
-	if liquidity_source_config.is_some() {
+	if liquidity_source_config.and_then(|lsc| lsc.lsps2_service.as_ref()).is_some() {
 		// Generally allow claiming underpaying HTLCs as the LSP will skim off some fee. We'll
 		// check that they don't take too much before claiming.
 		user_config.channel_config.accept_underpaying_htlcs = true;
+
+		// FIXME: When we're an LSPS2 client, set maximum allowed inbound HTLC value in flight
+		// to 100%. We should eventually be able to set this on a per-channel basis, but for
+		// now we just bump or default for all channels.
```
jkczyz, line 709: s/or/the ?

Same in commit message and PR description.

jkczyz, line 709: Should we note somewhere in user docs that this affects non-JIT channels?

jkczyz (approved):

PR #266, opened 2024-03-04
Improve connection management

tnull:
Based on #243.
Based on #244.

As a pre-factor to #256, we clean up our connection handling. Namely, we move it to a new `connection.rs` file and clean up the logic a bit, allowing us to drop the dependency on the `futures` crate.

Then, we introduce a `ConnectionManager` that handles pending connection attempts. Previously, concurrent calls to
`do_connect_peer`/`connect_peer_if_necessary` could result in multiple connections being opened, just to be closed as redundant shortly after. Here, we fix this behavior by ensuring only one connection attempt is inflight at any given point in time.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -6,64 +6,142 @@ use lightning::ln::msgs::SocketAddress;
 
 use bitcoin::secp256k1::PublicKey;
 
+use std::collections::hash_map::{self, HashMap};
 use std::net::ToSocketAddrs;
 use std::ops::Deref;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
-pub(crate) async fn connect_peer_if_necessary<L: Deref + Clone + Sync + Send>(
-	node_id: PublicKey, addr: SocketAddress, peer_manager: Arc<PeerManager>, logger: L,
-) -> Result<(), Error>
+pub(crate) struct ConnectionManager<L: Deref + Clone + Sync + Send>
 where
 	L::Target: Logger,
 {
-	if peer_manager.peer_by_node_id(&node_id).is_some() {
-		return Ok(());
-	}
-
-	do_connect_peer(node_id, addr, peer_manager, logger).await
+	pending_connections:
+		Mutex<HashMap<PublicKey, Vec<tokio::sync::oneshot::Sender<Result<(), Error>>>>>,
+	peer_manager: Arc<PeerManager>,
+	logger: L,
 }
 
-pub(crate) async fn do_connect_peer<L: Deref + Clone + Sync + Send>(
-	node_id: PublicKey, addr: SocketAddress, peer_manager: Arc<PeerManager>, logger: L,
-) -> Result<(), Error>
+impl<L: Deref + Clone + Sync + Send> ConnectionManager<L>
 where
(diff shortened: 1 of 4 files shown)


tests/integration_tests_rust.rs
```diff
@@ -327,3 +327,31 @@ fn do_connection_restart_behavior(persist: bool) {
...
+	let (_bitcoind, electrsd) = setup_bitcoind_and_electrsd();
+	let (node_a, node_b) = setup_two_nodes(&electrsd, false);
+
+	let node_a = Arc::new(node_a);
+	let node_b = Arc::new(node_b);
+
+	let node_id_b = node_b.node_id();
+	let node_addr_b = node_b.listening_addresses().unwrap().first().unwrap().clone();
+
+	std::thread::sleep(std::time::Duration::from_secs(1));
```
jkczyz, line 342: Could you add a comment on why the `sleep` is needed? Is there any way to avoid it?

jkczyz (commented):
Commits specific to this PR LGTM. Didn't review the commits from based PRs.

Also, in commit message for 5b56a8ba3117f73d1291947e1faff844d5fb8e0c:

s/task is responsible/task responsible

tests/integration_tests_rust.rs
jkczyz: Could we have an `is_listening` method, even if only for testing?

src/lib.rs
```diff
@@ -1811,10 +1813,6 @@ impl Drop for Node {
 	}
 }
 
-<<<<<<< HEAD
-async fn connect_peer_if_necessary(
-	node_id: PublicKey, addr: SocketAddress, peer_manager: Arc<PeerManager>,
-=======
```
jkczyz: Looks like some merge conflicts snuck in somewhere up the commit chain.

src/connection.rs
```diff
@@ -8,62 +8,141 @@ use bitcoin::secp256k1::PublicKey;
...
+			.to_socket_addrs()
+			.map_err(|e| {
+				log_error!(self.logger, "Failed to resolve network address: {}", e);
+				self.propagate_result_to_subscribers(&node_id, Err(Error::InvalidSocketAddress));
+				Error::InvalidSocketAddress
+			})?
+			.next()
+			.ok_or_else(|| {
+				self.propagate_result_to_subscribers(&node_id, Err(Error::ConnectionFailed));
+				Error::ConnectionFailed
```
jkczyz, line 70: Should we log? Why not `Error::InvalidSocketAddress` here, too?

src/connection.rs
```diff
@@ -8,62 +8,141 @@ use bitcoin::secp256k1::PublicKey;
...
 where
 	L::Target: Logger,
 {
-	if peer_manager.peer_by_node_id(&node_id).is_some() {
-		return Ok(());
-	}
-
-	do_connect_peer(node_id, addr, peer_manager, logger).await
+	pending_connections:
+		Mutex<Vec<(PublicKey, Vec<tokio::sync::oneshot::Sender<Result<(), Error>>>)>>,
```
jkczyz, line 19: Should we use as a `HashMap`?

jkczyz (approved):

Issue #193, opened 2023-11-15
Add BOLT12 support


tnull:
LDK 0.0.118 introduced BOLT 12 support. LDK Node should support sending and receiving via BOLT12, too.

PR #270, opened 2024-03-05
Refactor and modularize payment APIs

tnull:
Based on #243.
Based on #244.
Based on #266.

Another prefactor to #193, splitting this out to reduce rebase pain, e.g., with #141.

As we're about to introduce a bunch more payment API methods for BOLT12 support, we first clean up and declutter our payment APIs so far. 

What previously were individual methods on `Node` we now move to individual `PaymentHandler`s:

- `Bolt11PaymentHandler`
- `SpontaneousPaymentHandler`
- `OnchainPaymentHandler`

While we're at it, we slightly remove the API methods to shorten them, allowing to call them like so:

```rust
	let invoice = Bolt11Invoice::from_str("INVOICE_STR").unwrap();
	node.bolt11_payment().send(&invoice).unwrap();
```

Additionally, we move `PaymentStore` to a sub-module of `crate::payment` to clean up our top-level docs a bit.

### DIFF
--- a/bindings/ldk_node.udl
+++ b/bindings/ldk_node.udl
@@ -186,9 +186,9 @@ enum BuildError {
 
 [Enum]
 interface Event {
-	PaymentSuccessful(PaymentHash payment_hash, u64? fee_paid_msat);
-	PaymentFailed(PaymentHash payment_hash, PaymentFailureReason? reason);
-	PaymentReceived(PaymentHash payment_hash, u64 amount_msat);
+	PaymentSuccessful(PaymentId? payment_id, PaymentHash payment_hash, u64? fee_paid_msat);
+	PaymentFailed(PaymentId? payment_id, PaymentHash payment_hash, PaymentFailureReason? reason);
+	PaymentReceived(PaymentId? payment_id, PaymentHash payment_hash, u64 amount_msat);
 	ChannelPending(ChannelId channel_id, UserChannelId user_channel_id, ChannelId former_temporary_channel_id, PublicKey counterparty_node_id, OutPoint funding_txo);
 	ChannelReady(ChannelId channel_id, UserChannelId user_channel_id, PublicKey? counterparty_node_id);
 	ChannelClosed(ChannelId channel_id, UserChannelId user_channel_id, PublicKey? counterparty_node_id, ClosureReason? reason);
--- a/src/event.rs
+++ b/src/event.rs
@@ -46,6 +46,10 @@ use std::time::Duration;
 pub enum Event {
 	/// A sent payment was successful.
 	PaymentSuccessful {
+		/// A local identifier used to track the payment.
+		///
+		/// Will only be `None` for events serialized with LDK Node v0.2.1 or prior.
+		payment_id: Option<PaymentId>,
 		/// The hash of the payment.
 		payment_hash: PaymentHash,
 		/// The total fee which was spent at intermediate hops in this payment.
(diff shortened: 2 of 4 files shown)


jkczyz:
> I'm not sure: the API surface of covering all possible payment variations is rather large already, we're about to add substantially to it with BOLT12, and we eventually will also need to add, e.g., BOLT12 JIT variations. I think we eventually also want to provide helpers to generate and pay more unified interface (e.g., unified QR code, DNS payment addresses), but I'm not seeing how we can avoid the API getting a bit unwieldy if all variations keep living on a single object?

Maybe `Bolt11Protocol`? Though I'm a bit more inclined to use `Bolt11Payment` to align with the others. If you have the methods take `self` by value, then it really would be only one payment. Not sure how nice that works with bindings, however.

Another idea would be to divide by inbound and outbound instead of by protocol / mechanism.

jkczyz:
> You mean the `Node::X_payment` methods? No, for one we of course can't consume `Node`, but in bindings we can only return `Arc`s and never use `self`/`&mut self`. I'll however see to store/cache the corresponding `Arc`s in `Node` itself to reduce the number of clones on each call. I might also explore whether we can return a reference rather than an `Arc` in Rust, although I'm not entirely sure if that makes much of a difference.

Not the methods on `Node` but rather the methods on `Bolt11Payment`. But, right, I didn't notice that `bolt11_payment` returns an `Arc`. Why can't it return by value?

src/payment/payment_store.rs
```diff
@@ -82,6 +137,73 @@ impl_writeable_tlv_based_enum!(PaymentStatus,
 	(4, Failed) => {};
 );
 
+/// Represents the kind of a payment.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum PaymentKind {
```
jkczyz, line 142: Should we name this `PaymentPurpose` for a closer mapping to LDK?

src/payment/bolt11.rs
```diff
@@ -102,32 +102,39 @@ impl Bolt11Payment {
 				let amt_msat = invoice.amount_milli_satoshis().unwrap();
 				log_info!(self.logger, "Initiated sending {}msat to {}", amt_msat, payee_pubkey);
 
-				let payment = PaymentDetails {
-					preimage: None,
+				let kind = PaymentKind::Bolt11 {
 					hash: payment_hash,
+					preimage: None,
 					secret: payment_secret,
+				};
```
jkczyz: nit: can inline this

src/payment/payment_store.rs
```diff
@@ -142,22 +255,22 @@ where
...
 	}
 
 	pub(crate) fn insert(&self, payment: PaymentDetails) -> Result<bool, Error> {
 		let mut locked_payments = self.payments.lock().unwrap();
 
-		let hash = payment.hash.clone();
-		let updated = locked_payments.insert(hash.clone(), payment.clone()).is_some();
-		self.persist_info(&hash, &payment)?;
+		let id = payment.id.clone();
+		let updated = locked_payments.insert(id.clone(), payment.clone()).is_some();
```
jkczyz, line 267: nit: no need for `id` clones since it should be `Copy`

src/payment/payment_store.rs
```diff
@@ -142,22 +255,22 @@ where
...
+		self.persist_info(&id, &payment)?;
```
jkczyz, line 268: Can reference the original id here from `PaymentDetails`.

jkczyz (commented):
No major concerns left.

src/payment/payment_store.rs
jkczyz: No strong opinion. Though I wonder if "kind" is just another way of saying "type". I'm a little more inclined with using the latter, but happy to defer to you here.

jkczyz (commented):
LGTM. Please squash

Issue #255, opened 2024-02-21
Expose fees paid for outbound payments


tnull:
For outbound payments, we should expose the fees paid in `Event::PaymentSuccessful`, and possibly also in `PaymentDetails`, the latter would go hand-in-hand with the BOLT12 integration as it already will touch the data model there.

PR #271, opened 2024-03-05
Add `fee_paid_msat` to `PaymentSuccessful` event

jbesraa:
Resolves the first part of #255 adding the paid fee to the `PaymentSuccessful` event.

### DIFF
--- a/tests/common.rs
+++ b/tests/common.rs
@@ -80,6 +80,43 @@ macro_rules! expect_channel_ready_event {
 
 pub(crate) use expect_channel_ready_event;
 
+macro_rules! expect_payment_received_event {
+	($node: expr, $amount_msat: expr) => {{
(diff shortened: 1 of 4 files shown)


tests/common.rs
```diff
@@ -80,6 +80,43 @@ macro_rules! expect_channel_ready_event {
 
 pub(crate) use expect_channel_ready_event;
 
+macro_rules! expect_payment_received_event {
```
jbesraa, line 83: @tnull I added the following macros following the pattern in the file. for some reason, the macros I added are marked as "unused" although they are used in the multihop test. what am i missing?
Also, should I split the commit for adding those two, or its ok to be in the same commit?

jbesraa: "unused_macros: unused macro definition: `expect_payment_successful_event`"

PR #272, opened 2024-03-06
Introduce `status` method allowing to query the `Node`'s status 

### LINKED ISSUE #127: Expose Node status and statistics
We should provide a utility that returns some more information on the node status and statistics overall.

One example would be to which block height the node is currently synced.

tnull:
Fixes #127.

### DIFF
--- a/tests/integration_tests_rust.rs
+++ b/tests/integration_tests_rust.rs
@@ -289,7 +289,11 @@ fn do_connection_restart_behavior(persist: bool) {
 	let node_id_b = node_b.node_id();
 
 	let node_addr_b = node_b.listening_addresses().unwrap().first().unwrap().clone();
-	std::thread::sleep(std::time::Duration::from_secs(1));
+
+	while !node_b.status().is_listening {
+		std::thread::sleep(std::time::Duration::from_millis(10));
+	}
+
 	node_a.connect(node_id_b, node_addr_b, persist).unwrap();
 
 	let peer_details_a = node_a.list_peers().first().unwrap().clone();


src/types.rs
```diff
@@ -456,3 +457,20 @@ impl Default for ChannelConfig {
 		LdkChannelConfig::default().into()
 	}
 }
+
+/// The best known block as identified by its hash and height.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BestBlock {
```
jkczyz, line 463: Is this only needed because LDK's struct doesn't derive `Debug`?

src/lib.rs
```diff
@@ -1746,6 +1812,43 @@ impl<K: KVStore + Sync + Send + 'static> Drop for Node<K> {
...
+/// Represents the status of the [`Node`].
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct NodeStatus {
+	/// Indicates whether the [`Node`] is running.
+	pub is_running: bool,
+	/// Indicates whether the [`Node`] is listening for incoming connections on the addresses
+	/// configured via [`Config::listening_addresses`].
+	pub is_listening: bool,
+	/// The best block to which our Lightning wallet is currently synced.
+	pub current_best_block: BestBlock,
```
jkczyz, line 1829: Should we have something analogous for the on-chain wallet?

src/lib.rs
```diff
@@ -1746,6 +1812,43 @@ impl<K: KVStore + Sync + Send + 'static> Drop for Node<K> {
...
+	/// The timestamp, in seconds since start of the UNIX epoch, when the last rapid gossip sync
+	/// (RGS) snapshot we successfully applied was generated.
+	///
+	/// Will be `None` if RGS isn't configured or the snapshot hasn't been updated since the [`Node`] was initialized.
+	pub latest_rgs_snapshot_timestamp: Option<u64>,
+	/// The timestamp, in seconds since start of the UNIX epoch, when we last broadcasted a node
+	/// announcement.
+	///
+	/// Will be `None` if we have no public channels or we haven't broadcasted since the [`Node`] was initialized.
+	pub latest_node_announcement_broadcast_timestamp: Option<u64>,
```
jkczyz, line 1854: Would it make sense to use `Duration` for these?

jkczyz: Hmm... `Duration` is a timestamp when interpreted as relative to the Unix epoch. It just gives you more precision than `u64` seconds and may be more useful interface to work with. But fair enough regarding bindings.

jkczyz (approved):
LGTM. Seems like you'll need a rebase.

jkczyz (approved):

PR #273, opened 2024-03-06
Proposal: implement custom TLV records for spontaneous payments

rdmitr:
A very simple implementation of custom TLV records for keysend requests.


src/types.rs
```diff
@@ -456,3 +456,12 @@ impl Default for ChannelConfig {
 		LdkChannelConfig::default().into()
 	}
 }
+
+/// Custom TLV entry.
+pub struct TlvEntry {
```
jbesraa, line 461: In the protocol the TLV entry is defined with a length
https://github.com/lightning/bolts/blob/master/01-messaging.md#type-length-value-format

Whats the reason of not including it here?

jbesraa: I see that LDK is expecting type and value to set the tlv. I guess then the length is set by ldk when sending the message.
Maybe add a bit more info on the type and value describing the restrictions or characterizes of the props:
for example, for type:
`type identifiers below 2^16 are reserved for use in this specification. type identifiers greater than or equal to 2^16 are available for custom records.` (from the blip)

tests/common.rs
```diff
@@ -490,8 +490,11 @@ pub(crate) fn do_channel_full_cycle<K: KVStore + Sync + Send, E: ElectrumApi>(
 	// Test spontaneous/keysend payments
 	println!("\nA send_spontaneous_payment");
 	let keysend_amount_msat = 2500_000;
-	let keysend_payment_hash =
-		node_a.send_spontaneous_payment(keysend_amount_msat, node_b.node_id()).unwrap();
+	let tlv1 = TlvEntry { r#type: 131073, value: vec![0x00, 0x11, 0x22, 0x33] };
```
jbesraa, line 493: does this test validate that the counter-party node received the message?

src/lib.rs
```diff
@@ -1222,7 +1222,7 @@ impl<K: KVStore + Sync + Send + 'static> Node<K> {
 
 	/// Send a spontaneous, aka. "keysend", payment
 	pub fn send_spontaneous_payment(
-		&self, amount_msat: u64, node_id: PublicKey,
+		&self, amount_msat: u64, node_id: PublicKey, custom_tlvs: Vec<TlvEntry>,
```
jbesraa, line 1225: are the `custom_tlvs` required in order to send a spontaneous payment?

rolznz:
> Is there a use for this people want in the immediate future aside from the messages thing? Is the podcast streaming sats thing using the messages field or are there other fields it uses?

Yes, the main usecase is Podcasting 2.0: https://github.com/lightning/blips/blob/master/blip-0010.md

src/types.rs
rdmitr: > In the protocol the TLV entry is defined with a length https://github.com/lightning/bolts/blob/master/01-messaging.md#type-length-value-format
> 
> Whats the reason of not including it here?

The length of the data buffer is used for that

src/lib.rs
rdmitr: Not really required but may be useful. But I see the point, we'll probably take tnull's suggestion to base this change off PR 270; probably once it's merged

jbesraa:
should the `TlvEntry` be part of `LDK`? I think it would be nice to have there, and also change the function input 
`pub fn with_custom_tlvs(mut self, mut custom_tlvs: Vec<TlvEntry>) -> Result<Self, ()> {...}`. Happy to open an issue and resolve it if so

rolznz:
@tnull unfortunately the custom TLV records are needed (that's how the podcasting 2.0 data is passed)

rdmitr:
I screwed up merges on this one, and it's getting quite hairy; I will create a new PR and address all the comments.

PR #278, opened 2024-03-14
Return Peer's announcement info if available

benthecarman:


benthecarman:
Yeah the main thing we were going to use this for was getting the alias, maybe that'd be better to add here instead

PR #281, opened 2024-04-12
Improve on (potentially) blocking wallet behaviors

tnull:
~~Based on #141.~~

Due to our onchain wallet requiring us to hold a `Mutex` for the duration of the chain sync, many operations might be blocked for seconds at at time. 

Here, we make several changes that hopefully alleviate any blocking issues from holding the wallet lock(s). In particular we introduce a balance cache, drop our 'immediate retry sync' logic and drop the previous `wallet_lock` `Condvar` in favor of a pub/sub pattern that is beneficial in async runtime environments.


jbesraa (commented):
I tested this branch locally with an app with GUI that previously lagged when the wallet synced and I can confirm that with this PR things are much smoother

Issue #177, opened 2023-10-28
Explore integration of Payjoin/PDK


moneyball:
There is an opportunity to use Payjoin to make deposits into LDK Node use 1 txn instead of 2 when an on-chain txn funds a LN wallet/channel. Mutiny is doing this. We should explore it on LDK Node (probably medium priority behind LSP client, VSS, Electrum support).

DanGould:
[Payjoin V2](https://github.com/bitcoin/bips/pull/1483) in by PDK eliminates the need for tor #178

DanGould:
It's been released as a usable alpha with feature flag `v2`

DanGould:
v1 an v2 are backwards compatible so you should be able to use the instructions in [`send`](https://docs.rs/payjoin/latest/payjoin/send/index.html) and [`receive`](https://docs.rs/payjoin/latest/payjoin/receive/index.html) modules and follow the [payjoin-cli](https://github.com/DanGould/rust-payjoin/tree/master/payjoin-cli) reference implementation's `v2` feature codepath.

The `v2` module ought to be private in the next update.

PR #301, opened 2024-05-28
Allow node wallet to receive payjoin transactions

jbesraa:
fully addresses #177 
blocked by https://github.com/lightningdevkit/rust-lightning/pull/3024

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1017,8 +1017,6 @@ fn build_with_store_internal(
 	let mut payjoin_receiver = None;
 	if let Some(pj_config) = payjoin_config {
 		payjoin_handler = Some(Arc::new(PayjoinHandler::new(
-			Arc::clone(&tx_broadcaster),
-			Arc::clone(&logger),
 			pj_config.payjoin_relay.clone(),
 			Arc::clone(&tx_sync),
 			Arc::clone(&event_queue),
--- a/src/event.rs
+++ b/src/event.rs
@@ -188,6 +188,7 @@ pub enum PayjoinPaymentFailureReason {
 	Timeout,
 	TransactionFinalisationFailed,
 	InvalidReceiverResponse,
+	RequestFailed,
 }
 
 impl Readable for PayjoinPaymentFailureReason {
@@ -196,6 +197,7 @@ impl Readable for PayjoinPaymentFailureReason {
 			0 => Ok(Self::Timeout),
 			1 => Ok(Self::TransactionFinalisationFailed),
 			2 => Ok(Self::InvalidReceiverResponse),
+			3 => Ok(Self::RequestFailed),
 			_ => Err(DecodeError::InvalidValue),
 		}
 	}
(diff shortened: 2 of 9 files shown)


jbesraa:
Currently blocked by the failing CI https://github.com/lightningdevkit/ldk-node/actions/runs/9287478277/job/25556603253?pr=301, a tracking issue on rust bitcoin is already open here https://github.com/rust-bitcoin/rust-bitcoinconsensus/issues/32. 
@tnull I wonder if this is a deal breaker or we can introduce the `bitcoinconsensus` feature optionally for non-uniffi which mean the payjoin receiver wouldnt be available for bindings

jbesraa:
> > @tnull I wonder if this is a deal breaker or we can introduce the `bitcoinconsensus` feature optionally for non-uniffi which mean the payjoin receiver wouldnt be available for bindings
> 
> Hmm, unfortunately I think it's a bit of a deal breaker as we really want to keep the bindings and Rust as congruent as possible. Especially so as, in contrast to 'receive-and-open-channel', payjoin send/receive would be expected to be a client-side feature that mobile users might benefit from?
> 
> I mean if it really blocks your progress we could consider landing the code behind a `cfg(payjoin)` flag so it's already in the codebase and can be used, e.g., in tests. But I'd be hesitant to fully expose it in Rust and _not_ expose it in the bindings at all.
> 
> Two potential ways to go about it:
> 
> a) Push for a fix in `bitcoinconsensus`. b) Manually implement the necessary consensus checks, if they are not too many?

yup, working on option `a`.  Got in touch with BitcoinZavior to get some input from him about bindings, otherwise I am researching/experimenting with solutions.

Now that https://github.com/lightningdevkit/rust-lightning/pull/3024 is ready for reivew, this `bitcoinconsensus` binding issue is the only blocker for both regular payjoin receiver and receive-open-channel flows.

jbesraa:
@tnull Apparently I dont need the `bitcoinconsensus` feature after all.  

Previously I was checking if the transaction sent by the sender is broadcast-able, but the payjoin protocol requires that only if payjoin used in a non-interactive receiver, as mentioned here https://github.com/bitcoin/bips/blob/master/bip-0078.mediawiki#receivers-original-psbt-checklist. So the binding issue is longer blocking this.

jbesraa:
integration test already added here https://github.com/lightningdevkit/ldk-node/pull/301/files#diff-ebba69f49eedcd3dd6a656eff964c6184b6d29c02fbad6a49d3b41c87cc3de79

I am very close to open the third PR (payjoin->channel opening), and based on the work I have done there I expect a couple of minor changes here. Will let you know when its RFR.

jbesraa:
> @jbesraa We now upgraded LDK to 0.0.124/0.0.125 which shipped your 'unsafe funding' work, which unblocks this PR I think?
> 
> Do you still intend to finish these Payjoin PRs? Has there been any update on the `rust-payjoin` side w.r.t to the identifiers etc. we previously discussed?

Yup, I do. 

I see the `rust-payjoin` issue is still open https://github.com/payjoin/rust-payjoin/issues/336. How do you suggest approaching this? should we think about internal solution? or try to submit a pr to `rust-payjoin` ?

PR #303, opened 2024-05-30
Bump max inflight HTLC value for outbound private channels

tnull:
Based on #141.

- Firstly, we consolidate setting the default values of `UserConfig`. 

- Secondly, we override `max_inbound_htlc_value_in_flight_percent_of_channel` for private outbound channels to be 100% as users keep beeing confused why they aren't able to send the full channel value. LDK will likely default to this behavior soon anyways, at which point we'll be able to drop this manual override again.


valentinewallace (approved):
LGTM 👌

PR #307, opened 2024-06-11
Periodically archive fully resolved channel monitors

### LINKED ISSUE #300: Archive old `ChannelMonitor`s
LDK 0.0.123 added [`archive_fully_resolved_channel_monitors`](https://docs.rs/lightning/latest/lightning/chain/chainmonitor/struct.ChainMonitor.html#method.archive_fully_resolved_channel_monitors), we should call it occasionally.

tnull:
Fixes #300.

Previously, LDK was very conservative and kept channel monitors around ~forever or until the user manually decided to prune them. Recently it introduced the `ChainMonitor::archive_fully_resolved_monitors` method, which we now call periodically: every time a wallet sync succeeds, we check whether the latest archival height is 6 blocks in the past and
call `archive_fully_resolved_monitors`.

As this is not permanently persisted, we will always try to archive any pruned monitors when the first background sync after fresh initialization succeeds, ensuring we call it regularly also on
short-lived sessions, e.g, on mobile.

In the second and third commit, we fix previous omissions of `sync_wallets`: we now update the timestamps read by `NodeStatus` and also have it update the fee rate cache.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1124,7 +1124,8 @@ impl Node {
 		}
 	}
 
-	/// Manually sync the LDK and BDK wallets with the current chain state.
+	/// Manually sync the LDK and BDK wallets with the current chain state and update the fee rate
+	/// cache.
 	///
 	/// **Note:** The wallets are regularly synced in the background, which is configurable via
 	/// [`Config::onchain_wallet_sync_interval_secs`] and [`Config::wallet_sync_interval_secs`].
@@ -1142,6 +1143,7 @@ impl Node {
 		let archive_cman = Arc::clone(&self.channel_manager);
 		let sync_cmon = Arc::clone(&self.chain_monitor);
 		let archive_cmon = Arc::clone(&self.chain_monitor);
+		let fee_estimator = Arc::clone(&self.fee_estimator);
 		let sync_sweeper = Arc::clone(&self.output_sweeper);
 		let sync_logger = Arc::clone(&self.logger);
 		let confirmables = vec![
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -29,6 +29,9 @@ pub(crate) const DEFAULT_ESPLORA_SERVER_URL: &str = "https://blockstream.info/ap
 // The timeout after which we abandon retrying failed payments.
 pub(crate) const LDK_PAYMENT_RETRY_TIMEOUT: Duration = Duration::from_secs(10);
 
+// The interval (in block height) after which we archive pruned channel monitors.
+pub(crate) const LDK_CHANNEL_MONITOR_ARCHIVAL_INTERVAL: u32 = 6;
```
jbesraa, line 33: `chain_monitor::archive_fully_resolved_channel_monitors` needs `4032` blocks before archiving so not sure the docs are accurate.

jbesraa (approved):

src/config.rs
jbesraa: Yup, looks good

src/lib.rs
```diff
@@ -1486,3 +1543,19 @@ pub(crate) fn total_anchor_channels_reserve_sats(
...
+
+fn periodically_archive_fully_resolved_monitors(
+	channel_manager: Arc<ChannelManager>, chain_monitor: Arc<ChainMonitor>,
+	latest_channel_monitor_archival_height: Arc<RwLock<Option<u32>>>,
+) {
+	let mut latest_archival_height_lock = latest_channel_monitor_archival_height.write().unwrap();
+	let cur_height = channel_manager.current_best_block().height;
+	let should_archive = latest_archival_height_lock
+		.as_ref()
+		.map_or(true, |h| cur_height >= h + RESOLVED_CHANNEL_MONITOR_ARCHIVAL_INTERVAL);
```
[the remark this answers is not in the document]
jbesraa: Another benefit of running this more frequently is that `archive_..` will be counting blocks for the specific monitor earlier. i.e, `archive_..` will only start counting the two weeks time blocks if you call it.

PR #308, opened 2024-06-14
Allow to manually register and claim payments

### LINKED ISSUE #57: Consider adding `last_updated` field to `PaymentDetails`
This would allow the user to query for 'recent' payments based on a suitable filter.

In order to do this nicely we should however consider implementing `Writeable`/`Readable` for `SystemTime` upstream.

### LINKED ISSUE #280: Allow to manually register and claim payments
We should allow users to register interest in payments for which they got the preimage out-of-bounds. We'd then issue a `PaymentClaimable` event similar to the one in LDK and would allow them to manually claim such payments.

Things to figure out:

1) Should we keep a list of payment hashes for which we always issue `PaymentClaimable`, or issue `PaymentClaimable` if we can't claim with our default logic already
2) Can we avoid `PaymentClaimable` being issued if users don't expect it? Should we have a config flag? Or could we even always issue it but auto-fail the payment if it's issued and the user calls `event_handled` without claiming the payment?
3) If/how we want to track them in the payment store

tnull:
Fixes #280.
Fixes #57.

Previously, we wouldn't allow invoice generation for third-party payment hashes. However, as some users require this feature, we here allow them to generate invoices for which they supply the payment hashes, allowing them to retrieve the preimage out-of-bound. 

To this end, we add `receive_for_hash` and `receive_variable_amount_for_hash` variants to the `Bolt11Payment` handler. Then, when we receive the payment, we emit a new `PaymentClaimable` event that users need to call `claim_manually` for with the correct preimage. Note that this currently only implements the BOLT11 flow as for BOLT12 additions to LDK will be necessary that are pending the next release.

We also add a commit resolving #57, i.e., adding a `latest_update_timestamp` field to `PaymentDetails`  which allows to filter and sort payment based on how recent they are.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -598,18 +598,20 @@ where
 						..
 					} => {
 						let offer_id = payment_context.offer_id;
-						let payment = PaymentDetails {
-							id: payment_id,
-							kind: PaymentKind::Bolt12Offer {
-								hash: Some(payment_hash),
-								preimage: payment_preimage,
-								secret: Some(payment_secret),
-								offer_id,
-							},
-							amount_msat: Some(amount_msat),
-							direction: PaymentDirection::Inbound,
-							status: PaymentStatus::Pending,
+						let kind = PaymentKind::Bolt12Offer {
+							hash: Some(payment_hash),
+							preimage: payment_preimage,
+							secret: Some(payment_secret),
+							offer_id,
 						};
+
+						let payment = PaymentDetails::new(
+							payment_id,
+							kind,
+							Some(amount_msat),
+							PaymentDirection::Inbound,
+							PaymentStatus::Pending,
+						);
 
 						match self.payment_store.insert(payment) {
 							Ok(false) => (),
@@ -638,16 +640,18 @@ where
 					},
 					PaymentPurpose::SpontaneousPayment(preimage) => {
 						// Since it's spontaneous, we insert it now into our store.
-						let payment = PaymentDetails {
-							id: payment_id,
-							kind: PaymentKind::Spontaneous {
-								hash: payment_hash,
-								preimage: Some(preimage),
-							},
-							amount_msat: Some(amount_msat),
-							direction: PaymentDirection::Inbound,
-							status: PaymentStatus::Pending,
+						let kind = PaymentKind::Spontaneous {
+							hash: payment_hash,
+							preimage: Some(preimage),
 						};
+
+						let payment = PaymentDetails::new(
+							payment_id,
+							kind,
+							Some(amount_msat),
+							PaymentDirection::Inbound,
+							PaymentStatus::Pending,
+						);
 
 						match self.payment_store.insert(payment) {
 							Ok(false) => (),
--- a/src/payment/bolt11.rs
+++ b/src/payment/bolt11.rs
@@ -105,17 +105,18 @@ impl Bolt11Payment {
 				let amt_msat = invoice.amount_milli_satoshis().unwrap();
 				log_info!(self.logger, "Initiated sending {}msat to {}", amt_msat, payee_pubkey);
 
-				let payment = PaymentDetails {
-					id: payment_id,
-					kind: PaymentKind::Bolt11 {
-						hash: payment_hash,
-						preimage: None,
-						secret: payment_secret,
-					},
-					amount_msat: invoice.amount_milli_satoshis(),
-					direction: PaymentDirection::Outbound,
-					status: PaymentStatus::Pending,
+				let kind = PaymentKind::Bolt11 {
+					hash: payment_hash,
+					preimage: None,
+					secret: payment_secret,
 				};
+				let payment = PaymentDetails::new(
+					payment_id,
+					kind,
+					invoice.amount_milli_satoshis(),
+					PaymentDirection::Outbound,
+					PaymentStatus::Pending,
+				);
 
 				self.payment_store.insert(payment)?;
 
@@ -126,17 +127,18 @@ impl Bolt11Payment {
 				match e {
 					RetryableSendFailure::DuplicatePayment => Err(Error::DuplicatePayment),
 					_ => {
-						let payment = PaymentDetails {
-							id: payment_id,
-							kind: PaymentKind::Bolt11 {
-								hash: payment_hash,
-								preimage: None,
-								secret: payment_secret,
-							},
-							amount_msat: invoice.amount_milli_satoshis(),
-							direction: PaymentDirection::Outbound,
-							status: PaymentStatus::Failed,
+						let kind = PaymentKind::Bolt11 {
+							hash: payment_hash,
+							preimage: None,
+							secret: payment_secret,
 						};
+						let payment = PaymentDetails::new(
+							payment_id,
+							kind,
+							invoice.amount_milli_satoshis(),
+							PaymentDirection::Outbound,
+							PaymentStatus::Failed,
+						);
 
 						self.payment_store.insert(payment)?;
 						Err(Error::PaymentSendingFailed)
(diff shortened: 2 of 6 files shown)


src/event.rs
```diff
@@ -594,19 +594,21 @@ where
...
+							offer_id,
 						};
 
+						let payment = PaymentDetails::new(
+							payment_id,
+							kind,
+							Some(amount_msat),
+							PaymentDirection::Inbound,
+							PaymentStatus::Pending,
+						);
```
jkczyz, line 610: Any reason the nesting was reverted? Seems unrelated to the commit.

src/payment/bolt11.rs
```diff
@@ -254,62 +261,183 @@ impl Bolt11Payment {
...
+	/// Will check that the payment is know, and that the given preimage and claimable amount match
+	/// our expectations before attempting to claim the payment.
+	///
+	/// Upon success, a [`PaymentReceived`] event will be emitted.
+	///
+	/// [`receive_for_hash`]: Self::receive_for_hash
+	/// [`receive_variable_amount_for_hash`]: Self::receive_variable_amount_for_hash
+	/// [`PaymentClaimable`]: crate::Event::PaymentClaimable
+	/// [`PaymentReceived`]: crate::Event::PaymentReceived
+	pub fn claim_manually(
```
tvolk131, line 278: We should add a way to quickly fail the payment if the preimage cannot be obtained, rather than having to wait for invoice expiration.

src/event.rs
```diff
@@ -464,6 +464,25 @@ where
 			} => {
 				let payment_id = PaymentId(payment_hash.0);
 				if let Some(info) = self.payment_store.get(&payment_id) {
+					if info.direction == PaymentDirection::Outbound {
+						log_info!(
+							self.logger,
+							"Refused inbound payment with ID {}: circular payments are unsupported.",
```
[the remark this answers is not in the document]
tvolk131: Update: I talked with @joschisan about this, and it sounds like this short-circuiting happens in the lightning gateway and doesn't actually reach the underlying lightning implementation. I believe in the case I described above, the lightning implementation would view that the invoice never got paid. So this shouldn't be an issue.

tests/common/mod.rs
```diff
@@ -580,6 +608,43 @@ pub(crate) fn do_channel_full_cycle<E: ElectrumApi>(
 	assert_eq!(node_b.payment(&payment_id).unwrap().amount_msat, Some(determined_amount_msat));
 	assert!(matches!(node_b.payment(&payment_id).unwrap().kind, PaymentKind::Bolt11 { .. }));
 
+	// Test manually registered/claimed payments.
+	let invoice_amount_3_msat = 5532_000;
```
tvolk131, line 612: Nit: `5_532_000`

src/payment/store.rs
```diff
@@ -58,12 +75,17 @@ impl Writeable for PaymentDetails {
 
 impl Readable for PaymentDetails {
 	fn read<R: lightning::io::Read>(reader: &mut R) -> Result<PaymentDetails, DecodeError> {
+		let unix_time_secs = SystemTime::now()
+			.duration_since(UNIX_EPOCH)
+			.unwrap_or(Duration::from_secs(0))
```
tvolk131, line 80: Nit: `.unwrap_or_default()`

src/payment/store.rs
```diff
@@ -391,6 +415,11 @@ where
 				payment.status = status;
 			}
 
+			payment.latest_update_timestamp = SystemTime::now()
+				.duration_since(UNIX_EPOCH)
+				.unwrap_or(Duration::from_secs(0))
```
tvolk131, line 420: Nit: `.unwrap_or_default()`

src/payment/store.rs
```diff
@@ -34,6 +35,21 @@ pub struct PaymentDetails {
...
+}
+
+impl PaymentDetails {
+	pub(crate) fn new(
+		id: PaymentId, kind: PaymentKind, amount_msat: Option<u64>, direction: PaymentDirection,
+		status: PaymentStatus,
+	) -> Self {
+		let latest_update_timestamp = SystemTime::now()
+			.duration_since(UNIX_EPOCH)
+			.unwrap_or(Duration::from_secs(0))
```
tvolk131, line 49: Nit: `.unwrap_or_default()`

src/payment/bolt11.rs
```diff
@@ -254,62 +262,224 @@ impl Bolt11Payment {
...
+	///
+	/// We will register the given payment hash and emit a [`PaymentClaimable`] event once
+	/// the inbound payment arrives.
+	///
+	/// **Note:** users *MUST* handle this event and claim the payment manually via
+	/// [`claim_for_hash`] as soon as they have obtained access to the preimage of the given
+	/// payment hash.
+	///
+	/// [`PaymentClaimable`]: crate::Event::PaymentClaimable
+	/// [`claim_for_hash`]: Self::claim_for_hash
```
tvolk131, line 376: Do we want to also mention `fail_for_hash()` here?

src/payment/bolt11.rs
```diff
@@ -254,62 +262,224 @@ impl Bolt11Payment {
...
```
tvolk131, line 404: Do we want to also mention `fail_for_hash()` here?

tests/common/mod.rs
```diff
@@ -580,6 +608,43 @@ pub(crate) fn do_channel_full_cycle<E: ElectrumApi>(
...
+
+	let claimable_amount_msat = expect_payment_claimable_event!(
+		node_b,
+		manual_payment_id,
+		manual_payment_hash,
+		invoice_amount_3_msat
+	);
+	node_b
+		.bolt11_payment()
+		.claim_for_hash(manual_payment_hash, claimable_amount_msat, manual_preimage)
```
tvolk131, line 629: Is it worth testing another payment and calling `fail_for_hash()`?

tvolk131: Fair enough. Sounds good!

tvolk131 (approved):

src/event.rs
jkczyz: Mostly difficult to review as it doesn't seem relevant to the commit. Though I wasn't sure if this was a rebase issue since a prior PR used the nested variation.

src/payment/bolt11.rs
```diff
@@ -310,6 +313,40 @@ impl Bolt11Payment {
...
+	/// failed back, e.g., if the correct pre-image can't be retrieved in time before the claim
+	/// deadline has been reached.
+	///
+	/// Will check that the payment is known before failing the payment, and will return an error
+	/// otherwise.
+	///
+	/// [`receive_for_hash`]: Self::receive_for_hash
+	/// [`receive_variable_amount_for_hash`]: Self::receive_variable_amount_for_hash
+	/// [`PaymentClaimable`]: crate::Event::PaymentClaimable
+	pub fn fail_for_hash(&self, payment_hash: PaymentHash) -> Result<(), Error> {
```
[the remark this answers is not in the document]
jkczyz: Gotcha. Fine with keeping this as is.

jkczyz (commented):
Go ahead and squash. I'll take a final look in the morning.

tests/common/mod.rs
jkczyz: It would test that the storage status is updated from `Pending` to `Failed`, though.

src/payment/bolt11.rs
```diff
@@ -254,62 +262,233 @@ impl Bolt11Payment {
...
+	/// otherwise.
+	///
+	/// When claiming the payment has succeeded, a [`PaymentReceived`] event will be emitted.
+	///
+	/// [`receive_for_hash`]: Self::receive_for_hash
+	/// [`receive_variable_amount_for_hash`]: Self::receive_variable_amount_for_hash
+	/// [`PaymentClaimable`]: crate::Event::PaymentClaimable
+	/// [`PaymentReceived`]: crate::Event::PaymentReceived
+	pub fn claim_for_hash(
+		&self, payment_hash: PaymentHash, claimable_amount_msat: u64, preimage: PaymentPreimage,
```
tvolk131, line 282: Could we remove the `payment_hash` field and calculate it from the preimage?

jkczyz (approved):

PR #335, opened 2024-07-30
Account for `check-cfg`

tnull:
Recently, Rust 1.80 introduced automatic checking of `cfg` flags (see https://blog.rust-lang.org/2024/05/06/check-cfg.html). Here, we add all custom `cfg`s to the expected list


jkczyz (commented):
LGTM

PR #345, opened 2024-08-16
Allow honoring reserve in `send_all_to_address`

tnull:
Previously, `OnchainPayment::send_all_to_address` could only be used to fully drain the onchain wallet, i.e., would not retain any reserves.

Here, we try to introduce a `retain_reserves` bool that allows users to send all funds while honoring the configured on-chain reserves. While we're at it, we move the reserve checks for `send_to_address` also to the internal wallet's method, which makes the checks more accurate as they now are checked against the final transaction value, including transaction fees.

~This was requested by a user, but I'm a bit on the fence if we actually should move forward with it: for one, figuring out the spendable amount above the reserve is always gonna be inexact compared to draining the wallet.~

~Moreover, adding this to our API might send the wrong message of the reserve value being an exact value, while it's always on the safer side to maintain a larger reserve.~

### DIFF
--- a/tests/integration_tests_rust.rs
+++ b/tests/integration_tests_rust.rs
@@ -277,32 +277,74 @@ fn onchain_spend_receive() {
 	let addr_a = node_a.onchain_payment().new_address().unwrap();
 	let addr_b = node_b.onchain_payment().new_address().unwrap();
 
+	let premine_amount_sat = 1_100_000;
 	premine_and_distribute_funds(
 		&bitcoind.client,
 		&electrsd.client,
-		vec![addr_b.clone()],
-		Amount::from_sat(100000),
+		vec![addr_a.clone(), addr_b.clone()],
+		Amount::from_sat(premine_amount_sat),
 	);
 
 	node_a.sync_wallets().unwrap();
 	node_b.sync_wallets().unwrap();
-	assert_eq!(node_b.list_balances().spendable_onchain_balance_sats, 100000);
+	assert_eq!(node_a.list_balances().spendable_onchain_balance_sats, premine_amount_sat);
+	assert_eq!(node_b.list_balances().spendable_onchain_balance_sats, premine_amount_sat);
+
+	let channel_amount_sat = 1_000_000;
+	let reserve_amount_sat = 25_000;
+	open_channel(&node_b, &node_a, channel_amount_sat, true, &electrsd);
+	generate_blocks_and_wait(&bitcoind.client, &electrsd.client, 6);
+
+	node_a.sync_wallets().unwrap();
+	node_b.sync_wallets().unwrap();
+	expect_channel_ready_event!(node_a, node_b.node_id());
+	expect_channel_ready_event!(node_b, node_a.node_id());
+
+	let onchain_fee_buffer_sat = 1000;
+	let expected_node_a_balance = premine_amount_sat - reserve_amount_sat;
+	let expected_node_b_balance_lower =
+		premine_amount_sat - channel_amount_sat - reserve_amount_sat - onchain_fee_buffer_sat;
+	let expected_node_b_balance_upper =
+		premine_amount_sat - channel_amount_sat - reserve_amount_sat;
+	assert_eq!(node_a.list_balances().spendable_onchain_balance_sats, expected_node_a_balance);
+	assert!(node_b.list_balances().spendable_onchain_balance_sats > expected_node_b_balance_lower);
+	assert!(node_b.list_balances().spendable_onchain_balance_sats < expected_node_b_balance_upper);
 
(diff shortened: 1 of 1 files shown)


src/wallet.rs
```diff
@@ -249,30 +251,108 @@ where
...
-					psbt
+					(psbt, tx_details)
 				},
 				Err(err) => {
 					log_error!(self.logger, "Failed to create transaction: {}", err);
 					return Err(err.into());
 				},
 			};
 
+			// Check the reserve requirements (again) and return an error if they aren't met.
```
jkczyz, line 324: Not sure I follow. Why is a second check needed?

tests/integration_tests_rust.rs
```diff
@@ -279,7 +279,7 @@ fn onchain_spend_receive() {
 	assert!(node_b.list_balances().spendable_onchain_balance_sats < 100000);
 
 	let addr_b = node_b.onchain_payment().new_address().unwrap();
-	let txid = node_a.onchain_payment().send_all_to_address(&addr_b).unwrap();
+	let txid = node_a.onchain_payment().send_all_to_address(&addr_b, false).unwrap();
```
jkczyz, line 282: Could we additional test coverage?

src/wallet.rs
```diff
@@ -233,46 +237,115 @@ where
...
+						.enable_rbf();
+				},
+				OnchainSendType::SendAllRetainingReserve { cur_anchor_reserve_sats } => {
+					let spendable_amount_sats =
+						self.get_spendable_amount_sats(cur_anchor_reserve_sats).unwrap_or(0);
+					let mut tmp_tx_builder = locked_wallet.build_tx();
+					tmp_tx_builder
+						.add_recipient(address.script_pubkey(), spendable_amount_sats)
+						.fee_rate(fee_rate)
+						.enable_rbf();
```
jkczyz, line 265: Couldn't this result in a transaction where the entire `cur_anchor_reserve_sats` goes to fees? And thus there could be one less output, which would cause the fee estimation to be off?

Would the following solve this?

```rust
let change_address = locked_wallet.get_internal_address(AddressIndex::Peek(0));
tmp_tx_builder
	.drain_wallet()
	.drain_to(address.script_pubkey())
	.add_recipient(change_address, cur_anchor_reserve_sats)
	.fee_rate(fee_rate)
	.enable_rbf();
```

Though if `cur_anchor_reserve_sats` was exactly covered by one utxo, then an extra input would be used in the estimation. 🤔

src/wallet.rs
```diff
@@ -233,46 +237,115 @@ where
...
+								self.logger,
+								"Failed to create temporary transaction: {}",
+								err
+							);
+							return Err(err.into());
+						},
+					};
+
+					let estimated_tx_fee_sats =
+						tmp_tx_details.fee.unwrap_or(0).max(FEERATE_FLOOR_SATS_PER_KW as u64);
```
jkczyz, line 279: Shouldn't `FEERATE_FLOOR_SATS_PER_KW` be accounted for by the fee estimator when computing the `fee_rate` passed to the builder?

tests/integration_tests_rust.rs
```diff
@@ -271,26 +271,45 @@ fn start_stop_reinit() {
 #[test]
 fn onchain_spend_receive() {
 	let (bitcoind, electrsd) = setup_bitcoind_and_electrsd();
-	let chain_source = TestChainSource::Esplora(&electrsd);
+	let chain_source = TestChainSource::BitcoindRpc(&bitcoind);
```
jkczyz, line 274: Just checking to see if you want to change this. IIRC, there was some flakiness issues with `BitcoindRpc`?

tests/integration_tests_rust.rs
```diff
@@ -271,26 +271,45 @@ fn start_stop_reinit() {
...
+	open_channel(&node_b, &node_a, channel_amount_sat, true, &electrsd);
+	generate_blocks_and_wait(&bitcoind.client, &electrsd.client, 6);
+
+	node_a.sync_wallets().unwrap();
+	node_b.sync_wallets().unwrap();
+	expect_channel_ready_event!(node_a, node_b.node_id());
+	expect_channel_ready_event!(node_b, node_a.node_id());
+
+	let node_a_balance = premine_amount_sat - reserve_amount_sat;
+	let node_b_balance_lower = premine_amount_sat - channel_amount_sat - reserve_amount_sat - 1000;
```
jkczyz, line 304: Is this because `1000` sats is later sent? Might be worth making a constant for `1000` or adding a comment.

jkczyz (commented):
LGTM. Please squash.

jkczyz (approved):

PR #350, opened 2024-08-25
Refactor `ChannelConfig` / `MaxDustHTLCExposure`

### LINKED ISSUE #349: Refactor `ChannelConfig`
We should write our own Uniffi-compatible version of `enum MaxDustHLTCExposure`, which would allow us to drop the `Arc` and would expose the max dust exposure fields in `ChannelDetails`.

tnull:
Closes #349.

Previously, we chose to expose `ChannelConfig` as a Uniffi `interface`, providing accessor methods. Unfortunately this forced us to `Arc` it everywhere in the API, and also didn't allow to retrieve the currently set dust exposure limits. Here, we refactor our version of `ChannelConfig` to be a normal `struct` (Uniffi `dictionary`), and only expose the `MaxDustHTLCExposure` as an enum-`interface`.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -22,7 +22,7 @@ use lightning_transaction_sync::EsploraSyncClient;
 use bitcoin::secp256k1::PublicKey;
 use bitcoin::OutPoint;
 
-use std::sync::{Arc, Mutex, RwLock};
+use std::sync::{Arc, Mutex};
 
 pub(crate) type DynStore = dyn KVStore + Sync + Send;
 
@@ -279,7 +279,7 @@ pub struct ChannelDetails {
 	/// The largest value HTLC (in msat) we currently will accept, for this channel.
 	pub inbound_htlc_maximum_msat: Option<u64>,
 	/// Set of configurable parameters that affect channel operation.
-	pub config: Arc<ChannelConfig>,
+	pub config: ChannelConfig,
 }
 
 impl From<LdkChannelDetails> for ChannelDetails {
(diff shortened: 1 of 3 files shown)


src/types.rs
```diff
@@ -450,3 +422,49 @@ impl Default for ChannelConfig {
...
+
+/// Options for how to set the max dust exposure allowed on a channel.
+///
+/// See [`LdkChannelConfig::max_dust_htlc_exposure`] for details.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum MaxDustHTLCExposure {
+	/// This sets a fixed limit on the total dust exposure in millisatoshis.
+	///
+	/// Please refer to [`LdkMaxDustHTLCExposure`] for further details.
+	FixedLimitMsat {
```
jkczyz, line 434: Do we need to duplicate the units here?

src/types.rs
```diff
@@ -450,3 +422,49 @@ impl Default for ChannelConfig {
...
+		/// The fixed limit, in millisatoshis.
+		limit_msat: u64,
+	},
+	/// This sets a multiplier on the feerate (in sats/KW) to determine the maximum allowed dust
+	/// exposure.
+	///
+	/// Please refer to [`LdkMaxDustHTLCExposure`] for further details.
+	FeeRateMultiplier {
+		/// The applied fee rate multiplier.
+		multiplier: u64,
```
jkczyz, line 444: Should we include the units `_sats_per_kw`?

jkczyz: Yeah, thanks I misread that

jkczyz (commented):
LGTM. Please squash.

jkczyz (approved):

Issue #176, opened 2023-10-25
Allow to override fee rate for on-chain operations


tnull:
We should allow to override our fee estimation for methods like `Node::send_to_onchain_address`.

It would however make sense to expose this as an `Option<bitcoin::FeeRate>` in the API. To avoid immediately breaking API again, we should therefore wait until we could upgrade to `rust-bitcoin` 0.30 which introduced `FeeRate`.

jbesraa:
hey, I would like to work on this please.

It seems like `bdk` is using its own `FeeRate` in the tx builder, if we are using `bitcoin::FeeRate`  I guess we should convert from `bitcoin::FeeRate` to `bdk::FeeRate` as well?

PR #352, opened 2024-08-29
Refactor `FeeEstimator` to introduce local target variants

tnull:
.. previously we used LDK's `FeeEstimator` and `ConfirmationTarget` and ~misused some of the latter's variants for our non-Lightning operations.

Here, we introduce our own `FeeEstimator` and `ConfirmationTarget` allowing to add specific variants for `ChannelFunding` and `OnchainPayment`s, for example.

(This is can also be viewed as a prefactor to #176, which we'll do as part of / after the upcoming rust-bitcoin/BDK upgrade)

### DIFF
--- a/src/fee_estimator.rs
+++ b/src/fee_estimator.rs
@@ -2,9 +2,9 @@ use crate::config::FEE_RATE_CACHE_UPDATE_TIMEOUT_SECS;
 use crate::logger::{log_error, log_trace, Logger};
 use crate::{Config, Error};
 
-use lightning::chain::chaininterface::{
-	ConfirmationTarget, FeeEstimator, FEERATE_FLOOR_SATS_PER_KW,
-};
+use lightning::chain::chaininterface::ConfirmationTarget as LdkConfirmationTarget;
+use lightning::chain::chaininterface::FeeEstimator as LdkFeeEstimator;
+use lightning::chain::chaininterface::FEERATE_FLOOR_SATS_PER_KW;
 
 use bdk::FeeRate;
 use esplora_client::AsyncClient as EsploraClient;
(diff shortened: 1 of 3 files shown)


src/fee_estimator.rs
```diff
@@ -17,6 +17,26 @@ use std::ops::Deref;
 use std::sync::{Arc, RwLock};
 use std::time::Duration;
 
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
+pub(crate) enum ConfirmationTarget {
+	/// The default target for onchain payments.
+	OnchainPayment,
+	/// The target which we use for funding transactions.
```
jkczyz, line 24: s/which we use/used

src/fee_estimator.rs
```diff
@@ -61,23 +81,30 @@ where
...
 			let num_blocks = match target {
-				ConfirmationTarget::OnChainSweep => 6,
-				ConfirmationTarget::MinAllowedAnchorChannelRemoteFee => 1008,
-				ConfirmationTarget::MinAllowedNonAnchorChannelRemoteFee => 144,
-				ConfirmationTarget::AnchorChannelFee => 1008,
-				ConfirmationTarget::NonAnchorChannelFee => 12,
-				ConfirmationTarget::ChannelCloseMinimum => 144,
-				ConfirmationTarget::OutputSpendingFee => 12,
+				ConfirmationTarget::OnchainPayment => 6,
+				ConfirmationTarget::ChannelFunding => 12,
```
jkczyz, line 98: Seems like in practice the values are the same as before. Is this change to allow us to vary them in the future?

src/wallet.rs
```diff
@@ -43,7 +43,7 @@ enum WalletSyncStatus {
 	InProgress { subscribers: tokio::sync::broadcast::Sender<Result<(), Error>> },
 }
 
-pub struct Wallet<D, B: Deref, E: Deref, L: Deref>
+pub(crate) struct Wallet<D, B: Deref, E: Deref, L: Deref>
```
jkczyz, line 46: Visibility control for `Walle` and `WalletKeysManager` (below) seem unrelated to to the commits purpose.

jkczyz (commented):
LGTM. Please squash.

jkczyz (approved):

PR #355, opened 2024-09-05
Add copyright notices where they were missing

tnull:
Previously, only a few files included coypright headers. Here, we fix the omission and add the headers to all files where they were absent.


src/lib.rs
```diff
@@ -1,11 +1,9 @@
-// This file is Copyright its original authors, visible in version contror
-// history.
+// This file is Copyright its original authors, visible in version contror history.
```
jkczyz, line 1: s/contror/control

jkczyz (approved):

PR #358, opened 2024-09-10
Upgrade the World

tnull:
Closes bitcoindevkit/bdk#195.

... we update LDK, lightning-liquidity, BDK, rust-bitcoin, rust-esplora-client,
rust-electrum-client, etc.

Notably, we mostly maintain the functionalities here, any feature upgrades and refactors (e.g., adding additional chain data sources which we now can finally do) will happen in follow-up PRs.

Should be ready-for-review. The last commit can be dropped if BDK ships their next release during the review process, otherwise we'll (partially) revert it afterwards (see bitcoindevkit/bdk#359).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,14 +28,14 @@ panic = 'abort'     # Abort on panic
 default = []
 
 [dependencies]
-lightning = { version = "0.0.123", features = ["std"] }
-lightning-invoice = { version = "0.31.0" }
-lightning-net-tokio = { version = "0.0.123" }
-lightning-persister = { version = "0.0.123" }
-lightning-background-processor = { version = "0.0.123", features = ["futures"] }
-lightning-rapid-gossip-sync = { version = "0.0.123" }
-lightning-transaction-sync = { version = "0.0.123", features = ["esplora-async-https", "time"] }
-lightning-liquidity = { version = "=0.1.0-alpha.4", features = ["std"] }
+lightning = { version = "0.0.124", features = ["std"] }
+lightning-invoice = { version = "0.32.0" }
+lightning-net-tokio = { version = "0.0.124" }
+lightning-persister = { version = "0.0.124" }
+lightning-background-processor = { version = "0.0.124", features = ["futures"] }
+lightning-rapid-gossip-sync = { version = "0.0.124" }
+lightning-transaction-sync = { version = "0.0.124", features = ["esplora-async-https", "time"] }
+lightning-liquidity = { version = "0.1.0-alpha.5", features = ["std"] }
 
 #lightning = { git = "https://github.com/lightningdevkit/rust-lightning", branch="main", features = ["std"] }
 #lightning-invoice = { git = "https://github.com/lightningdevkit/rust-lightning", branch="main" }
@@ -55,18 +55,20 @@ lightning-liquidity = { version = "=0.1.0-alpha.4", features = ["std"] }
 #lightning-transaction-sync = { path = "../rust-lightning/lightning-transaction-sync", features = ["esplora-async"] }
 #lightning-liquidity = { path = "../lightning-liquidity", features = ["std"] }
 
-bdk = { version = "0.29.0", default-features = false, features = ["std", "async-interface", "use-esplora-async", "sqlite-bundled", "keys-bip39"]}
+bdk_chain = { version = "=0.19.0", default-features = false, features = ["std"] }
+bdk_esplora = { version = "=0.18.0", default-features = false, features = ["async-https-rustls"]}
+bdk_wallet = { version = "=1.0.0-beta.4", default-features = false, features = ["std", "keys-bip39"]}
 
 reqwest = { version = "0.11", default-features = false, features = ["json", "rustls-tls"] }
 rusqlite = { version = "0.28.0", features = ["bundled"] }
-bitcoin = "0.30.2"
+bitcoin = "0.32.2"
 bip39 = "2.0.0"
-bip21 = { version = "0.3.1", features = ["std"], default-features = false }
+bip21 = { version = "0.5", features = ["std"], default-features = false }
 
 rand = "0.8.5"
 chrono = { version = "0.4", default-features = false, features = ["clock"] }
-tokio = { version = "1.37", default-features = false, features = [ "rt-multi-thread", "time", "sync" ] }
-esplora-client = { version = "0.6", default-features = false }
+tokio = { version = "1.37", default-features = false, features = [ "rt-multi-thread", "time", "sync", "macros" ] }
+esplora-client = { version = "0.9", default-features = false }
 libc = "0.2"
 uniffi = { version = "0.26.0", features = ["build"], optional = true }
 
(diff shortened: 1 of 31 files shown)


src/balance.rs
```diff
@@ -81,6 +82,49 @@ pub enum LightningBalance {
 		/// The amount available to claim, in satoshis, excluding the on-chain fees which will be
 		/// required to do so.
 		amount_satoshis: u64,
+		/// The transaction fee we pay for the closing commitment transaction. This amount is not
+		/// included in the `amount_satoshis` value.
+		///
+		/// Note that if this channel is inbound (and thus our counterparty pays the commitment
+		/// transaction fee) this value will be zero. For channels created prior to LDK Node 0.4
+		/// the channel is always treated as outbound (and thus this value is never zero).
```
jkczyz, line 90: > the channel is always treated as outbound (and thus this value is never zero).

Not sure what is meant by this. Is because LDK Node only allowed outbound channels? If so, why do we need to say this?

src/balance.rs
```diff
@@ -81,6 +82,49 @@ pub enum LightningBalance {
 		/// The amount available to claim, in satoshis, excluding the on-chain fees which will be
 		/// required to do so.
```
jkczyz, line 83: Should this mention other amounts no longer included here?

src/fee_estimator.rs
```diff
@@ -164,7 +166,8 @@ where
 			ConfirmationTarget::OnchainPayment => 5000,
 			ConfirmationTarget::ChannelFunding => 1000,
 			ConfirmationTarget::Lightning(ldk_target) => match ldk_target {
-				LdkConfirmationTarget::OnChainSweep => 5000,
+				LdkConfirmationTarget::MaximumFeeEstimate => 8000,
```
jkczyz, line 167: What went into picking this number out of curiosity?

jkczyz (commented):
Almost through the LDK upgrade commits.

jkczyz (commented):
Reviewed up to but not including the first persistence commit.

Cargo.toml
```diff
@@ -55,18 +55,20 @@ lightning-liquidity = { version = "=0.1.0-alpha.4", features = ["std"] }
 #lightning-transaction-sync = { path = "../rust-lightning/lightning-transaction-sync", features = ["esplora-async"] }
 #lightning-liquidity = { path = "../lightning-liquidity", features = ["std"] }
 
-bdk = { version = "0.29.0", default-features = false, features = ["std", "async-interface", "use-esplora-async", "sqlite-bundled", "keys-bip39"]}
+bdk_chain = { version = "0.18.0", default-features = false, features = ["std"] }
+bdk_esplora = { version = "0.17.0", default-features = false, features = ["async-https-rustls"]}
+bdk_wallet = { version = "1.0.0-beta.2", default-features = false, features = ["std", "keys-bip39"]}
```
rustaceanrob, line 60: It may be worthwhile to upgrade to `1.0.0-beta.4` and the respective `bdk_chain` and `bdk_esplora` updates that were released today. I do not think there are any breaking changes, at least in the `bdk_wallet` API

src/builder.rs
```diff
@@ -532,36 +533,35 @@ fn build_with_store_internal(
...
-		database,
-	)
-	.map_err(|e| {
-		log_error!(logger, "Failed to set up wallet: {}", e);
-		BuildError::WalletSetupFailed
-	})?;
+	let descriptor = Bip84(xprv, KeychainKind::External);
+	let change_descriptor = Bip84(xprv, KeychainKind::Internal);
+	let mut wallet_persister =
+		KVStoreWalletPersister::new(Arc::clone(&kv_store), Arc::clone(&logger));
```
jkczyz, line 544: When will the first persistence occur?

src/builder.rs
```diff
@@ -583,15 +581,18 @@ fn build_with_store_internal(
...
+			(esplora_client, tx_sync, tx_broadcaster, fee_estimator)
 		},
 		None => {
 			// Default to Esplora client.
 			let server_url = DEFAULT_ESPLORA_SERVER_URL.to_string();
-			let tx_sync = Arc::new(EsploraSyncClient::new(server_url, Arc::clone(&logger)));
-			let blockchain =
-				EsploraBlockchain::from_client(tx_sync.client().clone(), BDK_CLIENT_STOP_GAP)
-					.with_concurrency(BDK_CLIENT_CONCURRENCY);
+			let mut client_builder = esplora_client::Builder::new(&server_url.clone());
```
jkczyz, line 589: `clone` not needed.

src/lib.rs
```diff
@@ -280,49 +280,44 @@ impl Node {
...
+						match wallet.sync().await {
+							Ok(()) => {
 								log_trace!(
-									sync_logger,
-									"Stopping background syncing on-chain wallet.",
-									);
-								return;
+								sync_logger,
+								"Background sync of on-chain wallet finished in {}ms.",
+								now.elapsed().as_millis()
```
jkczyz, line 304: Indentation off

src/lib.rs
```diff
@@ -280,49 +280,44 @@ impl Node {
...
-											err
-											)
-									}
-								}
+							Err(err) => {
+								log_error!(
+									sync_logger,
+									"Background sync of on-chain wallet failed: {}",
+									err
+									)
```
jkczyz, line 315: Indentation off

src/lib.rs
```diff
@@ -280,49 +280,44 @@ impl Node {
...
+				tokio::time::interval(Duration::from_secs(onchain_wallet_sync_interval_secs));
+			onchain_wallet_sync_interval
+				.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+			loop {
+				tokio::select! {
+					_ = stop_sync.changed() => {
+						log_trace!(
+							sync_logger,
+							"Stopping background syncing on-chain wallet.",
+							);
```
jkczyz, line 294: Indentation off

src/wallet/ser.rs
```diff
@@ -117,8 +141,8 @@ impl<'a> Writeable for ChangeSetSerWrapper<'a, BTreeSet<(ConfirmationBlockTime,
 		let len = BigSize(self.0.len() as u64);
 		len.write(writer)?;
 		for (time, txid) in self.0.iter() {
-			ChangeSetSerWrapper(time).write(writer)?;
-			txid.write(writer)?;
+			write_len_prefixed_field!(writer, ChangeSetSerWrapper(time))?;
+			write_len_prefixed_field!(writer, txid)?;
 		}
```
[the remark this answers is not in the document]
jkczyz: Hmm... right, on second thought we probably don't need to use `Iterable` or `WithoutLength`. But shouldn't we not write the length prefix for the `BTreeSet` size? Any `BTreeSet` or `BTreeMap` will always be written in the context of an `encode_tlv_stream`, so we'd end up writing the length twice. In LDK, we use `WithoutLength` to avoid this, but that's only because the serialization format predated the use of `encode_tlv_stream`.

src/wallet/ser.rs
```diff
@@ -117,8 +141,8 @@ impl<'a> Writeable for ChangeSetSerWrapper<'a, BTreeSet<(ConfirmationBlockTime,
 		let len = BigSize(self.0.len() as u64);
 		len.write(writer)?;
 		for (time, txid) in self.0.iter() {
-			ChangeSetSerWrapper(time).write(writer)?;
-			txid.write(writer)?;
+			write_len_prefixed_field!(writer, ChangeSetSerWrapper(time))?;
+			write_len_prefixed_field!(writer, txid)?;
```
jkczyz, line 145: When writing `BTreeSet` items, I guess only using length-prefixes means the item type can grow (e.g., by becoming a 3-tuple) but can never shrink or change a portion of the tuple. Maybe this is ok though since any change should probably result in using a different TLV type in the TLV stream containing the `BTreeSet`. Or should we treat the tuple itself as a type by writing it as a TLV stream?

src/wallet/mod.rs
```diff
@@ -274,6 +308,12 @@ where
 				},
 			}
 
+			let mut locked_persister = self.persister.lock().unwrap();
+			locked_wallet.persist(&mut locked_persister).map_err(|e| {
+				log_error!(self.logger, "Failed to persist wallet: {}", e);
+				Error::PersistenceFailed
+			})?;
```
jkczyz, line 315: If `psbt.extract_tx()` fails below, is there any implications of persisting here? Will the wallet think the inputs have been spent? Will it think a change output has been used when it really hasn't?

jkczyz:
> @jkczyz Let me know when I can squash.

Sure, go ahead!

src/wallet/mod.rs
jkczyz: If we never call `take_staged`, does that mean the `ChangeSet` returned by `staged` will continuously grow. I assume the purpose of it is to communicate what needs to be persisted, but it wasn't 100% clear from the docs. It mentions "committed" in some paces and "persisted" in others, though I assume those are equivalent?

jkczyz: Ah, so `take_staged` is basically "unstage".

jkczyz (approved):

PR #365, opened 2024-10-03
Refactor syncing and introduce `ChainSource`

### LINKED ISSUE #362: Support incremental wallet syncs
Post-1.0 `bdk_esplora` supports two methods of syncing: `full_scan` and `sync`. 

We should switch to 'usally' use incremental `sync`ing, and only `full_scan` if we suspect that we might be missing something, e.g., on first setting up a wallet from a seed or similar.

tnull:
Based on #358 
Closes #362 

We refactor our chain syncing and introduce a new `ChainSource` enum, which is a prerequisite for adding new chain data sources in a follow-up.

To this end, we first move the syncing logic to `ChainSource` and then introduce incremental syncing for Esplora which is always used except for initial syncs (now permanently persisted via a new `NodeMetrics` struct).

TODO:

- [x] Move Esplora-specific config options to fields in a tba. `EsploraConfig` object, whose defaults can be overridden in the `Builder::set_chain_source_esplora` method.

### DIFF
--- a/src/chain/mod.rs
+++ b/src/chain/mod.rs
@@ -6,8 +6,8 @@
 // accordance with one or both of these licenses.
 
 use crate::config::{
-	Config, BDK_CLIENT_CONCURRENCY, BDK_CLIENT_STOP_GAP, BDK_WALLET_SYNC_TIMEOUT_SECS,
-	FEE_RATE_CACHE_UPDATE_TIMEOUT_SECS, LDK_WALLET_SYNC_TIMEOUT_SECS,
+	Config, EsploraSyncConfig, BDK_CLIENT_CONCURRENCY, BDK_CLIENT_STOP_GAP,
+	BDK_WALLET_SYNC_TIMEOUT_SECS, FEE_RATE_CACHE_UPDATE_TIMEOUT_SECS, LDK_WALLET_SYNC_TIMEOUT_SECS,
 	RESOLVED_CHANNEL_MONITOR_ARCHIVAL_INTERVAL, TX_BROADCAST_TIMEOUT_SECS,
 	WALLET_SYNC_INTERVAL_MINIMUM_SECS,
 };
@@ -96,6 +96,7 @@ impl WalletSyncStatus {
 
 pub(crate) enum ChainSource {
 	Esplora {
+		sync_config: EsploraSyncConfig,
 		esplora_client: EsploraAsyncClient,
 		onchain_wallet: Arc<Wallet>,
 		onchain_wallet_sync_status: Mutex<WalletSyncStatus>,
@@ -112,9 +113,10 @@ pub(crate) enum ChainSource {
 
 impl ChainSource {
 	pub(crate) fn new_esplora(
-		server_url: String, onchain_wallet: Arc<Wallet>, fee_estimator: Arc<OnchainFeeEstimator>,
-		tx_broadcaster: Arc<Broadcaster>, kv_store: Arc<DynStore>, config: Arc<Config>,
-		logger: Arc<FilesystemLogger>, node_metrics: Arc<RwLock<NodeMetrics>>,
+		server_url: String, sync_config: EsploraSyncConfig, onchain_wallet: Arc<Wallet>,
+		fee_estimator: Arc<OnchainFeeEstimator>, tx_broadcaster: Arc<Broadcaster>,
+		kv_store: Arc<DynStore>, config: Arc<Config>, logger: Arc<FilesystemLogger>,
+		node_metrics: Arc<RwLock<NodeMetrics>>,
 	) -> Self {
 		let mut client_builder = esplora_client::Builder::new(&server_url);
 		client_builder = client_builder.timeout(DEFAULT_ESPLORA_CLIENT_TIMEOUT_SECS);
@@ -124,6 +126,7 @@ impl ChainSource {
 		let onchain_wallet_sync_status = Mutex::new(WalletSyncStatus::Completed);
 		let lightning_wallet_sync_status = Mutex::new(WalletSyncStatus::Completed);
 		Self::Esplora {
+			sync_config,
 			esplora_client,
 			onchain_wallet,
 			onchain_wallet_sync_status,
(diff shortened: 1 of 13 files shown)


src/chain/mod.rs
```diff
@@ -166,4 +180,57 @@ impl ChainSource {
...
+	}
+}
+
+impl Filter for ChainSource {
+	fn register_tx(&self, txid: &bitcoin::Txid, script_pubkey: &bitcoin::Script) {
+		{
+			match self {
+				Self::Esplora { tx_sync, .. } => tx_sync.register_tx(txid, script_pubkey),
+			}
+		}
```
jkczyz, line 590: Is this scope needed?

src/chain/mod.rs
```diff
@@ -161,4 +180,57 @@ impl ChainSource {
 			},
 		}
 	}
+
+	pub(crate) async fn sync_lightning_wallet(
+		&self, confirmables: Vec<&(dyn Confirm + Send + Sync)>,
```
jkczyz: For bitcoind, will this need `Listen`?

src/lib.rs
```diff
@@ -390,45 +386,40 @@ impl Node {
...
+					];
+					let now = Instant::now();
+
+					match chain_source.sync_lightning_wallet(confirmables).await {
+							Ok(()) => {
+								log_trace!(
+									sync_logger,
+									"Background sync of Lightning wallet finished in {}ms.",
+									now.elapsed().as_millis()
 									);
```
jkczyz: Indentation looks off.

src/fee_estimator.rs
```diff
@@ -198,3 +148,66 @@ where
...
+) -> FeeRate {
+	match target {
+		ConfirmationTarget::Lightning(
+			LdkConfirmationTarget::MinAllowedNonAnchorChannelRemoteFee,
+		) => {
+			let slightly_less_than_background = estimated_rate.to_sat_per_kwu() - 250;
+			FeeRate::from_sat_per_kwu(slightly_less_than_background)
+		},
+		_ => estimated_rate,
+	}
```
jkczyz, line 135: Will these helpers be called more than once? Just wondering what the rationale for splitting the logic across files. If the only difference later is how `convert_fee_rate` is called, seems that `update_fee_estimates` could be parameterized by a trait instead.

Otherwise, all the fee estimation logic is in the `chain` module instead of here, which seems wrong. A similar argument can be made for other code moved to `chain`, IMO (e.g., broadcasting).

jkczyz: Hmm... right, but it seems doable. For instance, the trait's method could take a list of confirmation targets and return an iterator over fee rates. This would keep the code fairly DRY, avoiding repeating much of the logic and keeping these functions internal to the `fee_estimation` module, IIUC.

jkczyz: Yeah, an internal enum but I believe it could be implemented on `EsploraAsyncClient`. You wouldn't need individual `ChainSource` objects.

jkczyz: Hmm... why couldn't the trait be internal to LDK Node?

jkczyz: To be clear, all I'm suggesting is parameterizing `OnchainFeeEstimator::update_fee_estimates` by a trait that can be implemented for both `EsploraAsyncClient` and `BitcoindRpcClient`. Then we wouldn't need to move all the code just to move it back later.

jkczyz: Hmm... or possibly returning some sort of async iterator? Alternatively, `OnchainFeeEstimator::update_fee_estimates` can be made into a macro and instantiated for each.

jkczyz: Right, I was thinking something like this:

```rust
pub(crate) async fn update_fee_estimates<F: FeeSource>(&self, fee_source: f) -> Result<(), Error> {
    let mut new_fee_rate_cache = HashMap::with_capacity(10);
    let confirmation_targets = vec![/* .. */];
    for (target, fee_rate) in fee_source.fee_rates(confirmation_targets) {
        let adjusted_fee_rate = /* .. */;
        new_fee_rate_cache.insert(target, adjusted_fee_rate);
    }

    if fee_estimator.set_fee_rate_cache(new_fee_rate_cache) { /* .. */ }
}
```

Where `FeeSource::fee_rates` returns an iterator. I'm handwaving the async part -- so fair enough if our MSRV prevents this -- but I'd think the different logic could be encapsulated in `FeeSource::fee_rates` implementations since it is given all targets. For esplora, it can fetch all at once using `get_fee_estimates`  -- so the trait method would need to be `async` as you said -- and return an iterator callinh `convert_fee_rate` on each item. For bitcoind, it can return an iterator that fetches them one-by-one.

Maybe there is a way to do it with a `Fn` trait parameter instead if using `aync_trait` is a problem? Something like: https://forum.dfinity.org/t/passing-async-function-or-closure-as-parameter/21785/9

Anyhow, happy to leave at is for now given the difficulties with `async` / blocking.

src/chain/mod.rs
```diff
@@ -578,16 +596,19 @@ impl Filter for ChainSource {
...
-	let mut latest_archival_height_lock = latest_channel_monitor_archival_height.write().unwrap();
+	kv_store: Arc<DynStore>, logger: Arc<FilesystemLogger>, node_metrics: Arc<RwLock<NodeMetrics>>,
+) -> Result<(), Error> {
+	let mut locked_node_metrics = node_metrics.write().unwrap();
 	let cur_height = channel_manager.current_best_block().height;
-	let should_archive = latest_archival_height_lock
+	let should_archive = locked_node_metrics
+		.latest_channel_monitor_archival_height
 		.as_ref()
 		.map_or(true, |h| cur_height >= h + RESOLVED_CHANNEL_MONITOR_ARCHIVAL_INTERVAL);
```
jkczyz: I guess unrelated to this PR, but it's unclear to me why we don't archive on every block. IIUC, this saying to do so every six blocks.

jkczyz (commented):
LGTM. Please squash.

jkczyz (approved):

PR #370, opened 2024-10-10
Add bitcoind RPC support

tnull:
Based on #365.

We add support for sourcing chain data from the `bitcoind` RPC interface. Ready for review.

TODO:
- [x] Figure out how to deal fully with `estimatesmartfee` in tests. Probably just do a simliar carve-out for non-mainnet as for Esplora, i.e., default to 253?
- [x] Figure out how to deal with explicit `sync_wallet` calls. No-op? Or still block until the next polling is finished?
- [x] Depends on https://github.com/lightningdevkit/rust-lightning/pull/3354, so we might temporarily need to copy/past `synchronize_listeners` over here.

~~- [ ] Implement UtxoSource / GossipVerifier (via `tokio` feature of `lightning-block-sync`)~~ (Bashing my head against the typechecker for a few hours is having me believe this is currently ~impossible with the upstream types, so moved to #377 / blocked on https://github.com/lightningdevkit/rust-lightning/issues/3369)

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ fn main() {
 LDK Node currently comes with a decidedly opinionated set of design choices:
 
 - On-chain data is handled by the integrated [BDK][bdk] wallet.
-- Chain data may currently be sourced from an [Esplora][esplora] server, while support for Electrum and `bitcoind` RPC will follow soon.
+- Chain data may currently be sourced from the Bitcoin Core RPC interface or an [Esplora][esplora] server, while support for Electrum will follow soon.
 - Wallet and channel state may be persisted to an [SQLite][sqlite] database, to file system, or to a custom back-end to be implemented by the user.
 - Gossip data may be sourced via Lightning's peer-to-peer network or the [Rapid Gossip Sync](https://docs.rs/lightning-rapid-gossip-sync/*/lightning_rapid_gossip_sync/) protocol.
 - Entropy for the Lightning and on-chain wallets may be sourced from raw bytes or a [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) mnemonic. In addition, LDK Node offers the means to generate and persist the entropy bytes to disk.


src/chain/mod.rs
```diff
@@ -903,6 +901,24 @@ impl ChainSource {
...
 
+					let fee_rate = match (fee_rate_update_res, config.network) {
+						(Ok(rate), _) => rate,
+						(Err(e), Network::Bitcoin) => {
+							// Strictly fail on mainnet.
+							log_error!(logger, "Failed to retrieve fee rate estimates: {}", e);
+							return Err(Error::FeerateEstimationUpdateFailed);
+						},
+						(Err(e), _) => {
+							// On regtest/testnet/signet we just skip, which will have us falling back to defaults.
```
jkczyz, line 912: Could you include _why_?

src/chain/mod.rs
```diff
@@ -915,6 +915,21 @@ impl ChainSource {
...
+							{
+								let mut locked_node_metrics = node_metrics.write().unwrap();
+								locked_node_metrics.latest_fee_rate_cache_update_timestamp =
+									unix_time_secs_opt;
+								write_node_metrics(
+									&*locked_node_metrics,
+									Arc::clone(&kv_store),
+									Arc::clone(&logger),
+								)?;
+							}
```
jkczyz, line 932: `LDKNode.swift` changes look out of place in d6b689494161164e296e50f70476b3eea44c25b9.

tests/integration_tests_rust.rs
```diff
@@ -28,42 +28,57 @@ use std::sync::Arc;
...
+}
+
+#[test]
+fn channel_full_cycle_bitcoind() {
+	let (bitcoind, electrsd) = setup_bitcoind_and_electrsd();
+	premine_dummy_blocks(&bitcoind.client, &electrsd.client);
+	let chain_source = TestChainSource::BitcoinRpc(&bitcoind);
+	let (node_a, node_b) = setup_two_nodes(&chain_source, false, true, false);
 	do_channel_full_cycle(node_a, node_b, &bitcoind.client, &electrsd.client, false, true, false);
 }
```
jkczyz, line 43: Why not expand the other tests for `BitcoinRpc`?

src/chain/bitcoind_rpc.rs
```diff
@@ -0,0 +1,53 @@
...
+
+use bitcoin::BlockHash;
+
+use base64::prelude::{Engine, BASE64_STANDARD};
+
+use std::sync::Arc;
+
+pub struct BitcoindRpcClient {
+	rpc_client: Arc<RpcClient>,
+}
```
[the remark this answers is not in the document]
jkczyz: Ah, I guess a later commit needs this. Just wondering if we need to re-implement `BlockSource`.

tests/integration_tests_rust.rs
jkczyz: Hmm... probably only tests where non-setup code depends on block (dis)connection notification (e.g., `onchain_spend_receive`). Though, I also wonder if we should use `BitcoinRpc` normally just to remove one moving part from the tests assuming that would lead to less flakiness.

jkczyz (commented):
LGTM modulo test changes

tests/integration_tests_rust.rs
jkczyz: Right, if the extra part is actually removing the underlying flakiness, maybe we should keep it. 😛 Though perhaps we need to make sure the (non-test) code is robust enough to retry appropriately.

jkczyz (commented):
LGTM, please squash

jkczyz (approved):

src/chain/mod.rs
```diff
@@ -201,9 +252,144 @@ impl ChainSource {
...
+						&**onchain_wallet as &(dyn Listen + Send + Sync),
+					),
+					(
+						channel_manager_best_block_hash,
+						&*channel_manager as &(dyn Listen + Send + Sync),
+					),
+					(sweeper_best_block_hash, &*output_sweeper as &(dyn Listen + Send + Sync)),
+				];
+
+				// TODO: Eventually we might want to see if we can synchronize `ChannelMonitor`s
```
enigbe, line 293: Apologies for the question post-merge but what would be the benefit of syncing the `ChannelMonitor`s before passing them to the `ChainMonitor`? There would still be a need to sync them via the monitor even if it was possible to sync before, right?

PR #392, opened 2024-10-28
Add `payment_preimage` field to `Event::PaymentSuccessful`

tnull:
.. we add this field for informative purposes.


enigbe (commented):
LGTM

Issue #309, opened 2024-06-17
Expose `logger` interface


tnull:
We should allow users to implement custom loggers, e.g., to implement their own rotation logic or to target a database backend.

enigbe:
Hi @tnull

Can you provide a bit more context about targeting a database backend? My initial thought is that the custom logger can save generated logs to any database server that can handle **`key`**(_`timestamp`_)-**`value`**(_`log`_) objects but I am not sure about the correctness of the data model and if `kv` storage is considered appropriate.

Additionally, the strategy I have thought to address this issue is to:
1. Create a `LoggerBuilder` that builds a `CustomLogger` which implements `Logger`,
2. Furnish the builder with methods to: 

    1. configure a database backend, 
    2. configure `logrotate` rotation schedule and/or configuration file(s), 
    3. etc...,

4. Default to the existing `CustomLogger::FilesystemLogger`  if custom build options are not configured,
5. Possibly build logger from config file,
6. Test against a database server.

```rust
enum CustomLogger {
    FilesystemLogger,
    DatabaseLogger
}
```

Please let me know if you think this approach is sound or needs refinement, and if I can start working on an implementation. Happy to get any feedback/recommendations from you.

enigbe:
Thanks for the responding @G8XSU 
I'll start work on this today.

enigbe:
Thanks for providing additional context @tnull. I understand the steps you have outlined and will contact you if I have any questions.

enigbe:
Thanks for the update. Switching here now.

PR #393, opened 2024-10-28
Add configurable logger

domZippilli:
## Description

Adds logger configuration to `ldk-node`.

## Motivation

The first impulse for this was to be able to write to stdout, as that's more desirable for some containerized infrastructure. But rather than add a specific stdout implementation, we thought it better to make this extensible to write to arbitrary destinations (for example, a network log ingester). This additionally has the benefit of letting you customize how the writing happens; for example the `tracing-appender` crate uses an `mpsc` for writers, with a single thread performing all the actual I/O off of the queue.

## Details

The PR has three commits:

   - [Rename FilesystemLogger to LdkNodeLogger.](https://github.com/lightningdevkit/ldk-node/pull/393/commits/4f82769974241f2b7ca8b233f128f02194a2a4f2) This renames the logger to an implementation-agnostic name. Hopefully this commit is easy to ignore when reviewing implementation.
   - [Make LdkNodeLogger technically customizable, retain default behavior.](https://github.com/lightningdevkit/ldk-node/pull/393/commits/2c54a18c5b74c654651d7e70bcf206c126b47aa2) This sets up the configurable logger, and then statically configures it to retain the existing behavior.
   - [Add customizable logging to configuration.](https://github.com/lightningdevkit/ldk-node/pull/393/commits/b0e7fef9098aee27622234a01604a9249ba38d03) This adds configuration points to the `Config` struct itself. Users of the `::default()` configuration should see no change in behavior.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -6,7 +6,9 @@
 // accordance with one or both of these licenses.
 
 use crate::chain::{ChainSource, DEFAULT_ESPLORA_SERVER_URL};
-use crate::config::{default_user_config, Config, EsploraSyncConfig, WALLET_KEYS_SEED_LEN};
+use crate::config::{
+	default_user_config, Config, EsploraSyncConfig, LoggingConfig, WALLET_KEYS_SEED_LEN,
+};
 
 use crate::connection::ConnectionManager;
 use crate::event::EventQueue;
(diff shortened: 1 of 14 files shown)


domZippilli:
Ah, I see. I was trying to avoid taking a dependency, but if you've already decided you want to use `log`, then that makes sense. We also wanted to avoid chasing generics all over the code, so the Fn trait objects were good, at least for the Rust side.

We originally did this in our local fork, as we can only see stdout logs right now. We can keep running with our fork until #309 is resolved.

As far as requirements I'd like to see for #309:
  - I'd like to be able to do structured logging. JSON, in our case, but full control over formatting would be best.
  - I'd like to be able to specify the device I'm writing to. Originally I did this with a `dyn Writer`, but for FFI you might need to do something else. At a minimum, I would need stdout.
  - I'd like to be able to specify how the writing happens. A non-blocking writer like I mentioned is important for high volume logging.

PR #400, opened 2024-11-08
chore: update rusqlite 0.28.0 -> 0.32.1

dpc:
rusqlite 0.28 is 2 years old at this point, and since cargo allows only one version of native dependencies, this creates conflicts with other dependencies that want to use sqlite.


dpc:
I see. [Rust 1.63](https://releases.rs/docs/1.63.0/) support ... :roll_eyes:

dpc:
If we were to use conditional compilation, then it should be possible to support multiple rusqlite version at the same time. People who need newer version could just opt-in using a cargo feature, while people who don't could just stick to defaults.

I need to find some time to try a PoC of that.

dpc:
Closing as redundant. See #403

PR #402, opened 2024-11-12
Add `.clippy.toml` to lint MSRV

rustaceanrob:
Allows clippy to catch and lint violations of the MSRV without further LSP configuration


jbesraa:
> Hmm, I'd rather consider setting the `rust-version` field in `Cargo.toml`, which, IIUC, [should also work](https://github.com/rust-lang/rust-clippy?tab=readme-ov-file#specifying-the-minimum-supported-rust-version)?

In SV2 project the `rust-version` was used and caused rust-analyzer to not work with some text editors unless you edit the config within the text editor to align the versions(or with `$ rustup use [MSRV]`), which becomes annoying if you switch between different projects with different MSRV's. Not sure about the `.clippy` file though as we are not using it.

jbesraa:
For me, if I configure rust-analyzer to use stable toolchain in vim and specify a different version in `Cargo.toml`, rust-analyzer wont work in vim. Other people in the team reported similar issue with different text editor also.

PR #418, opened 2024-12-04
Add LSPS1 client-side integration

tnull:
We add the capability to act as an LSPS1 / bLIP-51 client.

### DIFF
--- a/src/liquidity.rs
+++ b/src/liquidity.rs
@@ -223,11 +223,11 @@ where
 
 						match sender.send(response) {
 							Ok(()) => (),
-							Err(e) => {
+							Err(_) => {
 								log_error!(
 									self.logger,
-									"Failed to handle response from liquidity service: {:?}",
-									e
+									"Failed to handle response for request {:?} from liquidity service",
+									request_id
 								);
 							},
 						}
@@ -284,11 +284,11 @@ where
 
 						match sender.send(response) {
 							Ok(()) => (),
-							Err(e) => {
+							Err(_) => {
 								log_error!(
 									self.logger,
-									"Failed to handle response from liquidity service: {:?}",
-									e
+									"Failed to handle response for request {:?} from liquidity service",
+									request_id
 								);
 							},
 						}
@@ -342,11 +342,11 @@ where
 
 						match sender.send(response) {
 							Ok(()) => (),
-							Err(e) => {
+							Err(_) => {
 								log_error!(
 									self.logger,
-									"Failed to handle response from liquidity service: {:?}",
-									e
+									"Failed to handle response for request {:?} from liquidity service",
+									request_id
 								);
 							},
 						}
@@ -389,11 +389,11 @@ where
 
 						match sender.send(response) {
 							Ok(()) => (),
-							Err(e) => {
+							Err(_) => {
 								log_error!(
 									self.logger,
-									"Failed to handle response from liquidity service: {:?}",
-									e
+									"Failed to handle response for request {:?} from liquidity service",
+									request_id
 								);
 							},
 						}
@@ -441,11 +441,11 @@ where
 
 						match sender.send(response) {
 							Ok(()) => (),
-							Err(e) => {
+							Err(_) => {
 								log_error!(
 									self.logger,
-									"Failed to handle response from liquidity service: {:?}",
-									e
+									"Failed to handle response for request {:?} from liquidity service",
+									request_id
 								);
 							},
 						}


src/liquidity.rs
```diff
@@ -31,47 +35,126 @@ use std::time::Duration;
...
+		let liquidity_client_config =
+			Some(LiquidityClientConfig { lsps1_client_config, lsps2_client_config });
+
+		let liquidity_manager = Arc::new(LiquidityManager::new(
+			Arc::clone(&self.keys_manager),
+			Arc::clone(&self.channel_manager),
+			Some(Arc::clone(&self.chain_source)),
+			None,
+			None,
+			liquidity_client_config,
```
jkczyz, line 147: Why do both the `LiquidityManager` manager and each LSPS service need a copy of the config?

src/liquidity.rs
```diff
@@ -174,6 +194,175 @@ where
...
+				if let Some(lsps1_service) = self.lsps1_service.as_ref() {
+					if counterparty_node_id != lsps1_service.node_id {
+						debug_assert!(
+							false,
+							"Received response from unexpected LSP counterparty. This should never happen."
+							);
+						log_error!(
+							self.logger,
+							"Received response from unexpected LSP counterparty. This should never happen."
+							);
```
jkczyz, line 212: Could this happen from a misbehaving LSP?

src/liquidity.rs
```diff
@@ -282,6 +471,166 @@ where
...
+
+		let lsp_limits = self.lsps1_request_opening_params().await?.supported_options;
+		let channel_size_sat = lsp_balance_sat + client_balance_sat;
+
+		if channel_size_sat < lsp_limits.min_channel_balance_sat
+			|| channel_size_sat > lsp_limits.max_channel_balance_sat
+		{
+			log_error!(
+				self.logger,
+				"Requested channel size doesn't meet the LSP-provided limits (min: {}sat, max: {}sat).",
```
jkczyz, line 523: Should we log the requested size? Similarly in other places.

src/liquidity.rs
```diff
@@ -282,6 +471,166 @@ where
...
+			pending_create_order_requests_lock.insert(request_id, request_sender);
+		}
+
+		let response = tokio::time::timeout(
+			Duration::from_secs(LIQUIDITY_REQUEST_TIMEOUT_SECS),
+			request_receiver,
+		)
+		.await
+		.map_err(|e| {
+			log_error!(self.logger, "Liquidity request timed out: {}", e);
```
jkczyz, line 582: Should we log with request? Or do we expect that in `e`?

src/liquidity.rs
```diff
@@ -892,3 +895,113 @@ pub(crate) struct LSPS2BuyResponse {
...
 }
+
+/// A liquidity handler allowing to request channels via the [bLIP-51 / LSPS1] protocol.
+///
+/// Should be retrieved by calling [`Node::lsps1_liquidity`].
+///
+/// [bLIP-51 / LSPS1]: https://github.com/lightning/blips/blob/master/blip-0051.md
+/// [`Node::lsps1_liquidity`]: crate::Node::lsps1_liquidity
+#[derive(Clone)]
+pub struct LSPS1Liquidity {
```
jkczyz, line 965: Given LSPS2 channels are JIT, I suppose a similar struct won't be needed there? Perhaps the docs should mention how JIT channels are opened.

src/liquidity.rs
```diff
@@ -81,12 +185,185 @@ where
...
+						let response = LSPS1OpeningParamsResponse { supported_options };
+
+						match sender.send(response) {
+							Ok(()) => (),
+							Err(e) => {
+								log_error!(
+									self.logger,
+									"Failed to handle response from liquidity service: {:?}",
+									e
+								);
```
jkczyz, line 231: Will `e` contain the response?

src/liquidity.rs
jkczyz: Oh, I meant `LSPS1ClientConfig` and `LSPS2ClientConfig2`. Each config is cloned from the respective LSPS service, but the originals are never used anywhere else, AFAICT. Just wondering why we hang on to those inside each service. Maybe instead the config should be pulled out of the service and have the builder hold `Option<(LSPS1Service, LSPS1ClientConfig)>` if the config is not needed later. Then you won't have two copies of each config.

src/liquidity.rs
jkczyz: Ah, right, I was mistakenly thinking that it was part of the message.

src/liquidity.rs
jkczyz: Yeah, I'm fine with that though maybe include the `request_id` instead.

src/liquidity.rs
jkczyz: Yeah, though perhaps a single liquidity handler would be sufficient (i.e., no need for users to care which BLIP the method is defined in). I'm indifferent as to whether we should keep the method in the `Bolt11Payment` handler. But we should probably cross-link the docs where appropriate.

jkczyz: Right, I figure there at least be one abstraction for the grouping for LSP spec and method naming can differentiate as needed. But if there are competing standards inside that (or the move to BLIPS obsoletes the grouping), then maybe a separate one is needed.

src/liquidity.rs
jkczyz: SGTM. Ideally, we don't have multiple copies so they don't go out-of-sync if they somehow can change in the future.

PR #420, opened 2024-12-10
Add bLIP-52 / LSPS2 service-side support

tnull:
This adds initial bLIP-52 / LSPS2 service-side to LDK Node. When configured via `Builder::set_liquidity_provider_lsps2`, the node will serve JIT channels to clients.

In this initial version we have the user configure the detail fields via an `LSPS2ServiceConfig`, and do *not* track and retry honoring pending buy requests. This means that we will simply (silently due to lack of a feedback mechanism in the bLIP-52 / LSPS2 protocol) fail the flow if, e.g., we don't have sufficient liquidity or the user isn't connected to us at the time of handling the payment.

### DIFF
--- a/tests/integration_tests_rust.rs
+++ b/tests/integration_tests_rust.rs
@@ -8,13 +8,14 @@
 mod common;
 
 use common::{
-	do_channel_full_cycle, expect_channel_ready_event, expect_event, expect_payment_received_event,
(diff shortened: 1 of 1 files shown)


toneloc:
Thank you for ushering this along. I am excited for this to be included.

PR #426, opened 2025-01-03
Update BDK to 1.0.0 and LDK to 0.1.0

tnull:
We upgrade LDK and BDK to the latest releases.

PR #428, opened 2025-01-07
Start verifying gossip data via bitcoind RPC

### LINKED ISSUE #377: Start verifying gossip for the `bitcoind` RPC backend
Blocked on https://github.com/lightningdevkit/rust-lightning/issues/3369

We should verify the gossip we receive from the peer-to-peer network. Unfortunately this is not easily possible due to `GossipVerifier`'s type foo, so we'll have to wait until we shipped a more ergonomic API upstream.

tnull:
~~Based on #426.~~
Closes #377.

So far, we did not verify gossip data received via the peer-to-peer network. As a small fix in the latest LDK release allows us properly set the gossip verifier, we do so, and start verifying gossip when the bitcoind RPC gossip source is set.

### DIFF
--- a/src/gossip.rs
+++ b/src/gossip.rs
@@ -5,20 +5,23 @@
 // http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
 // accordance with one or both of these licenses.
 
+use crate::chain::ChainSource;
 use crate::config::RGS_SYNC_TIMEOUT_SECS;
-use crate::logger::{log_trace, FilesystemLogger, Logger};
-use crate::types::{GossipSync, Graph, P2PGossipSync, RapidGossipSync};
+use crate::logger::{log_error, log_trace, FilesystemLogger, Logger};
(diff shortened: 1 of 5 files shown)


src/gossip.rs
```diff
@@ -101,3 +125,30 @@ impl GossipSource {
 		}
 	}
 }
+
+pub(crate) struct RuntimeSpawner {
```
rustaceanrob, line 129: Could this potentially move up to `lib.rs`?  In the case of a future `ChainSource::Kyoto`, the enum variant could make use of a `RuntimeSpawner` for running the `kyoto::Node`. The type would be quite useful in that case but would not have much to do with the module `gossip`.

rustaceanrob: Yeah, perhaps this could be duplicated elsewhere. I see from the `builder.rs` the runtime is available to use, so that should be sufficient for me.

PR #432, opened 2025-01-16
Expose onchain transactions in store

### LINKED ISSUE #67: Expose on-chain transactions
Maybe/Probably blocked on BDK 1.0 as we still hope we can add event types for it.

See bitcoindevkit/bdk#36

tnull:
Closes #67 

Previously, onchain-transactions where not tracked in our `PaymentStore`.
Now, that the upgrade to BDK 1.0 is behind us, we take a stab at finally exposing them in our interface via `PaymentStore`. 

In the future we're looking to add dedicated `PaymentKind`s for channel fundings/channel closes (see #447), but for now all are exposed under the `Onchain` umbrella kind.

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -10,13 +10,16 @@ use persist::KVStoreWalletPersister;
 use crate::logger::{log_debug, log_error, log_info, log_trace, Logger, LdkLogger};
 
 use crate::fee_estimator::{ConfirmationTarget, FeeEstimator};
-use crate::payment::store::PaymentStore;
+use crate::payment::store::{ConfirmationStatus, PaymentStore};
+use crate::payment::{PaymentDetails, PaymentDirection, PaymentStatus};
 use crate::Error;
 
 use lightning::chain::chaininterface::BroadcasterInterface;
+use lightning::chain::channelmonitor::ANTI_REORG_DELAY;
 use lightning::chain::{BestBlock, Listen};
 
 use lightning::events::bump_transaction::{Utxo, WalletSource};
+use lightning::ln::channelmanager::PaymentId;
 use lightning::ln::inbound_payment::ExpandedKey;
 use lightning::ln::msgs::{DecodeError, UnsignedGossipMessage};
 use lightning::ln::script::ShutdownScript;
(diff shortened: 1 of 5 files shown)


src/wallet/mod.rs
```diff
@@ -133,6 +142,76 @@ where
...
+			let (payment_status, confirmation_status) = match wtx.chain_position {
+				bdk_chain::ChainPosition::Confirmed { anchor, .. } => {
+					let confirmation_height = anchor.block_id.height;
+					let cur_height = locked_wallet.latest_checkpoint().height();
+					let payment_status = if cur_height >= confirmation_height + ANTI_REORG_DELAY - 1
+					{
+						PaymentStatus::Succeeded
+					} else {
+						PaymentStatus::Pending
+					};
```
jkczyz, line 165: Say the transaction was re-orged out and was then RBF'ed. Is that possible? Would we have payment stuck in pending?

src/wallet/mod.rs
```diff
@@ -133,6 +142,76 @@ where
...
+			// `list_channels`. But for the latter the best approach is much less
+			// clear: for force-closes/HTLC spends we should be good querying
+			// `OutputSweeper::tracked_spendable_outputs`, but regular channel closes
+			// (i.e., `SpendableOutputDescriptor::StaticOutput` variants) are directly
+			// spent to a wallet address. The only solution I can come up with is to
+			// create and persist a list of 'static pending outputs' that we could use
+			// here to determine the `PaymentKind`, but that's not really satisfactory, so
+			// we're punting on it until we can come up with a better solution.
+			let kind = crate::payment::PaymentKind::Onchain { txid, status: confirmation_status };
+			let (sent, received) = locked_wallet.sent_and_received(&wtx.tx_node.tx);
```
jkczyz, line 189: Is change abstracted away from this API?

jkczyz: Couldn't the change be more than the amount sent, making it look like an inbound payment when it was really an outbound? Or am I misunderstanding?

src/wallet/mod.rs
jkczyz: Could we possibly detect that the UTXO has been spent and remove the pending payment?

jkczyz (commented):
Feel free to squash

jkczyz (approved):

PR #433, opened 2025-01-16
Implement `Listen::filtered_block_connected` for `Wallet`

rustaceanrob:
While `apply_block` works in the case of bitcoind RPC, chain sources that do not emit every block cannot make use of `apply_block`. Inserting the intermediate hashes allows `apply_block` to reconcile where to extend a block in the local chain representation.

ref: https://github.com/bitcoindevkit/bdk/blob/master/crates/wallet/src/test_utils.rs#L297

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -449,13 +449,32 @@ where
 	L::Target: Logger,
 {
 	fn filtered_block_connected(
-		&self, _header: &bitcoin::block::Header,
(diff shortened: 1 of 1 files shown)


rustaceanrob:
> so that we directly have some code to test this is working as expected?

Sure, this is hard to test without a chain source that can make use of it. Closing in favor of adding it as a commit in a future PR.

PR #434, opened 2025-01-17
Allow to override fee rates for onchain payments

### LINKED ISSUE #176: Allow to override fee rate for on-chain operations

tnull:
Fixes #176.

We allow to override our fee estimator in the `send_to_address` and `send_all_to_address` API methods. To this end, we implement a bindings-compatible wrapper around `bitcoin::FeeRate`.

### DIFF
--- a/src/payment/onchain.rs
+++ b/src/payment/onchain.rs
@@ -17,6 +17,24 @@ use bitcoin::{Address, Txid};
 
 use std::sync::{Arc, RwLock};
 
+#[cfg(not(feature = "uniffi"))]
+type FeeRate = bitcoin::FeeRate;
+#[cfg(feature = "uniffi")]
+type FeeRate = Arc<bitcoin::FeeRate>;
+
+macro_rules! maybe_map_fee_rate_opt {
+	($fee_rate_opt: expr) => {{
+		#[cfg(not(feature = "uniffi"))]
+		{
+			$fee_rate_opt
+		}
+		#[cfg(feature = "uniffi")]
+		{
+			$fee_rate_opt.map(|f| *f)
+		}
+	}};
+}
+
 /// A payment handler allowing to send and receive on-chain payments.
 ///
 /// Should be retrieved by calling [`Node::onchain_payment`].
@@ -50,9 +68,12 @@ impl OnchainPayment {
 	/// This will respect any on-chain reserve we need to keep, i.e., won't allow to cut into
 	/// [`BalanceDetails::total_anchor_channels_reserve_sats`].
 	///
+	/// If `fee_rate` is set it will be used on the resulting transaction. Otherwise we'll retrieve
+	/// a reasonable estimate from the configured chain source.
+	///
 	/// [`BalanceDetails::total_anchor_channels_reserve_sats`]: crate::BalanceDetails::total_anchor_channels_reserve_sats
 	pub fn send_to_address(
-		&self, address: &bitcoin::Address, amount_sats: u64,
+		&self, address: &bitcoin::Address, amount_sats: u64, fee_rate: Option<FeeRate>,
 	) -> Result<Txid, Error> {
 		let rt_lock = self.runtime.read().unwrap();
 		if rt_lock.is_none() {
@@ -63,7 +84,8 @@ impl OnchainPayment {
 			crate::total_anchor_channels_reserve_sats(&self.channel_manager, &self.config);
 		let send_amount =
 			OnchainSendAmount::ExactRetainingReserve { amount_sats, cur_anchor_reserve_sats };
-		self.wallet.send_to_address(address, send_amount)
+		let fee_rate_opt = maybe_map_fee_rate_opt!(fee_rate);
+		self.wallet.send_to_address(address, send_amount, fee_rate_opt)
 	}
 
 	/// Send an on-chain payment to the given address, draining the available funds.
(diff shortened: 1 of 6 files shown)


src/payment/onchain.rs
```diff
@@ -63,7 +72,32 @@ impl OnchainPayment {
...
-		self.wallet.send_to_address(address, send_amount)
+		self.wallet.send_to_address(address, send_amount, fee_rate)
+	}
+
+	/// Send an on-chain payment to the given address.
+	///
+	/// This will respect any on-chain reserve we need to keep, i.e., won't allow to cut into
+	/// [`BalanceDetails::total_anchor_channels_reserve_sats`].
+	///
+	/// If `fee_rate` is set it will used on the resulting transaction. Otherwise a reasonable
```
jkczyz, line 83: s/will used/will be used

jkczyz, line 83: Drop "a reasonable" or re-word for grammar?

src/payment/onchain.rs
```diff
@@ -77,9 +111,48 @@ impl OnchainPayment {
 	/// will try to send all spendable onchain funds, i.e.,
 	/// [`BalanceDetails::spendable_onchain_balance_sats`].
 	///
+	/// If `fee_rate` is set it will be used on the resulting transaction. Otherwise a reasonable
```
jkczyz, line 114: Likewise.

src/payment/onchain.rs
```diff
@@ -63,7 +72,32 @@ impl OnchainPayment {
...
+		let rt_lock = self.runtime.read().unwrap();
+		if rt_lock.is_none() {
+			return Err(Error::NotRunning);
+		}
+
+		let cur_anchor_reserve_sats =
+			crate::total_anchor_channels_reserve_sats(&self.channel_manager, &self.config);
+		let send_amount =
+			OnchainSendAmount::ExactRetainingReserve { amount_sats, cur_anchor_reserve_sats };
+		self.wallet.send_to_address(address, send_amount, fee_rate.map(|f| f.0))
```
jkczyz, line 100: Would it be worth DRYing up these method definitions with macro using a parameter for accessing the fee rate?

src/payment/onchain.rs
```diff
@@ -94,6 +167,6 @@ impl OnchainPayment {
 			OnchainSendAmount::AllDrainingReserve
 		};
 
-		self.wallet.send_to_address(address, send_amount)
+		self.wallet.send_to_address(address, send_amount, fee_rate.map(|f| f.0))
```
jkczyz, line 170: Likewise.

src/payment/onchain.rs
```diff
@@ -77,9 +111,48 @@ impl OnchainPayment {
...
+	/// This is useful if you have closed all channels and want to migrate funds to another
+	/// on-chain wallet.
+	///
+	/// Please note that if `retain_reserves` is set to `false` this will **not** retain any on-chain reserves, which might be potentially
+	/// dangerous if you have open Anchor channels for which you can't trust the counterparty to
+	/// spend the Anchor output after channel closure. If `retain_reserves` is set to `true`, this
+	/// will try to send all spendable onchain funds, i.e.,
+	/// [`BalanceDetails::spendable_onchain_balance_sats`].
+	///
+	/// If `fee_rate` is set it will be used on the resulting transaction. Otherwise a reasonable
```
jkczyz, line 149: Likewise.

src/uniffi_types.rs
```diff
@@ -345,3 +346,61 @@ impl UniffiCustomTypeConverter for NodeAlias {
 		obj.to_string()
 	}
 }
+
+/// Represents fee rate.
+///
+/// This is a simple wrapper around [`bitcoin::FeeRate`], only used for UniFFI bindings.
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub struct FeeRate(pub(crate) bitcoin::FeeRate);
```
jkczyz, line 354: Could you remind me what prevents using `bitcoin::FeeRate` directly in bindings?

src/wallet/mod.rs
```diff
@@ -239,9 +240,12 @@ where
 
 	pub(crate) fn send_to_address(
 		&self, address: &bitcoin::Address, send_amount: OnchainSendAmount,
+		fee_rate: Option<FeeRate>,
 	) -> Result<Txid, Error> {
+		// Use the set fee_rate or default to fee estimation.
 		let confirmation_target = ConfirmationTarget::OnchainPayment;
-		let fee_rate = self.fee_estimator.estimate_fee_rate(confirmation_target);
+		let fee_rate =
+			fee_rate.unwrap_or(self.fee_estimator.estimate_fee_rate(confirmation_target));
```
jkczyz, line 248: Let's use `unwrap_or_else`.

src/uniffi_types.rs
jkczyz: > Hmm, I'm not sure I fully understand the proposal, but AFAIU this would complicate things even further.

Basically, add a trait that has a `from_sat_per_vb` method and implement it for `FeeRate` when `#[cfg(feature = "uniffi")]`.

> I think I'm leaning towards using `bitcoin::FeeRate` in bindings, i.e., leave current HEAD state. Let me know if you'd be fine with that and if I can squash.

SGTM

src/payment/onchain.rs
```diff
@@ -17,6 +17,24 @@ use bitcoin::{Address, Txid};
...
+		#[cfg(not(feature = "uniffi"))]
+		{
+			$fee_rate_opt
+		}
+		#[cfg(feature = "uniffi")]
+		{
+			$fee_rate_opt.map(|f| *f)
+		}
+	}};
+}
```
jkczyz, line 36: Ah, I guess we could just inline this instead of using a macro? Was thinking a macro would be needed for the entire method.

jkczyz (approved):

Issue #310, opened 2024-06-17
Allow disabling wallet background syncing


tnull:
As Esplora syncing is heavily rate limited, we so far recommended to avoid manual syncing via `sync_wallets` and rather rely on the background syncing interval.

However, some users might have setups where they want to or have to rely on manually syncing the wallets. While they can currently bump the intervals to `u64::MAX`, we should consider supporting this explicitly by making the interval config settings `Option`s and disabling background syncing of the wallets/fee rate cache if the corresponding value is disabled.

PR #448, opened 2025-01-27
Include onchain transactions in events

### LINKED ISSUE #446: Expose on-chain transactions in events
In bitcoindevkit/bdk#432 we exposed on-chain transactions in the payment store.

We should also do this in events, although for the latter I'd prefer to wait until we can filter out channel-related transactions properly to avoid redudant events, e.g., on channel closure.

We also might want to wait until BDK finally includes on-chain events, which might allow us to make the logic a bit more lightweight.

Blocked on https://github.com/bitcoindevkit/bdk_wallet/issues/6 and  https://github.com/lightningdevkit/rust-lightning/issues/3548

tnull:
Closes #446.
Based on #432.

In #432 we exposed transactions in store. Here we take a stab at exposing them via events. However, to avoid emitting duplicate events for channel-related transactions, this is in draft until we have https://github.com/lightningdevkit/rust-lightning/issues/3566

### DIFF
--- a/bindings/kotlin/ldk-node-jvm/lib/src/test/kotlin/org/lightningdevkit/ldknode/LibraryTest.kt
+++ b/bindings/kotlin/ldk-node-jvm/lib/src/test/kotlin/org/lightningdevkit/ldknode/LibraryTest.kt
@@ -351,6 +351,24 @@ class LibraryTest {
         assert(spendableBalance1AfterClose < 100000u)
         assertEquals(102500uL, spendableBalance2AfterClose)
 
+        val externalAddress = bitcoinCli("getnewaddress")
+        val onchainTxid = node1.onchainPayment().sendToAddress(externalAddress, 10000u, null)
+        waitForTx(esploraEndpoint, onchainTxid)
+        mineAndWait(esploraEndpoint, 6u)
+        node1.syncWallets()
+
+        val onchainPaymentSuccessfulEvent = node1.waitNextEvent()
+        println("Got event: $onchainPaymentSuccessfulEvent")
+        when (onchainPaymentSuccessfulEvent) {
+            is Event.OnchainPaymentSuccessful -> {
+                assertEquals(onchainTxid, onchainPaymentSuccessfulEvent.txid)
+                assertEquals(10000000uL, onchainPaymentSuccessfulEvent.amountMsat)
+                assertTrue(onchainPaymentSuccessfulEvent.feePaidMsat != null)
+            }
+            else -> error("Expected successful on-chain payment event")
+        }
+        node1.eventHandled()
+
         assertTrue(logWriter1.getLogMessages().isNotEmpty())
         assertTrue(logWriter2.getLogMessages().isNotEmpty())
 
--- a/bindings/kotlin/ldk-node-jvm/lib/src/test/kotlin/org/lightningdevkit/ldknode/LibraryTest.kt
+++ b/bindings/kotlin/ldk-node-jvm/lib/src/test/kotlin/org/lightningdevkit/ldknode/LibraryTest.kt
@@ -226,6 +226,28 @@ class LibraryTest {
         node1.syncWallets()
         node2.syncWallets()
 
+        val onchainPaymentReceivedEvent1 = node1.waitNextEvent()
+        println("Got event: $onchainPaymentReceivedEvent1")
+        when (onchainPaymentReceivedEvent1) {
+            is Event.OnchainPaymentReceived -> {
+                assertEquals(txid1, onchainPaymentReceivedEvent1.txid)
+                assertEquals(100000000uL, onchainPaymentReceivedEvent1.amountMsat)
+            }
+            else -> error("Expected initial on-chain payment event")
+        }
+        node1.eventHandled()
+
+        val onchainPaymentReceivedEvent2 = node2.waitNextEvent()
+        println("Got event: $onchainPaymentReceivedEvent2")
+        when (onchainPaymentReceivedEvent2) {
+            is Event.OnchainPaymentReceived -> {
+                assertEquals(txid2, onchainPaymentReceivedEvent2.txid)
+                assertEquals(100000000uL, onchainPaymentReceivedEvent2.amountMsat)
+            }
+            else -> error("Expected initial on-chain payment event")
+        }
+        node2.eventHandled()
+
         val spendableBalance1 = node1.listBalances().spendableOnchainBalanceSats
         val spendableBalance2 = node2.listBalances().spendableOnchainBalanceSats
         val totalBalance1 = node1.listBalances().totalOnchainBalanceSats
--- a/src/data_store.rs
+++ b/src/data_store.rs
@@ -82,33 +82,43 @@ where
 	}
 
 	pub(crate) async fn insert_or_update(&self, object: SO) -> Result<bool, Error> {
+		self.insert_or_update_with(object, |updated, _| updated).await
+	}
+
+	/// Inserts `object` or merges it into an existing object, returning whether the store changed
+	/// and the effective object after the merge.
+	pub(crate) async fn insert_or_update_and_get(&self, object: SO) -> Result<(bool, SO), Error> {
+		self.insert_or_update_with(object, |updated, stored_object| {
+			(updated, stored_object.clone())
+		})
+		.await
+	}
+
+	async fn insert_or_update_with<R>(
+		&self, object: SO, result_fn: impl FnOnce(bool, &SO) -> R,
+	) -> Result<R, Error> {
 		let _guard = self.mutation_lock.lock().await;
 
 		let id = object.id();
-		let data_to_persist = {
+		let updated_object = {
 			let locked_objects = self.objects.lock().expect("lock");
 			if let Some(existing_object) = locked_objects.get(&id) {
 				let mut updated_object = existing_object.clone();
 				let updated = updated_object.update(object.to_update());
 				if updated {
-					Some(updated_object)
+					updated_object
 				} else {
-					None
+					return Ok(result_fn(false, existing_object));
 				}
 			} else {
-				Some(object)
+				object
 			}
 		};
 
-		match data_to_persist {
-			Some(updated_object) => {
-				self.persist(&updated_object).await?;
-				let mut locked_objects = self.objects.lock().expect("lock");
-				locked_objects.insert(id, updated_object);
-				Ok(true)
-			},
-			None => Ok(false),
-		}
+		self.persist(&updated_object).await?;
+		let mut locked_objects = self.objects.lock().expect("lock");
+		let stored_object = locked_objects.entry(id).insert_entry(updated_object).into_mut();
+		Ok(result_fn(true, stored_object))
 	}
 
 	pub(crate) async fn remove(&self, id: &SO::Id) -> Result<(), Error> {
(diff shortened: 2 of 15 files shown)


ovitrif:
> While the latest BDK 2.2 now shipped event (which would make emitting events easier on our end), we still need [lightningdevkit/rust-lightning#3566](https://github.com/lightningdevkit/rust-lightning/issues/3566) to properly classify the transactions in our store. So this remains blocked until the latter ships (likely with LDK 0.3).

**Is there a way we could get this into ldk-node sooner?**
To be honest we (Bitkit team), kind of relied on the expectation that the BDK 2.2 PR [#310 Return wallet events when applying updates](https://github.com/bitcoindevkit/bdk_wallet/pull/310) will be integrated in ldk-node soon enough (ie. this month or November); and we made our roadmapping around this expectation. Now I'm getting feedback that there is no way we can delay our release any further than when it's planned, and the due date is very soon; which leaves us with very few options:
1. Get help from you to get this into ldk-node very soon (November)
2. Collaborate with you to have someone from our team help with implementation efforts (for example by fixing [lightningdevkit/rust-lightning#3566](https://github.com/lightningdevkit/rust-lightning/issues/3566)), but then we will need a commitment that if we fix that, this PR takes priority and it is implemented shortly thereafter)
3. Do our own implementation for this

The team seems to lean towards favoring option **3**, which is not ideal IMHO, but the main argument is that we don't really have the time to go any other route.

The issue that I see from my PoV is that option **2** might even seem faster.

Could **2** be an option?

tnull:
Should be good for review now.

src/data_store.rs
```diff
@@ -403,6 +459,28 @@ mod tests {
...
+			store,
+			logger,
+		);
+
+		let supplied = MergingTestObject { id, data: [25u8; 3], preserved_data: [26u8; 3] };
+		let expected = MergingTestObject { data: supplied.data, ..existing };
+		assert_eq!(Ok((true, expected)), data_store.insert_or_update_and_get(supplied).await);
+
+		let unchanged = MergingTestObject { preserved_data: [27u8; 3], ..expected };
+		assert_eq!(Ok((false, expected)), data_store.insert_or_update_and_get(unchanged).await);
```
jkczyz, line 488: Add an assertion for when an insert occurs.

src/wallet/mod.rs
```diff
@@ -278,16 +278,22 @@ impl Wallet {
...
 
-					self.runtime.block_on(self.payment_store.insert_or_update(payment.clone()))?;
+					self.runtime.block_on(async {
+						let (updated, stored_payment) =
+							self.payment_store.insert_or_update_and_get(payment).await?;
 
-					if payment_status == PaymentStatus::Pending {
-						let pending_payment =
-							self.create_pending_payment_from_tx(payment, Vec::new());
+						if updated && payment_status == PaymentStatus::Succeeded {
```
jkczyz: If we crash after the `payment_store` write but before the event queue is persisted, `updated` will be false on the next sync upon restart and the event will be lost.

Also from Claude:

This `updated` gate is the only thing preventing re-emission after a restart (the reloaded store merges the re-derived payment as a no-op), and no test covers it: no restart test syncs and asserts `next_event() == None`. I wrote that test — it passes, so this is coverage-only.

src/wallet/mod.rs
```diff
@@ -278,16 +278,22 @@ impl Wallet {
...
 
-						self.runtime.block_on(
-							self.pending_payment_store.insert_or_update(pending_payment),
-						)?;
-					}
+						if payment_status == PaymentStatus::Pending {
+							let pending_payment =
+								self.create_pending_payment_from_tx(stored_payment, Vec::new());
+							self.pending_payment_store.insert_or_update(pending_payment).await?;
+						}
```
jkczyz: Directly from Claude:

Direct graduation leaves a pre-existing pending entry stale. Sequence: a tx confirms below `ANTI_REORG_DELAY`, so a pending entry is written recording its confirmation block; a shallow reorg then moves the tx into a different block; by the next batch sync the tx is ≥ `ANTI_REORG_DELAY` deep. That sync's `TxConfirmed` graduates the payment and emits an event with the new (canonical) block, but the pending entry still holds the old, now-orphaned block — and `ChainTipChanged` then merges that stale entry back in (`updated == true`), emitting a duplicate event carrying the orphaned block hash and regressing the stored confirmation to the orphaned block. Demonstrated by a failing integration test using a 1-block reorg; no restart needed. Remove/refresh the pending entry when graduating directly — with that fixed, emission is exactly-once for reorgs shallower than `ANTI_REORG_DELAY`, which is worth documenting on the event variants.

src/event.rs
```diff
@@ -271,6 +271,46 @@ pub enum Event {
...
+	/// It's guaranteed to have reached at least [`ANTI_REORG_DELAY`] confirmations.
+	///
+	/// [`ANTI_REORG_DELAY`]: lightning::chain::channelmonitor::ANTI_REORG_DELAY
+	OnchainPaymentSuccessful {
+		/// A local identifier used to track the payment.
+		payment_id: PaymentId,
+		/// The transaction identifier.
+		txid: Txid,
+		/// The value, in thousandths of a satoshi, that was sent.
+		amount_msat: u64,
```
jkczyz, line 290: Should we include a field for the fees paid?

src/event.rs
```diff
@@ -271,6 +271,46 @@ pub enum Event {
...
+		/// The hash of the block in which the transaction was confirmed.
+		block_hash: BlockHash,
+		/// The height at which the block was confirmed.
```
jkczyz: The height of the block in which the transaction was confirmed.

src/event.rs
```diff
@@ -271,6 +271,46 @@ pub enum Event {
...
+	OnchainPaymentReceived {
+		/// A local identifier used to track the payment.
+		payment_id: PaymentId,
+		/// The transaction identifier.
+		txid: Txid,
+		/// The value, in thousandths of a satoshi, that has been received.
+		amount_msat: u64,
+		/// The hash of the block in which the transaction was confirmed.
+		block_hash: BlockHash,
+		/// The height at which the block was confirmed.
```
jkczyz: The height of the block in which the transaction was confirmed.

src/event.rs
```diff
@@ -271,6 +271,46 @@ pub enum Event {
 		/// This will be `None` for events serialized by LDK Node v0.2.1 and prior.
 		reason: Option<ClosureReason>,
 	},
+	/// A sent on-chain payment was successful.
+	///
+	/// This is only emitted for wallet transactions which were not classified as channel
+	/// funding, splices, closes, sweeps, or other LDK-driven chain activity.
```
jkczyz, line 277: Claude:

Nit: two narrow cases can still emit these events for LDK-driven transactions and may deserve a doc sentence: (1) classification only happens when our own broadcaster handles a tx, so a closing tx broadcast solely by the counterparty that confirms while this node is offline is discovered by wallet sync unclassified and emits `OnchainPaymentReceived` for the close output; (2) records persisted by pre-classification versions deserialize with `tx_type: None`, so an old LDK-driven tx that only settles after the upgrade emits too. Normal counterparty-initiated coop closes are fine — both peers broadcast the closing tx, so it's classified.

tests/common/mod.rs
```diff
@@ -1546,12 +1621,18 @@ pub(crate) async fn do_channel_full_cycle<E: ElectrumApi>(
...
 
 	generate_blocks_and_wait(&bitcoind, electrsd, 6).await;
 	node_a.sync_wallets().unwrap();
 	node_b.sync_wallets().unwrap();
+	if !allow_0conf {
+		assert_eq!(
+			node_a.expect_onchain_payment_event(OnchainPaymentEvent::Received).await,
+			splice_out_txo.txid,
+		);
+	}
```
jkczyz, line 1632: Hmmm... Clause says we aren't using `InteractiveFunding` for 0-conf splices.

The gate is correct but subtle — worth a comment: with 0conf, LDK re-broadcasts the splice tx as `TransactionType::Funding` (classified → no event); without, the acceptor only broadcasts `InteractiveFunding`, which classification skips → event fires. The cross-variant inconsistency in the acceptor's record may be worth flagging upstream.

tests/integration_tests_rust.rs
jkczyz, line 3116: The CI failure is here according to Claude:

This ordering is what failed CI (`build (self-hosted, beta)`) once the new assertion below started depending on it: `send()` only queues the broadcast (drained by a background task), and `wait_for_tx` after mining also passes for a mempool-only tx — so the tx can miss the first of the 6 blocks, sit below `ANTI_REORG_DELAY` at the sync, and never produce the event (`expect_onchain_payment_event` can't recover; single 60s `next_event_async`, tests/common/mod.rs:426-434). Swap these two lines, as `onchain_send_receive` does.

tests/integration_tests_rust.rs
```diff
@@ -784,6 +829,8 @@ async fn reorged_onchain_payment_returns_to_unconfirmed() {
 	generate_blocks_and_wait(&bitcoind.client, &electrsd.client, 1).await;
 	node_a.sync_wallets().unwrap();
 	node_b.sync_wallets().unwrap();
+	assert_eq!(node_a.next_event(), None);
+	assert_eq!(node_b.next_event(), None);
```
jkczyz, line 1019: Claude:

After the reorg sync below, also assert no event, then re-mine to `ANTI_REORG_DELAY` and assert exactly one event fires. Extended that way, it currently fails due to the stale-pending-entry duplicate (src/wallet/mod.rs:289).

tests/integration_tests_rust.rs
jkczyz, line 3889: Claude:

Nothing tests that a restarted node doesn't re-emit. I wrote the test — it passes, so coverage-only.

bindings/kotlin/ldk-node-jvm/lib/src/test/kotlin/org/lightningdevkit/ldknode/LibraryTest.kt
```diff
@@ -226,6 +226,28 @@ class LibraryTest {
         node1.syncWallets()
         node2.syncWallets()
 
+        val onchainPaymentReceivedEvent1 = node1.waitNextEvent()
+        println("Got event: $onchainPaymentReceivedEvent1")
+        when (onchainPaymentReceivedEvent1) {
+            is Event.OnchainPaymentReceived -> {
+                assertEquals(txid1, onchainPaymentReceivedEvent1.txid)
+                assertEquals(100000000uL, onchainPaymentReceivedEvent1.amountMsat)
+            }
```
jkczyz, line 235: How about testing `OnchainPaymentSuccessful`?

jkczyz:
Tests mentioned in claude feedback can be found in https://github.com/jkczyz/ldk-node/commit/a7c0269c7e564e1f6bf3a1c2d9dc27e95a96d810

src/data_store.rs
tnull: Done

src/event.rs
tnull: Done.

bindings/kotlin/ldk-node-jvm/lib/src/test/kotlin/org/lightningdevkit/ldknode/LibraryTest.kt
tnull: Done.

tests/common/mod.rs
tnull: Done.

src/event.rs

src/event.rs

tests/integration_tests_rust.rs
tnull: Done.

tests/integration_tests_rust.rs
tnull: Done.

src/wallet/mod.rs
tnull: Yeah, we applied a somewhat related fix in #948. Maybe it makes sense to first land that then, so we can follow/reuse the exact same pattern here.

tnull:
Hmm, given the reclassification inconsistencies we're still facing the issue that counterparty spends might wrongfully trigger an onchain payment event.

I now explored an alternative approach over at https://github.com/tnull/ldk-node/tree/2025-01-include-onchain-transactions-in-events-alt where we (ab)use the `KeychainKind`, i.e., discern LDK spends and onchain payments via `Internal`/`External` descriptors. That seems like a hacky workaround and we'd also still get misfiring events for addresses already handed out before the upgrade. So not quite sure if it's worth it yet. Will hold a bit longer until we're clear on if we ship further improvements for the upcoming release.

tnull:
Moving this off the v0.8 milestone. I increasingly think we need to reconsider how we do onchain transaction classification, and until we do it's probably not feasible to keep piling on masses of hacky code to keep working around the shortcomings of the central architectural choices.

PR #450, opened 2025-01-30
Minor follow-ups to #407

tnull:
Based on #407 until that is merged.

We add some minor cleanups here.

(cc @enigbe)


enigbe (approved):
Thanks for adding this. 
I have reviewed and tested. This LGTM.

PR #453, opened 2025-01-31
Network graph update error logging

joostjager:


src/gossip.rs
```diff
@@ -110,9 +110,11 @@ impl GossipSource {
 							Error::GossipUpdateFailed
 						})?;
 
-						let new_latest_sync_timestamp = gossip_sync
-							.update_network_graph(&update_data)
-							.map_err(|_| Error::GossipUpdateFailed)?;
+						let new_latest_sync_timestamp =
+							gossip_sync.update_network_graph(&update_data).map_err(|e| {
+								log_trace!(logger, "Failed to update network graph with RGS data: {:?}", e);
```
joostjager, line 115: Not sure is {:?} is desired? Otherwise need to add Display to (nested) errors.

PR #454, opened 2025-01-31
Fix `log`-facade logging and clarify max level

tnull:
This is another small follow-up to #407.

Previously, the log facade logger would always use a `logger` target but then *also* add the correct module paht and line as part of the normal log messages. Here, we override the log target correctly.

Moreover, we previously would log all messages at the configured *maximum* level, not the level in the log record. We therefore also fix this and rename `level` to `max_log_level` everywhere to clear up any confusions regarding what the variable does.

(cc @enigbe)

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -112,7 +112,7 @@ impl Default for LiquiditySourceConfig {
 #[derive(Clone)]
 enum LogWriterConfig {
 	File { log_file_path: Option<String>, max_log_level: Option<LogLevel> },
-	Log(LogLevel),
+	Log { max_log_level: Option<LogLevel> },
 	Custom(Arc<dyn LogWriter>),
 }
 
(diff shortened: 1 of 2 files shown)


enigbe (approved):
Thanks for catching this @tnull.

Part of the conversation we had when reviewing #407 was to add a follow-up that addressed testing the writers in ways that didn't clutter the existing tests. I will be adding those tests shortly but I have tested this with a `log` facade writer locally and it logs entries correctly.

src/builder.rs
```diff
@@ -1326,7 +1326,10 @@ fn setup_logger(
 			Logger::new_fs_writer(log_file_path, max_log_level)
 				.map_err(|_| BuildError::LoggerSetupFailed)?
 		},
-		Some(LogWriterConfig::Log(log_level)) => Logger::new_log_facade(*log_level),
+		Some(LogWriterConfig::Log { max_log_level }) => {
+			let max_log_level = max_log_level.unwrap_or_else(|| DEFAULT_LOG_LEVEL);
+			Logger::new_log_facade(max_log_level)
```
joostjager, line 1331: One thing that I noticed about the log facade is that you can set log level GOSSIP, but that isn't visible anymore after this PR. The standard log only knows TRACE of course. Not a problem I think?

And perhaps some explanation on how the `RUST_LOG` env var interacts with this log level would be helpful too. It wasn't very clear to me how that works.

joostjager (approved):
Tested and looks identical to the file logger output now.

Issue #441, opened 2025-01-23
Switch to `MonitorUpdatingPersister`


tnull:
Especially since we now added remote storage via VSS, we should aim to reduce the persisted data and switch to use [`MonitorUpdatingPersister`](https://docs.rs/lightning/latest/lightning/util/persist/struct.MonitorUpdatingPersister.html).

arturgontijo:
Hey @tnull, is anybody working on this one?

It looks like a low hanging fruit from that `v0.6` issues list. Am I right?
Asking as I'm very new to this codebase =).

arturgontijo:
Ok, will give it a try...if things don't go smoothly I'll ping you guys on Discord =)
Thanks

PR #456, opened 2025-02-01
Use `MonitorUpdatingPersister`

arturgontijo:
This PR aims to solve #441 , making use of `MonitorUpdatingPersister` in the `ChainMonitor`.

Note: Not sure about a proper value for its `maximum_pending_updates` (set it to `10` for now).

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -973,7 +973,7 @@ fn build_with_store_internal(
 	let persister = Arc::new(Persister::new(
 		Arc::clone(&kv_store),
 		Arc::clone(&logger),
-		10, // (?)
+		100,
 		Arc::clone(&keys_manager),
 		Arc::clone(&keys_manager),
 		Arc::clone(&tx_broadcaster),
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -26,7 +26,7 @@ use crate::peer_store::PeerStore;
 use crate::tx_broadcaster::TransactionBroadcaster;
 use crate::types::{
 	ChainMonitor, ChannelManager, DynStore, GossipSync, Graph, KeysManager, MessageRouter,
-	OnionMessenger, PeerManager,
+	OnionMessenger, PeerManager, Persister,
 };
 use crate::wallet::persist::KVStoreWalletPersister;
 use crate::wallet::Wallet;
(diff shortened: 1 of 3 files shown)


src/io/test_utils.rs
```diff
@@ -81,54 +86,104 @@ pub(crate) fn do_read_write_remove_list_persist<K: KVStore + RefUnwindSafe>(kv_s
...
+
+	let persister_1 = MonitorUpdatingPersister::new(
+		store_1,
+		&chanmon_cfgs[1].logger,
+		persister_1_max_pending_updates,
+		&chanmon_cfgs[1].keys_manager,
+		&chanmon_cfgs[1].keys_manager,
+		&chanmon_cfgs[1].tx_broadcaster,
+		&chanmon_cfgs[1].fee_estimator,
+	);
```
enigbe, line 114: Given the duplication, consider extracting this into a helper function.

src/builder.rs
```diff
@@ -932,6 +923,38 @@ fn build_with_store_internal(
 		Arc::clone(&logger),
 	));
 
+	let persister = Arc::new(Persister::new(
+		Arc::clone(&kv_store),
+		Arc::clone(&logger),
+		10, // (?)
```
enigbe, line 929: I also don't know what value to recommend here, but pending when @tnull takes a look, you could define a constant to hold this value.

arturgontijo:
Hey @enigbe and @tnull, thanks for the initial review.

Regarding `test_utils.rs`, I kept it as a copy&paste but now from [rust-lightning/.../persist.rs](https://github.com/lightningdevkit/rust-lightning/blob/main/lightning/src/util/persist.rs#L1148).
But I totally agree that, as it is part of this repo now, we should make it better =)

I'll address all the suggestions and also agree that it can bring side effects that we are not anticipating yet.

arturgontijo:
> @arturgontijo We now shipped 0.5, so could consider to move forward with this PR soon. It also needs a rebase by now.

Cool, let me resolve that conflict and then we can discuss next steps.

src/builder.rs
domZippilli: FWIW I agree with you @tnull.

Our node ran on 100 for a long time and is 1000 now, but you would probably want 100 for most people. I dunno your design philosophy about what to make configurable, but it's probably a small number of power users who would want something other than 100.

domZippilli:
Why was this closed?

PR #460, opened 2025-02-05
Add missing argument to `README` example

tnull:


joostjager (approved):
utACK

I do wonder if it would be possible to put this code in a `.rs` file and compile it, to prevent deviations? Then link from the readme.

joostjager:
Rust doc tests do not check that it is valid code, or is it doing something like that?

Having it duplicated, is that necessary. Maybe lib.rs can just point to the readme or vv ?

PR #462, opened 2025-02-06
Upgrade `develop` to follow LDK `main` (bump to 4e32d852)

tnull:
We bump the new `develop` branch to follow LDK's `main` branch (as of 4e32d852).

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -124,7 +124,7 @@ impl LogWriter for Writer {
 
 				let log = format!(
 					"{} {:<5} [{}:{}] {}\n",
-					Utc::now().format("%Y-%m-%d %H:%M:%S"),
+					Utc::now().format("%Y-%m-%d %H:%M:%S%.3f"),
 					record.level.to_string(),
 					record.module_path,
 					record.line,


src/liquidity.rs
```diff
@@ -273,14 +275,8 @@ impl<L: Deref> LiquiditySource<L>
 where
 	L::Target: LdkLogger,
 {
-	pub(crate) fn set_peer_manager(&self, peer_manager: Arc<PeerManager>) {
```
martinsaposnic, line 277: when running tests, they fail with "Failed to handle LSPS2ServiceEvent as peer manager isn't available. This should never happen."

this method should not be deleted and should look like this:
```rust
pub(crate) fn set_peer_manager(&self, peer_manager: Arc<PeerManager>) {
	*self.peer_manager.write().unwrap() = Some(Arc::clone(&peer_manager));
}
```

and the builder should do 
```rust
liquidity_source.as_ref().map(|l| l.set_peer_manager(Arc::clone(&peer_manager)));
```

when building the liquidity_source

tnull: Ah, thanks, good catch!

src/lib.rs
```diff
@@ -521,6 +521,9 @@ impl Node {
 		let background_chan_man = Arc::clone(&self.channel_manager);
 		let background_gossip_sync = self.gossip_source.as_gossip_sync();
 		let background_peer_man = Arc::clone(&self.peer_manager);
+		let background_liquidity_man_opt =
+			self.liquidity_source.as_ref().map(|ls| Arc::clone(&ls.liquidity_manager()));
```
joostjager: Unnecessary clone?

src/payment/bolt12.rs
```diff
@@ -82,7 +83,7 @@ impl Bolt12Payment {
 		rand::thread_rng().fill_bytes(&mut random_bytes);
 		let payment_id = PaymentId(random_bytes);
 		let retry_strategy = Retry::Timeout(LDK_PAYMENT_RETRY_TIMEOUT);
-		let max_total_routing_fee_msat = None;
+		let route_params_config = RouteParametersConfig::default();
```
joostjager, line 87: Side-question: is there no cap to the routing fee paid?

src/wallet/mod.rs
```diff
@@ -757,6 +695,86 @@ impl WalletSource for Wallet {
 	}
 }
 
+impl Listen for Wallet {
```
joostjager, line 698: The move makes it hard to review. I think nothing changed for `Listen`?

src/chain/electrum.rs
```diff
@@ -424,12 +424,7 @@ impl ElectrumRuntimeClient {
 				Error::ConnectionFailed
 			})?,
 		);
-		let electrum_client_2 =
-			ElectrumClient::from_config(&server_url, electrum_config).map_err(|e| {
-				log_error!(logger, "Failed to connect to electrum server: {}", e);
-				Error::ConnectionFailed
-			})?;
-		let bdk_electrum_client = Arc::new(BdkElectrumClient::new(electrum_client_2));
+		let bdk_electrum_client = Arc::new(BdkElectrumClient::new(Arc::clone(&electrum_client)));
```
joostjager, line 427: I think there was something upstream that allows reuse now, or is this an independent clean up?

src/payment/mod.rs
```diff
@@ -22,87 +22,3 @@ pub use store::{
 	ConfirmationStatus, LSPFeeLimits, PaymentDetails, PaymentDirection, PaymentKind, PaymentStatus,
 };
 pub use unified_qr::{QrPaymentResult, UnifiedQrPayment};
-
-/// Represents information used to send a payment.
-#[derive(Clone, Debug, PartialEq)]
-pub struct SendingParameters {
```
joostjager, line 28: First renaming this to `RouteParametersConfig` (the ldk-node one) in a separate commit may be helpful with review here. For next time.

src/payment/bolt11.rs
```diff
@@ -118,29 +110,27 @@ impl Bolt11Payment {
...
-				.max_total_cltv_expiry_delta
-				.map(|d| route_params.payment_params.max_total_cltv_expiry_delta = d);
-			override_params.max_path_count.map(|p| route_params.payment_params.max_path_count = p);
-			override_params
-				.max_channel_saturation_power_of_half
-				.map(|s| route_params.payment_params.max_channel_saturation_power_of_half = s);
-		};
+		let payment_hash = PaymentHash(invoice.payment_hash().to_byte_array());
+		let payment_id = PaymentId(invoice.payment_hash().to_byte_array());
+		if let Some(payment) = self.payment_store.get(&payment_id) {
```
joostjager, line 105: Getting the payment twice?

src/payment/bolt11.rs
```diff
@@ -798,18 +769,41 @@ impl Bolt11Payment {
...
 		let invoice = maybe_deref(invoice);
+		let payment_params = PaymentParameters::from_bolt11_invoice(invoice);
 
-		let (_payment_hash, _recipient_onion, route_params) = bolt11_payment::payment_parameters_from_invoice(&invoice).map_err(|_| {
+		let amount_msat = invoice.amount_milli_satoshis().ok_or_else(|| {
 			log_error!(self.logger, "Failed to send probes due to the given invoice being \"zero-amount\". Please use send_probes_using_amount instead.");
 			Error::InvalidInvoice
 		})?;
 
+		let mut route_params =
```
joostjager, line 780: `route_params` vs `route_parameters` is a bit confusing. Maybe a more descriptive name is possible?

src/lib.rs
```diff
@@ -1204,6 +1205,7 @@ impl Node {
 				chain_source.update_fee_rate_estimates().await?;
 				chain_source.poll_and_update_listeners(sync_cman, sync_cmon, sync_sweeper).await?;
 			}
+			let _ = sync_sweeper_2.regenerate_and_broadcast_spend_if_necessary().await;
```
joostjager: Maybe just call clone above instead of having this _2 instance?

joostjager (commented):
A few small remarks.

Again I noticed how much I would have wanted the ldk pr(s) alongside for better context. I know this isn't the first or second time I say this, but I strongly believe that a single repo would be really helpful.

src/lib.rs
tnull: Ah, true, with all the rebases it changed slightly what we do here. We can drop one. Done.

src/chain/electrum.rs
tnull: Yes, https://github.com/bitcoindevkit/rust-electrum-client/pull/163

src/payment/mod.rs
tnull: Hmm, from my point of view that would have been very/even more confusing. But granted it's not super easy to review.

src/payment/bolt11.rs
tnull: Ugh, good catch. Snuck in during one of the many rebase rounds I think.

src/chain/electrum.rs
joostjager: Initiated by you, very nice.

src/payment/bolt11.rs
tnull: Hmm, no, I don't think so. FWIW, it's only confusing internally as we have `RouteParameters` and `RouteParametersConfig` now, but that's really more an LDK problem.

src/payment/mod.rs
joostjager: Yes, it isn't always easy to reach review optimality. I also agree that the rename commit is a bit weird, but it would extract a lot of mechanical change from the subsequent commit.

src/lib.rs
tnull: Sure, done.

src/payment/bolt11.rs
joostjager: I meant that there are two variables with ~ the same name in that fn.

martinsaposnic:
LGTM post merge.

Do we need a CONTRIBUTING.md explaning the new developer flow?

this is excellent, I have a few things to do in this repo and this makes it easier. Cheers!

src/chain/bitcoind.rs
```diff
@@ -1381,11 +1381,11 @@ impl Listen for ChainListener {
...
-	fn block_disconnected(&self, header: &bitcoin::block::Header, height: u32) {
-		self.onchain_wallet.block_disconnected(header, height);
-		self.channel_manager.block_disconnected(header, height);
-		self.chain_monitor.block_disconnected(header, height);
-		self.output_sweeper.block_disconnected(header, height);
+	fn blocks_disconnected(&self, fork_point_block: lightning::chain::BestBlock) {
+		self.onchain_wallet.blocks_disconnected(fork_point_block);
+		self.channel_manager.blocks_disconnected(fork_point_block);
+		self.chain_monitor.blocks_disconnected(fork_point_block);
+		self.output_sweeper.blocks_disconnected(fork_point_block);
```
martinsaposnic, line 1388: actually I missed this one. 

this is the only change that is breaking stuff for me locally. why was this changed from `block_disconnected` to `blocks_disconnected`?

martinsaposnic: nevermind, I was using an outdated ldk commit on ldk-node 🤦

PR #468, opened 2025-02-12
Fix amount calculation and track paid fees in `PaymentKind::Onchain`

tnull:
We recently started tracking onchain payments in the store. It turns out the calculated amount was slightly off as in the outbound case it would include the paid fees. Here, we fix this minor bug and start tracking the fees separately.

### DIFF
--- a/src/payment/store.rs
+++ b/src/payment/store.rs
@@ -42,6 +42,13 @@ pub struct PaymentDetails {
 	pub kind: PaymentKind,
 	/// The amount transferred.
 	pub amount_msat: Option<u64>,
+	/// The fees that were paid for this payment.
+	///
+	/// For Lightning payments, this will only be updated for outbound payments once they
+	/// succeeded.
+	///
+	/// Will be `None` for Lightning payments made with LDK Node v0.4.x and earlier.
+	pub fee_paid_msat: Option<u64>,
 	/// The direction of the payment.
 	pub direction: PaymentDirection,
 	/// The status of the payment.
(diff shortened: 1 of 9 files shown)


src/payment/store.rs
```diff
@@ -302,6 +302,8 @@ pub enum PaymentKind {
 		txid: Txid,
 		/// The confirmation status of this payment.
 		status: ConfirmationStatus,
+		/// The on-chain fees that were paid for this payment.
+		fee_msat: u64,
```
jkczyz, line 306: Any reason we track fees for on-chain payments but not for lightning payments? I suppose we could for outbound lightning payments, but IIUC we can't know for inbound payments.

Also, should we use sats here instead of msats? Or is this to align with msats used in `PaymentDetails`?

jkczyz: LGTM. Seems it would be useful to include.

tests/integration_tests_rust.rs
```diff
@@ -375,6 +376,29 @@ fn onchain_spend_receive() {
 		node_b.list_payments_with_filter(|p| matches!(p.kind, PaymentKind::Onchain { .. }));
 	assert_eq!(node_b_payments.len(), 3);
 
+	let payment_id = PaymentId(txid.to_byte_array());
+	let payment_a = node_a.payment(&payment_id).unwrap();
+	assert_eq!(payment_a.amount_msat, Some(amount_to_send_sats * 1000));
+	assert!(payment_a.fee_paid_msat > Some(0));
```
jkczyz, line 382: Could you assert that this is `None` when the payment is pending? Maybe add fee checks for lightning payments, too.

jkczyz: Oh, right. Should we check it earlier?

jkczyz (dismissed):

tests/common/mod.rs
```diff
@@ -148,6 +148,8 @@ macro_rules! expect_payment_successful_event {
 				if let Some(fee_msat) = $fee_paid_msat {
 					assert_eq!(fee_paid_msat, fee_msat);
 				}
+				let payment = $node.payment(&$payment_id.unwrap()).unwrap();
+				assert_eq!(payment.fee_paid_msat, fee_paid_msat);
```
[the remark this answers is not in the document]
jkczyz: Ah, sorry. I think that I assumed `expect_payment_successful_event` was only used for lightning payments. We could condition the assertion on the `PaymentKind` not being on-chain, maybe?

jkczyz (approved):

PR #469, opened 2025-02-12
Implement MigratableKVStore for VssStore

G8XSU:
* Implement MigratableKVStore for VssStore


enigbe (commented):
@G8XSU I have reviewed the implementation done here and it LGTM. I did some local testing to migrate from `VssStore` to `FilesystemStore` (also impl `MigratableKVStore`) and it works as it should.

Unrelated to this PR, but I noticed that the unit tests in `src/io/vss_store` require `vss-server` to be running at `TEST_VSS_BASE_URL`. Should we mock `vss-server` responses to keep them as unit tests, or should we move them to integration tests instead?

PR #471, opened 2025-02-13
Recalculate listeners' best block heights on each retry

tnull:
Previously, we would calculate each listener's best block height outside of the `loop`, which will have us retrying `init::synchronize_listeners` with stale data, potentially leading to panicking as we might try to connect blocks from the original height to listeners that might have been advanced by the previous itertation of the `loop` before hitting an error. Here, we mitigate this by simply retrieving the current best block height for each listener at the start of each iteration.

### DIFF
--- a/src/chain/mod.rs
+++ b/src/chain/mod.rs
@@ -281,48 +281,49 @@ impl ChainSource {
 					}
 				}
 
-				let channel_manager_best_block_hash =
-					channel_manager.current_best_block().block_hash;
-				let sweeper_best_block_hash = output_sweeper.current_best_block().block_hash;
-				let onchain_wallet_best_block_hash = onchain_wallet.current_best_block().block_hash;
-
-				let mut chain_listeners = vec![
-					(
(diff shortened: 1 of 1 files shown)


src/chain/mod.rs
```diff
@@ -281,48 +281,49 @@ impl ChainSource {
...
+						(
+							onchain_wallet_best_block_hash,
+							&**onchain_wallet as &(dyn Listen + Send + Sync),
+						),
+						(
+							channel_manager_best_block_hash,
+							&*channel_manager as &(dyn Listen + Send + Sync),
+						),
+						(sweeper_best_block_hash, &*output_sweeper as &(dyn Listen + Send + Sync)),
+					];
```
jkczyz, line 306: Discussed offline that making `current_best_block` a method on `Listen` would be useful. It simplifies the API and would avoid needing to allocate a `Vec` on each iteration.

jkczyz (approved):

PR #486, opened 2025-03-17
Add support for sourcing chain data from Electrum

### LINKED ISSUE #196: Add Electrum support
We will add Electrum support. 

- On the LDK side a prerequisite is https://github.com/lightningdevkit/rust-lightning/pull/2685. 
- On the BDK side we're blocked on #195

tnull:
Closes #196.

So far, we support Esplora and Bitcoind RPC chain sources. In this PR we add Electrum support based on the blocking `rust-electrum-client`, and it's `bdk_electrum` and `lightning-transaction-sync` counterparts. 

Due to the blocking nature of `rust-electrum-client` and as it directly connects to the server upon `Client::new` (see https://github.com/bitcoindevkit/rust-electrum-client/issues/166), we ended up wrapping the runtime-specific behavior in an `ElectrumRuntimeClient` object that is initialized and dropped in `Node::start` and `stop`, respectively.

~~One thing missing that we still need to consider is how we'd reestablish connections to the remote after they have been lost for one reason or another. IMO, that behavior should live in `rust-electrum-client` to avoid all users having to duplicate it, so it's pending resolution of https://github.com/bitcoindevkit/rust-electrum-client/issues/165~~

As we did with `bitcoind`-RPC,  Electrum support is tested by adding another `full_cycle` integration test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ fn main() {
 LDK Node currently comes with a decidedly opinionated set of design choices:
 
 - On-chain data is handled by the integrated [BDK][bdk] wallet.
-- Chain data may currently be sourced from the Bitcoin Core RPC interface or an [Esplora][esplora] server, while support for Electrum will follow soon.
+- Chain data may currently be sourced from the Bitcoin Core RPC interface, or from an [Electrum][electrum] or [Esplora][esplora] server.
 - Wallet and channel state may be persisted to an [SQLite][sqlite] database, to file system, or to a custom back-end to be implemented by the user.
 - Gossip data may be sourced via Lightning's peer-to-peer network or the [Rapid Gossip Sync](https://docs.rs/lightning-rapid-gossip-sync/*/lightning_rapid_gossip_sync/) protocol.
 - Entropy for the Lightning and on-chain wallets may be sourced from raw bytes or a [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) mnemonic. In addition, LDK Node offers the means to generate and persist the entropy bytes to disk.
@@ -72,6 +72,7 @@ The Minimum Supported Rust Version (MSRV) is currently 1.75.0.
 [rust_crate]: https://crates.io/
 [ldk]: https://lightningdevkit.org/
 [bdk]: https://bitcoindevkit.org/
+[electrum]: https://github.com/spesmilo/electrum-protocol
 [esplora]: https://github.com/Blockstream/esplora
 [sqlite]: https://sqlite.org/
 [rust]: https://www.rust-lang.org/


src/chain/electrum.rs
```diff
@@ -0,0 +1,303 @@
...
+
+		match timeout_fut.await {
+			Ok(res) => match res {
+				Ok(_) => {
+					log_trace!(self.logger, "Successfully broadcast transaction {}", txid);
+				},
+				Err(e) => {
+					log_error!(self.logger, "Failed to broadcast transaction {}: {}", txid, e);
+					log_trace!(
+						self.logger,
```
elnosh, line 190: For the broadcast method in the `BitcoindRpcClient`, it returns a `Result` and then logs the errors when called in `process_broadcast_queue` for `ChainSource`. Should this do that as well and return a `Result` and then log them in that method?

tests/integration_tests_rust.rs
```diff
@@ -43,6 +43,14 @@ fn channel_full_cycle() {
...
 }
 
+#[test]
+fn channel_full_cycle_electrum() {
+	let (bitcoind, electrsd) = setup_bitcoind_and_electrsd();
+	let chain_source = TestChainSource::Electrum(&electrsd);
+	let (node_a, node_b) = setup_two_nodes(&chain_source, false, true, false);
+	do_channel_full_cycle(node_a, node_b, &bitcoind.client, &electrsd.client, false, true, false);
+}
+
```
elnosh, line 53: would it be good to also add tests in `onchain_send_receive` for nodes using `TestChainSource::Electrum` since that functionality depends on the `ChainSource` used?

elnosh (commented):
I'm familiarizing myself with the codebase but I tried running a node locally with the changes here pointing to an electrum server and basic operations are working. I just have a couple of questions, if I may, to get a better understanding.

Not strictly related to the PR but maybe I can sneak it in, why is it that for bitcoin core rpc and electrum they are in their separate `bitcoind_rpc.rs` and `electrum.rs` files but for esplora it is all in the `mod.rs`? I see that that the `EsploraAsyncClient` is directly used without a wrapper but wanted to see what were the reasons for it (:

src/chain/mod.rs
```diff
@@ -236,13 +242,36 @@ impl ChainSource {
...
+					}
+				}
+
+				{
+					let registered_outputs_cache =
+						core::mem::take(&mut *registered_outputs_cache.lock().unwrap());
+					for output in registered_outputs_cache {
+						client.register_output(output)
+					}
+				}
```
enigbe, line 277: The `registered_txs_cache` and `registered_outputs_cache` are in-memory types. This means that every time we stop and then restart the `Node`, these fields will be empty. Doesn't this mean that any previously cached entries would be lost when we stop, and upon restart, we'd have no transactions or outputs to iterate through and then register?

src/chain/electrum.rs
```diff
@@ -0,0 +1,303 @@
...
+impl ElectrumRuntimeClient {
+	pub(crate) fn new(
+		server_url: String, runtime: Arc<tokio::runtime::Runtime>, config: Arc<Config>,
+		logger: Arc<Logger>,
+	) -> Result<Self, Error> {
+		let electrum_client = Arc::new(ElectrumClient::new(&server_url).map_err(|e| {
+			log_error!(logger, "Failed to connect to electrum server: {}", e);
+			Error::ConnectionFailed
+		})?);
+		let electrum_client_2 = ElectrumClient::new(&server_url).map_err(|e| {
```
joostjager, line 59: Perhaps something better exists than `_2`? Or make the first one `_1`?

Would you be able to use the same `ElectrumClient`?

src/builder.rs
```diff
@@ -976,6 +999,20 @@ fn build_with_store_internal(
 				Arc::clone(&node_metrics),
 			))
 		},
+		Some(ChainDataSourceConfig::Electrum { server_url, sync_config }) => {
+			let sync_config = sync_config.unwrap_or(ElectrumSyncConfig::default());
+			Arc::new(ChainSource::new_electrum(
```
joostjager, line 1004: (from commit msg, `Runtime` available)
>This might change in the future, but for now we do need this workaround.

What would be required for this potential change?

src/chain/mod.rs
```diff
@@ -167,6 +183,31 @@ impl ChainSource {
 		}
 	}
 
+	pub(crate) fn new_electrum(
+		server_url: String, sync_config: ElectrumSyncConfig, onchain_wallet: Arc<Wallet>,
+		fee_estimator: Arc<OnchainFeeEstimator>, tx_broadcaster: Arc<Broadcaster>,
+		kv_store: Arc<DynStore>, config: Arc<Config>, logger: Arc<Logger>,
+		node_metrics: Arc<RwLock<NodeMetrics>>,
+	) -> Self {
+		let electrum_runtime_client = RwLock::new(None);
```
joostjager, line 194: Add comment that this is initialized to None so that it can be instantiated when the runtime is available, and why?

src/chain/mod.rs
```diff
@@ -193,6 +234,35 @@ impl ChainSource {
 		}
 	}
 
+	pub(crate) fn start(&self, runtime: Arc<tokio::runtime::Runtime>) -> Result<(), Error> {
+		match self {
+			Self::Electrum { server_url, electrum_runtime_client, logger, .. } => {
```
joostjager: Is it the convention to not use an if statement here?

src/chain/mod.rs
```diff
@@ -128,6 +128,8 @@ pub(crate) enum ChainSource {
 		server_url: String,
 		sync_config: ElectrumSyncConfig,
 		electrum_runtime_client: RwLock<Option<ElectrumRuntimeClient>>,
+		registered_txs_cache: Mutex<Vec<(Txid, ScriptBuf)>>,
+		registered_outputs_cache: Mutex<Vec<WatchedOutput>>,
```
joostjager, line 132: The naming `cache` confused me a bit for the pre-start txs/outs. Perhaps there is a better name.

src/chain/mod.rs
```diff
@@ -304,73 +304,88 @@ impl ChainSource {
...
-
-					let lightning_wallet_sync_interval_secs = background_sync_config
-						.lightning_wallet_sync_interval_secs
-						.max(WALLET_SYNC_INTERVAL_MINIMUM_SECS);
-					let mut lightning_wallet_sync_interval = tokio::time::interval(
-						Duration::from_secs(lightning_wallet_sync_interval_secs),
-					);
-					lightning_wallet_sync_interval
-						.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+		macro_rules! start_tx_based_sync_loop {
```
joostjager, line 312: It seems ldk devs aren't all equally enthusiastic about macros. Some even actively refactoring them out.

Can this be expressed with a normal fn?

src/chain/electrum.rs
```diff
@@ -48,6 +51,48 @@ impl ElectrumRuntimeClient {
...
 	}
+
+	pub(crate) async fn broadcast(&self, tx: Transaction) {
+		let electrum_client = Arc::clone(&self.electrum_client);
+
+		let txid = tx.compute_txid();
+		let tx_bytes = tx.encode();
+
+		let spawn_fut =
+			self.runtime.spawn_blocking(move || electrum_client.transaction_broadcast(&tx));
```
joostjager, line 177: General question - is the thread pool for spawn_blocking large enough for everything that ldk_node may require (concurrently)?

src/chain/electrum.rs
```diff
@@ -0,0 +1,48 @@
...
+
+pub(crate) struct ElectrumRuntimeClient {
+	electrum_client: Arc<ElectrumClient>,
+	bdk_electrum_client: Arc<BdkElectrumClient<ElectrumClient>>,
+	tx_sync: Arc<ElectrumSyncClient<Arc<Logger>>>,
+	runtime: Arc<tokio::runtime::Runtime>,
+	logger: Arc<Logger>,
+}
+
+impl ElectrumRuntimeClient {
```
joostjager, line 50: Is something like `tokio::runtime::Handle::current()` useable to still allow init in the builder without a runtime passed in?

src/chain/mod.rs
```diff
@@ -658,7 +658,95 @@ impl ChainSource {
...
+						Error::WalletOperationFailed
+					})?;
+				}
+
+				// If this is our first sync, do a full scan with the configured gap limit.
+				// Otherwise just do an incremental sync.
+				let incremental_sync =
+					node_metrics.read().unwrap().latest_onchain_wallet_sync_timestamp.is_some();
+
+				macro_rules! get_and_apply_wallet_update {
```
joostjager, line 698: Macro necessary?

src/chain/mod.rs
```diff
@@ -1144,15 +1491,29 @@ impl ChainSource {
...
 impl Filter for ChainSource {
-	fn register_tx(&self, txid: &bitcoin::Txid, script_pubkey: &bitcoin::Script) {
+	fn register_tx(&self, txid: &Txid, script_pubkey: &Script) {
 		match self {
 			Self::Esplora { tx_sync, .. } => tx_sync.register_tx(txid, script_pubkey),
+			Self::Electrum { electrum_runtime_client, registered_txs_cache, .. } => {
+				match electrum_runtime_client.read().unwrap().as_ref() {
+					Some(client) => client.register_tx(txid, script_pubkey),
+					None => {
+						registered_txs_cache.lock().unwrap().push((*txid, script_pubkey.to_owned()))
```
joostjager, line 1501: Is it fully safe with the two locks?

src/chain/electrum.rs
```diff
@@ -0,0 +1,48 @@
...
+	pub(crate) fn new(
+		server_url: String, runtime: Arc<tokio::runtime::Runtime>, logger: Arc<Logger>,
+	) -> Result<Self, Error> {
+		let electrum_client = Arc::new(ElectrumClient::new(&server_url).map_err(|e| {
```
joostjager, line 55: You mention in the PR desc that conn reestablish is missing still. How would ldk-node behave when that happens?

joostjager (commented):
The question that I asked myself most during review is whether the way this PR deals with the runtime and start/stop is really the best option for now.

src/chain/mod.rs
joostjager: Something like queued, buffered or pending would - for me - match better what it is. But I guess that it depends on the definition of cache.

src/chain/electrum.rs
joostjager: Oh that's a lot. I had read somewhere it was 2*cpu and that seemed low.

src/chain/mod.rs
joostjager: Just asking the question to be sure. I think in `start` they are taken in the same order, so then it should be good?

src/chain/electrum.rs
joostjager: Could this be a reason to not yet merge? Perhaps a hard exit for now could help avoid users not being aware that they aren't watching the chain anymore.

src/builder.rs
joostjager: I think I misinterpreted 'this' in the commit message then. I thought it referred to getting the runtime available in the builder. Could perhaps be made explicit. Or could also be clear enough for people with enough understanding of the code and async.

src/chain/electrum.rs
joostjager: If it clears up the code, why wait for the refactoring? You'd need to revise the code that is now newly added. And that potential refactoring may never happen especially if it becomes bigger and bigger.

src/chain/mod.rs
joostjager: If you'd just do the await in the future in each arm, and then `Into` it into an `Update`, couldn't the rest just live in a function?

src/chain/electrum.rs
joostjager: Isn't it the case that in the main `start` just a single runtime is created and used everywhere?

src/chain/mod.rs
joostjager: Checked in more detail, and indeed seems safe. It would be better perhaps if this kind of analysis wasn't necessary. Nested locks are a risk in general I think. But not sure if there is a way to do it in rust.

src/chain/electrum.rs
joostjager: It's still not entirely clear to me how it could happen that the current handle is different from the one passed into the main start method. But I can see your argument that this is just how to works now.

src/chain/electrum.rs
joostjager: Can understand you'd rather not wait, but does seem important to know and document what's going to happen in the mean time before merging this PR?

src/chain/mod.rs
joostjager: How about using just one lock, locking a struct containing `electrum_runtime_client` and the pending txs/outs?

src/chain/mod.rs
joostjager: We tried it out offline. It seems that `into()` may be a workable way to do it after all.

src/chain/mod.rs
```diff
@@ -538,6 +679,118 @@ impl ChainSource {
...
+					})?;
+				}
+
+				// If this is our first sync, do a full scan with the configured gap limit.
+				// Otherwise just do an incremental sync.
+				let incremental_sync =
+					node_metrics.read().unwrap().latest_onchain_wallet_sync_timestamp.is_some();
+
+				fn apply_wallet_update(
+					update_res: Result<BdkUpdate, Error>, incremental_sync: bool, now: Instant,
```
joostjager, line 720: Can this work as a closure too with fewer params? Some of these parameters don't seem to be used further down in the function, so maybe there is no problem with move/borrow. Not sure how it works exactly with rust.

src/chain/electrum.rs
```diff
@@ -0,0 +1,303 @@
...
+		let mut batch = Batch::default();
+		let confirmation_targets = get_all_conf_targets();
+		for target in confirmation_targets {
+			let num_blocks = get_num_block_defaults_for_target(target);
+			batch.estimate_fee(num_blocks);
+		}
+
+		let spawn_fut = self.runtime.spawn_blocking(move || electrum_client.batch_call(&batch));
+
+		let timeout_fut = tokio::time::timeout(
```
joostjager, line 226: Is it really necessary to have these timeouts here and is there nothing in electrum client? It surprises me because apparently retries are implemented in the client code, but not timeouts?

src/chain/electrum.rs
```diff
@@ -0,0 +1,303 @@
...
+			// to continue on `signet`/`regtest` where we might not get estimates (or bogus
+			// values).
+			let fee_rate_btc_per_kvb = raw_fee_rate_btc_per_kvb
+				.as_f64()
+				.map_or(0.00001, |converted| converted.max(0.00001));
+
+			// Electrum, just like Bitcoin Core, gives us a feerate in BTC/KvB.
+			// Thus, we multiply by 25_000_000 (10^8 / 4) to get satoshis/kwu.
+			let fee_rate = {
+				let fee_rate_sat_per_kwu = (fee_rate_btc_per_kvb * 25_000_000.0).round() as u64;
```
joostjager, line 274: No risk here in rounding (down)?

src/chain/mod.rs
```diff
@@ -193,6 +243,63 @@ impl ChainSource {
...
+			} => {
+				let mut locked_client = electrum_runtime_client.write().unwrap();
+				let client = ElectrumRuntimeClient::new(
+					server_url.clone(),
+					runtime,
+					Arc::clone(&config),
+					Arc::clone(&logger),
+				)?;
+
+				// Apply any previously-cached `Filter` entries
```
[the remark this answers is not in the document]
joostjager: Just 'apply any pending entries', that's correct I think?

src/chain/mod.rs
```diff
@@ -193,6 +243,63 @@ impl ChainSource {
...
+				{
+					let pending_registered_txs =
+						core::mem::take(&mut *pending_registered_txs.lock().unwrap());
+					for (txid, script_pubkey) in pending_registered_txs {
+						client.register_tx(&txid, &script_pubkey);
+					}
+				}
+
+				{
```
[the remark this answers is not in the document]
joostjager: Or https://github.com/lightningdevkit/ldk-node/pull/486#discussion_r2024620408 ?

scripts/download_bitcoind_electrs.sh
```diff
@@ -1,3 +1,6 @@
+#!/bin/bash
+set -eox pipefail
```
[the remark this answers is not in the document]
joostjager: Ah, the -x prints the commands. That's def useful. And the rest of early exit flags makes sense ofc

src/chain/mod.rs
```diff
@@ -193,6 +243,63 @@ impl ChainSource {
...
+			},
+		}
+		Ok(())
+	}
+
+	pub(crate) fn stop(&self) {
+		match self {
+			Self::Electrum { electrum_runtime_client, .. } => {
+				let mut locked_client = electrum_runtime_client.write().unwrap();
+				*locked_client = None;
```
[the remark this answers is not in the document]
joostjager: It then recursively drops the inner electrum clients, which close the tcp connection?

src/chain/electrum.rs
joostjager: Isn't electrum different because it isn't an async library?

I think locking up worker threads can still happen because the async task may finish on the timeout, but the worker thread remains blocked? It isn't cancelled.

src/chain/mod.rs
```diff
@@ -122,6 +125,22 @@ pub(crate) enum ChainSource {
...
+		pending_registered_txs: Mutex<Vec<(Txid, ScriptBuf)>>,
+		pending_registered_outputs: Mutex<Vec<WatchedOutput>>,
+		onchain_wallet: Arc<Wallet>,
+		onchain_wallet_sync_status: Mutex<WalletSyncStatus>,
+		lightning_wallet_sync_status: Mutex<WalletSyncStatus>,
+		fee_estimator: Arc<OnchainFeeEstimator>,
+		tx_broadcaster: Arc<Broadcaster>,
+		kv_store: Arc<DynStore>,
+		config: Arc<Config>,
+		logger: Arc<Logger>,
```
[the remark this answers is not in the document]
joostjager: Talked about this offline. Could result in better code, but also requires rebase of the many commits in this PR and delay in landing the feature.

src/chain/mod.rs
```diff
@@ -538,6 +679,118 @@ impl ChainSource {
...
+								}
+								Ok(())
+							},
+							Err(e) => Err(e),
+						},
+						Err(e) => Err(e),
+					}
+				}
+
+				let cached_txs = onchain_wallet.get_cached_txs();
```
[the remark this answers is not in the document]
joostjager: Got it, thanks

src/chain/mod.rs
```diff
@@ -875,6 +1203,50 @@ impl ChainSource {
...
+						debug_assert!(
+							false,
+							"We should have started the chain source before updating fees"
+						);
+						return Err(Error::FeerateEstimationUpdateFailed);
+					};
+
+				let now = Instant::now();
+
+				let new_fee_rate_cache = electrum_client.get_fee_rate_cache_update().await?;
```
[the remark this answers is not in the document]
joostjager: I think in some cases it bubbles up before syncing the ln and onchain wallets, and then bubbles up further so that the syncing doesn't happen. Is that fine?

src/chain/electrum.rs
joostjager: Ah yes it's kwu. I was worried about somehow rounding down to zero. But I suppose with that precision it doesn't matter which way to round?

src/chain/electrum.rs
```diff
@@ -0,0 +1,303 @@
...
+		}
+
+		let mut new_fee_rate_cache = HashMap::with_capacity(10);
+		for (target, raw_fee_rate_btc_per_kvb) in
+			confirmation_targets.into_iter().zip(raw_estimates_btc_kvb.into_iter())
+		{
+			// Parse the retrieved serde_json::Value and fall back to 1 sat/vb (10^3 / 10^8 = 10^-5
+			// = 0.00001 btc/kvb) if we fail or it yields less than that. This is mostly necessary
+			// to continue on `signet`/`regtest` where we might not get estimates (or bogus
+			// values).
```
joostjager, line 266: Not get estimates means that it is in the `raw_estimates_btc_kvb` vec, but an empty value, or a non-float value (bogus)?

src/chain/electrum.rs
```diff
@@ -0,0 +1,303 @@
...
+			let fee_rate_btc_per_kvb = raw_fee_rate_btc_per_kvb
+				.as_f64()
+				.map_or(0.00001, |converted| converted.max(0.00001));
```
joostjager, line 269: Don't you want the minimum here?

src/chain/mod.rs
```diff
@@ -207,66 +314,41 @@ impl ChainSource {
...
-					fee_rate_update_interval.reset();
-					fee_rate_update_interval
-						.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
-
-					let lightning_wallet_sync_interval_secs = background_sync_config
-						.lightning_wallet_sync_interval_secs
-						.max(WALLET_SYNC_INTERVAL_MINIMUM_SECS);
-					let mut lightning_wallet_sync_interval = tokio::time::interval(
-						Duration::from_secs(lightning_wallet_sync_interval_secs),
+				if let Some(background_sync_config) = sync_config.background_sync_config.as_ref() {
```
joostjager, line 317: This extract could have been an independent commit. Non-blocking.

src/chain/mod.rs
```diff
@@ -538,6 +679,118 @@ impl ChainSource {
...
+					if let Some(client) = electrum_runtime_client.read().unwrap().as_ref() {
+						Arc::clone(client)
+					} else {
+						debug_assert!(
+							false,
+							"We should have started the chain source before syncing the onchain wallet"
+						);
+						return Err(Error::FeerateEstimationUpdateFailed);
+					};
+				let receiver_res = {
```
joostjager, line 701: This code block including the if below seems to be identical across all backends, and backend independent. Maybe ok for now, but avoiding code duplication may be better.

src/chain/mod.rs
```diff
@@ -637,6 +890,76 @@ impl ChainSource {
...
+				let sync_cman = Arc::clone(&channel_manager);
+				let sync_cmon = Arc::clone(&chain_monitor);
+				let sync_sweeper = Arc::clone(&output_sweeper);
+				let confirmables = vec![
+					sync_cman as Arc<dyn Confirm + Sync + Send>,
+					sync_cmon as Arc<dyn Confirm + Sync + Send>,
+					sync_sweeper as Arc<dyn Confirm + Sync + Send>,
+				];
+
+				let receiver_res = {
```
joostjager, line 921: A similar repeated block that also exists for esplora.

src/chain/mod.rs
```diff
@@ -875,6 +1203,50 @@ impl ChainSource {
...
+				let new_fee_rate_cache = electrum_client.get_fee_rate_cache_update().await?;
+				fee_estimator.set_fee_rate_cache(new_fee_rate_cache);
+
+				log_info!(
+					logger,
+					"Fee rate cache update finished in {}ms.",
+					now.elapsed().as_millis()
+				);
+
+				let unix_time_secs_opt =
```
joostjager, line 1236: Another candidate maybe for extraction. Some helper function that updates a timestamp to now and writes it out.

joostjager (commented):
Few final non-blocking things about code re-use.

src/chain/mod.rs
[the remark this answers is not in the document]
joostjager: I don't insist, just to make it consistent with the var rename.

src/chain/electrum.rs
joostjager: Is there a point though in having an async timeout wrapper if the spawn thread would hangs anyway?

src/chain/mod.rs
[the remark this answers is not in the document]
joostjager: If you trace up the call stack, I get to code like this:

```rust
ChainSource::Electrum { .. } => {
	chain_source.update_fee_rate_estimates().await?;
	chain_source
		.sync_lightning_wallet(sync_cman, sync_cmon, sync_sweeper)
		.await?;
	chain_source.sync_onchain_wallet().await?;
},
```

If the update of fee rates fails, the other syncs aren't executed either?

src/chain/mod.rs
joostjager: I mean the extraction of code into `start_tx_based_sync_loop`. In a prefactor commit, this could be done for just Esplora. And then later electrum is reusing it.

src/chain/mod.rs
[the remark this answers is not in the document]
joostjager: That's the question. Don't you want to sync the wallets even when something goes wrong with the fee update?

src/chain/mod.rs
joostjager: Still fine

src/chain/mod.rs
joostjager: Yes, this looks great! Also nice use of the enum that doesn't even contain pending txes if it is started. Educational for me!

src/chain/electrum.rs
joostjager: I see you also set retries now. I saw that it was 1 previously (the default). Did the earlier test with the wifi disconnect pass because that one retry happened after you already reconnected wifi?

If I understand this correctly, setting the timeout on the electrum client itself is essential to not end up with a potential graveyard of stuck threads? This might be important to add as a comment.

Also curious to hear why you think that retrying over and over again when a previous request is still stuck is a good idea? It will eventually exhaust the thread pool, right? And then stop independent processes within LDK too?

The original reason for the double timeout is because esplora-client turned out to be not trustworthy. But, as suggested above, isn't electrum different because it isn't an async library and it is unlikely for the timeout to fail? Looking at the sources, it seems that the timeout is set on a pretty low level for all calls.

joostjager (commented):
Nearly there. Just the one question about the double timeouts.

src/chain/electrum.rs
joostjager: Talked about this offline. Difficult to determine whether the double timeouts are net positive or not, but leaving them in for now.

enigbe (approved):
I have taken another look at this PR and now understand why you apply pending filter entries to the pending transaction and output vectors. The refactor that introduced `ElectrumRuntimeStatus` made it easier to reason about it as a state machine with the attendant events, actions, and state or internal transitions. 

state machine


From the snippet below, when we [start](https://github.com/lightningdevkit/ldk-node/blob/5abb42faec79e38c4775f7b107aaf372a5a2f604/src/lib.rs#L240-L245) the `Node` with the tokio runtime, it indicates that other background tasks do not get started before the `ChainSource` has started:

```rust
pub fn start_with_runtime(&self, runtime: Arc<tokio::runtime::Runtime>) -> Result<(), Error> {

        ... 

        // Start up any runtime-dependant chain sources (e.g. Electrum)

        self.chain_source.start(Arc::clone(&runtime)).map_err(|e| {

            log_error!(self.logger, "Failed to start chain syncing: {}", e);

            e

        })?;
        ...
       // Other background tasks
```

This suggests that it is unlikely that `Confirm`ables would register transactions and outputs prior the chain source starting and, in extension, prior to the runtime client being available (`ElectrumRuntimeStatus::Started`). Are there scenarios I am missing where this could happen? I'd appreciate any further clarification.

Otherwise, the PR looks good to me. Reviewing this has been quite educational. Thank you!

PR #490, opened 2025-03-17
Make Address optional for Opening Channels

npslaney:
Make address optional for open_channel and open_announced_channel if we're already connected to the peer

This takes care of the somewhat frequent case where when we're running a node, we want to open a new channel to a peer we're already connected to, it's a bit tedious to always have to bring the address.

Right now, just pulling the address from peer_manager and peer_store. Could also pull from the graph, but not sure if we'd need to go that far.


npslaney:
Yeah I see what you're saying on the first time thing. It's a nice to have but also able to be implemented at the user level. I'm good to close this.

Issue #381, opened 2024-10-17
Address remaining panic-on-persistence failures


tnull:
In #374  we fixed the panics in event handling, but there are still some others that are unaddressed. We still need to fix them, which is important for any kind of remote persistence, especially VSS.

Note that many of the -on-persistence-failure-panics are happening in upstream code or are blocked by upstream developments. So this is marked 'blocked on upstream' for now.

tnull:
In particular, this will require fully switching over to `KVStore` (i.e., including `ChannelMonitor` persistence) which we intend to  do for v0.8.

tnull:
This was mostly addressed by #919, but there are still remaining cases we'd need to address to fully resolve this. For the local ones we'd need some kind of wrapper for the node that cleanly shuts down and attempts restart whenever we hit a persistence failure (after retries for VSS, ofc). We also still have `UnrecoverableError` variants upstream that we'd need to drop to fully resolve this.

>   - P0, easy: make the background processor task return Result<(), io::Error> and make Node::stop() return Error::PersistenceFailed instead of panicking when final persistence fails.
>   - P1, not easy: running-node persistence failures should remain tracked for a restart/refactor design.
>   - P1, upstream-blocked panics:
>       - LDK documents UnrecoverableError as an immediate panic: https://github.com/lightningdevkit/rust-lightning/blob/3dfcc4cca1866c5e5d4d4eaf3b82e09584e2ce5c/lightning/src/chain/chainmonitor.rs#L102-L104
>       - Chain-sync monitor persistence failure panics: https://github.com/lightningdevkit/rust-lightning/blob/3dfcc4cca1866c5e5d4d4eaf3b82e09584e2ce5c/lightning/src/chain/chainmonitor.rs#L481-L503
>       - New monitor persistence UnrecoverableError panics: https://github.com/lightningdevkit/rust-lightning/blob/3dfcc4cca1866c5e5d4d4eaf3b82e09584e2ce5c/lightning/src/chain/chainmonitor.rs#L1124-L1137
>       - Monitor update persistence UnrecoverableError panics: https://github.com/lightningdevkit/rust-lightning/blob/3dfcc4cca1866c5e5d4d4eaf3b82e09584e2ce5c/lightning/src/chain/chainmonitor.rs#L1223-L1231
[...]

PR #498, opened 2025-03-19
Handle remaining panic on persistence failure

elnosh:
Closes https://github.com/lightningdevkit/ldk-node/issues/381

This is the only other `panic!` I could find related to a persistence failure. Could've used `?` there to return the `Err` but left the match to maintain the `log_error!` message.

There were other `panic!`s during event handling such as this one: https://github.com/lightningdevkit/ldk-node/blob/f0338d19256615088fabab2b6927d478ae3ec1a1/src/event.rs#L542-L555 but not related to a persistence failure so I left as is.


elnosh:
ok sorry about that and thank you for the clarification. Pushed other changes addressing comments from review

PR #526, opened 2025-04-18
feat: chain source rest sync with bitcoind

enigbe:


enigbe:
Thanks for the early feedback.

> 1. I'm not quite convinced we should rename the BitcoindRpc client. Wouldn't it be cleaner to just add a (granted, rather duplicative) BitcoindRest variant side-by-side, and then in a second step maybe see how we can DRY up the code. Mushing both variants together seems to result in a rather confusing API and code structure.
> 2. I'm a bit confused why we'd now still have RPC credentials on set_chain_source_bitcoind (and elsewhere in downstream from that), even if we want to use the REST backend? Is there any operation that still requires RPC access? Can't we just rely on REST alone?

I understand your concerns about the name and API changes I have introduced but here is a bit of context.

Adding a `ChainSource::BitcondRest` variant was my initial approach, too because I had erroneously thought Bitcoin Core's [REST interface](https://github.com/bitcoin/bitcoin/blob/master/doc/REST-interface.md) exposed (more or less) the same APIs as the RPC interface. However, the REST interface exposes a handful of APIs only for `GET` requests. It is not possible to [broadcast transactions](https://github.com/lightningdevkit/ldk-node/blob/1525255222cd7a48de3a9301a3053e8a4e3d1a86/src/chain/bitcoind_rpc.rs#L56-L60) with a pure REST variant as we do with the `BitcoindRpcClient`, There is no REST API to `sendrawtransaction`.
For that particular use case, given that we'd need a handle on an RPC client regardless, I thought it best to:
- Keep the `BitcoindRpcClient` as is.
- Introduce a configurable REST/RPC sync client. With this, users can decide if syncing should happen either via the REST or RPC APIs.
- Rename `ChainSource::BitcoindRpc` to `ChainSource::Bitcoind` as the latter leaves no ambiguity about the possibility of a REST variant.

Camillarhi:
Hi @enigbe,

I noticed your CI is failing. I had a similar issue, but it was resolved after rebasing onto the latest `dev` branch (which includes changes from https://github.com/lightningdevkit/ldk-node/pull/524).

Could you try rebasing and see if that fixes the problem?

Thanks

src/builder.rs
```diff
@@ -297,16 +310,70 @@ impl NodeBuilder {
...
+				rpc_user,
+				rpc_password,
+				sync_client_config,
+			}) => {
+				if rpc_host.is_empty()
+					|| *rpc_port == 0
+					|| rpc_user.is_empty()
+					|| rpc_password.is_empty()
+				{
+					panic!("RPC configuration is incomplete. Must fully configure via RPC first.");
```
chuksys, line 362: I think we could introduce a new `BuildError` variant like `IncompleteRPCConfig` and error out instead?

src/builder.rs
```diff
@@ -297,16 +310,70 @@ impl NodeBuilder {
...
+				}
+
+				*sync_client_config = Some(rest_client_config);
+			},
+			Some(cs) => {
+				panic!("This option is only valid for ChainSource::Bitcoind. Current chain source config: {cs:?}")
```
chuksys, line 368: Maybe we could introduce a new `BuildError` variant like `InvalidChainSource` and error out instead?

src/builder.rs
```diff
@@ -297,16 +310,70 @@ impl NodeBuilder {
...
+			},
+			None => {
+				panic!("Must configure via RPC first.")
```
chuksys, line 371: Here too I think we could error out by introducing and returning a new `BuildError` variant like `RPCNotConfigured`.

chuksys (commented):
Looking into this PR has helped me gain a good understanding of how chain sourcing works. I think introducing `BitcoindSyncClient` is a good idea as the user can still carry out syncing operations using the REST API while the RPC Client is still maintained to carry out operations not supported by the REST API. I have just a few suggestions with regards to error handling.

enigbe:
Thanks for the feedback @tnull. I revisited this last week and addressed concerns raised here.
> Aaah, okay, then the proposed approach makes more sense. I think it's still a bit awkward to have the mandatory RPC credentials on the `set_chain_source_bitcoind` method and _addtionally_ have the `BitcoindSyncConfig::Rpc` variant.
> 
This has been removed and we just keep a [config for REST sync](https://github.com/lightningdevkit/ldk-node/blob/a28d28fca6d594c8c0c7aaff30c4caaad8c53d5d/src/config.rs#L402-L407).

> While it then makes sense to reuse the client internally, it might be simpler to just keep the original `set_chain_source_bitcoind_rpc` method and just add a `set_chain_source_bitcoind_rest` variant, that also features the mandatory RPC credentials?

Agreed! This seems like a cleaner split. I've added a `set_chain_source_bitcoind_rest` variant accordingly. However, I think it'd be potentially confusing/awkward for callers to still pass RPC parameters to the REST function.  I opted for `set_chain_source_bitcoind_rest` to take only REST parameters and clarified comments on usage and dependency on the RPC variant.

If you'd still prefer we pass RPC parameters as a required input for the REST variant, happy to adjust.

enigbe:
> Hi @enigbe,
> 
> I noticed your CI is failing. I had a similar issue, but it was resolved after rebasing onto the latest `dev` branch (which includes changes from #524).
> 
> Could you try rebasing and see if that fixes the problem?
> 
> Thanks

Thanks for looking at this. I've rebased and have just the CLN integration tests failing. I tried reproducing the failing test locally but couldn't. I should give it another go today.

enigbe:
> Looking into this PR has helped me gain a good understanding of how chain sourcing works. I think introducing `BitcoindSyncClient` is a good idea as the user can still carry out syncing operations using the REST API while the RPC Client is still maintained to carry out operations not supported by the REST API. I have just a few suggestions with regards to error handling.

Thanks for the review. Added the suggestions recommended for error handling.

src/builder.rs
```diff
@@ -297,16 +327,72 @@ impl NodeBuilder {
...
+	/// Configures the [`Node`] instance to synchronize chain data from a Bitcoin Core REST endpoint.
+	///
+	/// This method enables chain data synchronization via Bitcoin Core's REST interface.
+	/// It must be called after [`Self::set_chain_source_bitcoind_rpc`] because REST is used only for chain
+	/// synchronization, while RPC is still required for other essential operations like transaction
+	/// broadcasting.
+	///
+	/// ## Parameters:
+	/// * `rest_host`, `rest_port` - Required parameters for the Bitcoin Core REST connection.
+	pub fn set_chain_source_bitcoind_rest(
```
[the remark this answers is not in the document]
enigbe: I'll refactor to include the recommended changes . I thought having to pass RPC config for REST would be confusing but I now agree that it would be clearer and less likely to misuse.

enigbe:
Please excuse the delay on this. I have addressed concerns from the last review. Investigating a flaky test but it's ready for another look.

Let me know what you think when you can review.

bindings/ldk_node.udl
```diff
@@ -34,6 +34,12 @@ dictionary ElectrumSyncConfig {
 	BackgroundSyncConfig? background_sync_config;
 };
 
+[Enum]
```
tnull: nit: Seems this was first added here, then dropped in a later commit?

src/chain/bitcoind/common.rs
tnull, line 1: Can we please do the code moves in a prefactor commit? This would make it immediately clear that nothing changed when reviewing via `git diff --color-moved`.

And to be honest I'm not entirely sure why we move all that code at all, if we just end up with the implementation for both `rpc` and `rest` in `mod.rs`?

src/chain/bitcoind/mod.rs
```diff
@@ -0,0 +1,533 @@
...
+use lightning_block_sync::rpc::{RpcClient, RpcError};
+use lightning_block_sync::BlockSource;
+
+use std::collections::HashMap;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Arc;
+
+pub mod common;
+
+pub enum BitcoindApiClient {
```
tnull: Why not just `BitcoindClient`?

enigbe: This has been renamed.

src/builder.rs
```diff
@@ -716,8 +764,15 @@ impl ArcedNodeBuilder {
 		self.inner.write().unwrap().set_chain_source_electrum(server_url, sync_config);
 	}
 
-	/// Configures the [`Node`] instance to source its chain data from the given Bitcoin Core RPC
-	/// endpoint.
+	/// Configures the [`Node`] instance to connect to a Bitcoin Core node via RPC.
+	///
+	/// This method establishes an RPC connection that enables all essential chain operations including
```
tnull, line 769: The docs here (and for `set_chain_source_bitcoind_rest`) differ from the ones above. Please make sure they are identical.

src/chain/bitcoind.rs
```diff
@@ -530,3 +820,26 @@ impl Listen for ChainListener {
 		self.output_sweeper.block_disconnected(header, height);
 	}
 }
+
+pub(crate) fn rpc_credentials(rpc_user: String, rpc_password: String) -> String {
+	BASE64_STANDARD.encode(format!("{}:{}", rpc_user, rpc_password))
+}
+
+pub(crate) fn endpoint(host: String, port: u16) -> HttpEndpoint {
+	HttpEndpoint::for_host(host.clone()).with_port(port)
```
tnull: This `clone` can be dropped.

src/chain/bitcoind.rs
```diff
@@ -272,35 +514,100 @@ impl BitcoindRpcClient {
...
+			},
+			BitcoindClient::Rest { rest_client, .. } => {
+				Box::pin(async move { rest_client.get_best_block().await })
+			},
+		}
 	}
 }
 
+#[derive(Debug, Clone)]
+pub(crate) struct MempoolEntry {
```
tnull: Why was this moved?

src/chain/bitcoind.rs
```diff
@@ -272,35 +514,100 @@ impl BitcoindRpcClient {
...
+				Box::pin(async move { rest_client.get_block(header_hash).await })
+			},
+		}
 	}
 
-	fn get_best_block(&self) -> AsyncBlockSourceResult<(BlockHash, Option<u32>)> {
-		Box::pin(async move { self.rpc_client.get_best_block().await })
+	fn get_best_block(
+		&self,
+	) -> lightning_block_sync::AsyncBlockSourceResult<(bitcoin::BlockHash, Option<u32>)> {
```
tnull: Why did we switch this from the imported `AsyncBlockSourceResult` to `lightning_block_sync::AsyncBlockSourceResult` here and elsewhere?

src/chain/bitcoind.rs
```diff
@@ -207,62 +422,89 @@ impl BitcoindRpcClient {
 	/// This method is an adapted version of `bdk_bitcoind_rpc::Emitter::mempool`. It emits each
 	/// transaction only once, unless we cannot assume the transaction's ancestors are already
 	/// emitted.
-	async fn get_mempool_transactions_and_timestamp_at_height(
+	pub(crate) async fn get_mempool_transactions_and_timestamp_at_height(
 		&self, best_processed_height: u32,
 	) -> std::io::Result<Vec<(Transaction, u64)>> {
-		let prev_mempool_time = self.latest_mempool_timestamp.load(Ordering::Relaxed);
-		let mut latest_time = prev_mempool_time;
+		macro_rules! get_mempool_transactions_and_timestamp_at_height {
```
tnull: Any chance we can make this a method or closure rather than a macro?

src/chain/bitcoind.rs
```diff
@@ -7,187 +7,402 @@
...
+			},
+			BitcoindClient::Rest { rpc_client, .. } => {
+				get_mempool_entry(Arc::clone(&rpc_client), txid).await
 			},
 		}
 	}
 
 	pub(crate) async fn update_mempool_entries_cache(&self) -> std::io::Result<()> {
-		let mempool_txids = self.get_raw_mempool().await?;
+		macro_rules! update_mempool_entries_cache {
```
tnull: Same here, it would be preferable not to use macros here.

src/chain/bitcoind.rs
```diff
@@ -7,187 +7,402 @@
...
 	}
 
+	/// Retrieves the raw mempool.
 	pub(crate) async fn get_raw_mempool(&self) -> std::io::Result<Vec<Txid>> {
-		let verbose_flag_json = serde_json::json!(false);
-		self.rpc_client
-			.call_method::<GetRawMempoolResponse>("getrawmempool", &[verbose_flag_json])
-			.await
-			.map(|resp| resp.0)
+		macro_rules! get_raw_mempool {
```
tnull: Same here.

src/chain/bitcoind.rs
```diff
@@ -7,187 +7,402 @@
...
-
-					match rpc_error_res {
-						Ok(rpc_error) => {
-							// Check if it's the 'not found' error code.
-							if rpc_error.code == -5 {
-								Ok(None)
-							} else {
-								Err(std::io::Error::new(std::io::ErrorKind::Other, rpc_error))
+
+		macro_rules! get_raw_transaction {
```
tnull: Same here.

src/chain/bitcoind.rs
```diff
@@ -7,187 +7,402 @@
...
+		}
 	}
 
+	/// Gets the mempool minimum fee rate.
 	pub(crate) async fn get_mempool_minimum_fee_rate(&self) -> std::io::Result<FeeRate> {
-		self.rpc_client
-			.call_method::<MempoolMinFeeResponse>("getmempoolinfo", &[])
-			.await
-			.map(|resp| resp.0)
+		macro_rules! get_mempool_minimum_fee_rate {
```
tnull: Same here.

src/chain/bitcoind.rs
enigbe: This has been addressed. I inadvertently qualified the type instead of importing it.

enigbe:
Hi @tnull,

This is ready for another look. As recommended, I have added:
- prefector commit where the previously introduced macros are refactored into helper methods in preparation for REST support.
- couple of fix-ups to address unwanted code move and type imports.

src/chain/bitcoind.rs
enigbe: Possibly moved this while rebasing but it's now addressed.

Issue #436, opened 2025-01-17
Add support for resolving HumanReadableNames


tnull:
We just upgraded to LDK 0.1, which added support for HRNs. We should add support for optionally(?) acting as a resolver for HRNs here, possibly first defaulting to off.

Issue #435, opened 2025-01-17
Add support for sending to Human Readable Names (BIP353)


tnull:
We just upgraded to LDK 0.1, which added support for HRNs. We should add support for sending-to HRNs here, maybe as a sub-module in `payments`, or to start out, under `Bolt12Payment`.

PR #528, opened 2025-04-22
Add support for sending to human-readable names (BIP 353)

chuksys:
This PR is a first step to add support for human-readable names as outlined in BIP 353 and implemented in LDK 0.1.

Refs #435 

## Changes

1.  A `send_to_human_readable_name` function was added under `Bolt12Payment`.
2.  A `HrnBolt12Offer` PaymentKind variant was added.
3. A `HrnParsingFailed` Error variant was added.
4. These new enum variants were added to the uniffi bindings.

I plan to follow up with a PR for #436 and another PR that will add HRN sending for on-chain payments.

### DIFF
--- a/bindings/ldk_node.udl
+++ b/bindings/ldk_node.udl
@@ -302,6 +302,7 @@ enum NodeError {
 	"InsufficientFunds",
 	"LiquiditySourceUnavailable",
 	"LiquidityFeeTooHigh",
+	"HrnParsingFailed",
 };
 
 dictionary NodeStatus {
@@ -405,6 +406,7 @@ interface PaymentKind {
 	Bolt12Offer(PaymentHash? hash, PaymentPreimage? preimage, PaymentSecret? secret, OfferId offer_id, UntrustedString? payer_note, u64? quantity);
 	Bolt12Refund(PaymentHash? hash, PaymentPreimage? preimage, PaymentSecret? secret, UntrustedString? payer_note, u64? quantity);
 	Spontaneous(PaymentHash hash, PaymentPreimage? preimage);
+	HrnBolt12Offer(HumanReadableName hrn);
 };
 
 [Enum]
@@ -763,3 +765,6 @@ typedef string OrderId;
 
 [Custom]
 typedef string DateTime;
+
+[Custom]
+typedef string HumanReadableName;
--- a/src/error.rs
+++ b/src/error.rs
@@ -120,6 +120,8 @@ pub enum Error {
 	LiquiditySourceUnavailable,
 	/// The given operation failed due to the LSP's required opening fee being too high.
 	LiquidityFeeTooHigh,
+	/// Parsing a Human-Readable Name has failed
+	HrnParsingFailed,
 }
 
 impl fmt::Display for Error {
@@ -192,6 +194,9 @@ impl fmt::Display for Error {
 			},
 			Self::LiquidityFeeTooHigh => {
 				write!(f, "The given operation failed due to the LSP's required opening fee being too high.")
+			},
+			Self::HrnParsingFailed => {
+				write!(f, "Failed to parse a human-readable name.")
 			},
 		}
 	}
--- a/src/payment/bolt12.rs
+++ b/src/payment/bolt12.rs
@@ -22,6 +22,8 @@ use lightning::offers::invoice::Bolt12Invoice;
 use lightning::offers::offer::{Amount, Offer, Quantity};
 use lightning::offers::parse::Bolt12SemanticError;
 use lightning::offers::refund::Refund;
+use lightning::onion_message::dns_resolution::HumanReadableName;
+use lightning::onion_message::messenger::Destination;
 use lightning::util::string::UntrustedString;
 
 use rand::RngCore;
@@ -252,6 +254,72 @@ impl Bolt12Payment {
 						Err(Error::PaymentSendingFailed)
 					},
 				}
+			},
+		}
+	}
+
+	/// Send a payment to an offer resolved from a human-readable name [BIP 353].
+	///
+	/// Paying to human-readable names makes it more intuitive to make payments for offers
+	/// as users can simply send payments to HRNs such as `user@example.com`.
+	///
+	/// This can be used to pay so-called "zero-amount" offers, i.e., an offer that leaves the
+	/// amount paid to be determined by the user.
+	///
+	/// `dns_resolvers` should be a list of node Destinations that are configured for dns resolution (as outlined in bLIP 32).
+	/// These nodes can be found by running a search through the `NetworkGraph` to find nodes that announce the
+	/// `dns_resolver` feature flag.
+	pub fn send_to_human_readable_name(
+		&self, name: &str, amount_msat: u64, dns_resolvers: Vec<Destination>,
+	) -> Result<PaymentId, Error> {
+		let rt_lock = self.runtime.read().unwrap();
+		if rt_lock.is_none() {
+			return Err(Error::NotRunning);
+		}
+
+		let hrn = HumanReadableName::from_encoded(&name).map_err(|_| Error::HrnParsingFailed)?;
+
+		let mut random_bytes = [0u8; 32];
+		rand::thread_rng().fill_bytes(&mut random_bytes);
+		let payment_id = PaymentId(random_bytes);
+		let retry_strategy = Retry::Timeout(LDK_PAYMENT_RETRY_TIMEOUT);
+		let max_total_routing_fee_msat = None;
+
+		match self.channel_manager.pay_for_offer_from_human_readable_name(
+			hrn.clone(),
+			amount_msat,
+			payment_id,
+			retry_strategy,
+			max_total_routing_fee_msat,
+			dns_resolvers,
+		) {
+			Ok(()) => {
+				log_info!(self.logger, "Initiated sending {} msats to {}", amount_msat, name);
+				let kind = PaymentKind::HrnBolt12Offer { hrn };
+				let payment = PaymentDetails::new(
+					payment_id,
+					kind,
+					Some(amount_msat),
+					None,
+					PaymentDirection::Outbound,
+					PaymentStatus::Pending,
+				);
+				self.payment_store.insert(payment)?;
+				Ok(payment_id)
+			},
+			Err(()) => {
+				log_error!(self.logger, "Failed to send payment to {}", name);
+				let kind = PaymentKind::HrnBolt12Offer { hrn };
+				let payment = PaymentDetails::new(
+					payment_id,
+					kind,
+					Some(amount_msat),
+					None,
+					PaymentDirection::Outbound,
+					PaymentStatus::Pending,
+				);
+				self.payment_store.insert(payment)?;
+				Err(Error::PaymentSendingFailed)
 			},
 		}
 	}
--- a/src/payment/store.rs
+++ b/src/payment/store.rs
@@ -16,6 +16,7 @@ use crate::Error;
 use lightning::ln::channelmanager::PaymentId;
 use lightning::ln::msgs::DecodeError;
 use lightning::offers::offer::OfferId;
+use lightning::onion_message::dns_resolution::HumanReadableName;
 use lightning::util::ser::{Readable, Writeable};
 use lightning::util::string::UntrustedString;
 use lightning::{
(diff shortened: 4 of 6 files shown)


src/payment/bolt12.rs
```diff
@@ -256,6 +258,72 @@ impl Bolt12Payment {
...
+	///
+	/// This can be used to pay so-called "zero-amount" offers, i.e., an offer that leaves the
+	/// amount paid to be determined by the user.
+	///
+	/// `dns_resolvers` should be a list of node Destinations that are configured for dns resolution (as outlined in bLIP 32).
+	/// These nodes can be found by running a search through the `NetworkGraph` to find nodes that announce the
+	/// `dns_resolver` feature flag.
+	pub fn send_to_human_readable_name(
+		&self, name: &str, amount_msat: u64, dns_resolvers: Vec<Destination>,
+	) -> Result<PaymentId, Error> {
```
chuksys, line 274: I should probably expose this method to uniffi bindings? 

Also, for `dns_resolvers` maybe I should add a method in graph.rs which helps find all nodes that have a certain feature flag. This could make it simpler for the user to find all `dns_resolver` nodes.

chuksys:
Thank you for the feedback @tnull. I will address the comments and push the updated changes as soon as possible.

chuksys:
> I will address the comments and push the updated changes as soon as possible.

Still actively working on this. Will push updates once I'm done (likely within the next week).

chuksys:
Hi @tnull, thanks once again for the pointer you gave me in our discord chat. I just pushed some updates that allow the user to set a list of default `node_id`s as the `Destination` for HRNs dns resolution. As advised, I would follow up with another PR later that would fully expose `Destination` from LDK to uniFFI bindings and allow using `BlindedPath`s as the `Destination` for HRNs dns resolution. I am yet to implement the method which allows the user to easily fetch `dns_resolver` nodes from the network_graph - I will work on that next and push when done. Thanks!

src/config.rs
```diff
@@ -167,6 +168,11 @@ pub struct Config {
 	/// **Note:** If unset, default parameters will be used, and you will be able to override the
 	/// parameters on a per-payment basis in the corresponding method calls.
 	pub sending_parameters: Option<SendingParameters>,
+	/// The dns_resolvers node_ids to be used for resolving Human-readable Names.
+	///
+	/// If set to `Some`, the values set will be used as dns_resolvers when sending to HRNs.
+	/// **Note:** If set to `None`, payments to HRNs will fail.
+	pub dns_resolvers_node_ids: Option<Vec<PublicKey>>,
```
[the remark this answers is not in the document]
chuksys: > Also, should we maybe only add this as part of an new `HumanReadableNamesConfig` object, as we may be adding addtional configs in the future?

I think this is a good idea. I'll work on this.

src/payment/bolt12.rs
```diff
@@ -256,6 +259,92 @@ impl Bolt12Payment {
...
+			hrn.clone(),
+			amount_msat,
+			payment_id,
+			retry_strategy,
+			max_total_routing_fee_msat,
+			destinations,
+		) {
+			Ok(()) => {
+				log_info!(self.logger, "Initiated sending {} msats to {}", amount_msat, name);
+				let kind = PaymentKind::Bolt12Offer {
```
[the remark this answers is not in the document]
chuksys: > I wonder if we should intercept the [`InvoiceReceived`](https://docs.rs/lightning/latest/lightning/events/enum.Event.html#variant.InvoiceReceived) event to at least set some of these fields when we get access to the invoice for this payment?

I will work on this.

> Also opened [lightningdevkit/rust-lightning#3779](https://github.com/lightningdevkit/rust-lightning/issues/3779) as it would be helpful to have a dedicated event that emits the offer after it has been retrieved.

Great! This would be really helpful.

src/payment/bolt12.rs
```diff
@@ -256,6 +259,92 @@ impl Bolt12Payment {
...
+	/// Send a payment to an offer resolved from a human-readable name [BIP 353].
+	///
+	/// Paying to human-readable names makes it more intuitive to make payments for offers
+	/// as users can simply send payments to HRNs such as `user@example.com`.
+	///
+	/// This can be used to pay so-called "zero-amount" offers, i.e., an offer that leaves the
+	/// amount paid to be determined by the user.
+	///
+	/// If `dns_resolvers_node_ids` in Config is set to `None`, this operation will fail.
+	pub fn send_to_human_readable_name(
```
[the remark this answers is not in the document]
chuksys: If we can intercept the `InvoiceReceived` event or have an event that is emitted after the offer is retrieved, I think we can have at least a test case built around any of these events. 

I think the end-to-end test would be the ideal, eventually.

src/payment/bolt12.rs
```diff
@@ -256,6 +259,92 @@ impl Bolt12Payment {
...
+					payer_note: None,
+					quantity: None,
+				};
+				let payment = PaymentDetails::new(
+					payment_id,
+					kind,
+					Some(amount_msat),
+					None,
+					PaymentDirection::Outbound,
+					PaymentStatus::Pending,
```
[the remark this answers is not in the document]
chuksys: Nice catch! It should be `Failed` here.

src/config.rs
```diff
@@ -181,10 +186,22 @@ impl Default for Config {
 			anchor_channels_config: Some(AnchorChannelsConfig::default()),
 			sending_parameters: None,
 			node_alias: None,
+			hrn_config: None,
 		}
 	}
 }
 
+/// Configuration options for Human-readable Names
```
tnull: Let's link to BIP353 here, too. Also, nit, let's stick with `Human-Readable Names`, i.e., consistent capitalization everywhere.

src/error.rs
```diff
@@ -193,6 +197,12 @@ impl fmt::Display for Error {
 			Self::LiquidityFeeTooHigh => {
 				write!(f, "The given operation failed due to the LSP's required opening fee being too high.")
 			},
+			Self::HrnParsingFailed => {
+				write!(f, "Failed to parse a human-readable name.")
+			},
+			Self::DnsResolversUnavailable => {
+				write!(f, "The given operation failed due to `dns-resolvers` not being configured in builder.")
```
tnull:
```suggestion
				write!(f, "The given operation failed due to DNS resolvers not being configured.")
```

src/payment/bolt12.rs
```diff
@@ -256,6 +259,92 @@ impl Bolt12Payment {
...
+		let dns_resolvers = match &self.config.dns_resolvers_node_ids {
+			Some(dns_resolvers) => Ok(dns_resolvers.clone()),
+			None => Err(Error::DnsResolversNotConfigured),
+		}?;
+
+		let destinations: Vec<Destination> =
+			dns_resolvers.into_iter().map(|public_key| Destination::Node(public_key)).collect();
+
+		match self.channel_manager.pay_for_offer_from_human_readable_name(
+			hrn.clone(),
```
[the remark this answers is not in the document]
tnull: Seems still there though?

src/payment/store.rs
```diff
@@ -404,7 +404,7 @@ pub enum PaymentKind {
 		/// The secret used by the payment.
 		secret: Option<PaymentSecret>,
 		/// The ID of the offer this payment is for.
-		offer_id: OfferId,
+		offer_id: Option<OfferId>,
```
tnull, line 411: Please describe when this field would be set / unset.

src/event.rs
```diff
@@ -1417,8 +1417,20 @@ where
 					);
 				}
 			},
-			LdkEvent::InvoiceReceived { .. } => {
-				debug_assert!(false, "We currently don't handle BOLT12 invoices manually, so this event should never be emitted.");
+			LdkEvent::InvoiceReceived { payment_id, invoice, context: _, responder: _ } => {
```
tnull: I don't think we'd ever actually see this emitted (see the previous `debug_assert` here), if we don't set [`manually_handle_bolt12_invoices`](https://docs.rs/lightning/latest/lightning/util/config/struct.UserConfig.html#structfield.manually_handle_bolt12_invoices). And if we do, we really need to handle this even properly, not only to update our `PaymentStore`.

src/payment/bolt12.rs
```diff
@@ -28,6 +30,20 @@ use std::num::NonZeroU64;
...
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
+#[cfg(not(feature = "uniffi"))]
+type HumanReadableName = LdkHumanReadableName;
+#[cfg(feature = "uniffi")]
+type HumanReadableName = Arc<crate::uniffi_types::HumanReadableName>;
+
+#[cfg(not(feature = "uniffi"))]
+pub fn maybe_convert_hrn(hrn: HumanReadableName) -> LdkHumanReadableName {
+	hrn.clone()
```
tnull: This `clone` can be dropped.

src/payment/bolt12.rs
```diff
@@ -254,6 +271,95 @@ impl Bolt12Payment {
...
+		match self.channel_manager.pay_for_offer_from_human_readable_name(
+			hrn.clone(),
+			amount_msat,
+			payment_id,
+			retry_strategy,
+			max_total_routing_fee_msat,
+			destinations,
+		) {
+			Ok(()) => {
+				log_info!(self.logger, "Initiated sending {} msats to {:?}", amount_msat, hrn);
```
tnull: Ah, let's print the hrn as `... ₿{}@{}", hrn.user(), hrn.doman()` for now, but I think we should implement `Display` for this upstream. Opened an issue for it: https://github.com/lightningdevkit/rust-lightning/issues/3823

src/event.rs
```diff
@@ -1417,8 +1417,24 @@ where
...
+					Ok(_) => {},
+					Err(e) => {
+						log_error!(self.logger, "Failed to access payment store: {}", e);
+						return Err(ReplayEvent());
+					},
+				};
+
+				let _ = self
+					.channel_manager
+					.send_payment_for_bolt12_invoice(&invoice, context.as_ref());
```
chuksys: I didn't handle the error here because I'm not sure if we want to return a `ReplayEvent` Error when we get a [Bolt12PaymentError::UnexpectedInvoice](https://docs.rs/lightning/latest/lightning/ln/channelmanager/enum.Bolt12PaymentError.html#variant.UnexpectedInvoice) or [Bolt12PaymentError::DuplicateInvoice](https://docs.rs/lightning/latest/lightning/ln/channelmanager/enum.Bolt12PaymentError.html#variant.DuplicateInvoice) - Plus since the success or failure of the payment, triggers [Event::PaymentSent](https://docs.rs/lightning/latest/lightning/events/enum.Event.html#variant.PaymentSent) or [Event::PaymentFailed](https://docs.rs/lightning/latest/lightning/events/enum.Event.html#variant.PaymentFailed) (which are already properly handled), I'm thinking this might be the way to go. Please let me know what you think.

src/payment/bolt12.rs
```diff
@@ -41,6 +43,18 @@ type Offer = Arc<crate::ffi::Offer>;
 type Refund = lightning::offers::refund::Refund;
 #[cfg(feature = "uniffi")]
 type Refund = Arc<crate::ffi::Refund>;
+type HumanReadableName = LdkHumanReadableName;
```
tnull: This seems to be missing a newline and `#[cfg(not(feature = "uniffi"))]` ?

src/payment/bolt12.rs
```diff
@@ -41,6 +43,18 @@ type Offer = Arc<crate::ffi::Offer>;
...
+#[cfg(feature = "uniffi")]
+type HumanReadableName = Arc<crate::uniffi_types::HumanReadableName>;
+
+#[cfg(not(feature = "uniffi"))]
+pub fn maybe_convert_hrn(hrn: HumanReadableName) -> LdkHumanReadableName {
```
tnull: Please reuse the generalized `maybe_*` methods now moved to `ffi/mod.rs`.

src/uniffi_types.rs
tnull, line 1: This file was renamed as part of #542, please don't reintroduce it here but rather make the required changes to `ffi/types.rs`.

chuksys:
> Seems there are still a few rebase issues.
> 
> Feel free to squash fixups. Btw, going forward it would be helpful to provide actual commit messages behind the `fixup!` prefix rather than just mirroring the feature commit's headline.

Thanks for the feedback. I've fixed the rebase issues. Going forward, I will ensure to provide commit messages for fixups.

bindings/ldk_node.udl
```diff
@@ -854,4 +871,4 @@ typedef string NodeAlias;
 typedef string OrderId;
 
 [Custom]
-typedef string DateTime;
+typedef string DateTime;
```
tnull: nit: Not sure why this change here was included. Please revert.

src/event.rs
tnull: Hmm, true, but we should at the very least log the error here I think.

PR #530, opened 2025-04-28
Cut v0.5.0 Rust release

tnull:
We cut the v0.5 Rust release. 

As we'll wait for LDK 0.1.3 to ship, the Kotlin/Swift binary releases will follow in a separate PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ldk-node"
-version = "0.5.0-rc.1"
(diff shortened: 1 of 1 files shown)


joostjager (approved):
One thing that I find a bit scary is Rust's auto-upgrade of patches. I saw ldk 0.1.0 is referenced in cargo.toml, which then I think auto-upgrades. But it makes the release non-deterministic? I suppose that's just the Rust way of doing it.

PR #540, opened 2025-05-16
Don't fail `send_all` retaining reserves for 0 channels 

### LINKED ISSUE #539: `send_all_to_address` with `retain_reserves` creates 0 amount output
In Fedimint, [we have a test that first closes all channels, then sends all of the funds onchain, then verifies that the node's balance is 0](https://github.com/fedimint/fedimint/blob/a3664bc390d9608b20fbdbc322e00a67b66c6959/gateway/integration_tests/src/main.rs#L595). 

When upgrading to `ldk-node v0.5.0`, this test fails when `retain_reserves` is set to true. Setting `retain_reserves` to false makes the test pass. We have not changed our config or our anchor reserve sats at all.

In LDK node's log, I see the following line:

```
2025-05-13 15:37:08 ERROR [ldk_node::wallet:370] Failed to create temporary transaction: Output below the dust limit: 0
```

Is this expected behavior? If I had to guess its trying to create a 0 amount output, because we have 0 channels and the "reserve" calculation dictates it needs to create an output, even though in this case it shouldn't. Let me know if I just have something configured incorrectly.

For now we will be setting `retain_reserves` to false so that our test passes.

tnull:
Fixes #539.

Previouly, `OnchainPayment::send_all_to_address` would fail in the
`retain_reserves` mode if the maintained reserves were below the dust
limit. Most notably this would happen if we had no channels open at all.

Here, we fix this by simply falling back to the draining case (not
considering reserves) if the anchor reserves are below dust. We also add
a unit test that would have caught this regression in the first place.

Note: Github diff really messes up the first commit which is mostly adding an `if` and re-indenting the previous codeblock. It's best reviewed with `git show --color-moved --ignore-all-space`

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -403,6 +403,10 @@ where
 						);
 						e
 					})?;
+
+					// 'cancel' the transaction to free up any used change addresses
+					locked_wallet.cancel_tx(&tmp_tx);
+
 					let estimated_spendable_amount = Amount::from_sat(
 						spendable_amount_sats.saturating_sub(estimated_tx_fee.to_sat()),
 					);


tests/integration_tests_rust.rs
```diff
@@ -496,6 +496,86 @@ fn onchain_send_receive() {
...
+	let txid = node_a.onchain_payment().send_all_to_address(&addr_b, true, None).unwrap();
+
+	wait_for_tx(&electrsd.client, txid);
+	generate_blocks_and_wait(&bitcoind.client, &electrsd.client, 6);
+
+	node_a.sync_wallets().unwrap();
+	node_b.sync_wallets().unwrap();
+	assert_eq!(node_a.list_balances().spendable_onchain_balance_sats, 0);
+	assert!(((premine_amount_sat * 2 - onchain_fee_buffer_sat)..=(premine_amount_sat * 2))
+		.contains(&node_b.list_balances().spendable_onchain_balance_sats));
```
m1sterc001guy, line 534: Maybe could use a comment, not immediately clear to me what this is validating.

m1sterc001guy (approved):
I have less context here but this does look like it will fix our issue :)

jkczyz (commented):
LGTM after squash and any comments to clarify the test expections.

jkczyz (approved):

Issue #504, opened 2025-03-21
Expose proper `Invoice`  / `Offer` types in bindings


tnull:
Previously, we'd expose `Bolt11Invoice`/`Offer`/`Refund`s as strings in bindings that get translated at time of use via `UniffiCustomTypeConverter`s. This has a number of drawbacks, maybe most notably that these types don't allow the user to validate any fields before paying them.

We should expose 'proper` types for these (i.e., likely we'd need to create newtype wrappers for at least some of them based on the `uniffi` feature flag) that allow retrieving all the important fields (amount, description, payer_note, etc).

PR #542, opened 2025-05-19
Feature: Add Bolt12 Uniffi Type Wrappers

alexanderwiederin:
Second PR for https://github.com/lightningdevkit/ldk-node/issues/504.

This PR adds full Offer-, Refund- and Bolt12Invoice- types for the FFI bindings. 

#### Changes:
- Type definitions in `ldk_node.udl` and `uniffi_types.rs`
- Implementation of UniffiType trait to standardize conversions between LDK types and their FFI-compatible wrappers in `uniffi_conversions.rs`
- Integration of wrapper types in payment handling methods in `bolt12.rs` and `unified_qr.rs`

#### Benefits:
- Direct access to Offer, Refund and Bolt12Invoice properties from FFI languages
- Consistent API across languages

#### Note:
Appreciate that this PR is big. Happy to break down into smaller parts if preferred.

### DIFF
--- a/bindings/ldk_node.udl
+++ b/bindings/ldk_node.udl
@@ -776,6 +776,31 @@ interface Refund {
 	string? payer_note();
 };
 
+interface Bolt12Invoice {
+	[Throws=NodeError, Name=from_str]
+	constructor([ByRef] string invoice_str);
+	PaymentHash payment_hash();
+	u64 amount_msats();
+	OfferAmount? amount();
+	PublicKey signing_pubkey();
+	u64 created_at();
+	u64? absolute_expiry_seconds();
+	u64 relative_expiry();
+	boolean is_expired();
+	string? description();
+	string? issuer();
+	string? payer_note();
+	sequence<u8>? metadata();
+	u64? quantity();
+	sequence<u8> signable_hash();
+	PublicKey payer_signing_pubkey();
+	PublicKey? issuer_signing_pubkey();
+	sequence<u8> chain();
+	sequence<sequence<u8>>? offer_chains();
+	sequence<Address> fallback_addresses();
+	sequence<u8> encode();
+};
+
 [Custom]
 typedef string Txid;
 
@@ -793,9 +818,6 @@ typedef string NodeId;
 
 [Custom]
 typedef string Address;
-
-[Custom]
-typedef string Bolt12Invoice;
 
 [Custom]
 typedef string OfferId;
--- a/src/ffi/types.rs
+++ b/src/ffi/types.rs
@@ -25,7 +25,6 @@ pub use crate::payment::{MaxTotalRoutingFeeLimit, QrPaymentResult, SendingParame
 pub use lightning::chain::channelmonitor::BalanceSource;
 pub use lightning::events::{ClosureReason, PaymentFailureReason};
 pub use lightning::ln::types::ChannelId;
-pub use lightning::offers::invoice::Bolt12Invoice;
 pub use lightning::offers::offer::OfferId;
 pub use lightning::routing::gossip::{NodeAlias, NodeId, RoutingFees};
 pub use lightning::util::string::UntrustedString;
@@ -56,6 +55,7 @@ use bitcoin::hashes::sha256::Hash as Sha256;
 use bitcoin::hashes::Hash;
 use bitcoin::secp256k1::PublicKey;
 use lightning::ln::channelmanager::PaymentId;
+use lightning::offers::invoice::Bolt12Invoice as LdkBolt12Invoice;
 use lightning::offers::offer::{Amount as LdkAmount, Offer as LdkOffer};
 use lightning::offers::refund::Refund as LdkRefund;
 use lightning::util::ser::Writeable;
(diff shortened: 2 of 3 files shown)


src/uniffi_types.rs
```diff
@@ -113,15 +114,90 @@ impl UniffiCustomTypeConverter for Address {
...
+	Bitcoin { amount_msats: u64 },
+	Currency { iso4217_code: String, amount: u64 },
+}
+
+impl From<LdkAmount> for OfferAmount {
+	fn from(ldk_amount: LdkAmount) -> Self {
+		match ldk_amount {
+			LdkAmount::Bitcoin { amount_msats } => OfferAmount::Bitcoin { amount_msats },
+			LdkAmount::Currency { iso4217_code, amount } => OfferAmount::Currency {
+				iso4217_code: iso4217_code.iter().map(|&b| b as char).collect(),
```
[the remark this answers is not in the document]
jkczyz: Hmmm... I guess we figured it was more of an application-level concern since more codes could be added in the future. Though I suppose those basic checks that you mentioned could be useful.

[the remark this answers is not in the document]
alexanderwiederin: Thanks! I think we can keep the current implementation as-is and let LDK handle the content validation then. What do you think?

src/uniffi_conversions.rs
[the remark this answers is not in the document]
alexanderwiederin: Cool! Moved to `ffi/mod.rs` to prevent collisions with the uniffi library - agree it's cleaner.

src/uniffi_conversions.rs
```diff
@@ -0,0 +1,111 @@
...
+
+		fn from_ldk(ldk_value: Self::LdkType) -> Self {
+			Bolt11InvoiceDescription::from(ldk_value)
+		}
+
+		fn as_ldk(&self) -> Self::LdkType {
+			match LdkBolt11InvoiceDescription::try_from(self) {
+				Ok(ldk_description) => ldk_description,
+				Err(e) => {
+					panic!("Failed to convert Bolt11InvoiceDescription to LDK type: {:?}", e)
```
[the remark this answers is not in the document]
alexanderwiederin: Deviated from your suggestion, but think I have something neat - let me know what you think! Thanks for the pointer.

src/uniffi_types.rs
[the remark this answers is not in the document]
tnull: Yeah, I think that should be fine, given that the upstream change is already underway.

src/uniffi_types.rs
```diff
@@ -301,20 +299,205 @@ impl From<LdkRefund> for Refund {
...
+		self.inner.fallbacks()
+	}
+
+	/// Writes `self` out to a `Vec<u8>`.
+	pub fn encode(&self) -> Vec<u8> {
+		self.inner.encode()
+	}
+}
+
+impl std::str::FromStr for Bolt12Invoice {
```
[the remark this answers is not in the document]
tnull: Hmm, AFAICT the resolution was to use the wire format for now. I'm not super happy about that, but okay.

src/ffi/conversions.rs
```diff
@@ -0,0 +1,70 @@
+mod uniffi_impl {
```
tnull: These new files need a licensing header, just as all the others.

src/ffi/conversions.rs
```diff
@@ -50,6 +51,20 @@ mod uniffi_impl {
 		}
 	}
 
+	impl UniffiType for Arc<Offer> {
```
tnull: Why are we implementing this on `Arc<Offer>`, not on `Offer`? I think we should do the latter if possible.

src/ffi/types.rs
```diff
@@ -113,15 +114,142 @@ impl UniffiCustomTypeConverter for Address {
...
+	pub fn id(&self) -> OfferId {
+		OfferId(self.inner.id().0)
+	}
+
+	/// Whether the offer has expired.
+	pub fn is_expired(&self) -> bool {
+		self.inner.is_expired()
+	}
+
+	/// A complete description of the purpose of the payment. Intended to be displayed to the user
```
tnull: nit: Please leave a whitespace comment line after the first sentence, i.e., put it on its own paragraph. This helps with readability and doc rendering.

src/ffi/conversions.rs
```diff
@@ -65,6 +65,20 @@ mod uniffi_impl {
 		}
 	}
 
+	impl UniffiType for Arc<Refund> {
```
tnull: Same here, can we implement this rather for `Refund` directly?

src/ffi/conversions.rs
alexanderwiederin: Thanks for this! 

Initially, I implemented these on the `Arc`s because that's how they are required to pass the ffi boundary. I looked into it again and now have a solution that Arc-allocates/-extracts the types in the conversion functions (i.e. `maybe_convert`, `maybe_wrap` etc.) and only does type conversion on the implementation methods. The benefit is that we can use standard traits and remove `ffi/conversions.rs` entirely. Downside is that the conversion functions become _slightly_ more complex.

I have implemented these changes on https://github.com/alexanderwiederin/ldk-node/commit/6b5f2974bbc82a1aa213e7f49c4e84033ab749b2 for better comparison. 

What do you think?

src/uniffi_types.rs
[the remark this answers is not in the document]
alexanderwiederin: Mmh, so what does this mean for us? We need to implement Writable and Readable for Bolt12Invoice?

src/ffi/conversions.rs
tnull: Looks a lot less invasive. However, there seem to be a few cases where we add unnecessary `clones` now that could lead to addtional memory usage and heap fragmentation. Can we avoid cloning when not necessary, and also enforce in the types of `maybe_convert` etc, that we'd always use `Arc::clone(&..)`, ie., that we'd only ever clone the reference, and otherwise would maintain noops or move semantics?

alexanderwiederin: Ok, found a solution that eliminates the need for cloning entirely and pushed the changes to this PR's branch. I have also renamed `maybe_try_convert` to `maybe_try_convert_enum` to be more specific about its purpose. Unlike other types that UniFFI typically wraps in an arc when crossing the boundary, enums are handled differently and are not wrapped. The new name makes expected argument cleaner. Thanks for the help!

src/ffi/types.rs
```diff
@@ -0,0 +1,1612 @@
...
+	fn from_custom(obj: Self) -> Self::Builtin {
+		obj.to_string()
+	}
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum OfferAmount {
+	Bitcoin { amount_msats: u64 },
+	Currency { iso4217_code: String, amount: u64 },
+}
```
enigbe, line 122: Curious about why `LdkAmount` is not used directly here. `OfferAmount` is basically the same struct except for the `iso4217_code` field. Is this because there is no support for fixed-sized arrays?

enigbe (commented):
Well done @alexanderwiederin,

I've reviewed this and just have a small question regarding fixed-sized arrays. The `ffi/conversion` approach looked good to me, but following your conversation with @tnull has highlighted some gaps in my understanding of Uniffi's internals especially with regards to heap allocation. I'll be looking into that further.

I've also tested this and it LGTM.

src/ffi/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+#[cfg(feature = "uniffi")]
+pub fn maybe_try_convert_enum<T, R>(wrapped_type: &T) -> Result<R, crate::error::Error>
+where
+	for<'a> R: TryFrom<&'a T, Error = crate::error::Error>,
+{
+	R::try_from(wrapped_type)
+}
+
+#[cfg(feature = "uniffi")]
+pub fn maybe_wrap<T>(ldk_type: impl Into<T>) -> std::sync::Arc<T> {
```
enigbe, line 30: Why is `ldk_type` `impl Into<T>` and not just `T`?

alexanderwiederin: Good question! `T` is not actually the type that is passed to `maybe_wrap`. `impl Into<T>` means that the argument we pass to `maybe_wrap` can be any type that implements `Into<T>` to convert itself into `T`. In our case `T` would be the UniFFI type that is needed at the call site.

For example, we pass LDK native types (like `lightning_invoice::Bolt11Invoice`), which gets converted to their corresponding UniFFI wrapper type (like `ffi::types::Bolt11Invoice`) using `.into()` before being wrapped in an Arc.

src/ffi/types.rs
alexanderwiederin: Yea, that's pretty much it. `OfferAmount` provides a binding-friendly representation by using String instead of a fixed-size byte array for iso4217_code. I also don't think you can automatically convert complex types across FFI language boundaries in general.

src/ffi/conversions.rs
alexanderwiederin: On further thought - wonder if we should rename `maybe_convert` to `maybe_unwrap`. Naming is more accurate with the `.as_ref()` call - no real conversion happening. What do you think?

tnull: > On further thought - wonder if we should rename `maybe_convert` to `maybe_unwrap`. Naming is more accurate with the `.as_ref()` call - no real conversion happening. What do you think?

I think I'm open to renaming it, but please not `maybe_unwrap`, as `unwrap` signals to me that we're .. `unwrap`ping, which is often potentially dangerous.

src/ffi/mod.rs
```diff
@@ -0,0 +1,47 @@
+// This file is Copyright its original authors, visible in version control history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
+
+#[cfg(feature = "uniffi")]
+pub mod types;
```
tnull: nit: I don't think this needs to be pub if we're re-exporting all types at this level anyways?

src/ffi/conversions.rs
alexanderwiederin: Renamed to `maybe_deref`. Thanks for the input!

bindings/ldk_node.udl
```diff
@@ -736,6 +736,69 @@ interface Bolt11Invoice {
 	PublicKey recover_payee_pub_key();
 };
 
+[Enum]
+interface OfferAmount {
+	Bitcoin(u64 amount_msats);
+	Currency(string iso4217_code, u64 amount);
+};
+
+interface Offer {
```
tnull, line 746: We'll also need to export the `Display` implementation here and below, similar to https://github.com/lightningdevkit/ldk-node/pull/573

alexanderwiederin: Ok, implemented for `Offer` and `Refund`. The LDK `Bolt12Invoice` does not implement Display so we can't apply the same implementation as for the others. What would you suggest?

tnull: Yeah, for `Bolt12Invoice` we'll probably leave as is for now.

Issue #425, opened 2025-01-03
Add new store for inbound payment metadata


tnull:
Prior to v0.1 LDK only offered `PaymentId`s for outbound payments, which led us to track inbound payments by an ID derived for the payment hash. This was always messy as the latter is only available for BOLT11 payments, resulting in BOLT11 payments being added to the payment store upon invoice creation, and BOLT12 payments only being added upon receiving the invoice request.

With https://github.com/lightningdevkit/rust-lightning/pull/3303 LDK added `PaymentId`s for inbound payments, so we should switch to use them in a backwards-compatible manner (i.e., probably still checking whether the old-style ID is available in the store for now). However, as this will mean we that we can't add the payment store entry upon invoice creation also for BOLT11 payments, we finally need a different solution to store additional metadata (invoices, but also JIT-channel fee limits) before receiving the payment. That is we need to add a `PaymentMetatdataStore` or similar.

moisesPompilio:
Opened the PR as a draft. It includes the implementation of PaymentMetadata and the plan for backward compatibility with the old payment persistence logic. It’s complete and works, but I’m not sure if the metadata structure is ideal — maybe some fields need to be added, removed, or the flow adjusted.

tnull:
Given that #948 doesn't require this anymore, and we also don't require a persistent store for JIT metadata since #899, we'll defer adding the metadata store/document stores until v0.9 as they aren't critical anymore.

PR #544, opened 2025-05-20
Introduce generalized `DataStore`

tnull:
We already have a 'store' pattern in a few places in our codebase (peer store, payment store), and for #425 we're about to add yet another one. To at least somewhat DRY up the storage logic, we here introduce a generalized `DataStore` and migrate our `PaymentStore` to it. This is a prefactor to #425.

(cc @moisesPompilio)

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -18,21 +18,23 @@ use crate::gossip::GossipSource;
 use crate::io::sqlite_store::SqliteStore;
 use crate::io::utils::{read_node_metrics, write_node_metrics};
 use crate::io::vss_store::VssStore;
+use crate::io::{
+	self, PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE, PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
+};
 use crate::liquidity::{
 	LSPS1ClientConfig, LSPS2ClientConfig, LSPS2ServiceConfig, LiquiditySourceBuilder,
 };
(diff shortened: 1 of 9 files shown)


src/data_store.rs
```diff
@@ -0,0 +1,274 @@
...
+		let objects =
+			Mutex::new(HashMap::from_iter(objects.into_iter().map(|obj| (obj.id(), obj))));
+		Self { objects, primary_namespace, secondary_namespace, kv_store, logger }
+	}
+
+	pub(crate) fn insert(&self, object: SO) -> Result<bool, Error> {
+		let mut locked_objects = self.objects.lock().unwrap();
+
+		self.persist(&object)?;
+		let updated = locked_objects.insert(object.id(), object).is_some();
```
jkczyz, line 71: Does the order matter here? The opposite order is used in `insert_or_update`, so there we could have the object in memory but fail persisting.

src/data_store.rs
```diff
@@ -0,0 +1,274 @@
...
+		Ok(updated)
+	}
+
+	pub(crate) fn insert_or_update(&self, object: &SO) -> Result<bool, Error> {
```
jkczyz, line 75: Could you add tests for `insert_or_update` as well?

jkczyz, line 75: Should we take `object` by value like is done in `insert`?

jkczyz (commented):
LGTM after squash.

jkczyz (approved):
Not sure if CI failure is flakiness.

PR #549, opened 2025-05-31
Support passing preimage for spontaneous payments

### LINKED ISSUE #535: Support passing preimage for spontaneous payments
Currently its generated at send-time https://github.com/lightningdevkit/ldk-node/blob/main/src/payment/spontaneous.rs#L82 but in some cases the app needs to be able to specify the preimage.

For example, NWC allows specifying a preimage as one of the request fields when doing a keysend payment: https://github.com/nostr-protocol/nips/blob/master/47.md#pay_keysend

aagbotemi:
This PR adds support for specifying custom preimages when sending spontaneous (keysend) payments.

## Solution
- Added `send_with_preimage()` method
- Added `send_with_preimage_and_custom_tlvs()` method for advanced use cases
- Modified internal `send_inner()` to accept optional preimage parameter and generate random preimage only when custom preimage is not provided.

An integration test is added to verify logic, and the new methods are added to the bindings.

Fixes #535

### DIFF
--- a/bindings/ldk_node.udl
+++ b/bindings/ldk_node.udl
@@ -213,6 +213,10 @@ interface SpontaneousPayment {
 	[Throws=NodeError]
 	PaymentId send_with_custom_tlvs(u64 amount_msat, PublicKey node_id, SendingParameters? sending_parameters, sequence<CustomTlvRecord> custom_tlvs);
 	[Throws=NodeError]
+	PaymentId send_with_preimage(u64 amount_msat, PublicKey node_id, PaymentPreimage preimage, SendingParameters? sending_parameters);
+	[Throws=NodeError]
+	PaymentId send_with_preimage_and_custom_tlvs(u64 amount_msat, PublicKey node_id, sequence<CustomTlvRecord> custom_tlvs, PaymentPreimage preimage, SendingParameters? sending_parameters);
+	[Throws=NodeError]
 	void send_probes(u64 amount_msat, PublicKey node_id);
 };
 
--- a/src/payment/spontaneous.rs
+++ b/src/payment/spontaneous.rs
@@ -57,7 +57,7 @@ impl SpontaneousPayment {
 	pub fn send(
 		&self, amount_msat: u64, node_id: PublicKey, sending_parameters: Option<SendingParameters>,
 	) -> Result<PaymentId, Error> {
-		self.send_inner(amount_msat, node_id, sending_parameters, None)
+		self.send_inner(amount_msat, node_id, sending_parameters, None, None)
 	}
 
 	/// Send a spontaneous payment including a list of custom TLVs.
@@ -65,19 +65,37 @@ impl SpontaneousPayment {
 		&self, amount_msat: u64, node_id: PublicKey, sending_parameters: Option<SendingParameters>,
 		custom_tlvs: Vec<CustomTlvRecord>,
 	) -> Result<PaymentId, Error> {
-		self.send_inner(amount_msat, node_id, sending_parameters, Some(custom_tlvs))
+		self.send_inner(amount_msat, node_id, sending_parameters, Some(custom_tlvs), None)
+	}
+
+	/// Send a spontaneous payment with custom preimage
+	pub fn send_with_preimage(
+		&self, amount_msat: u64, node_id: PublicKey, preimage: PaymentPreimage,
+		sending_parameters: Option<SendingParameters>,
+	) -> Result<PaymentId, Error> {
+		self.send_inner(amount_msat, node_id, sending_parameters, None, Some(preimage))
+	}
+
+	/// Send a spontaneous payment with custom preimage including a list of custom TLVs.
+	pub fn send_with_preimage_and_custom_tlvs(
+		&self, amount_msat: u64, node_id: PublicKey, custom_tlvs: Vec<CustomTlvRecord>,
+		preimage: PaymentPreimage, sending_parameters: Option<SendingParameters>,
+	) -> Result<PaymentId, Error> {
+		self.send_inner(amount_msat, node_id, sending_parameters, Some(custom_tlvs), Some(preimage))
 	}
 
 	fn send_inner(
 		&self, amount_msat: u64, node_id: PublicKey, sending_parameters: Option<SendingParameters>,
-		custom_tlvs: Option<Vec<CustomTlvRecord>>,
+		custom_tlvs: Option<Vec<CustomTlvRecord>>, preimage: Option<PaymentPreimage>,
 	) -> Result<PaymentId, Error> {
 		let rt_lock = self.runtime.read().unwrap();
 		if rt_lock.is_none() {
 			return Err(Error::NotRunning);
 		}
 
-		let payment_preimage = PaymentPreimage(self.keys_manager.get_secure_random_bytes());
+		let payment_preimage = preimage
+			.unwrap_or_else(|| PaymentPreimage(self.keys_manager.get_secure_random_bytes()));
+
 		let payment_hash = PaymentHash::from(payment_preimage);
 		let payment_id = PaymentId(payment_hash.0);
 
--- a/tests/integration_tests_rust.rs
+++ b/tests/integration_tests_rust.rs
@@ -19,8 +19,8 @@ use common::{
 use ldk_node::config::EsploraSyncConfig;
 use ldk_node::liquidity::LSPS2ServiceConfig;
 use ldk_node::payment::{
-	ConfirmationStatus, PaymentDirection, PaymentKind, PaymentStatus, QrPaymentResult,
-	SendingParameters,
+	ConfirmationStatus, PaymentDetails, PaymentDirection, PaymentKind, PaymentStatus,
+	QrPaymentResult, SendingParameters,
 };
 use ldk_node::{Builder, Event, NodeError};
 
@@ -29,8 +29,10 @@ use lightning::routing::gossip::{NodeAlias, NodeId};
 use lightning::util::persist::KVStore;
 
 use lightning_invoice::{Bolt11InvoiceDescription, Description};
+use lightning_types::payment::PaymentPreimage;
 
 use bitcoin::address::NetworkUnchecked;
+use bitcoin::hashes::sha256::Hash as Sha256Hash;
 use bitcoin::hashes::Hash;
 use bitcoin::Address;
 use bitcoin::Amount;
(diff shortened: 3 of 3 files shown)


src/payment/spontaneous.rs
```diff
@@ -57,27 +57,44 @@ impl SpontaneousPayment {
...
 		custom_tlvs: Vec<CustomTlvRecord>,
 	) -> Result<PaymentId, Error> {
-		self.send_inner(amount_msat, node_id, sending_parameters, Some(custom_tlvs))
+		self.send_inner(amount_msat, node_id, sending_parameters, Some(custom_tlvs), None)
+	}
+
+	/// Send a spontaneous with custom preimage
+	pub fn send_with_preimage(
+		&self, amount_msat: u64, node_id: PublicKey, sending_parameters: Option<SendingParameters>,
+		preimage: PaymentPreimage,
```
chuksys: I'm wondering if we should expand the purpose of `SendingParameters` a bit to allow passing of any custom parameters required for sending payments (not just routing and pathfinding parameters).

This approach would allow us to include preimage as an Option within `SendingParameters`, providing a single, coherent place for all payment-sending customizations.

A significant benefit of this change would be the ability to deprecate or drop specialized functions like `send_with_preimage` and `send_with_custom_tlvs` (in another PR), reducing overall boilerplate for users.

What do you think about this?

src/payment/spontaneous.rs
```diff
@@ -57,27 +57,44 @@ impl SpontaneousPayment {
...
 	/// Send a spontaneous payment including a list of custom TLVs.
 	pub fn send_with_custom_tlvs(
 		&self, amount_msat: u64, node_id: PublicKey, sending_parameters: Option<SendingParameters>,
 		custom_tlvs: Vec<CustomTlvRecord>,
 	) -> Result<PaymentId, Error> {
-		self.send_inner(amount_msat, node_id, sending_parameters, Some(custom_tlvs))
+		self.send_inner(amount_msat, node_id, sending_parameters, Some(custom_tlvs), None)
+	}
+
+	/// Send a spontaneous with custom preimage
```
chuksys: I think you missed the word "payment" here (after spontaneous) - that's if we'd still be adding this method.

chuksys (commented):
Just shared a few thoughts regarding the overall design of this API and its extensibility.

src/payment/spontaneous.rs
aagbotemi: I initially hesitated about mixing routing parameters with payment content. Your point about expanding `SendingParameters` to handle all payment customization makes sense for the long-term API design, and the benefit outweigh the conceptual mixing.

I'll refactor to move preimage into `SendingParameters` as an Option<PaymentPreimage>. I can drop the `send_with_preimage` and `send_with_preimage_and_custom_tlvs`, then `send_with_custom_tlvs` can be dropped in another PR. 

Thank you.

src/payment/spontaneous.rs
aagbotemi: Yeah, I missed that. Thank you. I am removing the method.

src/payment/spontaneous.rs
```diff
@@ -77,7 +77,10 @@ impl SpontaneousPayment {
 			return Err(Error::NotRunning);
 		}
 
-		let payment_preimage = PaymentPreimage(self.keys_manager.get_secure_random_bytes());
+		let payment_preimage = sending_parameters
```
chuksys: If we are doing this, I think we'll need to use `override_params` instead as that allows `send_inner` to select `sending_parameters` set in the config (if not passed into the `send` method directly).

chuksys (commented):
If we are going to use `sending_parameters` to pass in the custom preimage, I think we should allow the user to set it in config or pass it directly.

src/payment/spontaneous.rs
tnull: > I'm wondering if we should expand the purpose of SendingParameters a bit to allow passing of any custom parameters required for sending payments (not just routing and pathfinding parameters).

Ah, no, I don't think we should do that. Note that after LDK 0.2 release (cf. #462) we'll replace `SendingParameters` with LDK's `RouteParametersConfig` everywhere, so really shouldn't make any changes prior to it. Also, while we might be fine to expose the preimage for `keysend` payments, I don't think we'd like to allow users to override it for BOLT11/BOLT12 generally.

> I'll refactor to move preimage into SendingParameters as an Option.

Please revert this. I think we can either have separate `send` APIs as you previously suggested, or add the preimage as an optional field to the existing ones. Will take a look after the revert, but your previous approach might be fine.

src/payment/spontaneous.rs
tnull: No, I don't think we should do this, see above.

src/payment/spontaneous.rs
chuksys: > Ah, no, I don't think we should do that. Note that after LDK 0.2 release (cf. #462) we'll replace `SendingParameters` with LDK's `RouteParametersConfig` everywhere, so really shouldn't make any changes prior to it. Also, while we might be fine to expose the preimage for `keysend` payments, I don't think we'd like to allow users to override it for BOLT11/BOLT12 generally.

Thanks for the clarification!

bindings/ldk_node.udl
```diff
@@ -213,6 +213,10 @@ interface SpontaneousPayment {
 	PaymentId send_with_custom_tlvs(u64 amount_msat, PublicKey node_id, SendingParameters? sending_parameters, sequence<CustomTlvRecord> custom_tlvs);
 	[Throws=NodeError]
 	void send_probes(u64 amount_msat, PublicKey node_id);
+	[Throws=NodeError]
```
tnull, line 217: nit: Let's move these new entries to the correct location w.r.t. API, i.e. below `send_with_custom_tlvs`.

src/payment/spontaneous.rs
```diff
@@ -57,27 +57,44 @@ impl SpontaneousPayment {
...
 		custom_tlvs: Vec<CustomTlvRecord>,
 	) -> Result<PaymentId, Error> {
-		self.send_inner(amount_msat, node_id, sending_parameters, Some(custom_tlvs))
+		self.send_inner(amount_msat, node_id, sending_parameters, Some(custom_tlvs), None)
+	}
+
+	/// Send a spontaneous payment with custom preimage
+	pub fn send_with_preimage(
+		&self, amount_msat: u64, node_id: PublicKey, sending_parameters: Option<SendingParameters>,
+		preimage: PaymentPreimage,
```
tnull: If we add this as a new required parameter, let's move it before the optional parameters, i.e., `Option<SendingParameters>` (here and below).

aagbotemi: Alright, will fix it. Thank you.

src/ffi/mod.rs
```diff
@@ -27,10 +35,15 @@ where
...
-pub fn maybe_wrap<T>(ldk_type: impl Into<T>) -> std::sync::Arc<T> {
+pub fn maybe_wrap_arc<T>(ldk_type: impl Into<T>) -> std::sync::Arc<T> {
 	std::sync::Arc::new(ldk_type.into())
 }
 
+#[cfg(feature = "uniffi")]
+pub fn maybe_wrap<T>(ldk_type: impl Into<T>) -> T {
+	ldk_type.into()
+}
+
```
chuksys: I'm probably missing something here but I'm wondering why we need to "split" `maybe_wrap` into these 2 functions.

src/ffi/mod.rs
```diff
@@ -41,7 +54,19 @@ pub fn maybe_try_convert_enum<T>(value: &T) -> Result<&T, crate::error::Error> {
...
 
+#[cfg(not(feature = "uniffi"))]
+pub fn maybe_wrap_arc<T>(value: T) -> T {
+	value
+}
+
 #[cfg(not(feature = "uniffi"))]
 pub fn maybe_wrap<T>(value: T) -> T {
 	value
 }
```
chuksys: These 2 fns are basically the same, please is there a reason why we need both of them?

chuksys (commented):
Generally looks good to me - would appreciate some clarity on why we have `maybe_wrap` and `maybe_wrap_arc`.

src/ffi/mod.rs
aagbotemi: `maybe_wrap` and `maybe_wrap_arc` is to handle different type wrapping requirements depending on whether the uniffi feature is enabled. `maybe_wrap_arc` wraps the input type in a `std::sync::Arc<T>`. `maybe_wrap` performs a simple into() conversion without additional wrapping, used in cases where the type doesn’t need to be wrapped in an Arc.

src/ffi/mod.rs
aagbotemi: both `maybe_wrap_arc` and `maybe_wrap` are identical, simply returning the input value. The reason for keeping them as separate functions is to maintain a consistent API that aligns with their distinct behaviors when the `uniffi` feature is enabled

src/ffi/mod.rs
chuksys: From my understanding, I thought `maybe_wrap` (as it was before) already handled wrapping depending on whether the uniFFI feature flag is enabled - if enabled wrap in an Arc, if not return same value as passed in? 

I noticed `maybe_wrap` (this new version) is used to conditionally wrap the preimage - does this mean the preimage never needs to be wrapped in an Arc even when the uniFFI feature flag is enabled?

enigbe: In #542, the wrapper types introduced were for structs with a bit more complexity and state than `PaymentPreimage`. From my understanding, it makes sense to allocate them on the heap because they are also expensive to clone. `PaymentPreimage` is simpler, wraps 32 bytes, and should be cheaper to clone and doesn't need to be `Arc`'ed.

Also, when compiled with `uniffi`, there was a bit of conflict about converting `maybe_wrap` into a macro so it can handle the different cases (i.e. types needing heap allocation vs types that do not), however I think @tnull may not like the idea for reasons stated in #526. 

This change is easily revertible though if `Arc`ing is preferred.

bindings/ldk_node.udl
```diff
@@ -446,6 +450,10 @@ dictionary PaymentDetails {
 	u64 latest_update_timestamp;
 };
 
+dictionary PaymentPreimage {
+	bytes inner;
```
enigbe: I wonder about what would be the most appropriate description for `[u8;N]`: `bytes` vs `sequence<u8>`?

src/ffi/mod.rs
chuksys: > In #542, the wrapper types introduced were for structs with a bit more complexity and state than `PaymentPreimage`. From my understanding, it makes sense to allocate them on the heap because they are also expensive to clone. `PaymentPreimage` is simpler, wraps 32 bytes, and should be cheaper to clone and doesn't need to be `Arc`'ed.

Thanks for the clarification! This makes sense.

bindings/ldk_node.udl
aagbotemi: `bytes` represents fixed-size binary data like preimage better

src/ffi/types.rs
```diff
@@ -665,21 +669,64 @@ impl UniffiCustomTypeConverter for PaymentHash {
 	}
 }
 
-impl UniffiCustomTypeConverter for PaymentPreimage {
-	type Builtin = String;
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PaymentPreimage {
+	pub(crate) inner: Vec<u8>,
```
tnull: For consistency reasons, please just store an `LdkPaymentPreimage` as an `inner`.

src/ffi/types.rs
```diff
@@ -665,21 +669,64 @@ impl UniffiCustomTypeConverter for PaymentHash {
...
+				lightning::io::ErrorKind::InvalidData,
+				"Invalid payment preimage",
+			)
+		})?;
+
+		ldk_preimage.0.write(writer)
+	}
+}
+
+impl Readable for PaymentPreimage {
```
tnull: No, we really shouldn't roll our own custom serialization `PaymentPreimage` here, we *just* use the wrapper type in the API, for nothing else.

src/ffi/mod.rs
```diff
@@ -18,6 +18,14 @@ where
 	wrapped_type.as_ref().as_ref()
 }
 
+#[cfg(feature = "uniffi")]
+pub fn maybe_extract<T, R>(wrapped_type: T) -> Result<R, crate::error::Error>
```
tnull: I don't think we need this, as we could just reuse `maybe_try_convert_enum` (which could be renamed to `maybe_try_from`, maybe). 

In general, I'd prefer it if we wouldn't add any more wrapper helpers in this PR, I think (?) we should be able to make it work with what's there, no?

src/ffi/types.rs
aagbotemi: Alright. Thank you

src/ffi/types.rs
aagbotemi: Alright. Noted

src/ffi/mod.rs
aagbotemi: Okay, I will make use of the wrapper helper we currently have.

aagbotemi:
> When migrating `PaymentPreimage` to use its own wrapper type, it's important to _just_ use this type in the Uniffi API. It should just wrap the inner type and _not_ add any additional logic that might get stale over time. In particular, we def. shouldn't add specific serialization logic, as we'd still hold `LdkPaymentPreimage` internally, and just convert it at the API boundaries if necessary. To this end, please try to avoid adding more conversion wrapper methods if they are not really necessary (and I don't think they are in this case?).
> 
> It would be great if you could split this PR in two commits: a first commit introducing the Uniffi wrapper for `PaymentPreimage`, the second one exposing the new APIs for spontaneous payments.
> 
> Lastly, please try to write more conclusive commit messages that describe the changes, but also provide rationale for the changes. Feel free to refer to https://cbea.ms/git-commit/ for some guidance here. Thanks!


Thank you for the feedback @tnull. I will address the comments and push the updated changes.

enigbe:
> When migrating PaymentPreimage to use its own wrapper type, it's important to just use this type in the Uniffi API. It should just wrap the inner type and not add any additional logic that might get stale over time. In particular, we def. shouldn't add specific serialization logic, as we'd still hold LdkPaymentPreimage internally, and just convert it at the API boundaries if necessary.

@aagbotemi  initially tried this approach, representing `PaymentPreimage` as an interface/object that wraps `LdkPaymentPreimage`. However, because the `LightningBalance` enum has a variant struct with `PaymentPreimage` field, we are unable to just wrap the inner type as is because there is no support for interfaces/objects in an enum.

```bash
thread 'main' panicked at build.rs:10:59:
  called `Result::unwrap()` on an `Err` value: Objects cannot currently be used in enum variant data
``` 
I recommended representing `PaymentPreimage` as a dictionary to circumvent this error, but that means we have to add specific serialization for this, as has been done. 

We are currently using
```toml
uniffi = { version = "0.27.3", features = ["build"], optional = true }
```
and this error originates from this:
```rust
# uniffi_udl-0.27.3/src/converters/callables.rs
impl APIConverter<FieldMetadata> for weedle::argument::SingleArgument<'_> {
    fn convert(&self, ci: &mut InterfaceCollector) -> Result<FieldMetadata> {
        let type_ = ci.resolve_type_expression(&self.type_)?;
        if let Type::Object { .. } = type_ {
            bail!("Objects cannot currently be used in enum variant data");
        }
        if self.default.is_some() {
            bail!("enum interface variant fields must not have default values");
        }
        if self.attributes.is_some() {
            bail!("enum interface variant fields must not have attributes");
        }
        // TODO: maybe we should use our own `Field` type here with just name and type,
        // rather than appropriating record::Field..?
        Ok(FieldMetadata {
            name: self.identifier.0.to_string(),
            ty: type_,
            default: None,
            docstring: None,
        })
    }
}
```
but may have been addressed in ["0.29.3"](https://github.com/mozilla/uniffi-rs/blob/0116e54c5eaa0a52e67878c97a176db646b007e7/uniffi_udl/src/converters/callables.rs#L27-L43).

@tnull any reservations about updating `uniffi` to explore this? any suggestions on how to handle this without additional logic is welcome.

src/payment/bolt11.rs
```diff
@@ -539,6 +539,7 @@ impl Bolt11Payment {
 		} else {
 			None
 		};
+
```
tnull: nit: Spurious newline.

src/payment/store.rs
```diff
@@ -221,40 +221,40 @@ impl StorableObject for PaymentDetails {
 				},
 			}
 		}
-		if let Some(preimage_opt) = update.preimage {
+		if let Some(preimage_opt) = &update.preimage {
```
tnull: Not sure why these changes taking-by-reference were introduced, but I don't think we don't need any of them, especially if they'd require us to unnecessarily clone values?

src/payment/bolt11.rs
aagbotemi: Okay, noted

src/payment/store.rs
aagbotemi: I have fixed it.

PR #551, opened 2025-06-03
Upgrade to BDK 2.0

tnull:
We upgrade BDK to the latest release.

Draft for now as we have yet to wait for the actual BDK 2.0.0.

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -107,6 +107,16 @@ where
 		self.inner.lock().unwrap().tx_graph().full_txs().map(|tx_node| tx_node.tx).collect()
 	}
 
+	pub(crate) fn get_unconfirmed_txids(&self) -> Vec<Txid> {
+		self.inner
+			.lock()
+			.unwrap()
+			.transactions()
+			.filter(|t| t.chain_position.is_unconfirmed())
+			.map(|t| t.tx_node.txid)
+			.collect()
+	}
+
 	pub(crate) fn current_best_block(&self) -> BestBlock {
 		let checkpoint = self.inner.lock().unwrap().latest_checkpoint();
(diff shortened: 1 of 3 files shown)


notmandatory:
BDK 2.0 is now released, see release page for compatible `bdk` chain client versions: https://github.com/bitcoindevkit/bdk_wallet/releases/tag/wallet-2.0.0

src/wallet/mod.rs
```diff
@@ -136,11 +146,12 @@ where
...
 	}
 
-	pub(crate) fn apply_unconfirmed_txs(
-		&self, unconfirmed_txs: Vec<(Transaction, u64)>,
+	pub(crate) fn apply_mempool_txs(
+		&self, unconfirmed_txs: Vec<(Transaction, u64)>, evicted_txids: Vec<(Txid, u64)>,
 	) -> Result<(), Error> {
 		let mut locked_wallet = self.inner.lock().unwrap();
 		locked_wallet.apply_unconfirmed_txs(unconfirmed_txs);
+		locked_wallet.apply_evicted_txs(evicted_txids);
```
jkczyz, line 154: It wasn't clear from the `bdk_wallet` docs or release page that this is a new requirement. When upgrading, how do you typically see what changes are needed?

jkczyz:
FYI, `check-cln` is failing

src/wallet/mod.rs
tnull: In this case we knew that we were waiting on exactly that change, but I also went through all the commit history of all the BDK crates since the last upgrade to double-check I'm not overlooking something fundamental. So yeah, I agree the changelog could do a better job listing all the changes that need to be considered (cc @notmandatory).

PR #566, opened 2025-06-08
(draft) feat: add `PaymentFailureReason` to `PaymentStatus`

reez:


reez:
> > Probably some nuance I'm missing in my implementation, and a lot of spots I didn't use reason, but tried to use reason in the location where had the PaymentFailureReason to attach to the payment status.
> 
> Hmm, so at least in this PR we won't have a reason to set, as we mark payments as failed. I honestly also like the simplicity of the `PaymentStatus` enum previously, but if we'd want to go this way, we should probably introduce our own reason object that is a superset of LDK's `PaymentFailureReason` and have it cover all cases to always describe _why_ we failed the payment.

Ah yes, I didn't fully flesh it out once I got stuck at the swift bindings part, this makes sense.

I like the simplicity of `PaymentStatus` enum too, but also was running into spots where I wish I had some associated data/reason for when the case was a failure.

PR #572, opened 2025-06-09
Client trusts lsp

martinsaposnic:
Please just review the second commit https://github.com/lightningdevkit/ldk-node/commit/480a3ce54bfc0181dfa1c0ab5fb8568526cd92a8. The first commit is just compatibility changes so it works with rust-lightning

This PR uses the functionality introduced in this other PR https://github.com/lightningdevkit/rust-lightning/pull/3838, and creates 2 tests that demonstrates the feature

Changes:

- call store_funding_transaction when funding transaction is created
- call channel_needs_manual_broadcast to check if it needs to call funding_transaction_generated_manual_broadcast or funding_transaction_generated
- call lsps2_funding_tx_broadcast_safe when FundingTxBroadcastSafe is emitted
- a refactor on PaymentClaimable so it does not autoclaim (this will be changed in a future commit)
- a new fn receive_via_jit_channel_manual_claim so the client can actually do a manual claim with claim_for_hash, instead of auto claiming
- two beautiful tests that check the functionality
   - lsps2_client_trusts_lsp checks that the funding_tx is not broadcasted until the htlc is claimed
   - lsps2_lsp_trusts_client_but_client_does_not_claim checks that the funding_tx is broadcasted even though the client didn't claim the htlc


src/event.rs
```diff
@@ -512,31 +513,86 @@ where
...
-									err
-								)
-							},
-							Err(err) => {
-								log_error!(
-									self.logger,
-									"Failed to process funding transaction: {:?}",
-									err
+						self.liquidity_source.as_ref().map(|ls| {
+							ls.lsps2_store_funding_transaction(
```
tnull: Would we always need to store the funding transaction, even if we're not doing the full flow?

src/event.rs
```diff
@@ -556,295 +612,142 @@ where
...
 							..PaymentDetailsUpdate::new(payment_id)
 						};
-						match self.payment_store.update(&update) {
-							Ok(_) => return Ok(()),
-							Err(e) => {
-								log_error!(self.logger, "Failed to access payment store: {}", e);
-								return Err(ReplayEvent());
-							},
-						};
+						self.payment_store.update(&update).map_err(|e| {
```
tnull: There are a lot of unrelated changes happening in this commit. a) not sure if I agree that they are worth doing at all but b) always try to keep such changes in separate commits please.

src/event.rs
martinsaposnic: I moved it to inside the `if` statement, so it will store it only if `needs_manual_broadcast` is true 👍

src/event.rs
martinsaposnic: Sorry about that, I made a rebase and a cleanup. now this commit https://github.com/lightningdevkit/ldk-node/pull/572/commits/53a8174306dde161872491bd9b361ada13ec7744 will have the minimal changes to make the client_trusts_lsp flow work

martinsaposnic: > There are a lot of unrelated changes happening in this commit.

This was a refactor on the PaymentClaimable event to make the manual claim work on my tests.

That part is removed after the rebase, so we should be good now

src/liquidity.rs
```diff
@@ -134,6 +133,8 @@ pub struct LSPS2ServiceConfig {
 	pub min_payment_size_msat: u64,
 	/// The maximum payment size that we will accept when opening a channel.
 	pub max_payment_size_msat: u64,
+	/// Use the client trusts lsp model
+	pub client_trusts_lsp: bool,
```
enigbe, line 141: I'm trying to understand the `client_trusts_lsp` configuration here. According to [bLIP-52](https://github.com/lightning/blips/blob/b48e5db6864d1de6e4b6d71a73ad75569cbff20c/blip-0052.md?plain=1#L121-L125), I expect the LSP to dynamically switch from `lsp_trusts_client` to `client_trusts_lsp` mode upon detecting an attack, without requiring a restart.
However, this configuration appears to be static and set at node startup. If that's the case, what happens when:
1. A node initially running in `lsp_trusts_client` mode detects an attack
2. The node restarts with `client_trusts_lsp`: `true` to switch modes
3. There's an existing outbound JIT channel where the client expects the LSP to broadcast the funding transaction before sending the preimage to claim.

What are the potential consequences for that in-flight JIT channel during the mode transition?

src/liquidity.rs
```diff
@@ -292,6 +293,81 @@ where
...
+			return false;
+		}
+
+		// if we are in a client_trusts_lsp model, then we check if the LSP has an LSPS2 operation in progress
+		self.lsps2_service.as_ref().map_or(false, |_| {
+			let lsps2_service_handler = self.liquidity_manager.lsps2_service_handler();
+			if let Some(handler) = lsps2_service_handler {
+				handler
+					.channel_needs_manual_broadcast(user_channel_id, &counterparty_node_id)
+					.unwrap_or(false)
```
enigbe: Unsure about `unwrap_or` here. I think we should log the error and be concerned about its cause if `channel_needs_manual_broadcast` returns an `APIError::APIMisuseError`? Say we couldn't find a channel mapped to the provided `user_channel_id`, should we still go on to call `channel_manager.funding_transaction_generated` for a non-existent channel when handling `FundingGenerationReady`? I am confused about this.

enigbe (commented):
I have partially reviewed [53a81743](https://github.com/lightningdevkit/ldk-node/pull/572/commits/53a8174306dde161872491bd9b361ada13ec7744) (pending tests) and found it helpful in clarifying my understanding of LSPs. I have some initial comments/questions regarding the following:

### client_trusts_lsp configuration
- `client_trusts_lsp` is configured statically but I have assumed that based on [blip52](https://github.com/lightning/blips/blob/b48e5db6864d1de6e4b6d71a73ad75569cbff20c/blip-0052.md?plain=1#L121-L125), this could (should?) be dynamically configured without restarting the node when it detects an attack.
- Given the current implementation, what happens to in-flight outbound JIT channels (whose clients are awaiting funding transactions to be broadcast) during a trust mode transition, i.e. `lsp_trusts_client` to `client_trusts_lsp`?
### error silencing
I don't fully understand why possible `APIMisuseError` relating to `channel_needs_manual_broadcast` is silenced/ignored with `unwrap_or(false)`. Should we proceed with funding if we can't, for example, associate a channel with the `user_channel_id`?

src/liquidity.rs
martinsaposnic: just pushed a fixup that makes the flag dynamic

> There's an existing outbound JIT channel where the client expects the LSP to broadcast the funding transaction before sending the preimage to claim.

I just posted a question about this https://discord.com/channels/915026692102316113/994015949176963183/1397280758196080660

my current interpretation about this is that once lsps2.buy succeeds, that flag is part of the contract for this flow. the LSP can always abort and make you start over, but it cannot change the trust model mid negotiation

we will see what they respond on discord

src/liquidity.rs
martinsaposnic: this is also fixed in commit https://github.com/lightningdevkit/ldk-node/pull/572/commits/2cc52676933a3f99d880b2e304b8a4badff74692

thank you for the review!

src/liquidity.rs
martinsaposnic: > my current interpretation about this is that once lsps2.buy succeeds, that flag is part of the contract for this flow. the LSP can always abort and make you start over, but it cannot change the trust model mid negotiation

I created a test that shows how this works (test lsps2_in_flight_under_attack_switch)

tnull: > my current interpretation about this is that once lsps2.buy succeeds, that flag is part of the contract for this flow. the LSP can always abort and make you start over, but it cannot change the trust model mid negotiation

Yes, as also noted on Discord, I agree with that interpretation and even think that for now we can leave the flag to be statically determined at startup.

Issue #578, opened 2025-06-18
Consider allowing users to see transaction details before sending onchain payments


tnull:
We should consider to allow users to see transaction details before sending onchain payments. In particular, they might want to see the fee they are going to pay before they actually send the onchain payment.

We could consider an API where we allow to 'stage' a pending transaction first, and then have the user release/cancel it with another call. However, if we do this, we'll have to figure out when/how to prune pending transactions if the user doesn't end up releasing/cancelling them.

The latter part is especially important as we need to make sure that, once BDK allows us to, lock the used UTXOs when staging, and unlock them in the cancellation case, or after some timeout.

As this introduces quite some complication to our payment flow, I'm not fully sold on the idea, but it indeed seems like a pretty useful feature for users.

tnull:
> [@tnull](https://github.com/tnull) , I've been thinking about this. Do we need the full staged transaction complexity? The main goal seems to be showing users the fee before they commit. Could we instead expose a simpler `estimate_onchain_fee(amount, address)` API that:
> 
> * Calculates fee with current `feerates`
> * Returns the estimate
> 
> This avoids UTXO locking, timeouts, and state management. Fees are dynamic anyway, so staging doesn't guarantee accuracy if feerates change between preview and send. We could start with this simpler approach and only add transaction staging later if users specifically need to preview the complete transaction structure. What do you think?

Hmm, well, the issue there is that not only fees might change, but also your wallet state might change inbetween, meaning that the wallet might end up adding or removing a (e.g., change) output, which could considerably shift the fees in one or the other direction. 

Note that we'll also have the same issue for splicing (cf. https://github.com/lightningdevkit/ldk-node/pull/677#discussion_r2496930013), so working out an API design that works similarly in all cases would be preferable. That's why I'm hesitant to just 'ship the simple thing' right away.

f3r10:
Hi @tnull is this issue open to work on? It seems challenging. I am going to investigate further and see if I can come up with a viable approach.

PR #585, opened 2025-06-20
Add basic fee estimation to on-chain wallet

benthecarman:
Related to #578 

This adds basic fee estimation to the on-chain wallet. To not just completely copy-paste everything from `send_to_address` I refactored out a lot of the logic into its own function that `send_to_address` now uses. Then exposed the functionality and added it to tests.

### DIFF
--- a/tests/integration_tests_rust.rs
+++ b/tests/integration_tests_rust.rs
@@ -383,9 +383,19 @@ fn onchain_send_receive() {
 	);
 
 	let amount_to_send_sats = 54321;
+	let fee_estimate = node_a
+		.onchain_payment()
+		.estimate_send_to_address(&addr_a, amount_to_send_sats, None)
+		.unwrap();
 	let txid =
 		node_b.onchain_payment().send_to_address(&addr_a, amount_to_send_sats, None).unwrap();
 	wait_for_tx(&electrsd.client, txid);
+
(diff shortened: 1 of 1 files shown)


src/payment/onchain.rs
```diff
@@ -122,4 +122,63 @@ impl OnchainPayment {
...
+	/// Estimates the fee for sending an on-chain payment to the given address.
+	///
+	/// This will respect any on-chain reserve we need to keep, i.e., won't allow to cut into
+	/// [`BalanceDetails::total_anchor_channels_reserve_sats`].
+	///
+	/// If `fee_rate` is set it will be used for estimating the resulting transaction. Otherwise we'll retrieve
+	/// a reasonable estimate from the configured chain source.
+	///
+	/// [`BalanceDetails::total_anchor_channels_reserve_sats`]: crate::BalanceDetails::total_anchor_channels_reserve_sats
+	pub fn estimate_send_to_address(
```
tnull, line 135: Hmm, as discussed elsewhere, this is kind of an odd API as it is inherently TOCTOU race-y (based on fee rate updates, but also wallet state changes), and we don't offer any API that would allow to set exactly that net. fee  returned here. I think generally the 'staging' approach as described over at #578 would be preferable, but that of course also interacts with UTXO locking, etc.

Issue #587, opened 2025-06-25
bitcoind rpc gets stuck in an infinite loop if incorrect password is provided


rolznz:
This is logged continuously, and the whole LDK-node process seems to get stuck.

```
{"level":"error","line":512,"log_type":"LDK-node","module_path":"ldk_node::chain","msg":"Failed to synchronize chain listeners: BlockSourceError { kind: Persistent, error: Custom { kind: InvalidData, error: \"invalid response length: 0 bytes\" } }","time":"2025-06-25T12:23:59+07:00"}
```

tosynthegeek:
> [@tosynthegeek](https://github.com/tosynthegeek) indicated interest in working on this.

Yeah sure. Would love to pick this up

rolznz:
> Could you expand on what 'seems to get stuck' means?

It's somehow stopping me from being able to exit the process. I have to force-kill it. ctrl+c which normally works fine in our app, just does nothing.

PR #588, opened 2025-06-26
Implement Exponential Backoff for Transient Sync Errors

tosynthegeek:
This PR attempts to improve retry in the bitcoind RPC synchronization loop from #587 by replacing the existing linear backoff with a proper exponential backoff strategy. We could also add a maximum delay to prevent excessively long waits, but I am not sure if it would really help.

### DIFF
--- a/src/chain/mod.rs
+++ b/src/chain/mod.rs
@@ -425,6 +425,9 @@ impl ChainSource {
 					"Starting initial synchronization of chain listeners. This might take a while..",
 				);
 
+				let mut backoff = CHAIN_POLLING_INTERVAL_SECS;
+				const MAX_BACKOFF_SECS: u64 = 300;
+
 				loop {
 					let channel_manager_best_block_hash =
 						channel_manager.current_best_block().block_hash;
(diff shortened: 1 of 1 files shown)


src/chain/mod.rs
```diff
@@ -504,8 +506,23 @@ impl ChainSource {
 
 						Err(e) => {
 							log_error!(logger, "Failed to synchronize chain listeners: {:?}", e);
-							tokio::time::sleep(Duration::from_secs(CHAIN_POLLING_INTERVAL_SECS))
-								.await;
+							backoff_delay_multiplier += 1;
```
tnull: I don't think we should let the backoff delay grow infinitely. Please include a sane upper bound.

src/chain/mod.rs
```diff
@@ -504,8 +506,23 @@ impl ChainSource {
...
+									logger,
+									"Transient error. Retrying sync in {} seconds (backoff multiplier: {}).",
+									backoff_delay,
+									backoff_delay_multiplier
+								);
+								tokio::time::sleep(Duration::from_secs(backoff_delay)).await;
+							} else {
+								log_error!(
+									logger,
+									"Failed to synchronize chain listeners: {:?}, e"
```
tnull: This doesn't compile. Also, as discussed elsewhere, we at least need to error out. But technically, we should keep retrying currently as otherwise we'd need to add panic here as permanently being unable to connect to the chain source is an unrecoverable error.

src/chain/mod.rs
tosynthegeek: Got it, I am adding a max backoff delay of 300 secs for this

src/chain/mod.rs
tosynthegeek: > This doesn't compile. Also, as discussed elsewhere, we at least need to error out. But technically, we should keep retrying currently as otherwise we'd need to add panic here as permanently being unable to connect to the chain source is an unrecoverable error.

@tnull As you suggested that panicking isnt best and not being able to error out I stuck with retrying but with the delay of 300 secs. Exponential backoff from `CHAIN_POLLING_INTERVAL_SECS` was only introduced for transient errors

PR #591, opened 2025-07-01
Bump `uniffi` dependency to v0.28.3

tnull:
We bump our `uniffi` dependency to v0.28.3 to unlock some of the nicer features `uniffi` added since the previously-used v0.27.3. However, we can't bump it further to v0.29.3, as we have users requiring compatibility with `uniffi-bindgen-go`, which only supports v0.28.3 at the time of writing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -84,7 +84,7 @@ esplora-client = { version = "0.12", default-features = false, features = ["toki
 esplora-client_0_11 = { package = "esplora-client", version = "0.11", default-features = false, features = ["tokio", "async-https-rustls"] }
 electrum-client = { version = "0.23.1", default-features = true }
 libc = "0.2"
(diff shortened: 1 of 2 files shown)


enigbe (approved):

tankyleo (approved):
Not familiar with `uniffi` myself, reproduced the diff on my machine, things seem to be working, and also ran successfully `scripts/uniffi_bindgen_generate_python.sh` and `scripts/uniffi_bindgen_generate_kotlin.sh`.

Likely I am missing some dependencies for `scripts/uniffi_bindgen_generate_swift.sh` and `scripts/uniffi_bindgen_generate_kotlin_android.sh`

Also double checked `uniffi-bindgen-go` and indeed the README says "Currently, uniffi-bindgen-go uses uniffi-rs version 0.28.3".

PR #592, opened 2025-07-07
Make shutdown procedure more robust

tnull:
These are a few improvements to our shutdown logic. Most notably, we now switch our 'shutdown' signal for the background event processor from using channels to just keeping the `JoinHandle`, which also gets rid of a potential race condition. Moreover, we attempt to clear the broadcast queue one last time before we shut down.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -147,9 +147,7 @@ use types::{
 };
 pub use types::{ChannelDetails, CustomTlvRecord, PeerDetails, UserChannelId};
 
-#[cfg(tokio_unstable)]
-use logger::log_trace;
-use logger::{log_debug, log_error, log_info, LdkLogger, Logger};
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -670,6 +667,17 @@ impl Node {
 		// Disconnect all peers.
 		self.peer_manager.disconnect_all_peers();
 
+		// Ensure we process the broadcast queue one last time, as we may have generated
+		// force-closures during peer disconnection.
```
wpaulino: If we're force closing due to a disconnection, then we never broadcast the funding transaction to begin with, so we shouldn't need to broadcast a commitment transaction. I assume the behavior you were seeing will end up getting fixed by https://github.com/lightningdevkit/rust-lightning/pull/3881.

tnull: Ah, okay. I expected this but wasn't sure if we could lean on it. Will drop the commit then.

PR #594, opened 2025-07-08
Adds `InboundChannelsConfig` configuration

a-mpch:
Fixes partially #502 

## Context

When running a personal or business node would like to control how peers request channels either blocking all possible channels, announced channels or some channels given a minimum.

This is useful as operator to take care of how liquidity moves. Even more, if we acquired a channel LSP for specific reasons and don't want that liquidity to move.

Other example is when running a operative node (i.e, exchange / wallet) don't want to be full of tiny channels that could affect route finding.

## Solution

Added a `InboundChannelsConfig` configuration on `config.rs` with two parameters to reject channels or set a minimum channel size. In case that node runner wants to block all possible channels, it would mean to set the `minimum_channel_size` to a greater number.

## Questions
I was thinking that better approach is to have a "channels_config" and inside would have inbound (or policy / limits) and also de Anchors channel config.

### DIFF
--- a/bindings/ldk_node.udl
+++ b/bindings/ldk_node.udl
@@ -13,11 +13,17 @@ dictionary Config {
 	u64 probing_liquidity_limit_multiplier;
 	AnchorChannelsConfig? anchor_channels_config;
 	SendingParameters? sending_parameters;
+	InboundChannelsConfig inbound_channels_config;
 };
 
 dictionary AnchorChannelsConfig {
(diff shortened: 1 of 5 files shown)


a-mpch:
No problem, maybe I could increase the scope of this to help on the LSP interface so it could be solved in a better way. 

Said so, wouldn't LSP nodes would like to reject most channels except the ones that LSP opens as liquidity request or collaboratively open with their peers? Maybe I'm missing something from the spec?

PR #596, opened 2025-07-18
Add option to include headers in Esplora config

benthecarman:
This is useful if the esplora server has a form of authentication in front of it


src/builder.rs
```diff
@@ -294,9 +295,27 @@ impl NodeBuilder {
...
+		self.chain_data_source_config = Some(ChainDataSourceConfig::Esplora {
+			server_url,
+			headers: Default::default(),
+			sync_config,
+		});
+		self
+	}
+
+	/// Configures the [`Node`] instance to source its chain data from the given Esplora server.
+	/// The given `headers` will be included in all requests to the Esplora server, typically used for
```
tnull, line 309: nit: Can we add a whitespace line after the first sentence here (and below, ofc)?
```suggestion
	///
	/// The given `headers` will be included in all requests to the Esplora server, typically used for
```

PR #598, opened 2025-07-24
Clean up and modularize chain syncing logic

tnull:
We intended to modularize and clean up our chain sync since quite a while. Here, we finally do it by introducing a new `ChainSourceKind` type and moving the respective syncing logic to new `EsploraChainSource`, `ElectrumChainSource`, and `BitcoindChainSource` objects in the respective modules.

This is a rather large refactor, but 99% of the changes are code moves.

Additional cleanup steps / minor fixes will be coming up, but given the PR size I'll probably open a follow-up for them.

### DIFF
--- a/src/chain/mod.rs
+++ b/src/chain/mod.rs
@@ -347,9 +347,7 @@ impl ChainSource {
 			}
 		}
 	}
-}
 
-impl ChainSource {
 	// Synchronize the onchain wallet via transaction-based protocols (i.e., Esplora, Electrum,
 	// etc.)
 	pub(crate) async fn sync_onchain_wallet(&self) -> Result<(), Error> {
@@ -367,9 +365,7 @@ impl ChainSource {
 			},
 		}
 	}
-}
 
-impl ChainSource {
 	// Synchronize the Lightning wallet via transaction-based protocols (i.e., Esplora, Electrum,
 	// etc.)
 	pub(crate) async fn sync_lightning_wallet(
@@ -394,9 +390,7 @@ impl ChainSource {
 			},
 		}
 	}
-}
 
-impl ChainSource {
 	pub(crate) async fn poll_and_update_listeners(
 		&self, channel_manager: Arc<ChannelManager>, chain_monitor: Arc<ChainMonitor>,
 		output_sweeper: Arc<Sweeper>,
@@ -419,9 +413,7 @@ impl ChainSource {
 			},
 		}
 	}
-}
 
-impl ChainSource {
 	pub(crate) async fn update_fee_rate_estimates(&self) -> Result<(), Error> {
 		match &self.kind {
 			ChainSourceKind::Esplora(esplora_chain_source) => {
@@ -435,9 +427,7 @@ impl ChainSource {
 			},
 		}
 	}
-}
 
-impl ChainSource {
 	pub(crate) async fn process_broadcast_queue(&self) {
 		match &self.kind {
 			ChainSourceKind::Esplora(esplora_chain_source) => {


src/chain/mod.rs
```diff
@@ -697,118 +650,8 @@ impl ChainSource {
 	// etc.)
 	pub(crate) async fn sync_onchain_wallet(&self) -> Result<(), Error> {
 		match &self.kind {
-			ChainSourceKind::Esplora {
```
joostjager: I really don't know how to review a diff like this. But I think it is possible to improve by keeping the moved code close to where it was removed in a first commit, and then doing a clean move in a second.

image

I think it is too much to ask from a reviewer to see changes like this in the diff as it is.

tnull: Hmm, will play around with that a bit. Maybe it makes sense to first move everything into individual `impl` blocks beneath the original method before joining all in the new location.

tnull: I now went ahead and introduced intermittent `impl` blocks, and split the commits introducing the new chain source types and moving them to the dedicated modules. All of this should make for a considerable improvement, especially when reviewed via `git diff --color-moved --ignore-all-space`. I think there are some areas where `git diff` is still tripping, but I honestly don't know how to improve things further there. 

Now pushed this new rewrite, which amounted to an net diff that is almost 0 (just moved a single helper method to a more reasonable location in the process).

src/chain/mod.rs
```diff
@@ -1514,84 +1537,87 @@ impl ChainSource {
...
+									// Log 400 at lesser level, as this often just means bitcoind already knows the
+									// transaction.
+									// FIXME: We can further differentiate here based on the error
+									// message which will be available with rust-esplora-client 0.7 and
+									// later.
 									log_trace!(
-										logger,
-										"Successfully broadcast transaction {}",
-										txid
+										self.logger,
```
joostjager: Glad you did this somewhat artificial commit struct. Very easy to ascertain that it's still the same.

src/chain/electrum.rs
```diff
@@ -42,7 +42,83 @@ const BDK_ELECTRUM_CLIENT_BATCH_SIZE: usize = 5;
 const ELECTRUM_CLIENT_NUM_RETRIES: u8 = 3;
 const ELECTRUM_CLIENT_TIMEOUT_SECS: u8 = 10;
 
-pub(crate) struct ElectrumRuntimeClient {
+pub(super) enum ElectrumRuntimeStatus {
```
joostjager: Maybe the diff would improve a bit if you left this at roughly the same location as where it was.

src/chain/bitcoind.rs
```diff
@@ -5,27 +5,578 @@
 // http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
 // accordance with one or both of these licenses.
 
-use crate::types::{ChainMonitor, ChannelManager, Sweeper, Wallet};
+use super::WalletSyncStatus;
```
joostjager, line 8: commit msg typo 'bitoind'

joostjager (approved):
This PR shows a really nice decomposition / clean up. I think it was worth it.

The only real comment/question that I have is whether `ChainSource` is still needed? It is now mainly a pass through and I was wondering if that could be replaced by a trait?

src/chain/bitcoind.rs
tnull: Fixed.

src/chain/electrum.rs
tnull: I'm not sure I moved this much, but rather moved other things around that have `git` conclude that this was moved. Not sure it's worth experimenting further, I already did a few iterations until I reached the current state^^

joostjager: Yes, fair. Current state is already 90% of the win, if not 100% given diff limits.

PR #599, opened 2025-07-31
Refactor chain sourcing logic to make result propagation more robust

tnull:
Previously, we might have overlooked some cases that would exit the syncing methods and bubble up an error via `?` instead of propagating to all subscribers.

Here, we split out the code to inner methods to ensure we always propagate.

Should be easy to review via `git diff --color-moved --ignore-all-space`

### DIFF
--- a/src/chain/bitcoind.rs
+++ b/src/chain/bitcoind.rs
@@ -5,7 +5,7 @@
 // http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
 // accordance with one or both of these licenses.
 
-use super::WalletSyncStatus;
+use super::{periodically_archive_fully_resolved_monitors, WalletSyncStatus};
 
 use crate::config::{
 	BitcoindRestClientConfig, Config, FEE_RATE_CACHE_UPDATE_TIMEOUT_SECS, TX_BROADCAST_TIMEOUT_SECS,
@@ -340,7 +340,7 @@ impl BitcoindChainSource {
 		let chain_listener = ChainListener {
 			onchain_wallet: Arc::clone(&self.onchain_wallet),
 			channel_manager: Arc::clone(&channel_manager),
-			chain_monitor,
+			chain_monitor: Arc::clone(&chain_monitor),
 			output_sweeper,
 		};
 		let mut spv_client =
@@ -355,6 +355,14 @@ impl BitcoindChainSource {
 					now.elapsed().unwrap().as_millis()
 				);
 				*self.latest_chain_tip.write().unwrap() = Some(tip);
+
+				periodically_archive_fully_resolved_monitors(
+					Arc::clone(&channel_manager),
+					chain_monitor,
+					Arc::clone(&self.kv_store),
+					Arc::clone(&self.logger),
+					Arc::clone(&self.node_metrics),
+				)?;
 			},
 			Ok(_) => {},
 			Err(e) => {


src/chain/electrum.rs
```diff
@@ -245,9 +257,7 @@ impl ElectrumChainSource {
 			)?;
 		}
 
-		self.lightning_wallet_sync_status.lock().unwrap().propagate_result_to_subscribers(res);
-
-		res
+		Ok(())
```
joostjager: Not propagate an error from `sync_confirmables`?

src/chain/bitcoind.rs
```diff
@@ -306,6 +306,18 @@ impl BitcoindChainSource {
 			})?;
 		}
 
+		let res = self
+			.poll_and_update_listeners_inner(channel_manager, chain_monitor, output_sweeper)
+			.await;
+
+		self.wallet_polling_status.lock().unwrap().propagate_result_to_subscribers(res);
+		res
```
joostjager, line 315: nit: elsewhere you had a newline here

src/chain/bitcoind.rs
```diff
@@ -400,20 +394,11 @@ impl BitcoindChainSource {
...
-		match write_res {
-			Ok(()) => (),
-			Err(e) => {
-				log_error!(self.logger, "Failed to persist node metrics: {}", e);
-				let res = Err(Error::PersistenceFailed);
-				self.wallet_polling_status.lock().unwrap().propagate_result_to_subscribers(res);
-				return res;
-			},
-		}
+		)?;
```
joostjager, line 410: Logging  and error mapping no longer necessary?

src/chain/bitcoind.rs
```diff
@@ -329,9 +329,7 @@ impl BitcoindChainSource {
 				},
 				Err(e) => {
 					log_error!(self.logger, "Failed to poll for chain data: {:?}", e);
-					let res = Err(Error::TxSyncFailed);
-					self.wallet_polling_status.lock().unwrap().propagate_result_to_subscribers(res);
```
joostjager, line 321: The cleanup here, does that mean that the previous commit would send redundant updates?

src/chain/bitcoind.rs
```diff
@@ -354,6 +354,14 @@ impl BitcoindChainSource {
 					now.elapsed().unwrap().as_millis()
 				);
 				*self.latest_chain_tip.write().unwrap() = Some(tip);
+
+				periodically_archive_fully_resolved_monitors(
```
joostjager, line 359: Context question: channel manager already knows the best block and I assume it is updated somehow, but still we need an external trigger here to archive? It seems responsibility is now with two units instead of one.

src/chain/electrum.rs
tnull: Oh, good catch. Fixed.

src/chain/bitcoind.rs
tnull: Fixed.

src/chain/bitcoind.rs
tnull: No, we don't do it in the other cases, and will just log the error after bubbling up anyways.

src/chain/bitcoind.rs
tnull: Not quite sure what you mean? We previously propagated manually whenever erroring out. But now we can just bubble up the error as we'll propagate to subscribers in the wrapping method.

src/chain/bitcoind.rs
tnull: So far we left when/how to archive to the users in LDK. I guess we could consider making it part of the background processor eventually, but I don't think `ChannelManager` could do it (in absence of `KVStore` access, etc)

src/chain/bitcoind.rs
joostjager: Yes, but before the change in this commit (this specific commit), was the propagate method being called twice? I am wondering if all commits in the stack are leaving the code in a correct state.

src/chain/bitcoind.rs
joostjager: Ok. I was confused a bit about the two sources of block information being used (here the connection event, and then the height in cm). But if this is how it is done for now, it's ok.

src/chain/bitcoind.rs
tnull: Ah, yes, that might be true. Happy to squash the cleanup if you prefer.

joostjager (approved):
LGTM. One comment outstanding about the stability of an intermediate commit, but non blocking.

src/chain/bitcoind.rs
tnull: Now went ahead and did so

PR #604, opened 2025-08-08
Add reorg test

### LINKED ISSUE #575: Property-based reorg tests
We should add tests that assert that we continue to operate correct after reorgs happen. 

To this end, we should employ the `proptest` crate to setup randomized scenarios and assert that certain invariants hold independently when blocks are connected/reorged, and when syncs are happening. These scenarios should be tested for all chain sources (Esplora, Electrum, bitcoind RPC).

moisesPompilio:
This PR introduces functional tests to validate node behavior under blockchain reorganization scenarios, specifically when opening and closing channels.

- **Channel open reorg**: Ensures the node maintains channel consistency after a reorg occurring immediately after channel creation.
- **Channel close reorg**: Validates both normal close and force close operations when a reorg happens during channel closure, confirming that state remains consistent.

Additionally, the premine_and_distribute_funds helper was optimized. Previously, when many addresses were funded, transactions were sent sequentially, waiting for each to appear in the mempool before sending the next. This was slow. The implementation now uses sendmany from Bitcoin Core to send all outputs in a single transaction, improving speed for all tests relying on this function.

Close #575


tests/prop_test.rs
tnull, line 1: Let's rename this `reorg_tests.rs`

tests/reorg_test.rs
```diff
@@ -0,0 +1,219 @@
...
+		config_node!(chain_source_electrsd, anchor_channels),
+		config_node!(chain_source_bitcoind, anchor_channels),
+		config_node!(chain_source_esplora, anchor_channels),
+	];
+
+	let (bitcoind, electrs) = (&bitcoind.client, &electrsd.client);
+	macro_rules! reorg {
+		($reorg_depth: expr) => {{
+			invalidate_blocks(bitcoind, $reorg_depth);
+			generate_blocks_and_wait(bitcoind, electrs, $reorg_depth);
```
tnull: I want to note that operating on the same chain and generating/invalidating blocks is only safe/working because `proptest`s are run sequentially as it seems. Fine for this case, but maybe we should inline `init_setup_test_reorg` to avoid confusion in the future if other tests would try to reuse the same helper?

tests/reorg_test.rs
```diff
@@ -0,0 +1,219 @@
...
+
+proptest! {
+	#![proptest_config(ProptestConfig::with_cases(5))]
+	#[test]
+	fn reorg_test(
+		reorg_depth in 1..=6usize,
+		force_close in prop::bool::ANY,
+	) {
+
+	let (bitcoind, electrsd) = init_setup_test_reorg();
```
tnull: nit: Seems all of this method needs one more indentation to the right. `rustfmt` didn't catch this as it doesn't operate on macro code.

tests/reorg_test.rs
```diff
@@ -0,0 +1,219 @@
...
+			},
+			_ => panic!("Unexpected balance state for node_hub!"),
+			}
+
+			assert!(node.list_balances().lightning_balances.len() < 2);
+
+			assert!(node.list_balances().pending_balances_from_channel_closures.len() > 0);
+			match node.list_balances().pending_balances_from_channel_closures[0] {
+				PendingSweepBalance::BroadcastAwaitingConfirmation { .. } => {},
+				_ => println!("Unexpected balance state!"),
```
tnull: Should we panic here and below rather than just printing that we reached an unexpected state?

tests/reorg_test.rs
```diff
@@ -0,0 +1,219 @@
...
+
+	generate_blocks_and_wait(bitcoind, electrs, 6);
+	sync_wallets!();
+
+	reorg!(reorg_depth);
+	sync_wallets!();
+
+	// Check balance after close channel
+	nodes.iter().for_each(|node| {
+			assert!(node.list_balances().spendable_onchain_balance_sats > amount - 7000);
```
tnull: Where is this 7000 number coming from? Mind introducing an appropriately named variable for it?

tests/reorg_test.rs
```diff
@@ -0,0 +1,219 @@
...
+					},
+					other => panic!("Unexpected event: {:?}", other),
+				}
+			}
+			user_channels
+		}};
+	}
+
+	let mut node_channels_id = HashMap::new();
+	for index in 0..nodes.len() {
```
tnull: nit: Rather than iterating over the index, can we use iterators? (here and elswhere)
```suggestion
	for node in nodes.iter() {
```

They are usually a bit easier to read.

moisesPompilio: Yeah, I replaced some for index loops with clearer iterator expressions.

tests/reorg_test.rs
moisesPompilio: Done! It’s the approximate tx fee to prevent amount check failures.

tests/reorg_test.rs
moisesPompilio: Yeah, had to remove it for tests, but added it back now.

tests/reorg_test.rs
moisesPompilio: done!

tests/reorg_test.rs
moisesPompilio: You’re right. I initially created `init_setup_test_reorg` when building the test, thinking I’d need many proptest rounds. Since reorgs don’t require that many cases, the performance impact is negligible now, so I’ve removed the helper.

tests/reorg_test.rs
```diff
@@ -0,0 +1,193 @@
...
+
+		let (bitcoind, electrs) = (&bitcoind.client, &electrsd.client);
+		macro_rules! reorg {
+			($reorg_depth: expr) => {{
+				invalidate_blocks(bitcoind, $reorg_depth);
+				generate_blocks_and_wait(bitcoind, electrs, $reorg_depth);
+			}};
+		}
+
+		let amount = 2_100_000;
```
tnull: For numerical amounts and fees please *always*  append a `_sat`/`_msat`/`_btc` suffix to make it explicit which unit we're dealing in (here and everywhere else).

moisesPompilio: Done!

Issue #534, opened 2025-04-29
Support BDK's `evicted_at` in our bitcoind RPC mempool syncing logic


tnull:
When upgrading to BDK 2.0, we'll have to replicate the functionality of https://github.com/bitcoindevkit/bdk/pull/1857 for our code.

This is merely a reminder to do this so we don't forget.

PR #605, opened 2025-08-08
Fix evicted transation

moisesPompilio:
This PR fixes an issue where `BitcoinD` failed to detect when a transaction was removed from the mempool, causing inconsistencies compared to `Electrum` and `Esplora.`

The root cause was an endianness mismatch in the `evicted_txids` check: one txid was in **little-endian** and the other in **big-endian**, making proper matching nearly impossible. The fix ensures consistent txid comparison by normalizing endianness.

Additionally, this PR adds an RBF functional test to verify that the node correctly:

- Detects a Replace-By-Fee event. 
- Replaces the original transaction in the mempool.
- Uses the updated transaction to open a channel and allocate the correct balance to the user.


src/chain/bitcoind.rs
```diff
@@ -1121,7 +1122,12 @@ impl BitcoindClient {
 		let mempool_entries_cache = mempool_entries_cache.lock().await;
 		let evicted_txids = unconfirmed_txids
 			.into_iter()
-			.filter(|txid| mempool_entries_cache.contains_key(txid))
+			.filter(|txid| {
+				let mut bytes = txid.to_byte_array();
```
tnull: I don't think we should fix this here, but when we first parse in the `Txid`, i.e., before we even add it to the cache.

While you're doing this, mind adding a few end-to-end tests for all the `JsonResponse` de/serialization, i.e., asserting that we get the same objects when we encode and decode `GetRawMempoolResponse`,`GetMempoolEntryResponse`, etc. These might also good candidates for simple proptests.

moisesPompilio: Got it, thanks! I’ll make those changes and add the tests as suggested.

moisesPompilio:
I updated the implementation to replace `bitcoin::consensus::encode::deserialize_hex` and `bitcoin::consensus::encode::serialize_hex` with `Txid::from_str` and `Txid::to_string`, aligning the behavior with the official [rust-bitcoin `Txid` documentation](https://docs.rs/bitcoin/latest/bitcoin/struct.Txid.html), where `fmt::Display` outputs the reversed-byte hex string expected by RPC. I also modified the logic for removing unconfirmed transactions so that when a transaction is no longer in the mempool, it is marked as evicted and removed from the node's balance count if it has not been included in a block. I think this change may solve [issue #534](https://github.com/lightningdevkit/ldk-node/issues/534), although I’m not entirely certain.  

I also added property-based tests to ensure that `JsonResponse` serialization and deserialization are fully consistent for `GetRawMempoolResponse`, `GetMempoolEntryResponse`, `GetRawTransactionResponse`, `FeeResponse`, and `MempoolMinFeeResponse`. These round-trip checks guarantee that encoded data can be decoded back into identical Rust objects.  

Additionally, I expanded the RBF functional tests to cover two scenarios:  
1. **RBF transactions in the mempool before confirmation** — ensuring the node detects the replacement, updates balances, and removes outdated outputs.  
2. **RBF transactions confirmed directly in a block** — verifying that the node correctly marks the replacement as confirmed and removes the obsolete unconfirmed transaction.  

For this, I introduced `distribute_funds_unconfirmed` to distribute funds without confirming them, and `MultiNodeTestSetup` to simplify multi-node orchestration, wallet synchronization, balance validation, RBF preparation, and fee bumping.

src/chain/bitcoind.rs
```diff
@@ -1109,7 +1109,7 @@ impl BitcoindClient {
 		let mempool_entries_cache = mempool_entries_cache.lock().await;
 		let evicted_txids = unconfirmed_txids
 			.into_iter()
-			.filter(|txid| mempool_entries_cache.contains_key(txid))
+			.filter(|txid| !mempool_entries_cache.contains_key(txid))
```
tnull, line 1116: Aaah, good catch. Mind renaming `unconfirmed_txids` to `bdk_mempool_txids` or `bdk_unconfirmed_txids`. I think a bit cleaner naming would have likely avoided that bug in the first place.

src/chain/bitcoind.rs
```diff
@@ -754,7 +754,7 @@ impl BitcoindClient {
 	async fn get_raw_transaction_rpc(
 		rpc_client: Arc<RpcClient>, txid: &Txid,
 	) -> std::io::Result<Option<Transaction>> {
-		let txid_hex = bitcoin::consensus::encode::serialize_hex(txid);
+		let txid_hex = txid.to_string();
```
tnull, line 760: Ugh, classic mistake. Thanks for catching this.

tests/common/mod.rs
```diff
@@ -1074,6 +1110,170 @@ pub(crate) fn do_channel_full_cycle<E: ElectrumApi>(
 	println!("\nB stopped");
 }
 
+pub(crate) struct MultiNodeTestSetup {
```
tnull: This is a lot of boilerplate, and would use a lot of CI resources for no added benefit. I think it would be preferable if the `test_rbf` tests would just reuse the existing test utilities. FWIW, to test RBF works as expected you'll only need to setup a single node (maybe one per chain source, as we do for the `full_cycle`) test, i.e., should just be able to use `setup_node`.

src/chain/bitcoind.rs
moisesPompilio: Done!

tests/common/mod.rs
moisesPompilio: Yeah, I initially made that struct for a different approach, but it turned out unnecessary. I removed it and now use one node per chain type, keeping the test lighter.

src/chain/bitcoind.rs
```diff
@@ -1411,3 +1411,164 @@ impl std::fmt::Display for HttpError {
...
+				version: bitcoin::transaction::Version(version),
+				input: inputs,
+				output: outputs,
+				lock_time: bitcoin::absolute::LockTime::from_consensus(lock_time),
+			}
+		}
+	}
+
+	proptest! {
+		#![proptest_config(proptest::test_runner::Config::with_cases(250))]
```
tnull: While good for test coverage initially, 250 seems like a high number if we run it every time we push to CI, no? Maybe 10 or 20 seems more reasonable?

tests/integration_tests_rust.rs
```diff
@@ -670,6 +673,217 @@ fn onchain_wallet_recovery() {
...
+		() => {
+			bump_fee_amount_sat = tx.vsize() as u64;
+			let attempts = 5;
+			for _attempt in 0..attempts {
+				bump_fee!();
+
+				match bitcoind.send_raw_transaction(&tx) {
+					Ok(res) => {
+						// Mine a block immediately so the transaction is confirmed
+						// before any node identifies it as a transaction that was in the mempool.
```
tnull: Hmm, depending on the chain source that might be race-y, no? In particular, for bitcoind RPC we might happen to poll the mempool just before the block gets mined?

src/chain/bitcoind.rs
moisesPompilio: Good point. Reducing from 250 to 20 only saves a few milliseconds, but I agree 250 is excessive for CI. 20 runs seems more reasonable for these scenarios.

tests/integration_tests_rust.rs
moisesPompilio: Yes, race conditions are possible, but they don’t affect the test outcome. When `is_insert_block` is true, there are three possible scenarios:

- When the node checks the mempool, it sees the RBF transaction. The old transaction is removed to allow the RBF. I haven’t triggered this in my tests, but it wouldn’t break anything.
- When the node checks the mempool, it finds it empty, either because the transaction is already in a block or the block is being processed. Old unconfirmed transactions are removed and the correct transaction is picked from the block. This scenario is rare but safe.
- When the node checks confirmed transactions in the block, it sees the RBF transaction already included. The old unconfirmed transaction is removed. This is the most common scenario in the test and effectively tests the case where the node first encounters the RBF transaction via the block rather than the mempool.

tests/integration_tests_rust.rs
```diff
@@ -670,6 +673,217 @@ fn onchain_wallet_recovery() {
...
+
+			let mut option_fee_output_index = None;
+			for (index, output) in tx.output.iter().enumerate() {
+				if !scripts_buf.contains(&output.script_pubkey) {
+					option_fee_output_index = Some(index);
+					break;
+				}
+			}
+			fee_output_index = option_fee_output_index.expect(
+				"No output available for fee pumping. Need at least one output not being modified.",
```
tnull: nit: `bump`?

tests/integration_tests_rust.rs
```diff
@@ -670,6 +673,217 @@ fn onchain_wallet_recovery() {
...
+						if is_insert_block {
+							generate_blocks_and_wait(bitcoind, electrs, 1);
+						}
+						let new_txid = res.0.parse().unwrap();
+						wait_for_tx(electrs, new_txid);
+						break;
+					},
+					Err(_) => {
+						if _attempt == attempts - 1 {
+							panic!("Failed to pump fee after {} attempts", attempts);
```
tnull: nit: `bump`?

tests/integration_tests_rust.rs
```diff
@@ -670,6 +673,217 @@ fn onchain_wallet_recovery() {
...
+
+	let (bitcoind, electrs) = (&bitcoind.client, &electrsd.client);
+	premine_blocks(bitcoind, electrs);
+
+	// Helpers declaration before starting the test
+	let all_addrs =
+		nodes.iter().map(|node| node.onchain_payment().new_address().unwrap()).collect::<Vec<_>>();
+	let amount_sat = 2_100_000;
+	let mut txid;
+	macro_rules! distribute_funds_all_nodes {
```
tnull, line 715: Can we make all these helpers methods rather than macros? Macros have serveral drawbacks, in particular they're much harder to debug in case of a failure as the compiler won't output the correct line numbers then.

tests/integration_tests_rust.rs
moisesPompilio: Done!

tests/integration_tests_rust.rs
moisesPompilio: Done!

tests/integration_tests_rust.rs
moisesPompilio: Yes, though I think `distribute_funds_all_nodes` and `validate_balances` can stay as macros. The first just calls a function, and in the second the main failure point is `assert_eq!`, so keeping them as `macro_rules!` helps keep the helper functions cleaner.

tnull: Hmm I would generally prefer to inline them or use methods, but won't hold this PR up over it.

PR #606, opened 2025-08-11
Bump MSRV to rustc 1.85

tnull:
We generally align our MSRV with Debian's stable channel. Debian 13 'Trixie' was just released, shipping rustc 1.85. We therefore bump our MSRV on the `main` branch here.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -18,7 +18,7 @@ jobs:
         toolchain: [
           stable,
           beta,
-          1.75.0, # Our MSRV
+          1.85.0, # Our MSRV
           ]
         include:
           - toolchain: stable
@@ -29,7 +29,7 @@ jobs:
             platform: macos-latest
           - toolchain: stable
             platform: windows-latest
-          - toolchain: 1.75.0
+          - toolchain: 1.85.0
             msrv: true
     runs-on: ${{ matrix.platform }}
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -42,11 +42,9 @@ jobs:
...
         if: matrix.check-fmt
         run: rustup component add rustfmt && cargo fmt --all -- --check
-      - name: Pin packages to allow for MSRV
-        if: matrix.msrv
-        run: |
-          cargo update -p home --precise "0.5.9" --verbose # home v0.5.11 requires rustc 1.81 or newer
-          cargo update -p idna_adapter --precise "1.1.0" --verbose # idna_adapter 1.2 switched to ICU4X, requiring 1.81 and newer
+#      - name: Pin packages to allow for MSRV
+#        if: matrix.msrv
+#        run:
```
jkczyz: Should we drop this?

.github/workflows/rust.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         toolchain: [
           stable,
           beta,
-          1.75.0, # Our MSRV
+          1.85.0, # Our MSRV
```
jkczyz, line 21: Given we were already at 1.75.0, I take it is ok that we are ahead of `rust-bitcoin` and `rust-lightning`? IIRC, those used 1.63.0 as the MSRV. We're we just ahead back we had some other dependency that required a higher MSRV?

tnull: Yes, it's mostly the HTTP stack depedencies that had us at 1.75. However, I generally like the concept of following Debian stable rustc (which it's pretty conservative).

.github/workflows/rust.yml
tnull: We could, I mainly left it in because I expect that some dependencies will require pinning again soon. Now dropped it.

PR #608, opened 2025-08-12
Add `receive_via_jit_channel_for_hash()` method

### LINKED ISSUE #597: `receive_via_jit_channel()` with custom hash
To implement swap-in functionality with a JIT channel (i.e., swapping an on-chain payment to Lightning by opening a new channel), I need a method similar to `receive_via_jit_channel()` that supports a custom hash. The existing `receive_for_hash()` method is insufficient because it does not allow for skimmed fees.

### Open a channel to LSP
An alternative to this functionality would be:
1. Receiving on-chain funds on the Lightning node.
2. Opening a channel to the LSP.

However, this approach has the following uncertainties:
- It is unclear whether the LSP will support zero-conf channels in this scenario.
- Without dual funding, the Lightning node will not obtain any inbound liquidity.

### LSPS1
Request a channel from LSP. This approach requires the client to trust the LSP, which might be sometimes unacceptable (e.g. Breez).

andrei-21:
This method will allow to implement swap-in functionality with a JIT channel (i.e., swapping an on-chain payment to Lightning by opening a new channel).

Closes #597.

### DIFF
--- a/src/payment/bolt11.rs
+++ b/src/payment/bolt11.rs
@@ -362,8 +362,17 @@ impl Bolt11Payment {
 		}
 
 		if let Some(details) = self.payment_store.get(&payment_id) {
-			if let Some(expected_amount_msat) = details.amount_msat {
-				if claimable_amount_msat < expected_amount_msat {
+			// For payments requested via `receive*_via_jit_channel_for_hash()`
+			// `skimmed_fee_msat` held by LSP must be taken into account.
+			let skimmed_fee_msat = match details.kind {
+				PaymentKind::Bolt11Jit {
+					counterparty_skimmed_fee_msat: Some(skimmed_fee_msat),
+					..
+				} => skimmed_fee_msat,
+				_ => 0,
+			};
+			if let Some(invoice_amount_msat) = details.amount_msat {
+				if claimable_amount_msat < invoice_amount_msat - skimmed_fee_msat {
 					log_error!(
 						self.logger,
 						"Failed to manually claim payment {} as the claimable amount is less than expected",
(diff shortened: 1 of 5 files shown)


src/payment/bolt11.rs
```diff
@@ -578,6 +578,44 @@ impl Bolt11Payment {
...
+	/// receive it via a newly created just-in-time (JIT) channel.
+	///
+	/// When the returned invoice is paid, the configured [LSPS2]-compliant LSP will open a channel
+	/// to us, supplying just-in-time inbound liquidity.
+	///
+	/// If set, `max_total_lsp_fee_limit_msat` will limit how much fee we allow the LSP to take for opening the
+	/// channel to us. We'll use its cheapest offer otherwise.
+	///
+	/// We will register the given payment hash and emit a [`PaymentClaimable`] event once
+	/// the inbound payment arrives.
```
tnull: Mind adding here that we still check that the LSP-withheld fees are what we expect *before* issuing the event, i.e., users don't have to double-check that before manually claiming.

src/payment/bolt11.rs
```diff
@@ -360,8 +360,15 @@ impl Bolt11Payment {
 		}
 
 		if let Some(details) = self.payment_store.get(&payment_id) {
+			let skimmed_fee_msat = match details.kind {
```
tnull, line 367: nit: Mind adding a comment here to describe when this would be relevant?

src/payment/bolt11.rs
```diff
@@ -360,8 +360,15 @@ impl Bolt11Payment {
...
+				PaymentKind::Bolt11Jit {
+					counterparty_skimmed_fee_msat: Some(skimmed_fee_msat),
+					..
+				} => skimmed_fee_msat,
+				_ => 0,
+			};
 			if let Some(expected_amount_msat) = details.amount_msat {
```
tnull: nit: Let's rename the variable here, as "the expected amount" will be the net amount, i.e., with `skimmed_fee_msat` already subtracted.

PR #621, opened 2025-08-19
Static invoice server

joostjager:
Wires up async payments functionality in ldk-node. This PR only covers the static invoice creation part. In a follow up, the full mechanism will be put in place to hold an htlc sender-side until the receiver comes online again.

### DIFF
--- a/bindings/ldk_node.udl
+++ b/bindings/ldk_node.udl
@@ -13,6 +13,7 @@ dictionary Config {
 	u64 probing_liquidity_limit_multiplier;
 	AnchorChannelsConfig? anchor_channels_config;
 	RouteParametersConfig? route_parameters;
+	boolean async_payment_services_enabled;
 };
 
 dictionary AnchorChannelsConfig {
@@ -209,6 +210,12 @@ interface Bolt12Payment {
 	Bolt12Invoice request_refund_payment([ByRef]Refund refund);
 	[Throws=NodeError]
 	Refund initiate_refund(u64 amount_msat, u32 expiry_secs, u64? quantity, string? payer_note);
+	[Throws=NodeError]
+	Offer receive_async();
+	[Throws=NodeError]
+	void set_paths_to_static_invoice_server(bytes paths);
+	[Throws=NodeError]
+	bytes blinded_paths_for_async_recipient(bytes recipient_id);
 };
 
 interface SpontaneousPayment {
@@ -311,6 +318,8 @@ enum NodeError {
 	"InsufficientFunds",
 	"LiquiditySourceUnavailable",
 	"LiquidityFeeTooHigh",
+	"InvalidBlindedPaths",
+	"AsyncPaymentServicesDisabled",
 };
 
 dictionary NodeStatus {
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1455,7 +1455,7 @@ fn build_with_store_internal(
 		Arc::clone(&channel_manager),
 		message_router,
 		Arc::clone(&channel_manager),
-		IgnoringMessageHandler {},
+		Arc::clone(&channel_manager),
 		IgnoringMessageHandler {},
 		IgnoringMessageHandler {},
 	));
--- a/src/config.rs
+++ b/src/config.rs
@@ -179,6 +179,8 @@ pub struct Config {
 	/// **Note:** If unset, default parameters will be used, and you will be able to override the
 	/// parameters on a per-payment basis in the corresponding method calls.
 	pub route_parameters: Option<RouteParametersConfig>,
+	/// Whether to enable the static invoice service to support async payment reception for clients.
+	pub async_payment_services_enabled: bool,
 }
 
 impl Default for Config {
@@ -193,6 +195,7 @@ impl Default for Config {
 			anchor_channels_config: Some(AnchorChannelsConfig::default()),
 			route_parameters: None,
 			node_alias: None,
+			async_payment_services_enabled: false,
 		}
 	}
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -120,6 +120,10 @@ pub enum Error {
 	LiquiditySourceUnavailable,
 	/// The given operation failed due to the LSP's required opening fee being too high.
 	LiquidityFeeTooHigh,
+	/// The given blinded paths are invalid.
+	InvalidBlindedPaths,
+	/// Asynchronous payment services are disabled.
+	AsyncPaymentServicesDisabled,
 }
 
 impl fmt::Display for Error {
@@ -192,6 +196,10 @@ impl fmt::Display for Error {
 			},
 			Self::LiquidityFeeTooHigh => {
 				write!(f, "The given operation failed due to the LSP's required opening fee being too high.")
+			},
+			Self::InvalidBlindedPaths => write!(f, "The given blinded paths are invalid."),
+			Self::AsyncPaymentServicesDisabled => {
+				write!(f, "Asynchronous payment services are disabled.")
 			},
 		}
 	}
--- a/src/event.rs
+++ b/src/event.rs
@@ -6,7 +6,6 @@
 // accordance with one or both of these licenses.
 
 use crate::types::{CustomTlvRecord, DynStore, PaymentStore, Sweeper, Wallet};
-
 use crate::{
 	hex_utils, BumpTransactionEventHandler, ChannelManager, Error, Graph, PeerInfo, PeerStore,
 	UserChannelId,
@@ -19,6 +18,7 @@ use crate::fee_estimator::ConfirmationTarget;
 use crate::liquidity::LiquiditySource;
 use crate::logger::Logger;
 
+use crate::payment::asynchronous::static_invoice_store::StaticInvoiceStore;
 use crate::payment::store::{
 	PaymentDetails, PaymentDetailsUpdate, PaymentDirection, PaymentKind, PaymentStatus,
 };
@@ -27,7 +27,7 @@ use crate::io::{
 	EVENT_QUEUE_PERSISTENCE_KEY, EVENT_QUEUE_PERSISTENCE_PRIMARY_NAMESPACE,
 	EVENT_QUEUE_PERSISTENCE_SECONDARY_NAMESPACE,
 };
-use crate::logger::{log_debug, log_error, log_info, LdkLogger};
+use crate::logger::{log_debug, log_error, log_info, log_trace, LdkLogger};
 
 use crate::runtime::Runtime;
 
@@ -458,6 +458,7 @@ where
 	runtime: Arc<Runtime>,
 	logger: L,
 	config: Arc<Config>,
+	static_invoice_store: Option<StaticInvoiceStore>,
 }
 
 impl<L: Deref + Clone + Sync + Send + 'static> EventHandler<L>
@@ -470,8 +471,9 @@ where
 		channel_manager: Arc<ChannelManager>, connection_manager: Arc<ConnectionManager<L>>,
 		output_sweeper: Arc<Sweeper>, network_graph: Arc<Graph>,
 		liquidity_source: Option<Arc<LiquiditySource<Arc<Logger>>>>,
-		payment_store: Arc<PaymentStore>, peer_store: Arc<PeerStore<L>>, runtime: Arc<Runtime>,
-		logger: L, config: Arc<Config>,
+		payment_store: Arc<PaymentStore>, peer_store: Arc<PeerStore<L>>,
+		static_invoice_store: Option<StaticInvoiceStore>, runtime: Arc<Runtime>, logger: L,
+		config: Arc<Config>,
 	) -> Self {
 		Self {
 			event_queue,
@@ -487,6 +489,7 @@ where
 			logger,
 			runtime,
 			config,
+			static_invoice_store,
 		}
 	}
 
@@ -1494,11 +1497,55 @@ where
 			LdkEvent::OnionMessagePeerConnected { .. } => {
 				debug_assert!(false, "We currently don't support onion message interception, so this event should never be emitted.");
 			},
-			LdkEvent::PersistStaticInvoice { .. } => {
-				debug_assert!(false, "We currently don't support static invoice persistence, so this event should never be emitted.");
+
+			LdkEvent::PersistStaticInvoice {
+				invoice,
+				invoice_slot,
+				recipient_id,
+				invoice_persisted_path,
+			} => {
+				if let Some(store) = self.static_invoice_store.as_ref() {
+					match store
+						.handle_persist_static_invoice(invoice, invoice_slot, recipient_id)
+						.await
+					{
+						Ok(_) => {
+							self.channel_manager.static_invoice_persisted(invoice_persisted_path);
+						},
+						Err(e) => {
+							log_error!(self.logger, "Failed to persist static invoice: {}", e);
+							return Err(ReplayEvent());
+						},
+					};
+				}
 			},
-			LdkEvent::StaticInvoiceRequested { .. } => {
-				debug_assert!(false, "We currently don't support static invoice persistence, so this event should never be emitted.");
+			LdkEvent::StaticInvoiceRequested { recipient_id, invoice_slot, reply_path } => {
+				if let Some(store) = self.static_invoice_store.as_ref() {
+					let invoice =
+						store.handle_static_invoice_requested(&recipient_id, invoice_slot).await;
+
+					match invoice {
+						Ok(Some(invoice)) => {
+							if let Err(e) =
+								self.channel_manager.send_static_invoice(invoice, reply_path)
+							{
+								log_error!(self.logger, "Failed to send static invoice: {:?}", e);
+							}
+						},
+						Ok(None) => {
+							log_trace!(
+								self.logger,
+								"No static invoice found for recipient {} and slot {}",
+								hex_utils::to_string(&recipient_id),
+								invoice_slot
+							);
+						},
+						Err(e) => {
+							log_error!(self.logger, "Failed to retrieve static invoice: {}", e);
+							return Err(ReplayEvent());
+						},
+					}
+				}
 			},
 			LdkEvent::FundingTransactionReadyForSigning { .. } => {
 				debug_assert!(false, "We currently don't support interactive-tx, so this event should never be emitted.");
--- a/src/io/mod.rs
+++ b/src/io/mod.rs
@@ -73,3 +73,8 @@ pub(crate) const BDK_WALLET_TX_GRAPH_KEY: &str = "tx_graph";
 pub(crate) const BDK_WALLET_INDEXER_PRIMARY_NAMESPACE: &str = "bdk_wallet";
 pub(crate) const BDK_WALLET_INDEXER_SECONDARY_NAMESPACE: &str = "";
 pub(crate) const BDK_WALLET_INDEXER_KEY: &str = "indexer";
+
+/// [`StaticInvoice`]s will be persisted under this key.
+///
+/// [`StaticInvoice`]: lightning::offers::static_invoice::StaticInvoice
+pub(crate) const STATIC_INVOICE_STORE_PRIMARY_NAMESPACE: &str = "static_invoices";
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -136,6 +136,7 @@ use gossip::GossipSource;
 use graph::NetworkGraph;
 use io::utils::write_node_metrics;
 use liquidity::{LSPS1Liquidity, LiquiditySource};
+use payment::asynchronous::static_invoice_store::StaticInvoiceStore;
 use payment::{
 	Bolt11Payment, Bolt12Payment, OnchainPayment, PaymentDetails, SpontaneousPayment,
 	UnifiedQrPayment,
@@ -498,6 +499,12 @@ impl Node {
 			Arc::clone(&self.logger),
 		));
 
+		let static_invoice_store = if self.config.async_payment_services_enabled {
+			Some(StaticInvoiceStore::new(Arc::clone(&self.kv_store)))
+		} else {
+			None
+		};
+
 		let event_handler = Arc::new(EventHandler::new(
 			Arc::clone(&self.event_queue),
 			Arc::clone(&self.wallet),
@@ -509,6 +516,7 @@ impl Node {
 			self.liquidity_source.clone(),
 			Arc::clone(&self.payment_store),
 			Arc::clone(&self.peer_store),
+			static_invoice_store,
 			Arc::clone(&self.runtime),
 			Arc::clone(&self.logger),
 			Arc::clone(&self.config),
@@ -818,6 +826,7 @@ impl Node {
 		Bolt12Payment::new(
 			Arc::clone(&self.channel_manager),
 			Arc::clone(&self.payment_store),
+			Arc::clone(&self.config),
 			Arc::clone(&self.is_running),
 			Arc::clone(&self.logger),
 		)
@@ -831,6 +840,7 @@ impl Node {
 		Arc::new(Bolt12Payment::new(
 			Arc::clone(&self.channel_manager),
 			Arc::clone(&self.payment_store),
+			Arc::clone(&self.config),
 			Arc::clone(&self.is_running),
 			Arc::clone(&self.logger),
 		))
--- a/src/payment/asynchronous/mod.rs
+++ b/src/payment/asynchronous/mod.rs
@@ -1,0 +1,9 @@
+// This file is Copyright its original authors, visible in version control history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
+// accordance with one or both of these licenses.
+
+mod rate_limiter;
+pub(crate) mod static_invoice_store;
--- a/src/payment/asynchronous/static_invoice_store.rs
+++ b/src/payment/asynchronous/static_invoice_store.rs
@@ -1,0 +1,277 @@
+// This file is Copyright its original authors, visible in version control history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
+// accordance with one or both of these licenses.
+
+//! Store implementation for [`StaticInvoice`]s.
+
+use crate::hex_utils;
+use crate::io::STATIC_INVOICE_STORE_PRIMARY_NAMESPACE;
+use crate::payment::asynchronous::rate_limiter::RateLimiter;
+use crate::types::DynStore;
+
+use bitcoin::hashes::sha256::Hash as Sha256;
+use bitcoin::hashes::Hash;
+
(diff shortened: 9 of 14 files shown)


src/config.rs
```diff
@@ -179,6 +180,10 @@ pub struct Config {
 	/// **Note:** If unset, default parameters will be used, and you will be able to override the
 	/// parameters on a per-payment basis in the corresponding method calls.
 	pub route_parameters: Option<RouteParametersConfig>,
+
+	/// The [`BlindedMessagePath`]s that we will use as an async recipient to interactively build [`Offer`]s with a
+	/// static invoice server, so the server can serve [`StaticInvoice`]s to payers on our behalf when we're offline.
+	pub static_invoice_server_paths: Option<Vec<BlindedMessagePath>>,
```
tnull: Hmm, assuming this might not end up being the only config option regarding async payments, we should probably move this to an `AsyncPaymentConfig` sub-config, similar to what we do for anchor channels, for example.

Also note that we'll need to expose all of these values in Uniffi, too, so we'll need to see whether `BlindedMessagePath` is easy enough to expose, or if we need a simpler local type or bindings-compatible type wrapper.

src/builder.rs
```diff
@@ -544,6 +550,15 @@ impl NodeBuilder {
 		Ok(self)
 	}
 
+	/// Sets the [`BlindedMessagePath`]s that we will use as an async recipient to interactively build [`Offer`]s with a
+	/// static invoice server, so the server can serve [`StaticInvoice`]s to payers on our behalf when we're offline.
+	pub fn set_paths_to_static_invoice_server(
```
tnull: Note this will also need to be added on the `uniffi` variant of the builder.

src/lib.rs
```diff
@@ -1575,6 +1576,15 @@ impl Node {
 				Error::PersistenceFailed
 			})
 	}
+
+	/// Retrieve an [`Offer`] for receiving async payments as an often-offline recipient. Will only return an offer if
+	/// [`Builder::set_paths_to_static_invoice_server`] was called and we succeeded in interactively building a
+	/// [`StaticInvoice`] with the static invoice server.
+	///
+	/// Useful for posting offers to receive payments later, such as posting an offer on a website.
+	pub fn get_async_receive_offer(&self) -> Result<Offer, ()> {
```
tnull: We def. wouldn't want to add this here. Rather, it either should be part of the `Bolt12Payment` API, or even would need to be a separate, parallel `AsyncPayment` API module.

src/config.rs
joostjager: Can we not expose this to Uniffi initially? To keep scope down.

tnull: > Can we not expose this to Uniffi initially? To keep scope down.

No, especially since this is meant to be client-side API and it's part of the `Config` now, we need to expose it in Uniffi.

joostjager: Can't we conditionally disable that? I mean for MVP we don't need to go full out on bindings.

tnull: Not sure what you mean by "full out on", given it's not that much work? 

We generally aim to keep up feature parity for bindings use, also as a) it otherwise will get hard to follow which feature is available where and b) people tend to never get back to such things once PRs are merged.

joostjager: My thinking is that extra work (mapping / simplifying types) specifically for binding may not be worth it for the MVP.

src/rate_limiter.rs
```diff
@@ -0,0 +1,34 @@
+use std::collections::HashMap;
+use std::time::{Duration, Instant};
+
+pub(crate) struct RateLimiter {
+	users: HashMap<Vec<u8>, Instant>,
+	min_interval: Duration,
+}
+
+impl RateLimiter {
```
joostjager: Keeping it simple here

src/static_invoice_store.rs
```diff
@@ -0,0 +1,71 @@
...
+use bitcoin::hashes::{sha256, Hash};
+use lightning::{offers::static_invoice::StaticInvoice, util::ser::Writeable};
+use std::{
+	sync::{Arc, Mutex},
+	time::Duration,
+};
+
+pub(crate) struct StaticInvoiceStore {
+	kv_store: Arc<DynStore>,
+	rate_limiter: Mutex<RateLimiter>,
```
joostjager: Needed to add `Mutex` here even though I don't think it is used concurrently. Otherwise `handle_event` needed to be made mutable, the Arc higher up no longer working, etc.

src/static_invoice_store.rs
```diff
@@ -0,0 +1,78 @@
...
+	) -> Result<(), lightning::io::Error> {
+		let mut limiter = limiter.lock().unwrap();
+		if !limiter.allow(recipient_id) {
+			Err(lightning::io::Error::new(lightning::io::ErrorKind::Other, "Rate limit exceeded"))
+		} else {
+			Ok(())
+		}
+	}
+
+	pub(crate) async fn handle_static_invoice_requested(
```
joostjager: Async, prepare for async kv store.

src/lib.rs
joostjager: Moved to bolt12payment api

src/error.rs
```diff
@@ -120,6 +120,10 @@ pub enum Error {
 	LiquiditySourceUnavailable,
 	/// The given operation failed due to the LSP's required opening fee being too high.
 	LiquidityFeeTooHigh,
+	/// The given blinded paths are invalid.
+	InvalidBlindedPaths,
+	/// The requested operation failed.
+	OperationFailed,
```
joostjager: It seemed to me that some internal error-like value is useful. `AsyncPaymentOperationFailed` wouldn't add much, because the caller already knows they are calling an async payments method.

joostjager:
Made rate limiter a bit less simplistic by adding a leaky bucket. Otherwise the second call of two consecutive calls immediately fails.

src/error.rs
tnull: Hmm, I'd prefer to add such a non-descript error type. Can we express better *what* failed in the name?

src/lib.rs
```diff
@@ -170,6 +175,8 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex, RwLock};
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
+use crate::static_invoice_store::StaticInvoiceStore;
```
tnull: nit: Please move this up to the other `crate` imports (and drop `crate`).

src/lib.rs
```diff
@@ -1476,6 +1486,39 @@ impl Node {
 				Error::PersistenceFailed
 			})
 	}
+
+	/// Sets the [`BlindedMessagePath`]s that we will use as an async recipient to interactively build [`Offer`]s with a
+	/// static invoice server, so the server can serve [`StaticInvoice`]s to payers on our behalf when we're offline.
+	///
+	/// [`Offer`]: lightning::offers::offer::Offer
+	/// [`StaticInvoice`]: lightning::offers::static_invoice::StaticInvoice
+	pub fn set_paths_to_static_invoice_server(&self, paths: Vec<u8>) -> Result<(), Error> {
```
tnull: For consistency, these setters/getters should like be configured on `Builder`, not on the `Node` object.

This also means that you'll need to add an internal `AsyncPaymentConfig` (or similar) object to store the configured values until they're applied on `build`.

src/lib.rs
```diff
@@ -1476,6 +1486,39 @@ impl Node {
...
+	}
+
+	/// [`BlindedMessagePath`]s for an async recipient to communicate with this node and interactively
+	/// build [`Offer`]s and [`StaticInvoice`]s for receiving async payments.
+	///
+	/// [`Offer`]: lightning::offers::offer::Offer
+	/// [`StaticInvoice`]: lightning::offers::static_invoice::StaticInvoice
+	pub fn blinded_paths_for_async_recipient(
+		&self, recipient_id: Vec<u8>,
+	) -> Result<Vec<u8>, Error> {
```
tnull: In the Rust API, we def. still want to return the actual `BlindedMessagePaths`, just for the `uniffi` interface it's okay to fallback to a `bytes` for now.

bindings/ldk_node.udl
```diff
@@ -157,6 +157,10 @@ interface Node {
 	boolean verify_signature([ByRef]sequence<u8> msg, [ByRef]string sig, [ByRef]PublicKey pkey);
 	[Throws=NodeError]
 	bytes export_pathfinding_scores();
+	[Throws=NodeError]
+	void set_paths_to_static_invoice_server(sequence<u8> paths);
+	[Throws=NodeError]
+	sequence<u8> blinded_paths_for_async_recipient(sequence<u8> recipient_id);
```
tnull: Let's use `bytes` rather than `sequence<u8>` (cf. https://mozilla.github.io/uniffi-rs/latest/types/builtin_types.html)

src/event.rs
```diff
@@ -1494,11 +1498,45 @@ where
...
+					.static_invoice_store
+					.handle_static_invoice_requested(recipient_id, invoice_slot)
+					.await;
+
+				match invoice {
+					Ok(Some(invoice)) => {
+						if let Err(_) =
+							self.channel_manager.send_static_invoice(invoice, reply_path)
+						{
+							log_error!(self.logger, "Failed to send static invoice");
```
tnull: Can we log the error here?

src/payment/bolt12.rs
```diff
@@ -450,4 +450,19 @@ impl Bolt12Payment {
 
 		Ok(maybe_wrap(refund))
 	}
+
+	/// Retrieve an [`Offer`] for receiving async payments as an often-offline recipient. Will only return an offer if
```
tnull: nit: As a rule of thumb, please give the initial sentence its own paragraph and start the second sentence (here and elsewhere) in a second paragraph. This reads better *and* displays better in rendered list views.

src/payment/bolt12.rs
```diff
@@ -450,4 +450,19 @@ impl Bolt12Payment {
...
+	/// [`Node::set_paths_to_static_invoice_server`] was called and we succeeded in interactively building a
+	/// [`StaticInvoice`] with the static invoice server.
+	///
+	/// Useful for posting offers to receive payments later, such as posting an offer on a website.
+	///
+	/// [`Node::set_paths_to_static_invoice_server`]: crate::Node::set_paths_to_static_invoice_server
+	/// [`StaticInvoice`]: lightning::offers::static_invoice::StaticInvoice
+	pub fn get_async_receive_offer(&self) -> Result<Offer, Error> {
```
tnull: Let's try to align this better with the Rust naming scheme and/or the other methods here.

Maybe simply `pub fn receive_async()`?

src/rate_limiter.rs
tnull, line 1: This seems like it will just be used for the static invoice store. Let's move this to `crate::payments::async::utils` or `crate::payments::async::rate_limiter` or similar.

src/static_invoice_store.rs
tnull, line 1: Especially given that we also have the `rate_limiter.rs`, let's move this to `crate::payments::async::store`

src/static_invoice_store.rs
```diff
@@ -0,0 +1,90 @@
+use crate::{hex_utils, rate_limiter::RateLimiter, types::DynStore};
+use bitcoin::hashes::{sha256, Hash};
```
tnull: nit: Let's import `sha256::Hash as Sha256` to keep it consistent with the rest of the codebase.

src/static_invoice_store.rs
```diff
@@ -0,0 +1,90 @@
+use crate::{hex_utils, rate_limiter::RateLimiter, types::DynStore};
```
tnull: nit: Please give these imports their own lines (only group at the highest/module level), same below and elsewhere.

src/static_invoice_store.rs
```diff
@@ -0,0 +1,90 @@
...
+	request_rate_limiter: Mutex<RateLimiter>,
+	persist_rate_limiter: Mutex<RateLimiter>,
+}
+
+impl StaticInvoiceStore {
+	// Static invoices are stored at "static_invoices/<sha256(recipient_id)>/<invoice_slot>".
+	//
+	// Example:
+	// static_invoices/039058c6f2c0cb492c533b0a4d14ef77cc0f78abccced5287d84a1a2011cfb81/001f
+	const PRIMARY_NAMESPACE: &str = "static_invoices";
```
tnull: Please move the `NAMESPACE` constant to `src/io/mod.rs` and follow the pattern there.

src/static_invoice_store.rs
```diff
@@ -0,0 +1,90 @@
...
+		invoice.write(&mut buf)?;
+
+		self.kv_store.write(Self::PRIMARY_NAMESPACE, &secondary_namespace, &key, buf)
+	}
+
+	fn get_storage_location(invoice_slot: u16, recipient_id: Vec<u8>) -> (String, String) {
+		let hash = sha256::Hash::hash(&recipient_id).to_byte_array();
+		let secondary_namespace = hex_utils::to_string(&hash);
+
+		let key = hex_utils::to_string(&invoice_slot.to_be_bytes());
```
tnull: I wonder if it would make sense to just write the (base-10) integer for `invoice_slot`?

joostjager: I chose the hex one because it is more compact. Do you suggest base-10 for easier debugging?

src/config.rs
joostjager: Bindings added

src/static_invoice_store.rs
tnull: > I chose the hex one because it is more compact. Do you suggest base-10 for easier debugging?

Yeah, especially in `FilesystemStore`, but also when persisted otherwise the integer could be more human readable and would match the actual invoice_slot you'd see in the logs.

src/error.rs
joostjager: Multiple possibilities:
* Serialization of the paths failed
* Recipient id too long
* No blinded path possible

Let me know which ones you'd want to bring out. Distinguishing between the nr 2 and 3 requires a change in LDK.

src/lib.rs
joostjager: I am not sure if anyone ever looks at these imports, but sorted does look cleaner. Done.

src/error.rs
tnull: Hmm, could we use the `InvalidBlindedPaths` as an umbrella for now?

src/lib.rs
joostjager: I purposefully didn't put them on the builder, because the paths can change. On channel manager, it also isn't a constructor-style parameter.

Note that `set_paths_to_static_invoice_server` and `blinded_paths_for_async_recipient` aren't used on the same node. And for `blinded_paths_for_async_recipient`, the node must be running already to construct the blinded paths.

src/lib.rs
joostjager: Why would it be required to know the details of the blinded paths for the Rust API and not for the other APIs? Not sure which use that unlocks.

src/payment/bolt12.rs
joostjager: Yes, I agree. Fixed.

src/payment/bolt12.rs
joostjager: `receive_async` could suggest that it is an async. executed method, but maybe that's just my own bias. I agree that it matches the existing `receive` method better. It is inconsistent with the naming in channel manager. Made the change.

src/rate_limiter.rs
joostjager: Moved this into `payment`. A sub-folder `async` feels a bit too much to me. Not that many files in `payment`?

src/static_invoice_store.rs
joostjager: Moved to (just) `payment`

src/static_invoice_store.rs
joostjager: Converted to zero-padded decimal.

src/error.rs
joostjager: Ok, done

src/lib.rs
tnull: I do. :)

src/lib.rs
tnull: > I purposefully didn't put them on the builder, because the paths can change. On channel manager, it also isn't a constructor-style parameter.

Hmm, but having these setters/getters here seems kind of arbitrary. Maybe we should at least move them to the `Bolt12Payment` proxy object then? Also, as noted elsewhere, let's take the `BlindedMessagePath`s directly in the Rust API, and just use `Vec<u8>` in bindings.

src/lib.rs
tnull: > Why would it be required to know the details of the blinded paths for the Rust API and not for the other APIs? Not sure which use that unlocks.

Users might want to expect what they're paying to. It's fine to take a shortcut here for bindings, but please use the full Rust type in the Rust API.

src/rate_limiter.rs
tnull: I mean, I'd prefer to have it in a sub-module where it's clear what files belong together. But I won't block the PR on it, can also clean it up myself afterwards.

src/static_invoice_store.rs
tnull: > Moved to (just) `payment`

See above, would still prefer a submodule.

src/static_invoice_store.rs
tnull: Fine by me.

src/lib.rs
joostjager: This is not about what they are paying to. It's the path that the receiver can use to contact the static invoice server. Because it is blinded, they still don't know where that server is potentially?

src/rate_limiter.rs
joostjager: I don't think it is necessary and also wouldn't want to call it 'not clean'. To me these structuring decisions are highly subjective. Someone else might think that spontaneous payments also need their own folder for example.

But I don't mind creating the extra level though.

src/lib.rs
joostjager: Done

src/lib.rs
tnull: We try to use the proper Rust types where possible, and I really don't see why we shouldn't in this case. In case of `set_paths_to_static_invoice_server` users could theoretically handcraft a blinded path somehow if their service implementation doesn't exactly meet our serialization and try to use it. Here we'd just match that API type.

src/rate_limiter.rs
tnull: Alright, feel free to leave as is for now.

joostjager: `async` is a keyword, so made it `asynchronous`

src/lib.rs
joostjager: I think consistency between languages is more important. Especially if the handcrafted blinded path is speculative. But no blocker, I've updated the code with distinct method signatures.

src/event.rs
```diff
@@ -5,8 +5,8 @@
 // http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
 // accordance with one or both of these licenses.
 
+use crate::payment::asynchronous::static_invoice_store::StaticInvoiceStore;
```
tnull: nit: Please move this down to the other `crate::payment` imports.

src/event.rs
```diff
@@ -1494,11 +1497,46 @@ where
...
+				invoice_persisted_path,
+			} => {
+				if let Some(store) = self.static_invoice_store.as_ref() {
+					match store
+						.handle_persist_static_invoice(invoice, invoice_slot, recipient_id)
+						.await
+					{
+						Ok(_) => {},
+						Err(e) => {
+							log_error!(self.logger, "Failed to persist static invoice: {}", e);
```
tnull, line 1516: I think we need to return `Err(ReplayEvent())` here and below in case of persistence failure to be able to retry?

src/io/mod.rs
```diff
@@ -73,3 +73,8 @@ pub(crate) const BDK_WALLET_TX_GRAPH_KEY: &str = "tx_graph";
 pub(crate) const BDK_WALLET_INDEXER_PRIMARY_NAMESPACE: &str = "bdk_wallet";
 pub(crate) const BDK_WALLET_INDEXER_SECONDARY_NAMESPACE: &str = "";
 pub(crate) const BDK_WALLET_INDEXER_KEY: &str = "indexer";
+
+// Static invoices will be persisted at "static_invoices/<sha256(recipient_id)>/<invoice_slot>".
+//
+// Example: static_invoices/039058c6f2c0cb492c533b0a4d14ef77cc0f78abccced5287d84a1a2011cfb81/001f
```
tnull: This is out-of-date now that we don't serialize `invoice_slot` as hex?

src/io/mod.rs
```diff
@@ -73,3 +73,8 @@ pub(crate) const BDK_WALLET_TX_GRAPH_KEY: &str = "tx_graph";
...
+pub(crate) const STATIC_INVOICES_PRIMARY_NAMESPACE: &str = "static_invoices";
```
tnull:
```suggestion
pub(crate) const STATIC_INVOICE_STORE_PRIMARY_NAMESPACE: &str = "static_invoices";
```

src/io/mod.rs
tnull: Also, let's move this example to the static invoice store, as this is just the primary namespace const. We *should* however cross-link to the the respective LDK docs here so that readers know what 'static invoices' are in the first place.

src/payment/asynchronous/mod.rs
tnull, line 1: This and the other new files need a module description and a copyright header.

src/payment/asynchronous/static_invoice_store.rs
```diff
@@ -0,0 +1,265 @@
...
+use bitcoin::hashes::Hash;
+
+use lightning::{offers::static_invoice::StaticInvoice, util::ser::Writeable};
+
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+
+pub(crate) struct StaticInvoiceStore {
+	kv_store: Arc<DynStore>,
+	request_rate_limiter: Mutex<RateLimiter>,
```
tnull, line 25: Do we really need two independent rate limiters here or would a single one not suffice?

src/payment/asynchronous/static_invoice_store.rs
```diff
@@ -0,0 +1,265 @@
...
+		let key = format!("{:05}", invoice_slot);
+		(secondary_namespace, key)
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use std::{sync::Arc, time::Duration};
+
+	use bitcoin::{
```
tnull, line 118: nit: As mentioned elsewhere previously, module-level grouping of imports would be preferred.

I'm starting to wonder if we should go 'bleeding edge' here and introduce a `nightly` rustfmt CI job, which would indeed support dealing with the import groups for us:

```
group_imports = "StdExternalCrate"
reorder_imports = true
imports_granularity = "Module"
```

src/payment/bolt12.rs
```diff
@@ -54,15 +57,16 @@ pub struct Bolt12Payment {
...
 	is_running: Arc<RwLock<bool>>,
+	async_payment_services_enabled: bool,
 	logger: Arc<Logger>,
 }
 
 impl Bolt12Payment {
 	pub(crate) fn new(
 		channel_manager: Arc<ChannelManager>, payment_store: Arc<PaymentStore>,
-		is_running: Arc<RwLock<bool>>, logger: Arc<Logger>,
+		async_payment_services_enabled: bool, is_running: Arc<RwLock<bool>>, logger: Arc<Logger>,
```
tnull: Hmm, let's just forward the full `Config` as we might need it for other things in the future, too?

src/payment/bolt12.rs
```diff
@@ -450,4 +454,89 @@ impl Bolt12Payment {
 
 		Ok(maybe_wrap(refund))
 	}
+
+	/// Retrieve an [`Offer`] for receiving async payments as an often-offline recipient.
+	///
+	/// Will only return an offer if [`Bolt12Payment::set_paths_to_static_invoice_server`] was called and we succeeded
+	/// in interactively building a [`StaticInvoice`] with the static invoice server.
+	///
+	/// Useful for posting offers to receive payments later, such as posting an offer on a website.
```
tnull, line 463: Mind adding an 'experimental feature' note here and on the other main APIs?

src/payment/asynchronous/static_invoice_store.rs
```diff
@@ -0,0 +1,265 @@
...
+					if e.kind() == lightning::io::ErrorKind::NotFound {
+						Ok(None)
+					} else {
+						Err(e)
+					}
+				},
+			)
+	}
+
+	pub(crate) async fn handle_persist_static_invoice(
```
tnull, line 89: Do we need to check whether we actually know/have a channel open with the recipient? Or is all that fully handled on the LDK side already?

src/payment/asynchronous/static_invoice_store.rs
joostjager: I thought that it wouldn't be nice if a recipient can't update their invoice because someone else is keeping the limited active with requests.

src/payment/asynchronous/static_invoice_store.rs
joostjager: That would be great. FIxing import groups isn't something that humans should be required to do, especially because some people don't even care for them.

src/payment/bolt12.rs
joostjager: I can do that, but in what regard is it experimental? Added

src/event.rs
joostjager: Good point. Also `static_invoice_persisted` shouldn't be called in that case. Fixed here and below.

src/io/mod.rs
joostjager: Decimal invoice slot, example moved, LDK link added.

src/io/mod.rs
joostjager: Not sure if I like this better. It's not the store that is stored, but the invoices. I don't really mind either. Changed.

src/payment/asynchronous/mod.rs
joostjager: Added

src/payment/asynchronous/static_invoice_store.rs
joostjager: Discussed offline. At this point, I think it is part of the integration of ldk-node into the wider system to ensure that `blinded_paths_for_async_recipient` isn't called without restrictions.

src/payment/bolt12.rs
joostjager: I didn't want to add a dependency to the full object if a bool suffices. Also makes it more clear what part of config matters. Maybe we don't need things in the future. But no strong objection, so changed.

src/event.rs
```diff
@@ -1494,11 +1497,48 @@ where
...
+
+					match invoice {
+						Ok(Some(invoice)) => {
+							if let Err(e) =
+								self.channel_manager.send_static_invoice(invoice, reply_path)
+							{
+								log_error!(self.logger, "Failed to send static invoice: {:?}", e);
+							}
+						},
+						Ok(None) => {},
```
tnull: Should we log here, or do we think this would be very spammy? Also, does this point to a gap in the protocol that we don't seem to have a way to communicate to the client that a lookup failed?

joostjager: I think there's already more per-onion message trace logging, so spamming the logs is possible already. I've added a trace statement. At this stage it may be helpful in debugging production issues.

You could consider the lack of a failure onion message a gap in the protocol. Onion messages are problematic anyway because they may be dropped without notice.

A similar gap exists for the `held_htlc_available` message. The sender won't get an ack indicating whether that message was processed. If that one gets dropped, the sender lsp will hold the htlc waiting for a release that will never come.

PR #626, opened 2025-08-21
chore: update rusqlite to 0.37.0

dpc:
Update rusqlite.


dpc:
No idea if there are any MSRV, or other considerations. Just wanted to see if it will work and how hard it will be.

PR #627, opened 2025-08-22
Upgrade VSS encryption and obfuscation scheme

tnull:
Based on https://github.com/lightningdevkit/vss-rust-client/pull/40

In this PR we account for `vss-client` being renamed to `vss-client-ng`, upgrade to the latest v0.4.0 release, as well as  fix minor issues with the data encryption and key obfuscation scheme currently employed by `VssStore`. 

As these are breaking changes, we'll also include a migration procedure as part of this PR.

### DIFF
--- a/tests/integration_tests_vss.rs
+++ b/tests/integration_tests_vss.rs
@@ -130,3 +130,69 @@ async fn vss_v0_schema_backwards_compatibility() {
 
 	node_new.stop().unwrap();
 }
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn vss_node_restart() {
+	let (bitcoind, electrsd) = common::setup_bitcoind_and_electrsd();
+	let esplora_url = format!("http://{}", electrsd.esplora_url.as_ref().unwrap());
+	let vss_base_url = std::env::var("TEST_VSS_BASE_URL").unwrap();
+
+	let rand_suffix: String =
+		(0..7).map(|_| rng().sample(rand::distr::Alphanumeric) as char).collect();
+	let store_id = format!("restart_test_{}", rand_suffix);
+	let storage_path = common::random_storage_path().to_str().unwrap().to_owned();
+	let seed_bytes = [42u8; 64];
+
+	// Setup initial node and fund it.
+	let (expected_balance_sats, expected_node_id) = {
+		let mut builder = Builder::new();
+		builder.set_network(bitcoin::Network::Regtest);
+		builder.set_storage_dir_path(storage_path.clone());
+		builder.set_entropy_seed_bytes(seed_bytes);
+		builder.set_chain_source_esplora(esplora_url.clone(), None);
+		let node = builder
+			.build_with_vss_store_and_fixed_headers(
+				vss_base_url.clone(),
+				store_id.clone(),
+				HashMap::new(),
+			)
+			.unwrap();
+
+		node.start().unwrap();
+		let addr = node.onchain_payment().new_address().unwrap();
+		common::premine_and_distribute_funds(
+			&bitcoind.client,
+			&electrsd.client,
+			vec![addr],
+			bitcoin::Amount::from_sat(100_000),
+		)
+		.await;
+		node.sync_wallets().unwrap();
+
+		let balance = node.list_balances().spendable_onchain_balance_sats;
+		assert!(balance > 0);
+		let node_id = node.node_id();
+
+		node.stop().unwrap();
+		(balance, node_id)
+	};
+
+	// Verify node can be restarted from VSS backend.
+	let mut builder = Builder::new();
+	builder.set_network(bitcoin::Network::Regtest);
+	builder.set_storage_dir_path(storage_path);
+	builder.set_entropy_seed_bytes(seed_bytes);
+	builder.set_chain_source_esplora(esplora_url, None);
+
+	let node = builder
+		.build_with_vss_store_and_fixed_headers(vss_base_url, store_id, HashMap::new())
+		.unwrap();
+
+	node.start().unwrap();
+	node.sync_wallets().unwrap();
+
+	assert_eq!(expected_node_id, node.node_id());
+	assert_eq!(expected_balance_sats, node.list_balances().spendable_onchain_balance_sats);
+
+	node.stop().unwrap();
+}


src/io/vss_store.rs
```diff
@@ -321,9 +325,10 @@ impl Drop for VssStore {
 }
 
 struct VssStoreInner {
+	schema_version: VssSchemaVersion,
```
jkczyz, line 354: Don't have a strong opinion, but would it be worth using a trait to avoid all the conditional logic?

src/io/vss_store.rs
```diff
@@ -105,6 +111,21 @@ impl VssStore {
 			header_provider.clone(),
 		);
 
+		let runtime_handle = internal_runtime.handle();
+		let schema_store_id = store_id.clone();
+		let schema_key_obfuscator = KeyObfuscator::new(obfuscation_master_key);
+		let (schema_version, blocking_client) = tokio::task::block_in_place(move || {
```
jkczyz: Ah, I guess we can't if we are determining at runtime. Could maybe still have separate structs V0 and V1 structs with static methods to keep the call sites cleaner.

Cargo.toml
```diff
@@ -65,7 +65,8 @@ serde = { version = "1.0.210", default-features = false, features = ["std", "der
 serde_json = { version = "1.0.128", default-features = false, features = ["std"] }
 log = { version = "0.4.22", default-features = false, features = ["std"]}
 
-vss-client = "0.3"
+#vss-client = "0.3"
+vss-client = { git = "https://github.com/tnull/vss-rust-client", rev = "03ca9d99f70387aabec225020e46434cda8d18ff" }
```
tankyleo: nit: any chance we can point to the new release directly instead of modifying this twice IIRC ?

tankyleo: can we skip this and go straight to `vss-client-ng` ?

src/io/vss_store.rs
```diff
@@ -365,17 +370,45 @@ impl VssStoreInner {
 	fn build_obfuscated_key(
 		&self, primary_namespace: &str, secondary_namespace: &str, key: &str,
 	) -> String {
-		let obfuscated_key = self.key_obfuscator.obfuscate(key);
-		if primary_namespace.is_empty() {
-			obfuscated_key
+		if self.schema_version == VssSchemaVersion::V1 {
+			let obfuscated_prefix =
+				self.build_obfuscated_prefix(primary_namespace, secondary_namespace);
```
tankyleo: your call: in this function we already branch on the schema version to determine the prefix, so could help us clean things up here ?

src/io/vss_store.rs
```diff
@@ -44,6 +44,13 @@ type CustomRetryPolicy = FilteredRetryPolicy<
 	Box<dyn Fn(&VssError) -> bool + 'static + Send + Sync>,
 >;
 
+enum VssSchemaVersion {
```
tankyleo, line 50: nit: typos in the commit message

Cargo.toml
tankyleo: can we go straight to `vss-client-ng` here ?

src/io/vss_store.rs
tankyleo, line 398: nit: this also branches on the version here, your call if you want to do something about this :)

src/io/vss_store.rs
```diff
@@ -170,13 +168,14 @@ impl KVStoreSync for VssStore {
...
 		// TODO: We could drop the timeout here once we ensured vss-client's Retry logic always
 		// times out.
-		let spawned_fut = internal_runtime.spawn(async move {
-			tokio::time::timeout(VSS_IO_TIMEOUT, fut).await.map_err(|_| {
-				let msg = "VssStore::read timed out";
-				Error::new(ErrorKind::Other, msg)
-			})
-		});
-		self.runtime.block_on(spawned_fut).expect("We should always finish")?
+		tokio::task::block_in_place(move || {
```
tankyleo: why do we need the `block_in_place` here ?

src/io/vss_store.rs
```diff
@@ -98,9 +98,9 @@ impl VssStore {
 			derive_data_encryption_and_obfuscation_keys(&vss_seed);
 		let key_obfuscator = KeyObfuscator::new(obfuscation_master_key);
 		let retry_policy = ExponentialBackoffRetryPolicy::new(Duration::from_millis(10))
-			.with_max_attempts(10)
-			.with_max_total_delay(Duration::from_secs(15))
-			.with_max_jitter(Duration::from_millis(10))
+			.with_max_attempts(100)
+			.with_max_total_delay(Duration::from_secs(180))
+			.with_max_jitter(Duration::from_millis(100))
```
tankyleo: would be cool to get the motivation for these, is CI more stable with these much bigger values ?

Cargo.toml
```diff
@@ -65,8 +65,8 @@ serde = { version = "1.0.210", default-features = false, features = ["std", "der
 serde_json = { version = "1.0.128", default-features = false, features = ["std"] }
 log = { version = "0.4.22", default-features = false, features = ["std"]}
 
-#vss-client = "0.3"
-vss-client = { git = "https://github.com/tnull/vss-rust-client", rev = "03ca9d99f70387aabec225020e46434cda8d18ff" }
+#vss-client-ng = "0.3"
+vss-client-ng = { git = "https://github.com/tnull/vss-client", rev = "98ac5e171ef1ab970bbc58cd995ffc1e615421d1" }
```
tankyleo: can we go straight to `0.4.0` here ?

src/io/vss_store.rs
```diff
@@ -317,7 +330,10 @@ impl Drop for VssStore {
 
 struct VssStoreInner {
 	schema_version: VssSchemaVersion,
-	client: VssClient<CustomRetryPolicy>,
+	blocking_client: VssClient<CustomRetryPolicy>,
+	// A secondary client that will only be used for async persistence via `KVStore`, to ensure TCP
+	// connections aren't shared between our outer and the internal runtime.
```
tankyleo, line 357: Trying to understand the motivation for this, someone could currently be using the sync and the async API at the same time?

src/io/vss_store.rs
```diff
@@ -653,6 +674,114 @@ fn retry_policy() -> CustomRetryPolicy {
 		}) as _)
 }
 
+// FIXME: This returns the used client as currently `VssClient`'s `RetryPolicy`s aren't `Clone`. So
+// we're forced to take the owned client and return it to be able to reuse the same connection
+// later.
+async fn determine_and_write_schema_version(
+	client: VssClient<CustomRetryPolicy>, store_id: String, data_encryption_key: [u8; 32],
```
tankyleo: Don't quite see the need for ownership here, I was able to get things to compile while taking a reference here and removing the `move` of the closure where this gets called.

src/io/vss_store.rs
tnull: Hmm, not sure if duplicating all the struct boilerplate and passing in all the fields currently living on `self` would be much cleaner?

src/io/vss_store.rs
tnull: Thanks, fixed.

Cargo.toml
tnull: Hmm, I was about to write 'No, as there are more breaking API changes that are addressed in the following commits', but then I realized by now it's only the crate renaming changes (previously it has also been the `Retry` changes). 

I therefore now added a fixup commit that will have us simply import `vss-client-ng` v0.4 as `vss-client` directly here, and dropped the renaming commit altogether to avoid the unnecessary churn.

Cargo.toml
tnull: See above, fixed above.

src/io/vss_store.rs
tnull: We basically always need to use `block_in_place` when we're about to block in a tokio runtime context (in this case by calling `block_on`), as it tells any potential outer runtime context here to move any other tasks that run on the same thread to a different thread. So much for the theory, in practice it's often not sufficient to avoid weird deadlocks etc, but that's a different story.

Here, maybe it's noteworthy that it only 'appears' that we added `block_in_place`, but we were also using it before, as we included it `crate::runtime::Runtime::block_on`. It's just because here we use a 'bare' tokio `Runtime` that we wrap `block_on` explictly.

src/io/vss_store.rs
tnull: The motivation is that we basically need to try as hard as we can to avoid hitting a persistence error and hence a panic, or maybe worse, a hanging node in an undefined state.

src/io/vss_store.rs
tnull: It's not even 'at the same time', but even in our use case we currently still use `KVStoreSync` to read ~all data on startup, then continue with the same `VssStore` but `KVStore` for some of the background processing, while other calls there are still going through `KVStoreSync`. As we only use the internal runtime for the `KVStoreSync` case, it could happen that we start a TCP connection on one runtime and then try to reuse it via the same `VssClient`/`reqwest::Client` on the other, which could lead to deadlocks as we could run into cases where one runtime is blocked waiting on the other, which in turn never gets woken as the TCP socket would wake up the other runtime when data is ready. 

TLDR: Over the last few weeks we were running into blocking issues where write tasks would simply never return, likely due deadlocking between tasks. One workaround for this was the bump to `reqwest` 0.12 in `VssClient`, the use of the total timeout value there which in combination with the `RetryPolicy` logic here seems to be robust, i.e., in case we'd hit such a blocking task we would just eventually abandon it and retry. The *other* mitigation strategy is introducing different clients to make sure such deadlock scenarios would be avoided. Hopefully both combined are sufficiently reliably for this release. After the release we intend to remove the reliance on `KVStoreSync` entirely, and hence also the internal runtime / split-client logic here.

src/io/vss_store.rs
tnull: Ah, good catch. Not sure why I was under the impression that we needed to `move` there. Added a fixup.

src/io/vss_store.rs
tankyleo: Thanks ok yea didn't catch that `block_in_place` was already getting called before.

src/io/vss_store.rs
```diff
@@ -44,11 +44,15 @@ type CustomRetryPolicy = FilteredRetryPolicy<
...
 >;
 
+#[derive(Debug, PartialEq)]
 enum VssSchemaVersion {
 	// The initial schema version.
 	// This used an empty `aad` and unobfuscated `primary_namespace`/`secondary_namespace`s in the
 	// stored key.
 	V0,
+	// The second deployed schema version.
+	// Here we started to obfuscate the primary and secondary namespaces and the obfuscated `store_key` (`obfuscate(primary_namespace#secondary_namespace)#obfuscate(key)`) is now used as `aad` for encryption, ensuring that the encrypted blobs commit to the key they're stored under.
```
tankyleo: nit: newlines needed here and in the commit message. Also the commit message has a clunky "we the" :)

src/io/vss_store.rs
```diff
@@ -365,17 +370,45 @@ impl VssStoreInner {
...
+			format!("{}#{}", primary_namespace, secondary_namespace)
 		}
 	}
 
 	fn extract_key(&self, unified_key: &str) -> io::Result<String> {
-		let mut parts = unified_key.splitn(3, '#');
-		let (_primary_namespace, _secondary_namespace) = (parts.next(), parts.next());
+		let mut parts = if self.schema_version == VssSchemaVersion::V1 {
+			let mut parts = unified_key.splitn(2, '#');
+			let _obfuscated_namespace = parts.next();
```
tankyleo, line 427: nit: `skip` ? Also applies in the other branch

src/io/vss_store.rs
```diff
@@ -653,6 +674,114 @@ fn retry_policy() -> CustomRetryPolicy {
...
+
+		let response = client.list_key_versions(&request).await.map_err(|e| {
+			let msg = format!("Failed to determine schema version: {}", e);
+			Error::new(ErrorKind::Other, msg)
+		})?;
+
+		let wallet_data_present = !response.key_versions.is_empty();
+		if wallet_data_present {
+			// If the wallet data is present, it means we're not running for the first time.
+			Ok((VssSchemaVersion::V0, client))
```
tankyleo: Let's also write the version at the schema version key in this case ? So we don't have to keep checking the bdk wallet prefix to determine we are at V0.

src/io/vss_store.rs
tankyleo: Thanks ok sounds good

src/io/vss_store.rs
tnull: Yeah, skip would likely do the same thing, but this is a pre-existing pattern that I intentionally left untouched to make it clear we didn't change anything about the scheme.

src/io/vss_store.rs
tnull: I considered that but opted against it as it gets messy, i.e., it's a bit weird that then we'd need to write the `V0` value with the `V1` encryption scheme, i.e., it would be the only `V1` value in the entire store. Do you feel strongly about writing it here too? Note that we have very few V0 users out there yet.

src/io/vss_store.rs
tnull: Added fixups to introduce line wrapping and newlines. Also amended the commit message.

tnull:
Rebased after #689 has been merged.

src/io/vss_store.rs
tnull: Not sure I understand what you're asking here?

src/io/vss_store.rs
tankyleo: sounds good let's leave it as is thank you for the background

src/io/vss_store.rs
tankyleo: The function `fn build_obfuscated_prefix` contains an `if else` on the `VssSchemaVersion`, and the caller here `fn build_obfuscated_key` also has a `if else` on the `VssSchemaVersion`

tnull: Yes? Not sure I see why this is an issue?

tankyleo: no issue just a nit all good :)

tankyleo (commented):
reviewed the rebase, good to squash

tnull:
> reviewed the rebase, good to squash

Issue #367, opened 2024-10-08
Rebroadcast/bumping of on-chain wallet transactions


tnull:
Post-#358 we should look into better tracking of on-chain transactions (also do #67), and implement rebroadcasting / bumping for them.

Anyitechs:
> > Hi [@tnull](https://github.com/tnull), I'll like to work on this. Is anyone currently working on it?
> 
> I'm happy for you to pick it up, but AFAICT you already have a few tasks assigned? Maybe it would be better to drive them to completion first before picking up new ones?

Yes, true. Thank you.

Camillarhi:
Hello @tnull, I would like to pick this up, can I work on this?

Camillarhi:
> > Hello [@tnull](https://github.com/tnull), I would like to pick this up, can I work on this?
> 
> Sure! Please feel free to reach out via Discord if you have further questions/want to clarify the context.

Thanks! I will

Issue #452, opened 2025-01-30
Find way to drop RBF'd `Pending` payments


tnull:
In https://github.com/lightningdevkit/ldk-node/pull/432 we introduced tracking of onchain payments, which are marked as `PaymentStatus::Pending` until reaching `ANTI_REORG_DELAY` confirmations.

However, that means that if a previously tracked payment get's RBF'd the original might be marked as `Pending` forever in our store. We should find a way to detect double-spends and eventually drop the original entry.

This should possibly happen in conjunction with #367.

tnull:
Was addressed in #628 / #658 and follow-ups.

PR #628, opened 2025-08-23
Enhance onchain transaction management

Camillarhi:
This PR enhances on-chain transaction management:

1. **Rebroadcast/bumping of on-chain wallet transactions**

2. **Handle RBF'd Pending payments**

### Changes

- Added background job for rebroadcasting unconfirmed onchain transactions with max attempt limit
- Implemented RBF (Replace-by-Fee) functionality allowing users to bump fees on outbound unconfirmed transactions

### Related Issues

- Fixes https://github.com/lightningdevkit/ldk-node/issues/367

### DIFF
--- a/bindings/ldk_node.udl
+++ b/bindings/ldk_node.udl
@@ -277,6 +277,8 @@ interface OnchainPayment {
 	Txid send_to_address([ByRef]Address address, u64 amount_sats, FeeRate? fee_rate);
 	[Throws=NodeError]
 	Txid send_all_to_address([ByRef]Address address, boolean retain_reserve, FeeRate? fee_rate);
+	[Throws=NodeError]
+	Txid bump_fee_rbf(PaymentId payment_id, FeeRate? fee_rate);
 };
 
 interface FeeRate {
--- a/src/payment/onchain.rs
+++ b/src/payment/onchain.rs
@@ -10,6 +10,7 @@
 use std::sync::{Arc, RwLock};
 
 use bitcoin::{Address, Txid};
+use lightning::ln::channelmanager::PaymentId;
 
 use crate::config::Config;
 use crate::error::Error;
@@ -119,5 +120,21 @@ impl OnchainPayment {
 
 		let fee_rate_opt = maybe_map_fee_rate_opt!(fee_rate);
 		self.wallet.send_to_address(address, send_amount, fee_rate_opt)
+	}
+
+	/// Attempt to bump the fee of an unconfirmed transaction using Replace-by-Fee (RBF).
+	///
+	/// This creates a new transaction that replaces the original one, increasing the fee by the
+	/// specified increment to improve its chances of confirmation.
+	///
+	/// The new transaction will have the same outputs as the original but with a
+	/// higher fee, resulting in faster confirmation potential.
+	///
+	/// Returns the [`Txid`] of the new replacement transaction if successful.
+	pub fn bump_fee_rbf(
+		&self, payment_id: PaymentId, fee_rate: Option<FeeRate>,
+	) -> Result<Txid, Error> {
+		let fee_rate_opt = maybe_map_fee_rate_opt!(fee_rate);
+		self.wallet.bump_fee_rbf(payment_id, fee_rate_opt)
 	}
 }
--- a/src/payment/pending_payment_store.rs
+++ b/src/payment/pending_payment_store.rs
@@ -84,10 +84,11 @@ impl StorableObjectUpdate<PendingPaymentDetails> for PendingPaymentDetailsUpdate
 
 impl From<&PendingPaymentDetails> for PendingPaymentDetailsUpdate {
 	fn from(value: &PendingPaymentDetails) -> Self {
-		Self {
-			id: value.id(),
-			payment_update: Some(value.details.to_update()),
-			conflicting_txids: Some(value.conflicting_txids.clone()),
-		}
+		let conflicting_txids = if value.conflicting_txids.is_empty() {
+			None
+		} else {
+			Some(value.conflicting_txids.clone())
+		};
+		Self { id: value.id(), payment_update: Some(value.details.to_update()), conflicting_txids }
 	}
 }
--- a/src/payment/store.rs
+++ b/src/payment/store.rs
@@ -291,6 +291,15 @@ impl StorableObject for PaymentDetails {
 			}
 		}
 
+		if let Some(tx_id) = update.txid {
+			match self.kind {
+				PaymentKind::Onchain { ref mut txid, .. } => {
+					update_if_necessary!(*txid, tx_id);
+				},
+				_ => {},
+			}
+		}
+
 		if updated {
 			self.latest_update_timestamp = SystemTime::now()
 				.duration_since(UNIX_EPOCH)
@@ -540,6 +549,7 @@ pub(crate) struct PaymentDetailsUpdate {
 	pub direction: Option<PaymentDirection>,
 	pub status: Option<PaymentStatus>,
 	pub confirmation_status: Option<ConfirmationStatus>,
+	pub txid: Option<Txid>,
 }
 
 impl PaymentDetailsUpdate {
@@ -555,6 +565,7 @@ impl PaymentDetailsUpdate {
 			direction: None,
 			status: None,
 			confirmation_status: None,
+			txid: None,
 		}
 	}
 }
@@ -570,9 +581,9 @@ impl From<&PaymentDetails> for PaymentDetailsUpdate {
 			_ => (None, None, None),
 		};
 
-		let confirmation_status = match value.kind {
-			PaymentKind::Onchain { status, .. } => Some(status),
-			_ => None,
+		let (confirmation_status, txid) = match &value.kind {
+			PaymentKind::Onchain { status, txid, .. } => (Some(*status), Some(*txid)),
+			_ => (None, None),
 		};
 
 		let counterparty_skimmed_fee_msat = match value.kind {
@@ -593,6 +604,7 @@ impl From<&PaymentDetails> for PaymentDetailsUpdate {
 			direction: Some(value.direction),
 			status: Some(value.status),
 			confirmation_status,
+			txid,
 		}
 	}
 }
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -12,6 +12,7 @@ use std::sync::{Arc, Mutex};
 
 use bdk_chain::spk_client::{FullScanRequest, SyncRequest};
 use bdk_wallet::descriptor::ExtendedDescriptor;
+use bdk_wallet::error::{BuildFeeBumpError, CreateTxError};
 use bdk_wallet::event::WalletEvent;
 #[allow(deprecated)]
 use bdk_wallet::SignOptions;
@@ -29,8 +30,9 @@ use bitcoin::{
 	Address, Amount, FeeRate, OutPoint, ScriptBuf, Transaction, TxOut, Txid, WPubkeyHash, Weight,
 	WitnessProgram, WitnessVersion,
 };
-
-use lightning::chain::chaininterface::BroadcasterInterface;
+use lightning::chain::chaininterface::{
+	BroadcasterInterface, INCREMENTAL_RELAY_FEE_SAT_PER_1000_WEIGHT,
+};
 use lightning::chain::channelmonitor::ANTI_REORG_DELAY;
 use lightning::chain::{BestBlock, Listen};
 use lightning::events::bump_transaction::{Input, Utxo, WalletSource};
@@ -259,16 +261,23 @@ impl Wallet {
 						confirmation_status,
 					);
 
-					let pending_payment =
-						self.create_pending_payment_from_tx(payment.clone(), Vec::new());
+					self.payment_store.insert_or_update(payment.clone())?;
 
-					self.payment_store.insert_or_update(payment)?;
-					self.pending_payment_store.insert_or_update(pending_payment)?;
+					if payment_status == PaymentStatus::Pending {
+						let pending_payment =
+							self.create_pending_payment_from_tx(payment, Vec::new());
+
+						self.pending_payment_store.insert_or_update(pending_payment)?;
+					}
 				},
 				WalletEvent::ChainTipChanged { new_tip, .. } => {
-					// Get all on-chain payments that are Pending
 					let pending_payments: Vec<PendingPaymentDetails> =
 						self.pending_payment_store.list_filter(|p| {
+							debug_assert!(
+								p.details.status == PaymentStatus::Pending,
+								"Non-pending payment {:?} found in pending store",
+								p.details.id,
+							);
 							p.details.status == PaymentStatus::Pending
 								&& matches!(p.details.kind, PaymentKind::Onchain { .. })
 						});
@@ -307,7 +316,16 @@ impl Wallet {
 							.collect();
 
 						if !txs_to_broadcast.is_empty() {
-							let tx_refs: Vec<&Transaction> = txs_to_broadcast.iter().collect();
+							let tx_refs: Vec<(
+								&Transaction,
+								lightning::chain::chaininterface::TransactionType,
+							)> =
+								txs_to_broadcast
+									.iter()
+									.map(|tx| {
+										(tx, lightning::chain::chaininterface::TransactionType::Sweep { channels: vec![] })
+									})
+									.collect();
 							self.broadcaster.broadcast_transactions(&tx_refs);
 							log_info!(
 								self.logger,
@@ -335,23 +353,31 @@ impl Wallet {
 					self.payment_store.insert_or_update(payment)?;
 					self.pending_payment_store.insert_or_update(pending_payment)?;
 				},
-				WalletEvent::TxReplaced { txid, conflicts, tx, .. } => {
-					let payment_id = self
-						.find_payment_by_txid(txid)
-						.unwrap_or_else(|| PaymentId(txid.to_byte_array()));
+				WalletEvent::TxReplaced { txid, conflicts, .. } => {
+					let Some(payment_id) = self.find_payment_by_txid(txid) else {
+						log_error!(
+							self.logger,
+							"Could not find payment for replaced transaction {}. Skipping.",
+							txid
+						);
+						continue;
+					};
 
 					// Collect all conflict txids
-					let conflict_txids: Vec<Txid> =
+					let mut conflict_txids: Vec<Txid> =
 						conflicts.iter().map(|(_, conflict_txid)| *conflict_txid).collect();
 
-					let payment = self.create_payment_from_tx(
-						locked_wallet,
-						txid,
+					conflict_txids.push(txid);
+					// The payment already exists in the store at this point: `bump_fee_rbf` updates
+					// the payment store with the replacement txid before the next sync cycle, so we
+					// can safely fetch it here.
+					debug_assert!(
+						self.payment_store.get(&payment_id).is_some(),
+						"Payment {:?} expected in store during WalletEvent::TxReplaced but not found",
 						payment_id,
-						&tx,
-						PaymentStatus::Pending,
-						ConfirmationStatus::Unconfirmed,
 					);
+					let payment =
+						self.payment_store.get(&payment_id).ok_or(Error::InvalidPaymentId)?;
 					let pending_payment_details = self
 						.create_pending_payment_from_tx(payment.clone(), conflict_txids.clone());
 
@@ -1025,6 +1051,201 @@ impl Wallet {
 		}
 
 		None
+	}
+
+	#[allow(deprecated)]
+	pub(crate) fn bump_fee_rbf(
+		&self, payment_id: PaymentId, fee_rate: Option<FeeRate>,
+	) -> Result<Txid, Error> {
+		let payment = self.payment_store.get(&payment_id).ok_or_else(|| {
+			log_error!(self.logger, "Payment {} not found in payment store", payment_id);
+			Error::InvalidPaymentId
+		})?;
+
+		if let PaymentKind::Onchain { status, .. } = &payment.kind {
+			match status {
+				ConfirmationStatus::Confirmed { .. } => {
+					log_error!(
+						self.logger,
+						"Transaction {} is already confirmed and cannot be replaced via RBF",
+						payment_id
+					);
+					return Err(Error::InvalidPaymentId);
+				},
+				ConfirmationStatus::Unconfirmed => {},
+			}
+		}
+
+		if payment.direction != PaymentDirection::Outbound {
+			log_error!(
+				self.logger,
+				"Cannot RBF payment {}: only outbound payments can be replaced",
+				payment_id
+			);
+			return Err(Error::InvalidPaymentId);
+		}
+
+		let txid = match &payment.kind {
+			PaymentKind::Onchain { txid, .. } => *txid,
+			_ => {
+				log_error!(
+					self.logger,
+					"Payment {} is not an on-chain payment, cannot be replaced via RBF",
+					payment_id
+				);
+				return Err(Error::InvalidPaymentId);
+			},
+		};
+
+		let mut locked_wallet = self.inner.lock().unwrap();
+
+		debug_assert!(
+			locked_wallet.tx_details(txid).is_some(),
+			"Transaction {} expected in wallet but not found",
+			txid,
+		);
+		let old_tx = locked_wallet
+			.tx_details(txid)
+			.ok_or_else(|| {
+				log_error!(self.logger, "Transaction {} not found in wallet", txid);
+				Error::InvalidPaymentId
+			})?
+			.tx
+			.deref()
+			.clone();
+
+		let old_fee_rate = locked_wallet.calculate_fee_rate(&old_tx).map_err(|e| {
+			log_error!(self.logger, "Failed to calculate fee rate of transaction {}: {}", txid, e);
+			Error::WalletOperationFailed
+		})?;
+
+		// BIP 125 requires the replacement to pay a higher fee rate than the original.
+		// The minimum increase is the incremental relay fee.
+		let min_required_fee_rate_sat_per_kwu =
+			old_fee_rate.to_sat_per_kwu() + INCREMENTAL_RELAY_FEE_SAT_PER_1000_WEIGHT as u64;
+
+		let confirmation_target = ConfirmationTarget::OnchainPayment;
+		let estimated_fee_rate =
+			fee_rate.unwrap_or_else(|| self.fee_estimator.estimate_fee_rate(confirmation_target));
+
+		// Use the higher of minimum RBF requirement or current network estimate
+		let final_fee_rate_sat_per_kwu =
+			min_required_fee_rate_sat_per_kwu.max(estimated_fee_rate.to_sat_per_kwu());
+		let final_fee_rate = FeeRate::from_sat_per_kwu(final_fee_rate_sat_per_kwu);
+
+		let mut psbt = {
+			let mut builder = locked_wallet.build_fee_bump(txid).map_err(|e| {
+				log_error!(self.logger, "BDK fee bump failed for {}: {:?}", txid, e);
+				match e {
+					BuildFeeBumpError::TransactionNotFound(_) => Error::InvalidPaymentId,
+					BuildFeeBumpError::TransactionConfirmed(_) => {
+						log_error!(self.logger, "Payment {} is already confirmed", payment_id);
+						Error::InvalidPaymentId
+					},
+					BuildFeeBumpError::IrreplaceableTransaction(_) => {
+						Error::OnchainTxCreationFailed
+					},
+					BuildFeeBumpError::FeeRateUnavailable => Error::FeerateEstimationUpdateFailed,
+					BuildFeeBumpError::UnknownUtxo(_) => Error::OnchainTxCreationFailed,
+					BuildFeeBumpError::InvalidOutputIndex(_) => Error::OnchainTxCreationFailed,
+				}
+			})?;
+
+			builder.fee_rate(final_fee_rate);
+
+			match builder.finish() {
+				Ok(psbt) => Ok(psbt),
+				Err(CreateTxError::FeeRateTooLow { required: required_fee_rate }) => {
+					log_info!(self.logger, "BDK requires higher fee rate: {}", required_fee_rate);
+
+					// BDK may require a higher fee rate than our estimate due to
+					// differences in UTXO selection or transaction weight calculations.
+					// We cap the retry at 1.5x our target fee rate as a safety bound
+					// to avoid overpaying.
+					let max_allowed_fee_rate = FeeRate::from_sat_per_kwu(
+						final_fee_rate_sat_per_kwu.saturating_mul(3).saturating_div(2),
+					);
+					if required_fee_rate > max_allowed_fee_rate {
+						log_error!( self.logger, "BDK required fee rate {} exceeds sanity cap {} (1.5x our estimate) for tx {}", required_fee_rate, max_allowed_fee_rate, txid );
+						return Err(Error::InvalidFeeRate);
+					}
+
+					let mut builder = locked_wallet.build_fee_bump(txid).map_err(|e| {
+						log_error!(self.logger, "BDK fee bump retry failed for {}: {:?}", txid, e);
+						Error::InvalidFeeRate
+					})?;
+
+					builder.fee_rate(required_fee_rate);
+					builder.finish().map_err(|e| {
+						log_error!(
+							self.logger,
+							"Failed to finish PSBT with required fee rate: {:?}",
+							e
+						);
+						Error::InvalidFeeRate
+					})
+				},
+				Err(e) => {
+					log_error!(self.logger, "Failed to create fee bump PSBT: {:?}", e);
+					Err(Error::InvalidFeeRate)
+				},
+			}?
+		};
+
+		match locked_wallet.sign(&mut psbt, SignOptions::default()) {
+			Ok(finalized) => {
+				if !finalized {
+					log_error!(self.logger, "Failed to finalize signing for fee bump of {}", txid);
+					return Err(Error::OnchainTxCreationFailed);
+				}
+			},
+			Err(err) => {
+				log_error!(
+					self.logger,
+					"Failed to sign fee bump transaction for {}: {}",
+					txid,
+					err
+				);
+				return Err(err.into());
+			},
+		}
+
+		let mut locked_persister = self.persister.lock().unwrap();
+		locked_wallet.persist(&mut locked_persister).map_err(|e| {
+			log_error!(self.logger, "Failed to persist wallet after fee bump of {}: {}", txid, e);
+			Error::PersistenceFailed
+		})?;
+
+		let fee_bumped_tx = psbt.extract_tx().map_err(|e| {
+			log_error!(self.logger, "Failed to extract fee bump transaction for {}: {}", txid, e);
+			e
+		})?;
+
+		let new_txid = fee_bumped_tx.compute_txid();
+
+		self.broadcaster.broadcast_transactions(&[(
+			&fee_bumped_tx,
+			lightning::chain::chaininterface::TransactionType::Sweep { channels: vec![] },
+		)]);
+
+		let new_payment = self.create_payment_from_tx(
+			&locked_wallet,
+			new_txid,
+			payment.id,
+			&fee_bumped_tx,
+			PaymentStatus::Pending,
+			ConfirmationStatus::Unconfirmed,
+		);
+
+		let pending_payment_store =
+			self.create_pending_payment_from_tx(new_payment.clone(), Vec::new());
+
+		self.pending_payment_store.insert_or_update(pending_payment_store)?;
+		self.payment_store.insert_or_update(new_payment)?;
+
+		log_info!(self.logger, "RBF successful: replaced {} with {}", txid, new_txid);
+
+		Ok(new_txid)
 	}
 }
 
--- a/tests/integration_tests_rust.rs
+++ b/tests/integration_tests_rust.rs
@@ -14,7 +14,7 @@ use std::sync::Arc;
 use bitcoin::address::NetworkUnchecked;
 use bitcoin::hashes::sha256::Hash as Sha256Hash;
 use bitcoin::hashes::Hash;
-use bitcoin::{Address, Amount, ScriptBuf};
+use bitcoin::{Address, Amount, ScriptBuf, Txid};
 use common::logging::{init_log_logger, validate_log_entry, MultiNodeLogger, TestLogWriter};
 use common::{
 	bump_fee_and_broadcast, distribute_funds_unconfirmed, do_channel_full_cycle,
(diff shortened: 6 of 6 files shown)


bindings/ldk_node.udl
```diff
@@ -13,13 +13,20 @@ dictionary Config {
...
+	RebroadcastPolicy rebroadcast_policy;
 };
 
 dictionary AnchorChannelsConfig {
 	sequence<PublicKey> trusted_peers_no_reserve;
 	u64 per_channel_reserve_sats;
 };
 
+dictionary RebroadcastPolicy {
+	u64 min_rebroadcast_interval;
```
tnull: If you introduce any user-facing fields, please make sure to always include the unit in the variable name.

bindings/ldk_node.udl
```diff
@@ -13,13 +13,20 @@ dictionary Config {
...
 	SendingParameters? sending_parameters;
+	RebroadcastPolicy rebroadcast_policy;
 };
 
 dictionary AnchorChannelsConfig {
 	sequence<PublicKey> trusted_peers_no_reserve;
 	u64 per_channel_reserve_sats;
 };
 
+dictionary RebroadcastPolicy {
```
tnull: Honestly not sure if we should expose this to the user at all, or if we should simply implement sane defaults, mod possibly allowing to disable auto-rebroadcast.

src/payment/onchain.rs
```diff
@@ -120,4 +120,31 @@ impl OnchainPayment {
...
 	}
+
+	/// Manually trigger a rebroadcast of a specific transaction according to the default policy.
+	///
+	/// This is useful if you suspect a transaction may not have propagated properly through the
+	/// network and you want to attempt to rebroadcast it immediately rather than waiting for the
+	/// automatic background job to handle it.
+	///
+	/// updating the attempt count and last broadcast time for the transaction in the payment store.
+	pub fn rebroadcast_transaction(&self, txid: Txid) -> Result<(), Error> {
```
tnull: I think these APIs should take `PaymentId` rather than a `Txid`, given it's the identifier we use everywhere else in the API.

src/payment/onchain.rs
```diff
@@ -120,4 +120,31 @@ impl OnchainPayment {
...
+	/// This creates a new transaction that replaces the original one, increasing the fee by the
+	/// specified increment to improve its chances of confirmation. The original transaction must
+	/// be signaling RBF replaceability for this to succeed.
+	///
+	/// `fee_bump_increment` specifies the additional fee amount in satoshis to add to the
+	/// transaction. The new transaction will have the same outputs as the original but with a
+	/// higher fee, resulting in faster confirmation potential.
+	///
+	/// Returns the Txid of the new replacement transaction if successful.
+	pub fn bump_fee_rbf(&self, txid: Txid, fee_bump_increment: u64) -> Result<Txid, Error> {
```
tnull: As noted above, please always make sure to include units in variable names, in this case `fee_bump_increment`. 

That said, not sure if we should such add an API that allows user misuse if they'd pick a too-high or too low fee bump increment vs. creating an API implementing fee bumping based on the current fee estimation.

src/payment/store.rs
```diff
@@ -451,6 +475,9 @@ impl_writeable_tlv_based_enum!(PaymentKind,
 	(0, Onchain) => {
 		(0, txid, required),
 		(2, status, required),
+		(4, raw_tx, option),
+		(10, last_broadcast_time, option),
+		(12, broadcast_attempts, required),
```
tnull: Note that LDK's serialization format follows the ['it's okay to be odd'](https://github.com/lightning/bolts/blob/master/00-introduction.md#its-ok-to-be-odd) rule. That means that new serialized fields added after the fact always need to be odd and `option`al to ensure forwards/backwards compatibility (even though we currently only guarantee the latter in LDK Node).

src/payment/onchain.rs
Camillarhi: Thanks for the feedback. You're right, letting the user set the exact fee value is risky and can lead to setting a too-high or too-low bump fee. I'll rework the API to handle the bump internally based on the current fee estimation. I will also do nothing if the current estimation has dropped lower than the original fee.

src/payment/store.rs
Camillarhi: Thanks for catching this.  I'll revert the changes and make the new fields optional to ensure compatibility.

bindings/ldk_node.udl
Camillarhi: Thanks for the feedback. I've set auto-rebroadcast to be enabled by default for reliability, but added a config option to disable it for users who need more control, while using the set sane default values.

src/wallet/mod.rs
```diff
@@ -568,6 +660,325 @@ where
...
+			status: PaymentStatus::Pending,
+			latest_update_timestamp: SystemTime::now()
+				.duration_since(UNIX_EPOCH)
+				.unwrap_or(Duration::from_secs(0))
+				.as_secs(),
+		};
+
+		self.payment_store.remove(&payment_id)?;
+
+		self.payment_store.insert_or_update(payment_details)?;
```
moisesPompilio: I'm not fully sure about this flow: since with RBF we can’t guarantee which transaction will eventually confirm (usually the last one, but not always), removing the previous payment here might cause issues. What happens if the earlier tx ends up confirming instead of the latest — would the wallet balance be deducted while the store still shows the payment as pending? If so, could this lead to a confusing UX for the user?

Camillarhi: Excellent catch. You're right to be concerned about the potential for state inconsistency.

However, the update_payment_store method acts as a reconciler that syncs the payment store with the actual state of the BDK wallet. Since the payment store is designed to be a reflection of the wallet's state.

src/config.rs
```diff
@@ -179,6 +182,16 @@ pub struct Config {
...
+	/// This will determine whether to automatically rebroadcast unconfirmed transactions
+	/// (e.g., channel funding or sweep transactions).
+	///
+	/// If enabled, the node will periodically attempt to rebroadcast any unconfirmed transactions to
+	/// increase propagation and confirmation likelihood. This is helpful in cases where transactions
+	/// were dropped by the mempool or not widely propagated.
+	///
+	/// Defaults to `true`. Disabling this may be desired for privacy-sensitive use cases or low-bandwidth
+	/// environments, but may result in slower or failed confirmations if transactions are not re-announced.
+	pub auto_rebroadcast_unconfirmed_tx: Option<bool>,
```
tnull: Why not just make this a `bool`? What would `Some(false)` mean compared to `None`?

src/builder.rs
```diff
@@ -738,6 +738,22 @@ impl NodeBuilder {
...
+	///
+	/// If enabled (default), the node will periodically rebroadcast unconfirmed transactions to increase
+	/// the likelihood of confirmation. This helps recover from mempool drops or poor initial propagation.
+	///
+	/// Disabling this may be desired for privacy or bandwidth-constrained environments, but could result
+	/// in slower or failed confirmations if transactions are not re-announced.
+	///
+	/// Default: `true`
+	pub fn set_auto_rebroadcast_unconfirmed(
+		&mut self, enable: bool,
```
tnull: That's an odd API. If we're positive we don't want to expose the `RebroadcastPolicy` for now, let's just remove the setter on the builder entirely and users that want to disable the auto-rebroadcasting can just use the `bool` on `Config`.

src/config.rs
```diff
@@ -534,6 +548,45 @@ impl From<MaxDustHTLCExposure> for LdkMaxDustHTLCExposure {
...
+	/// - `1.5`: 50% increase each attempt
+	/// - `2.0`: 100% increase (doubling) each attempt
+	///
+	/// **Recommended values**: 1.2-2.0
+	pub backoff_factor: f32,
+}
+
+impl Default for RebroadcastPolicy {
+	fn default() -> Self {
+		Self { min_rebroadcast_interval: 300, max_broadcast_attempts: 24, backoff_factor: 1.5 }
```
tnull: Can we make these default values constants?

src/config.rs
```diff
@@ -534,6 +548,45 @@ impl From<MaxDustHTLCExposure> for LdkMaxDustHTLCExposure {
...
+/// to ensure they remain in mempools and eventually get confirmed.
+#[derive(Clone, Debug)]
+pub struct RebroadcastPolicy {
+	/// Minimum time between rebroadcast attempts in seconds.
+	///
+	/// This prevents excessive network traffic by ensuring a minimum delay
+	/// between consecutive rebroadcast attempts.
+	///
+	/// **Recommended values**: 60-600 seconds (1-10 minutes)
+	pub min_rebroadcast_interval: u64,
```
tnull: Please append `_secs` to this field to make it very clear what the unit is.

src/lib.rs
```diff
@@ -402,6 +402,33 @@ impl Node {
...
+					tokio::select! {
+						_ = stop_rebroadcast.changed() => {
+							log_debug!(
+								rebroadcast_logger,
+								"Stopping rebroadcasting unconfirmed transactions."
+							);
+							return;
+						}
+						_ = interval.tick() => {
+							if let Err(e) = rebroadcast_wallet.rebroadcast_unconfirmed_transactions() {
```
tnull: Hmm, please make sure all commits in the PR build and test individually (this doesn't build as `rebroadcast_unconfirmed_transactions` is only added in the next commit). It would be preferable to have multiple commits, but if the only way to make them build is to have all of them in one huge commit, so be it.

bindings/ldk_node.udl
```diff
@@ -13,6 +13,7 @@ dictionary Config {
 	u64 probing_liquidity_limit_multiplier;
 	AnchorChannelsConfig? anchor_channels_config;
 	SendingParameters? sending_parameters;
+	boolean? auto_rebroadcast_unconfirmed_tx
```
tnull: This is missing a `;` at the end.

src/payment/onchain.rs
```diff
@@ -120,4 +121,30 @@ impl OnchainPayment {
...
+	/// Manually trigger a rebroadcast of a specific transaction according to the default policy.
+	///
+	/// This is useful if you suspect a transaction may not have propagated properly through the
+	/// network and you want to attempt to rebroadcast it immediately rather than waiting for the
+	/// automatic background job to handle it.
+	///
+	/// updating the attempt count and last broadcast time for the transaction in the payment store.
+	pub fn rebroadcast_transaction(&self, payment_id: PaymentId) -> Result<(), Error> {
+		self.wallet.rebroadcast_transaction(payment_id)?;
+		Ok(())
```
tnull: nit: Can just drop the `?` and `Ok(())`, no?

src/payment/store.rs
```diff
@@ -353,6 +371,12 @@ pub enum PaymentKind {
 		txid: Txid,
 		/// The confirmation status of this payment.
 		status: ConfirmationStatus,
+		/// The raw transaction for rebroadcasting
+		raw_tx: Option<Vec<u8>>,
```
tnull: Should this just hold the `Transaction` rather than raw bytes?

src/payment/store.rs
```diff
@@ -451,6 +475,9 @@ impl_writeable_tlv_based_enum!(PaymentKind,
 	(0, Onchain) => {
 		(0, txid, required),
 		(2, status, required),
+		(4, raw_tx, option),
```
tnull: Not strictly necessary in LDK Node, but generally in LDK serialization fields added after the fact need to be *odd*, i.e., 1/3/5 in this case.

src/wallet/mod.rs
```diff
@@ -200,35 +202,66 @@ where
...
-				let direction = PaymentDirection::Inbound;
-				let amount_msat = Some(
-					received.to_sat().saturating_sub(sent.to_sat().saturating_sub(fee.to_sat()))
-						* 1000,
-				);
-				(direction, amount_msat)
-			};
 
-			let fee_paid_msat = Some(fee.to_sat() * 1000);
+			// Get existing payment or create new one
```
tnull: I'm not quite sure why we need these changes and the additional store access here? `insert_or_update` should do the trick, no?

src/wallet/mod.rs
tnull: We might potentially still end up with multiple entries for the same payment though, right, as we'd not drop the RBF entry once the original transaction confirms? There would also be no 'history' of all the bumps that happened.

Camillarhi: Yes, having multiple entries is a possibility. Regarding dropping the entry once the original confirms, that's what I plan on doing in this issue [#452 ](https://github.com/lightningdevkit/ldk-node/issues/452).

As for the history of bumps, a solution could be introducing another status `replaced` with a replacement ID that indicates the transaction was bumped and replaced.

src/wallet/mod.rs
Camillarhi: Yes, it should. I will revert on this

src/config.rs
Camillarhi: Thanks, this has been changed to `bool?`

src/builder.rs
Camillarhi: Thanks for pointing this out, the setter has been removed, and the users can use the `bool` on the `config`

src/config.rs
Camillarhi: Yes, this has been updated to use constant values

src/payment/store.rs
Camillarhi: Thanks, I have updated this to use the next odd values, which are 1,3, and 5

src/wallet/mod.rs
moisesPompilio: I think one way to simplify tracking RBF transactions is to add two fields to each payment: `is_rbf` (a boolean indicating whether the transaction is part of an RBF sequence) and `origin_payment_id` (optional, pointing to the original payment in the sequence). The original transaction would have `is_rbf = true` and `origin_payment_id = None`. Each RBF attempt would also have `is_rbf = true` and set `origin_payment_id` to the original transaction’s ID. If a new RBF is created from an existing RBF, it can reuse the same origin_payment_id, linking it back to the original.

With this setup, handling confirmations becomes straightforward:

- If an RBF transaction confirms, find the original transaction via `origin_payment_id` and clean up all other RBFs linked to it.
- When a transaction with is_rbf = true and origin_payment_id = None (i.e., an original transaction) confirms, find all RBF transactions with origin_payment_id equal to its ID and remove them.

This keeps data consistent and ensures that extra lookups are only needed when `is_rbf` is true, making RBF handling explicit and efficient.

For example, with A as the original transaction and B, C, D as RBFs:
```mermaid
graph TD
    A[A<br/>is_rbf=true<br/>origin=None]
    B[B<br/>is_rbf=true<br/>origin=A.payment_id]
    C[C<br/>is_rbf=true<br/>origin=A.payment_id]
    D[D<br/>is_rbf=true<br/>origin=A.payment_id]

    A --> B
    A --> C
    A --> D
```

moisesPompilio: I think there’s a design consideration for transactions that won’t confirm because an RBF replaced them. When an RBF transaction confirms, the original and any other RBFs in the sequence become invalid. In my opinion, a more user-friendly approach is to mark them as failed and reference the txid of the confirming transaction, which makes it clear why these transactions didn’t confirm while keeping the history intact.

Camillarhi: Thank you for this feedback. This aligns well with the introduction of a `replacement id` as mentioned above.

Regarding retaining the history of bumped transactions and marking them as `failed` when any confirms,  this could lead to increasing clutter of failed transactions if RBF is used frequently.

My proposed approach during node sync and store updates is when a confirmed transaction is encountered, if it's an RBF transaction, we can confirm that transaction and clear/remove the history of bumps from the store. This prevents unnecessary accumulation of failed transaction records while maintaining proper transaction state tracking.

src/payment/store.rs
Camillarhi: Yes, it should. It has been updated to hold `Transaction` instead

Camillarhi:
> @Camillarhi Excuse the delay here - last few weeks have been super busy pushing towards an LDK 0.2 beta release. Will see to get back to review on this in a proper ASAP, it might take another week or so though.
> 
> In the meantime I want to note that recently there has been some movement towards #446 as BDK now added events (see [bitcoindevkit/bdk_wallet#6](https://github.com/bitcoindevkit/bdk_wallet/issues/6) / [bitcoindevkit/bdk_wallet#310](https://github.com/bitcoindevkit/bdk_wallet/pull/310)) which will ship as part of the next BDK 2.2 release. So I do wonder if we should generally _first_ move to update our payment store based on the events emitted from syncing (also allowing for #448 to land), and then basing the changes here on top. This might be in particular interesting as [bitcoindevkit/bdk_wallet#310](https://github.com/bitcoindevkit/bdk_wallet/pull/310) included a `TxReplaced` event which might allow us to detect and track (all?) RBF transactions belonging to a 'payment' more easily.
> 
> Any thoughts on this?

Thanks for the update, no problem at all on the review delay. I’ll go through the links you mentioned and follow up with thoughts

src/wallet/mod.rs
```diff
@@ -307,9 +319,15 @@ impl Wallet {
...
-						.create_pending_payment_from_tx(payment.clone(), conflict_txids.clone());
+					let pending_payment_details = self.create_pending_payment_from_tx(
+						payment.clone(),
+						conflict_txids.clone(),
+						None,
+						None,
+						None,
+					);
 
+					self.payment_store.insert_or_update(payment)?;
```
tnull: Hmm, this seems ~unrelated? Is this a fix for something we overlooked/omitted before or do we need this suddenly? If it's the former, this should be in a separate commit.

src/config.rs
```diff
@@ -561,6 +579,49 @@ pub enum AsyncPaymentsRole {
 	Server,
 }
 
+/// Policy for controlling transaction rebroadcasting behavior.
+///
+/// Determines the strategy for resending unconfirmed transactions to the network
+/// to ensure they remain in mempools and eventually get confirmed.
+#[derive(Clone, Debug)]
+pub struct RebroadcastPolicy {
```
tnull: I still don't think we need all this extra boilerplate for something as simple as rebroadcasting. Let's just rebroadcast once a block and be done with it.

src/payment/pending_payment_store.rs
```diff
@@ -20,11 +20,20 @@ pub struct PendingPaymentDetails {
 	pub details: PaymentDetails,
 	/// Transaction IDs that have replaced or conflict with this payment.
 	pub conflicting_txids: Vec<Txid>,
+	/// The raw transaction for rebroadcasting
+	pub raw_tx: Option<Transaction>,
```
tnull: This is a lot of extra data stored. When would we reliably drop that data again? What happens if this never confirms, will we just keep this entry forever in our store?

src/wallet/mod.rs
```diff
@@ -978,6 +1008,114 @@ impl Wallet {
 
 		None
 	}
+
+	pub(crate) fn rebroadcast_unconfirmed_transactions(&self) -> Result<(), Error> {
```
tnull: As mentioned above, I think all this logic is waaaay to complicated. We should just rebroadcast once a block and be done with it, really doesn't require any user-facing API really, as they need to assume that once-broadcast transaction will remain in/return to the mempool anyways.

src/payment/pending_payment_store.rs
```diff
@@ -86,7 +86,10 @@ impl StorableObject for PendingPaymentDetails {
 		}
 
 		if let Some(new_conflicting_txids) = &update.conflicting_txids {
-			update_if_necessary!(self.conflicting_txids, new_conflicting_txids.clone());
+			// Don't overwrite existing conflicts with an empty list
```
tnull: Not quite sure I immediately understand why this is happening?

src/wallet/mod.rs
```diff
@@ -311,9 +312,11 @@ impl Wallet {
 					let conflict_txids: Vec<Txid> =
 						conflicts.iter().map(|(_, conflict_txid)| *conflict_txid).collect();
 
+					// Use the last transaction id in the conflicts as the new txid
+					let new_txid = conflicts.last().map(|(_, new_tx)| *new_tx).unwrap_or(txid);
```
tnull: Hmm, this seems very unreliable? I don't think it's part of the API guarantees that it's always the last `Txid` that will be our original Txid? Can we lean on that or do we need more explicit APIs from BDK here?

src/wallet/mod.rs
```diff
@@ -1116,6 +1119,145 @@ impl Wallet {
...
+					BuildFeeBumpError::FeeRateUnavailable => Error::InvalidPaymentId,
+					_ => Error::InvalidFeeRate,
+				}
+			})?;
+
+			builder.fee_rate(estimated_fee_rate);
+
+			match builder.finish() {
+				Ok(psbt) => Ok(psbt),
+				Err(CreateTxError::FeeRateTooLow { required }) => {
```
tnull: Can we name this `required_fee_rate` or similar?

src/wallet/mod.rs
```diff
@@ -1116,6 +1119,145 @@ impl Wallet {
...
+					log_info!(self.logger, "BDK requires higher fee rate: {}", required);
+
+					// Safety check
+					const MAX_REASONABLE_FEE_RATE_SAT_VB: u64 = 1000;
```
tnull: Where did we get this limit from? Also, usually sat/vb is a float value and we tend to use sat/kwu in LDK. Would be great to keep it as consistent as possible and do such calculations in sat/kwu too.

tests/integration_tests_rust.rs
```diff
@@ -685,8 +685,111 @@ async fn onchain_wallet_recovery() {
 }
 
 #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
-async fn test_rbf_via_mempool() {
-	run_rbf_test(false).await;
+async fn onchain_fee_bump_rbf() {
```
tnull: Can we leave the original test in place and add (a lot more) test coverage via proptests or similarly randomized test scenarios? It seems that with RBF so much could go wrong that we need to add considerable test coverage before moving on.

src/wallet/mod.rs
Camillarhi: This should be in the second commit. I'll fix that.

src/config.rs
Camillarhi: Makes sense, I'll remove the `RebroadcastPolicy` entirely along with the `backoff/attempt-tracking` logic. I'll trigger the rebroadcast from the `ChainTipChanged` event from BDK in `update_payment_store()` instead of the timer-based background task, so it naturally happens once per block.

src/payment/pending_payment_store.rs
Camillarhi: The guard against empty lists was meant to avoid accidentally clearing existing conflict data. I'll clean up the call sites to pass `None` when there are no conflicts and remove this extra check.

src/wallet/mod.rs
Camillarhi: From my tests, it happened to be the last one, but looking at the implementation, `direct_conflicts` used to get the conflicts for a replaced txid, comes from BDK's `tx_graph` and doesn't guarantee any ordering. I think this needs to be fixed upstream in BDK. What do you think is the best approach here?

src/payment/pending_payment_store.rs
Camillarhi: I stored `raw_tx` here so the rebroadcast logic could just call `broadcast_transactions()` without having to lock and query the BDK wallet. But as you have pointed out, there's no cleanup path if the tx never confirms. I'll drop it from `PendingPaymentDetails` and pull the transaction from BDK when needed instead.

src/wallet/mod.rs
Camillarhi: I'll use sat/kwu for consistency with LDK conventions.

The limit itself is a safety check to prevent setting an extremely high fee rate when bumping. Since BDK's build_fee_bump can return a required minimum fee that's higher than our estimated rate, I wanted to cap how high we'd go. I'm open to adjusting it or removing it entirely if you think the check isn't needed.

src/wallet/mod.rs
tnull: > From my tests, it happened to be the last one, but looking at the implementation, `direct_conflicts` used to get the conflicts for a replaced txid, comes from BDK's `tx_graph` and doesn't guarantee any ordering. I think this needs to be fixed upstream in BDK. What do you think is the best approach here?

Hmm, mind opening an issue with them, maybe they can guarantee an ordering here, so that indeed the last `Vec` entry is always guaranteed to be the latest version that is kept in the mempool?

Camillarhi: Sure, I will open one

src/wallet/mod.rs
Camillarhi: After reconsidering, I've removed the safety check entirely as I don't think it is necessary because BDK's `build_fee_bump()` already validates BIP 125 requirements internally and only returns `FeeRateTooLow` with a legitimate required rate. I've updated the code to trust BDK's required fee rate if it fails, and also keeping other calculations in sat/kWU for consistency with LDK.

src/wallet/mod.rs
tnull: It seems to me that pending https://github.com/bitcoindevkit/bdk/issues/2112 our best option here is to track a list of the original previous outpoints on the pending payment entry, and when we then get a replacement we can then manually iterate through all and see if the replacement's spends matches one of them?

Camillarhi: That approach wouldn't really provide a solution for this use case. I basically just need the new `txid` for when the transaction is rebroadcast.
As discussed elsewhere, I'll try another approach and see if that gives me the intended result.

Camillarhi: Update: Resolved this by fetching payment details, since the replacement already updates the stored state with the new txid. Thanks

src/wallet/mod.rs
```diff
@@ -29,6 +29,7 @@ use bitcoin::{
 	Address, Amount, FeeRate, OutPoint, ScriptBuf, Transaction, TxOut, Txid, WPubkeyHash, Weight,
 	WitnessProgram, WitnessVersion,
 };
+
```
tnull: nit: Changes like this are unrelated and in this case will be reverted by the weekly `cargo +nightly fmt` job anyways.

src/wallet/mod.rs
```diff
@@ -244,31 +245,54 @@ impl Wallet {
 					self.pending_payment_store.insert_or_update(pending_payment)?;
 				},
 				WalletEvent::ChainTipChanged { new_tip, .. } => {
-					// Get all payments that are Pending with Confirmed status
+					// Get all on-chain payments that are Pending
 					let pending_payments: Vec<PendingPaymentDetails> =
 						self.pending_payment_store.list_filter(|p| {
 							p.details.status == PaymentStatus::Pending
```
tnull, line 281: Hmm,  wouldn't we expect all in the `Pending` store to be pending? In fact, should we add a debug assert in 1-2 places that ensures any entries are marked pending, and any other entries are removed from the pending-payment store?

src/wallet/mod.rs
```diff
@@ -244,31 +245,54 @@ impl Wallet {
...
+						let txs_to_broadcast: Vec<Transaction> = unconfirmed_outbound_txids
+							.iter()
+							.filter_map(|txid| {
+								locked_wallet.tx_details(*txid).map(|d| (*d.tx).clone())
+							})
+							.collect();
+
+						if !txs_to_broadcast.is_empty() {
+							let tx_refs: Vec<&Transaction> = txs_to_broadcast.iter().collect();
+							self.broadcaster.broadcast_transactions(&tx_refs);
```
tnull, line 329: This will need a rebase now as the `BroadcasterInterface` changed on `main` (note that the  'real' introduction of `TransactionType` etc will be in #79).

src/wallet/mod.rs
```diff
@@ -323,14 +326,9 @@ impl Wallet {
...
-					let payment = self.create_payment_from_tx(
-						locked_wallet,
-						txid,
-						payment_id,
-						&tx,
-						PaymentStatus::Pending,
-						ConfirmationStatus::Unconfirmed,
-					);
+					// We fetch payment details here since the replacement has updated the stored state
+					let payment =
```
tnull, line 379: Hmm, but that would mean we expect this `get` to always return `Some` at this point, right? Can we add a `debug_assert` checking that it does?

src/wallet/mod.rs
```diff
@@ -1002,6 +1000,152 @@ impl Wallet {
...
+			return Err(Error::InvalidPaymentId);
+		}
+
+		let txid = match &payment.kind {
+			PaymentKind::Onchain { txid, .. } => *txid,
+			_ => return Err(Error::InvalidPaymentId),
+		};
+
+		let old_tx =
+			locked_wallet.tx_details(txid).ok_or(Error::InvalidPaymentId)?.tx.deref().clone();
```
tnull: Hmm, same here, we expect this to never fail - can we add a `debug_assert`?

src/wallet/mod.rs
```diff
@@ -1002,6 +1000,152 @@ impl Wallet {
...
+			Error::InvalidPaymentId
+		})?;
+		let old_fee_rate_sat_per_kwu = old_fee_rate.to_sat_per_kwu();
+
+		// BIP 125 requires the replacement to pay a higher fee rate than the original.
+		// The minimum increase is the incremental relay fee.
+		let min_required_fee_rate_sat_per_kwu =
+			old_fee_rate_sat_per_kwu + INCREMENTAL_RELAY_FEE_SAT_PER_1000_WEIGHT as u64;
+
+		let confirmation_target = ConfirmationTarget::OnchainPayment;
```
tnull, line 1127: See above, we might want to allow for an override target fee rate like we do for `send_to_address` etc?

src/payment/onchain.rs
```diff
@@ -120,4 +121,18 @@ impl OnchainPayment {
...
+	///
+	/// This creates a new transaction that replaces the original one, increasing the fee by the
+	/// specified increment to improve its chances of confirmation. The original transaction must
+	/// be signaling RBF replaceability for this to succeed.
+	///
+	/// The new transaction will have the same outputs as the original but with a
+	/// higher fee, resulting in faster confirmation potential.
+	///
+	/// Returns the Txid of the new replacement transaction if successful.
+	pub fn bump_fee_rbf(&self, payment_id: PaymentId) -> Result<Txid, Error> {
```
tnull: Do we need an optional fee-rate override parameter here, similar to what we have in `send_to_address`?

src/wallet/mod.rs
```diff
@@ -1002,6 +1000,152 @@ impl Wallet {
...
+				Ok(psbt) => Ok(psbt),
+				Err(CreateTxError::FeeRateTooLow { required: required_fee_rate }) => {
+					log_info!(self.logger, "BDK requires higher fee rate: {}", required_fee_rate);
+
+					let mut builder = locked_wallet.build_fee_bump(txid).map_err(|e| {
+						log_error!(self.logger, "BDK fee bump retry failed for {}: {:?}", txid, e);
+						Error::InvalidFeeRate
+					})?;
+
+					builder.fee_rate(required_fee_rate);
```
tnull, line 1178: Hmm, should we apply an upper-bound to the `require_fee_rate`, i.e., ensure it's still in some margin of what we expected? Maybe we even want to apply some upper-bound (relative to the original rate?) to our estimate to ensure we don't spend all our funds on fees if the chain source's fee estimation endpoint is lying to us?

src/wallet/mod.rs
Camillarhi: Yes, this is true. All records in `PendingStore` should be marked as pending. I will also add the debug assert

src/wallet/mod.rs
Camillarhi: Sure, I will rebase. Also, did you mean #791 ?

tnull: > Also, did you mean #791 ?

Ah, yes, indeed!

src/wallet/mod.rs
Camillarhi: Yes, that is fine, I will go ahead and add that

src/payment/onchain.rs
Camillarhi: Yes, I will go ahead and include it

src/wallet/mod.rs
Camillarhi: Yeah, I thought about this too, that's why I computed a `min_required_fee_rate_sat_per_kwu` above:

```
let min_required_fee_rate_sat_per_kwu =
	old_fee_rate_sat_per_kwu + INCREMENTAL_RELAY_FEE_SAT_PER_1000_WEIGHT as u64;
```

Then take the max of that and the `estimated_fee_rate`. Based on the tests so far, the `final_fee_rate` is what ends up being used in most cases, the difference between it and `required_fee_rate` has been around 2-3 sat/kwu. I think it'd be pretty rare for the chain source estimate to actually win out, but an upper-bound sanity check on it is reasonable to add

src/wallet/mod.rs
tnull: Hmm seems we hitting some of the new `debug_asserts` in `test_rbf_via_direct_block_insertion`:

```
Payment b25fa25838fc1654906457e86846d380996624ff8750c4910049be4350710f1c expected in store during WalletEvent::TxReplaced but not found
```

Might be good to understand what's happening here. Possibly some of the fixes in #769 could also be related, but not sure.

tnull (commented):
Seems CI is unhappy/hanging due to some of the `debug_assert`s getting hit now.

src/wallet/mod.rs
tnull: > Based on the tests so far, the final_fee_rate is what ends up being used in most cases, the difference between it and required_fee_rate has been around 2-3 sat/kwu. I think it'd be pretty rare for the chain source estimate to actually win out, but an upper-bound sanity check on it is reasonable to add

Did you do tests on mainnet as well? Note that testing such things on `regtest` or even `testnet`/`signet` will lead to bogus values.

Camillarhi: Yes, this is correct. I will introduce an upper-bound sanity check.

src/wallet/mod.rs
Camillarhi: Thanks, I will check out what's happening

tnull:
Might also need a rebase now after https://github.com/lightningdevkit/ldk-node/pull/769 landed.

src/wallet/mod.rs
Camillarhi: @tnull, I introduced a sanity check at 1.5x our target fee rate to avoid overpaying

src/payment/onchain.rs
```diff
@@ -120,4 +121,21 @@ impl OnchainPayment {
 		let fee_rate_opt = maybe_map_fee_rate_opt!(fee_rate);
 		self.wallet.send_to_address(address, send_amount, fee_rate_opt)
 	}
+
+	/// Attempt to bump the fee of an unconfirmed transaction using Replace-by-Fee (RBF).
+	///
+	/// This creates a new transaction that replaces the original one, increasing the fee by the
+	/// specified increment to improve its chances of confirmation. The original transaction must
```
tnull: I think we can drop `The original transaction must be signaling RBF replaceability for this to succeed.` as this is given for all of our payments, no?

src/payment/onchain.rs
```diff
@@ -120,4 +121,21 @@ impl OnchainPayment {
...
+	/// be signaling RBF replaceability for this to succeed.
+	///
+	/// The new transaction will have the same outputs as the original but with a
+	/// higher fee, resulting in faster confirmation potential.
+	///
+	/// Returns the Txid of the new replacement transaction if successful.
```
tnull: nit: Tick `Txid`.

src/wallet/mod.rs
```diff
@@ -286,6 +295,11 @@ impl Wallet {
 									payment.details.status = PaymentStatus::Succeeded;
 									self.payment_store.insert_or_update(payment.details)?;
 									self.pending_payment_store.remove(&payment_id)?;
+									debug_assert!(
```
tnull: I this `debug_assert ` shouldn't be necessary, if anything it might be race-y if our `KVStore` view is somehow temporarily inconsistent.

src/wallet/mod.rs
```diff
@@ -335,7 +358,7 @@ impl Wallet {
 					self.payment_store.insert_or_update(payment)?;
 					self.pending_payment_store.insert_or_update(pending_payment)?;
 				},
-				WalletEvent::TxReplaced { txid, conflicts, tx, .. } => {
+				WalletEvent::TxReplaced { txid, conflicts, .. } => {
 					let payment_id = self
 						.find_payment_by_txid(txid)
 						.unwrap_or_else(|| PaymentId(txid.to_byte_array()));
```
tnull: Ah, it seems we shouldn't fallback here to `PaymentId(txid.to_byte_array())` but rather log and skip processing if we can't find the respective payment entry?

Then, not sure if we should use `find_payment_by_txid` at all if we now assume that the payment entry will be in our store?

src/wallet/mod.rs
```diff
@@ -335,7 +358,7 @@ impl Wallet {
 					self.payment_store.insert_or_update(payment)?;
 					self.pending_payment_store.insert_or_update(pending_payment)?;
 				},
-				WalletEvent::TxReplaced { txid, conflicts, tx, .. } => {
+				WalletEvent::TxReplaced { txid, conflicts, .. } => {
```
tnull, line 356: I do wonder if this also handles the case of *inbound* replaced transactions correctly?

src/wallet/mod.rs
```diff
@@ -344,14 +367,14 @@ impl Wallet {
 					let conflict_txids: Vec<Txid> =
 						conflicts.iter().map(|(_, conflict_txid)| *conflict_txid).collect();
 
-					let payment = self.create_payment_from_tx(
-						locked_wallet,
-						txid,
+					// We fetch payment details here since the replacement has updated the stored state
```
tnull: I'm confused by this comment: what logic exactly is this referring to - where does the replacement update the store, i.e., why can we be sure it has been updated here? It seems our processing order would ensure we always process *this* logic first, not `TxUnconfirmed`, if this is what this comment is referring to?

src/wallet/mod.rs
```diff
@@ -1026,6 +1049,175 @@ impl Wallet {
...
+			locked_wallet.tx_details(txid).is_some(),
+			"Transaction {} expected in wallet but not found",
+			txid,
+		);
+		let old_tx =
+			locked_wallet.tx_details(txid).ok_or(Error::InvalidPaymentId)?.tx.deref().clone();
+
+		let old_fee_rate = locked_wallet.calculate_fee_rate(&old_tx).map_err(|e| {
+			log_error!(self.logger, "Failed to calculate fee rate of transaction {}: {}", txid, e);
+			Error::InvalidPaymentId
```
tnull: This probably should use a different error variant.

src/wallet/mod.rs
```diff
@@ -1026,6 +1049,175 @@ impl Wallet {
...
+
+		let mut locked_wallet = self.inner.lock().unwrap();
+
+		debug_assert!(
+			locked_wallet.tx_details(txid).is_some(),
+			"Transaction {} expected in wallet but not found",
+			txid,
+		);
+		let old_tx =
+			locked_wallet.tx_details(txid).ok_or(Error::InvalidPaymentId)?.tx.deref().clone();
```
tnull: Please try to always log when returning an error, also in this instance.

src/wallet/mod.rs
```diff
@@ -1026,6 +1049,175 @@ impl Wallet {
...
+
+		let old_fee_rate = locked_wallet.calculate_fee_rate(&old_tx).map_err(|e| {
+			log_error!(self.logger, "Failed to calculate fee rate of transaction {}: {}", txid, e);
+			Error::InvalidPaymentId
+		})?;
+		let old_fee_rate_sat_per_kwu = old_fee_rate.to_sat_per_kwu();
```
tnull: I think we can drop the intermediate variable here, as `old_fee_rate.to_sat_per_kwu()` is basically same length.

src/wallet/mod.rs
tnull: Hmm, thanks. So now I still wonder if should ever spend more than the explicit override fee rate set by the user. Generally, maybe this auto-retry logic is already to complicated, and it might be easier to just bubble-up the `FeeRateTooLow` error and have the user use an explicit override fee rate of their choice to adjust the target upwards if they'd ever hit it?

src/wallet/mod.rs
```diff
@@ -1002,6 +1049,175 @@ impl Wallet {
...
+			&locked_wallet,
+			new_txid,
+			payment.id,
+			&fee_bumped_tx,
+			PaymentStatus::Pending,
+			ConfirmationStatus::Unconfirmed,
+		);
+
+		let pending_payment_store =
+			self.create_pending_payment_from_tx(new_payment.clone(), Vec::new());
```
tnull, line 1241: Wouldn't we want to track the original `Txid` as a conflict here?

src/payment/onchain.rs
Camillarhi: Yes, we can drop that. All our payments signal RBF, so it's not worth documenting.

src/wallet/mod.rs
Camillarhi: The comment refers to `bump_fee_rbf` , It updates the payment store with the new replacement `txId` before the next wallet sync. So by the time the sync runs and emits `TxReplaced`, the payment is already updated in the store. I'll reword the comment to make this clearer.

src/wallet/mod.rs
Camillarhi: Agreed, we should not fall back to `PaymentId(txid.to_byte_array())` here. However, we should still use `find_payment_by_txid` because the `txid` in `TxReplaced` is the old replaced txid, `find_payment_by_txid` looks up the `pending_payment_store` to map it back to the correct payment id (either directly or via `conflicting_txids`). If no payment id is found, we should log and skip processing.

src/wallet/mod.rs
tnull: Ah, but this really only holds if the replacement was initiated by us then. How do we handle replacements for inbound payments for example?

src/wallet/mod.rs
Camillarhi: Yes, the logic handles inbound transactions correctly as well. I will also expand the integration test to verify the inbound side.

src/wallet/mod.rs
Camillarhi: For inbound replacements, since we already have the unconfirmed transaction in the store, it relies on using `find_payment_by_txid` to look up the `PaymentID` by matching with the `txid` or `conflicts` and updates the payment.

src/wallet/mod.rs
Camillarhi: `bump_fee_rbf` just creates the initial pending entry, and the empty vec gets treated as `None` during update, so it won't overwrite existing conflicts. The conflict tracking is handled in the `TxReplaced` event. I've also updated `conflict_txids` to include the replaced txid, as this is not returned as part of the conflicts from bdk_wallet, so no intermediate txid gets lost across consecutive bumps.

src/wallet/mod.rs
Camillarhi: How about we only do the auto-retry when the user didn't provide an explicit fee rate? If they passed a specific rate and BDK rejects it, we bubble up the error and let them adjust. If they passed `None`, we handle it internally with the 1.5x sanity cap.

tnull: Sounds good!

src/wallet/mod.rs
```diff
@@ -311,23 +353,31 @@ impl Wallet {
...
+						"Payment {:?} expected in store during WalletEvent::TxReplaced but not found",
 						payment_id,
-						&tx,
-						PaymentStatus::Pending,
-						ConfirmationStatus::Unconfirmed,
 					);
+					let payment =
+						self.payment_store.get(&payment_id).ok_or(Error::InvalidPaymentId)?;
 					let pending_payment_details = self
 						.create_pending_payment_from_tx(payment.clone(), conflict_txids.clone());
```
tnull, line 382: nit: I think we can avoid this `clone` now.

src/wallet/mod.rs
tnull: Let's do this in a follow-up!

tests/integration_tests_rust.rs
```diff
@@ -2462,3 +2462,161 @@ async fn persistence_backwards_compatibility() {
 
 	node_new.stop().unwrap();
 }
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn onchain_fee_bump_rbf() {
+	let (bitcoind, electrsd) = setup_bitcoind_and_electrsd();
+	let chain_source = TestChainSource::Esplora(&electrsd);
```
tnull, line 2469: Ah, actually, in the follow-up, please also switch this to use `random_chain_source` and run it a few times locally to make sure the test (and our logic changes ofc) works independently of the chosen chain source. @Camillarhi

tests/integration_tests_rust.rs
```diff
@@ -2462,3 +2462,161 @@ async fn persistence_backwards_compatibility() {
...
+		Err(NodeError::InvalidPaymentId),
+		node_a.onchain_payment().bump_fee_rbf(payment_id, None)
+	);
+
+	// Successful fee bump
+	let new_txid = node_b.onchain_payment().bump_fee_rbf(payment_id, None).unwrap();
+	wait_for_tx(&electrsd.client, new_txid).await;
+
+	// Sleep to allow for transaction propagation
+	tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
tnull, line 2520: I think we should also be able to avoid this manual sleeps here and below as `wait_for_tx` should wait until the `new_txid` shows up?

Camillarhi: Yes, we should I will also address this in the follow-up PR

PR #632, opened 2025-09-10
Add onion mailbox for async recipients

joostjager:
Adds functionality for LSPs to hold onion messages for their clients until they come back online.

Depends on https://github.com/lightningdevkit/rust-lightning/pull/4046

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -7,9 +7,9 @@
 
 use crate::chain::ChainSource;
 use crate::config::{
-	default_user_config, may_announce_channel, AnnounceError, BitcoindRestClientConfig, Config,
-	ElectrumSyncConfig, EsploraSyncConfig, DEFAULT_ESPLORA_SERVER_URL, DEFAULT_LOG_FILENAME,
-	DEFAULT_LOG_LEVEL, WALLET_KEYS_SEED_LEN,
+	default_user_config, may_announce_channel, AnnounceError, AsyncPaymentsRole,
+	BitcoindRestClientConfig, Config, ElectrumSyncConfig, EsploraSyncConfig,
+	DEFAULT_ESPLORA_SERVER_URL, DEFAULT_LOG_FILENAME, DEFAULT_LOG_LEVEL, WALLET_KEYS_SEED_LEN,
 };
 
 use crate::connection::ConnectionManager;
@@ -27,6 +27,7 @@ use crate::liquidity::{
 };
 use crate::logger::{log_error, LdkLogger, LogLevel, LogWriter, Logger};
 use crate::message_handler::NodeCustomMessageHandler;
+use crate::payment::asynchronous::om_mailbox::OnionMessageMailbox;
 use crate::peer_store::PeerStore;
 use crate::runtime::Runtime;
 use crate::tx_broadcaster::TransactionBroadcaster;
@@ -191,6 +192,8 @@ pub enum BuildError {
 	LoggerSetupFailed,
 	/// The given network does not match the node's previously configured network.
 	NetworkMismatch,
+	/// The role of the node in an asynchronous payments context is not compatible with the current configuration.
+	AsyncPaymentsConfigMismatch,
 }
 
 impl fmt::Display for BuildError {
@@ -219,6 +222,12 @@ impl fmt::Display for BuildError {
 			Self::NetworkMismatch => {
 				write!(f, "Given network does not match the node's previously configured network.")
 			},
+			Self::AsyncPaymentsConfigMismatch => {
+				write!(
+					f,
+					"The async payments role is not compatible with the current configuration."
+				)
+			},
 		}
 	}
 }
@@ -240,6 +249,7 @@ pub struct NodeBuilder {
 	gossip_source_config: Option<GossipSourceConfig>,
 	liquidity_source_config: Option<LiquiditySourceConfig>,
 	log_writer_config: Option<LogWriterConfig>,
+	async_payments_role: Option<AsyncPaymentsRole>,
 	runtime_handle: Option<tokio::runtime::Handle>,
 }
 
@@ -266,6 +276,7 @@ impl NodeBuilder {
 			liquidity_source_config,
 			log_writer_config,
 			runtime_handle,
+			async_payments_role: None,
 		}
 	}
 
@@ -544,6 +555,21 @@ impl NodeBuilder {
 		Ok(self)
 	}
 
+	/// Sets the role of the node in an asynchronous payments context.
+	///
+	/// See <https://github.com/lightning/bolts/pull/1149> for more information about the async payments protocol.
+	pub fn set_async_payments_role(
+		&mut self, role: Option<AsyncPaymentsRole>,
+	) -> Result<&mut Self, BuildError> {
+		if let Some(AsyncPaymentsRole::Server) = role {
+			may_announce_channel(&self.config)
+				.map_err(|_| BuildError::AsyncPaymentsConfigMismatch)?;
+		}
+
+		self.async_payments_role = role;
+		Ok(self)
+	}
+
 	/// Builds a [`Node`] instance with a [`SqliteStore`] backend and according to the options
 	/// previously configured.
 	pub fn build(&self) -> Result<Node, BuildError> {
@@ -700,6 +726,7 @@ impl NodeBuilder {
 			self.chain_data_source_config.as_ref(),
 			self.gossip_source_config.as_ref(),
 			self.liquidity_source_config.as_ref(),
+			self.async_payments_role,
 			seed_bytes,
 			runtime,
 			logger,
@@ -732,6 +759,7 @@ impl NodeBuilder {
 			self.chain_data_source_config.as_ref(),
 			self.gossip_source_config.as_ref(),
 			self.liquidity_source_config.as_ref(),
+			self.async_payments_role,
 			seed_bytes,
 			runtime,
 			logger,
@@ -989,6 +1017,13 @@ impl ArcedNodeBuilder {
 		self.inner.write().unwrap().set_node_alias(node_alias).map(|_| ())
 	}
 
+	/// Sets the role of the node in an asynchronous payments context.
+	pub fn set_async_payments_role(
+		&self, role: Option<AsyncPaymentsRole>,
+	) -> Result<(), BuildError> {
+		self.inner.write().unwrap().set_async_payments_role(role).map(|_| ())
+	}
+
 	/// Builds a [`Node`] instance with a [`SqliteStore`] backend and according to the options
 	/// previously configured.
 	pub fn build(&self) -> Result<Arc<Node>, BuildError> {
@@ -1082,8 +1117,9 @@ impl ArcedNodeBuilder {
 fn build_with_store_internal(
 	config: Arc<Config>, chain_data_source_config: Option<&ChainDataSourceConfig>,
 	gossip_source_config: Option<&GossipSourceConfig>,
-	liquidity_source_config: Option<&LiquiditySourceConfig>, seed_bytes: [u8; 64],
-	runtime: Arc<Runtime>, logger: Arc<Logger>, kv_store: Arc<DynStore>,
+	liquidity_source_config: Option<&LiquiditySourceConfig>,
+	async_payments_role: Option<AsyncPaymentsRole>, seed_bytes: [u8; 64], runtime: Arc<Runtime>,
+	logger: Arc<Logger>, kv_store: Arc<DynStore>,
 ) -> Result<Node, BuildError> {
 	optionally_install_rustls_cryptoprovider();
 
@@ -1378,8 +1414,14 @@ fn build_with_store_internal(
 			100;
 	}
 
-	if config.async_payment_services_enabled {
-		user_config.accept_forwards_to_priv_channels = true;
+	if let Some(role) = async_payments_role {
+		match role {
+			AsyncPaymentsRole::Server => {
+				user_config.accept_forwards_to_priv_channels = true;
+				user_config.enable_htlc_hold = true;
+			},
+			AsyncPaymentsRole::Client => user_config.hold_outbound_htlcs_at_next_hop = true,
+		}
 	}
 
 	let message_router =
@@ -1452,17 +1494,32 @@ fn build_with_store_internal(
 	}
 
 	// Initialize the PeerManager
-	let onion_messenger: Arc<OnionMessenger> = Arc::new(OnionMessenger::new(
-		Arc::clone(&keys_manager),
-		Arc::clone(&keys_manager),
-		Arc::clone(&logger),
-		Arc::clone(&channel_manager),
-		message_router,
-		Arc::clone(&channel_manager),
-		Arc::clone(&channel_manager),
-		IgnoringMessageHandler {},
-		IgnoringMessageHandler {},
-	));
+	let onion_messenger: Arc<OnionMessenger> =
+		if let Some(AsyncPaymentsRole::Server) = async_payments_role {
+			Arc::new(OnionMessenger::new_with_offline_peer_interception(
+				Arc::clone(&keys_manager),
+				Arc::clone(&keys_manager),
+				Arc::clone(&logger),
+				Arc::clone(&channel_manager),
+				message_router,
+				Arc::clone(&channel_manager),
+				Arc::clone(&channel_manager),
+				IgnoringMessageHandler {},
+				IgnoringMessageHandler {},
+			))
+		} else {
+			Arc::new(OnionMessenger::new(
+				Arc::clone(&keys_manager),
+				Arc::clone(&keys_manager),
+				Arc::clone(&logger),
+				Arc::clone(&channel_manager),
+				message_router,
+				Arc::clone(&channel_manager),
+				Arc::clone(&channel_manager),
+				IgnoringMessageHandler {},
+				IgnoringMessageHandler {},
+			))
+		};
 	let ephemeral_bytes: [u8; 32] = keys_manager.get_secure_random_bytes();
 
 	// Initialize the GossipSource
@@ -1649,6 +1706,12 @@ fn build_with_store_internal(
 		},
 	};
 
+	let om_mailbox = if let Some(AsyncPaymentsRole::Server) = async_payments_role {
+		Some(Arc::new(OnionMessageMailbox::new()))
+	} else {
+		None
+	};
+
 	let (stop_sender, _) = tokio::sync::watch::channel(());
 	let (background_processor_stop_sender, _) = tokio::sync::watch::channel(());
 	let is_running = Arc::new(RwLock::new(false));
@@ -1681,6 +1744,8 @@ fn build_with_store_internal(
 		is_running,
 		is_listening,
 		node_metrics,
+		om_mailbox,
+		async_payments_role,
 	})
 }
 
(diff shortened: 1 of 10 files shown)


src/lib.rs
```diff
@@ -1491,6 +1497,11 @@ impl Node {
 				Error::PersistenceFailed
 			})
 	}
+
+	#[allow(missing_docs)]
+	pub fn om_mailbox_is_empty(&self) -> bool {
```
joostjager: I wanted to expose this to the integration test, but not sure how to do it without introducing an extra cfg flag

joostjager:
Rebased after merge of #621

src/config.rs
```diff
@@ -330,6 +330,8 @@ pub(crate) fn default_user_config(config: &Config) -> UserConfig {
 		user_config.channel_handshake_limits.force_announced_channel_preference = true;
 	}
 
+	user_config.enable_htlc_hold = true;
```
tnull: I think we only want to set this if `async_payment_services_enabled`?

src/event.rs
```diff
@@ -1491,22 +1497,32 @@ where
 
 				self.bump_tx_event_handler.handle_event(&bte).await;
 			},
-			LdkEvent::OnionMessageIntercepted { .. } => {
-				debug_assert!(false, "We currently don't support onion message interception, so this event should never be emitted.");
+			LdkEvent::OnionMessageIntercepted { peer_node_id, message } => {
+				self.om_mailbox.onion_message_intercepted(peer_node_id, message);
```
tnull: Should we still log an error and do nothing here if `!config.async_payment_services_enabled`?

src/builder.rs
```diff
@@ -1645,6 +1664,8 @@ fn build_with_store_internal(
 		},
 	};
 
+	let om_mailbox = Arc::new(OnionMessageMailbox::new());
```
tnull: Can we make this an `Option` that will only be set if `async_payment_services_enabled`?

src/lib.rs
tnull: Well, that's why they are integration / black box tests, they test the public API. Would be good to drop this method. If you want to test internal functionality, you'd need to write a unit test.

src/om_mailbox.rs
```diff
@@ -0,0 +1,107 @@
+use std::{
```
tnull: As always, please fix imports.

src/om_mailbox.rs
```diff
@@ -0,0 +1,107 @@
...
+	collections::{HashMap, VecDeque},
+	sync::Mutex,
+};
+
+use lightning::ln::msgs::OnionMessage;
+
+pub(crate) struct OnionMessageMailbox {
+	map: Mutex<HashMap<bitcoin::secp256k1::PublicKey, VecDeque<lightning::ln::msgs::OnionMessage>>>,
```
tnull: Please just import `PublicKey` and `OnionMessage`, as this type signature is huge.

src/om_mailbox.rs
```diff
@@ -0,0 +1,107 @@
...
+}
+
+impl OnionMessageMailbox {
+	const MAX_MESSAGES_PER_PEER: usize = 100;
```
tnull: Hmm, these 100/100 values seems relatively restrictive and still results in ~300 MB of OM data in memory, AFAIU. 

Maybe it would be worth to at least go for a 30/300 ratio, so that we allow for more potential peers?

src/om_mailbox.rs
```diff
@@ -0,0 +1,107 @@
...
+			queue.pop_front();
+		}
+		queue.push_back(message);
+
+		// Enforce a peers limit. If exceeded, evict the peer with the longest queue.
+		if map.len() > Self::MAX_PEERS {
+			let peer_to_remove = map
+				.iter()
+				.max_by_key(|(_, queue)| queue.len())
+				.map(|(peer, _)| peer.clone())
```
tnull: I think `PublicKey` is `Copy`.

src/om_mailbox.rs
```diff
@@ -0,0 +1,107 @@
...
+pub(crate) struct OnionMessageMailbox {
+	map: Mutex<HashMap<bitcoin::secp256k1::PublicKey, VecDeque<lightning::ln::msgs::OnionMessage>>>,
+}
+
+impl OnionMessageMailbox {
+	const MAX_MESSAGES_PER_PEER: usize = 100;
+	const MAX_PEERS: usize = 100;
+
+	pub fn new() -> Self {
+		Self { map: Mutex::new(HashMap::new()) }
```
tnull: Would it make sense to pre-alloc some `capacity` from the start to avoid constant reallocations?

src/om_mailbox.rs
```diff
@@ -0,0 +1,107 @@
...
+				.map(|(peer, _)| peer.clone())
+				.unwrap();
+
+			map.remove(&peer_to_remove);
+		}
+	}
+
+	pub(crate) fn onion_message_peer_connected(
+		&self, peer_node_id: bitcoin::secp256k1::PublicKey,
+	) -> Vec<OnionMessage> {
```
tnull, line 39: Can we have this return an `Option` instead of bubbling up an empty `Vec`?

src/om_mailbox.rs
tnull, line 1: This should probably also be moved to `payment/asynchronous` now?

src/lib.rs
joostjager: I don't think it is so binary, but I understand your pov. As far as I know, there isn't really another way to detect that the onion message is intercepted and waiting. Would you want to use a (blind) sleep instead?

src/om_mailbox.rs
joostjager: That seems unnecessarily loose? Basically there would be two return values (None and empty Vec) that are the same thing?

src/om_mailbox.rs
joostjager: As this is triggered by a network message, this seems to be a negligible optimization?

src/builder.rs
joostjager: I am not a fan of this type of optionality as there resources involved with an empty onion mailbox are negligible. But made the change.

src/event.rs
joostjager: Added trace log

src/lib.rs
joostjager: Using sleep 3s now.

src/om_mailbox.rs
joostjager: Any of those constant can be right or wrong, no strong opinion about it. Changed.

src/om_mailbox.rs
joostjager: Added MAX_PEERS capacity.

src/om_mailbox.rs
joostjager: Indeed. Made it `*peer`

src/om_mailbox.rs
joostjager: OM mailbox isn't specific to async payments. But moved anyway.

src/om_mailbox.rs
tnull: Fair enough.

src/om_mailbox.rs
tnull: Right now it is. But granted, at some point it may become handy for other LSP services once they support BOLT12. We'll see.

src/config.rs
```diff
@@ -181,6 +181,13 @@ pub struct Config {
 	pub route_parameters: Option<RouteParametersConfig>,
 	/// Whether to enable the static invoice service to support async payment reception for clients.
 	pub async_payment_services_enabled: bool,
+	/// If this is set to true, then if we as an often-offline payer receive a [`StaticInvoice`] to
```
tnull: Given the context here I don't fully grasp what this flag is for and the docs' lingo is very static-invoice/async-payment specific. Note that in LDK Node we *don't* want to just forward all the LDK config knobs but rather make opinionated decisions on sane defaults. Can we simplify the config options and docs so that a user who doesn't know anything about the feature can understand when to set or not set it? 

Do we ever want to enable both `async_payment_services_enabled` and `hold_outbound_htlcs_at_next_hop`? If not, how would we enforce only one is set? Also, given that the docs mention this is for unannounced nodes only, we might want to enforce this via the `may_announce_channel` check.

It would btw, also help if we could either make the two bools similarly named or move them to `AsyncPaymentService`/`AsyncPaymentClient` sub-configs (or even just allow to set them via the `Builder` as we do for the Liquidity service / clients) to make it clearer what they are for and how they should be used.

src/event.rs
```diff
@@ -1491,11 +1497,24 @@ where
...
+				}
 			},
-			LdkEvent::OnionMessagePeerConnected { .. } => {
-				debug_assert!(false, "We currently don't support onion message interception, so this event should never be emitted.");
+			LdkEvent::OnionMessagePeerConnected { peer_node_id } => {
+				if let Some(om_mailbox) = self.om_mailbox.as_ref() {
+					let messages = om_mailbox.onion_message_peer_connected(peer_node_id);
+
+					for message in messages {
+						let _ = self.onion_messenger.forward_onion_message(message, &peer_node_id);
```
tnull: Should we log errors at TRACE level here?

src/config.rs
joostjager: Discussed offline. Moved booleans into an enum on builder.

src/event.rs
joostjager: Added

src/builder.rs
```diff
@@ -704,6 +713,7 @@ impl NodeBuilder {
 			runtime,
 			logger,
 			Arc::new(vss_store),
+			self.async_payments_role,
```
tnull: Move this up  to the other config objects (`self...`)?

src/builder.rs
```diff
@@ -736,6 +746,7 @@ impl NodeBuilder {
 			runtime,
 			logger,
 			kv_store,
+			self.async_payments_role,
```
tnull: Move this to the other config objects (`self...`)?

src/builder.rs
```diff
@@ -241,6 +242,7 @@ pub struct NodeBuilder {
 	liquidity_source_config: Option<LiquiditySourceConfig>,
 	log_writer_config: Option<LogWriterConfig>,
 	runtime_handle: Option<tokio::runtime::Handle>,
+	async_payments_role: Option<AsyncPaymentsRole>,
```
tnull, line 252: Move this to the other config objects?

src/builder.rs
```diff
@@ -989,6 +1000,11 @@ impl ArcedNodeBuilder {
 		self.inner.write().unwrap().set_node_alias(node_alias).map(|_| ())
 	}
 
+	/// Sets the role of the node in an asynchronous payments context.
+	pub fn set_async_payments_role(&self, role: Option<AsyncPaymentsRole>) {
+		_ = self.inner.write().unwrap().set_async_payments_role(role)
```
tnull:
```suggestion
		self.inner.write().unwrap().set_async_payments_role(role);
```

src/builder.rs
```diff
@@ -544,6 +547,12 @@ impl NodeBuilder {
 		Ok(self)
 	}
 
+	/// Sets the role of the node in an asynchronous payments context.
+	pub fn set_async_payments_role(&mut self, role: Option<AsyncPaymentsRole>) -> &mut Self {
+		self.async_payments_role = role;
```
tnull, line 569: Should we check here that we only set `Server` if the `may_announce_channel` requirements are met (ie. set NodeAlias and listening addresses)?

src/builder.rs
```diff
@@ -1084,6 +1100,7 @@ fn build_with_store_internal(
 	gossip_source_config: Option<&GossipSourceConfig>,
 	liquidity_source_config: Option<&LiquiditySourceConfig>, seed_bytes: [u8; 64],
 	runtime: Arc<Runtime>, logger: Arc<Logger>, kv_store: Arc<DynStore>,
+	async_payments_role: Option<AsyncPaymentsRole>,
```
tnull: Move this to the other config objects?

src/builder.rs
```diff
@@ -1649,6 +1687,12 @@ fn build_with_store_internal(
 		},
 	};
 
+	let om_mailbox = if let Some(AsyncPaymentsRole::Server) = async_payments_role {
```
tnull, line 1709: nit: Using `Option::map` might be cleaner here?

src/builder.rs
```diff
@@ -544,6 +547,12 @@ impl NodeBuilder {
 		Ok(self)
 	}
 
+	/// Sets the role of the node in an asynchronous payments context.
```
tnull: Can we link the spec PR here and maybe also in other relevant docs so users have a pointer to understand what 'async payments' is referring to here?

src/config.rs
```diff
@@ -537,6 +534,17 @@ impl From<MaxDustHTLCExposure> for LdkMaxDustHTLCExposure {
 	}
 }
 
+#[derive(Debug, Clone, Copy)]
+/// The role of the node in an asynchronous payments context.
+pub enum AsyncPaymentsRole {
+	/// Node acts a client in an async payments context. This means that if possible, it will instruct its peers to hold
+	/// htlcs for it, so that it can go offline.
+	Client,
+	/// Node acts as a server in an async payments context. This means that it will hold async payments htlcs and onion
```
tnull:
```suggestion
	/// Node acts as a server in an async payments context. This means that it will hold async payments HTLCs and onion
```

src/config.rs
```diff
@@ -537,6 +534,17 @@ impl From<MaxDustHTLCExposure> for LdkMaxDustHTLCExposure {
 	}
 }
 
+#[derive(Debug, Clone, Copy)]
+/// The role of the node in an asynchronous payments context.
+pub enum AsyncPaymentsRole {
+	/// Node acts a client in an async payments context. This means that if possible, it will instruct its peers to hold
+	/// htlcs for it, so that it can go offline.
```
tnull:
```suggestion
	/// HTLCs for it, so that it can go offline.
```

src/config.rs
```diff
@@ -537,6 +534,17 @@ impl From<MaxDustHTLCExposure> for LdkMaxDustHTLCExposure {
 	}
 }
 
+#[derive(Debug, Clone, Copy)]
+/// The role of the node in an asynchronous payments context.
```
tnull: Please link the spec PR here, too.

src/lib.rs
```diff
@@ -126,6 +126,7 @@ pub use builder::BuildError;
 pub use builder::NodeBuilder as Builder;
 
 use chain::ChainSource;
+use config::AsyncPaymentsRole;
```
tnull: Please move this into the `config::` group below.

src/payment/asynchronous/om_mailbox.rs
```diff
@@ -0,0 +1,99 @@
+use std::collections::{HashMap, VecDeque};
+use std::sync::Mutex;
+
+use bitcoin::secp256k1::PublicKey;
+use lightning::ln::msgs::OnionMessage;
+
+pub(crate) struct OnionMessageMailbox {
+	map: Mutex<HashMap<bitcoin::secp256k1::PublicKey, VecDeque<lightning::ln::msgs::OnionMessage>>>,
```
tnull:
```suggestion
	map: Mutex<HashMap<PublicKey, VecDeque<OnionMessage>>>,
```

src/payment/asynchronous/om_mailbox.rs
```diff
@@ -0,0 +1,99 @@
...
+		if map.len() > Self::MAX_PEERS {
+			let peer_to_remove =
+				map.iter().max_by_key(|(_, queue)| queue.len()).map(|(peer, _)| *peer).unwrap();
+
+			map.remove(&peer_to_remove);
+		}
+	}
+
+	pub(crate) fn onion_message_peer_connected(
+		&self, peer_node_id: bitcoin::secp256k1::PublicKey,
```
tnull:
```suggestion
		&self, peer_node_id: PublicKey,
```

tests/common/mod.rs
```diff
@@ -310,6 +310,13 @@ pub(crate) fn setup_two_nodes(
 
 pub(crate) fn setup_node(
 	chain_source: &TestChainSource, config: TestConfig, seed_bytes: Option<Vec<u8>>,
+) -> TestNode {
+	setup_node_for_async_payments(chain_source, config, seed_bytes, None)
```
tnull, line 314: Hmm, it's a bit weird to call into the async-payment specific method always here, but probably fine by now. We have conflicting changes in open PRs I believe, so will be forced to revisit that with a proper refactor either way.

joostjager: I didn't want to add the async role to all the calls to `setup_node`. But open to suggestions.

tnull: Yeah, no, as mentioned, just leave as is for now, and then we can see where we land once we add more `setup_node` variants (soon).

src/builder.rs
joostjager: Assuming you mean above runtime

src/builder.rs
joostjager: Spec PR linked. Not sure about other docs that. Eventually the spec PR should cover it all?

src/builder.rs
joostjager: Added. And added `AsyncPaymentsConfigMismatch` build error for this.

src/builder.rs
joostjager: Now returning BuildError

src/builder.rs
joostjager: I don't think it works with map, because for `Client` I still need to return None?

src/builder.rs
```diff
@@ -547,10 +555,18 @@ impl NodeBuilder {
 		Ok(self)
 	}
 
-	/// Sets the role of the node in an asynchronous payments context.
-	pub fn set_async_payments_role(&mut self, role: Option<AsyncPaymentsRole>) -> &mut Self {
+	/// Sets the role of the node in an asynchronous payments context. See https://github.com/lightning/bolts/pull/1149
```
tnull: nit: The plain URLs yield some warnings during doc generation:

```
warning: this URL is not a hyperlink
   --> src/builder.rs:558:73
    |
558 |     /// Sets the role of the node in an asynchronous payments context. See https://github.com/lightning/bolts/pull/1149
    |                                                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: bare URLs are not automatically turned into clickable links
    = note: `#[warn(rustdoc::bare_urls)]` on by default
help: use an automatic link instead
    |
558 |     /// Sets the role of the node in an asynchronous payments context. See <https://github.com/lightning/bolts/pull/1149>
    |                                                                            +                                            +

warning: this URL is not a hyperlink
   --> src/config.rs:538:67
    |
538 | /// The role of the node in an asynchronous payments context. See https://github.com/lightning/bolts/pull/1149 for more
    |                                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: bare URLs are not automatically turned into clickable links
help: use an automatic link instead
    |
538 | /// The role of the node in an asynchronous payments context. See <https://github.com/lightning/bolts/pull/1149> for more
    |                                                                   +                                            +
```

Also, while you're at it, please put the `See ...` sentence in its own second paragraph (here and below) to improve doc rendering.

PR #633, opened 2025-09-12
Require both `KVStore` and `KVStoreSync` implementations, switch BP to be fully-async

tnull:
As an intermediary step towards making our IO fully async, we now require any store to implement both `KVStore` and `KVStoreSync`, which allows us to switch over to the fully-async background processor and take further migration steps bit-by-bit when we make more and more of the core codebase `async`.

To this end, we refactor `VssStore` and `SqliteStore` to implement `KVStore`

TODOs:

- [x] Implement `KVStore` for `TestStore` upstream to fix tests
- [x] Implement write-order tracking for the VSS `KVStore` implementation

.. draft until then.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ default = []
 #lightning-types = { version = "0.2.0" }
 #lightning-invoice = { version = "0.33.0", features = ["std"] }
 #lightning-net-tokio = { version = "0.1.0" }
-#lightning-persister = { version = "0.1.0" }
+#lightning-persister = { version = "0.1.0", features = ["tokio"] }
 #lightning-background-processor = { version = "0.1.0" }
 #lightning-rapid-gossip-sync = { version = "0.1.0" }
 #lightning-block-sync = { version = "0.1.0", features = ["rest-client", "rpc-client", "tokio"] }
@@ -44,7 +44,7 @@ default = []
 #lightning-types = { git = "https://github.com/lightningdevkit/rust-lightning", branch = "main" }
 #lightning-invoice = { git = "https://github.com/lightningdevkit/rust-lightning", branch = "main", features = ["std"] }
 #lightning-net-tokio = { git = "https://github.com/lightningdevkit/rust-lightning", branch = "main" }
-#lightning-persister = { git = "https://github.com/lightningdevkit/rust-lightning", branch = "main" }
+#lightning-persister = { git = "https://github.com/lightningdevkit/rust-lightning", branch = "main", features = ["tokio"] }
 #lightning-background-processor = { git = "https://github.com/lightningdevkit/rust-lightning", branch = "main" }
 #lightning-rapid-gossip-sync = { git = "https://github.com/lightningdevkit/rust-lightning", branch = "main" }
 #lightning-block-sync = { git = "https://github.com/lightningdevkit/rust-lightning", branch = "main", features = ["rest-client", "rpc-client", "tokio"] }
@@ -56,7 +56,7 @@ lightning = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "
 lightning-types = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "3e21ba37a133977d4247e86f25df983b39326994" }
 lightning-invoice = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "3e21ba37a133977d4247e86f25df983b39326994", features = ["std"] }
 lightning-net-tokio = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "3e21ba37a133977d4247e86f25df983b39326994" }
-lightning-persister = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "3e21ba37a133977d4247e86f25df983b39326994" }
+lightning-persister = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "3e21ba37a133977d4247e86f25df983b39326994", features = ["tokio"] }
 lightning-background-processor = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "3e21ba37a133977d4247e86f25df983b39326994" }
 lightning-rapid-gossip-sync = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "3e21ba37a133977d4247e86f25df983b39326994" }
 lightning-block-sync = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "3e21ba37a133977d4247e86f25df983b39326994", features = ["rest-client", "rpc-client", "tokio"] }
@@ -68,7 +68,7 @@ lightning-macros = { git = "https://github.com/lightningdevkit/rust-lightning", 
 #lightning-types = { path = "../rust-lightning/lightning-types" }
 #lightning-invoice = { path = "../rust-lightning/lightning-invoice", features = ["std"] }
 #lightning-net-tokio = { path = "../rust-lightning/lightning-net-tokio" }
-#lightning-persister = { path = "../rust-lightning/lightning-persister" }
+#lightning-persister = { path = "../rust-lightning/lightning-persister", features = ["tokio"] }
 #lightning-background-processor = { path = "../rust-lightning/lightning-background-processor" }
 #lightning-rapid-gossip-sync = { path = "../rust-lightning/lightning-rapid-gossip-sync" }
 #lightning-block-sync = { path = "../rust-lightning/lightning-block-sync", features = ["rest-client", "rpc-client", "tokio"] }
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -31,7 +31,7 @@ use lightning::routing::scoring::{
 };
 use lightning::sign::{EntropySource, NodeSigner};
 use lightning::util::persist::{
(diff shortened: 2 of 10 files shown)


src/io/vss_store.rs
```diff
@@ -54,29 +63,76 @@ impl VssStore {
...
+	}
+
+	// Same logic as for the obfuscated keys below, but just for locking, using the plaintext keys
+	fn build_locking_key(
+		&self, primary_namespace: &str, secondary_namespace: &str, key: &str,
+	) -> String {
+		if primary_namespace.is_empty() {
+			key.to_owned()
+		} else {
+			format!("{}#{}#{}", primary_namespace, secondary_namespace, key)
```
joostjager, line 73: Is it necessary to make a string out of this? It doesn't need to be mapped to a filename like for fs_store, so maybe it can also simply be a tuple?

src/io/vss_store.rs
```diff
@@ -135,6 +216,9 @@ struct VssStoreInner {
 	store_id: String,
 	storable_builder: StorableBuilder<RandEntropySource>,
 	key_obfuscator: KeyObfuscator,
+	// Per-key locks that ensures that we don't have concurrent writes to the same file. The lock also encapsulates the
```
joostjager: Not a file

src/io/vss_store.rs
```diff
@@ -215,87 +304,100 @@ impl VssStoreInner {
...
-			let msg = format!(
-				"Failed to decode data read from key {}/{}/{}: {}",
-				primary_namespace, secondary_namespace, key, e
-			);
-			Error::new(ErrorKind::Other, msg)
-		})?;
-
-		Ok(self.storable_builder.deconstruct(storable)?.0)
+
+		self.execute_locked_read(locking_key, async move || {
```
joostjager: I think we don't need the lock for reading?

src/io/vss_store.rs
tnull: I think so, as the `HashMap` needs to hold an owned value. We could have it be a `(String, String, String)`, but that's worse.

src/io/vss_store.rs
tnull: Dropped.

src/io/vss_store.rs
tnull: Done.

src/io/vss_store.rs
joostjager: Is that worse? The string concatenation looks a bit unnecessary. Or make it a struct that is used as the key?

tnull: > Is that worse? The string concatenation looks a bit unnecessary. Or make it a struct that is used as the key?

It at least requires three individual allocations instead of one? I.e. more clutter on the heap, and probably also some slowdown? 

FWIW, I mirrored what we do for the obfuscated key. Unfortunately it's not super easy to just reuse that.

joostjager: Not ideal indeed, the heap allocs

src/io/sqlite_store/mod.rs
```diff
@@ -60,6 +63,76 @@ impl SqliteStore {
...
+		})
+	}
+	fn write(
+		&self, primary_namespace: &str, secondary_namespace: &str, key: &str, buf: Vec<u8>,
+	) -> Pin<Box<dyn Future<Output = Result<(), io::Error>> + Send>> {
+		let primary_namespace = primary_namespace.to_string();
+		let secondary_namespace = secondary_namespace.to_string();
+		let key = key.to_string();
+		let inner = Arc::clone(&self.inner);
+		let fut = tokio::task::spawn_blocking(move || {
```
joostjager, line 120: During final look, I am now again wondering if this function is actually preserving order?

src/io/sqlite_store/mod.rs
```diff
@@ -60,6 +63,76 @@ impl SqliteStore {
...
+			inner.read_internal(&primary_namespace, &secondary_namespace, &key)
+		});
+		Box::pin(async move {
+			fut.await.unwrap_or_else(|e| {
+				let msg = format!("Failed to IO operation due join error: {}", e);
+				Err(io::Error::new(io::ErrorKind::Other, msg))
+			})
+		})
+	}
+	fn write(
```
joostjager, line 111: newlines above, here and elsewhere

src/io/sqlite_store/mod.rs
tnull: Hmm, good point, it seems it would depend on how tokio exactly schedules the blocking tasks. I considered a few other options, but now simply also followed the tried and true 'write version locking' approach here, as we already use that in VSS and FS stores.

src/io/sqlite_store/mod.rs
tnull: Fixed.

src/io/sqlite_store/mod.rs
joostjager: Hmm but isn't this completely unnecessary because sqlite has its own global lock? With FS and VSS there is the actual possibility of parallel execution.

tnull: > Hmm but isn't this completely unnecessary because sqlite has its own global lock? With FS and VSS there is the actual possibility of parallel execution.

Well, I first thought so, too, but I think the issue is that tokio gives no guarantee in which order spawned tasks are executed/scheduled. I.e., AFAICT it could happen that we spawn to writes `w1, w2` but the task for `w2` gets polled first, acquiring the `Mutex` first. Same goes for the case where multiple writes wait on the same connection lock: say `w1` currently holds the `Mutex` and two more writes `w2, w3` would get get queued, AFAIU there is no guarantee that when `w1` drops the lock `w2` always acquires the lock next. 

TLDR: it seems we unfortunately need to do the `version` dance here, too. Maybe there's an easier mechanism in the SQLite case (e.g., `prepare` should technically take care of that, but we can't lock the connection `Mutex` outside of the spawned task as the guard is not `Send`) we could lean on to guarantee the ordered writes, but applying the same approach seemed simplest for now?

joostjager: I just refuse to believe that we need that version dance really for this problem. Isn't `block_in_place` meant for this? If we ensure that the write has happened before the fn returns, it doesn't matter that during that execution there may be other writes that get processed in a certain order?

tnull: Hmm? How do you imagine `block_in_place` to work here? `block_in_place` takes a future and drives it to completion, i.e., makes it a blocking operation. For the async `KVStore` we however exactly need a future, not a blocking operation, which is what `spawn_blocking` does for us.

joostjager: Does it? I saw

`pub fn block_in_place<F, R>(f: F) -> R where F: FnOnce() -> R`

tnull: > Does it? I saw
> 
> `pub fn block_in_place<F, R>(f: F) -> R where F: FnOnce() -> R`

Sorry, please replace `block_in_place` with `block_on` above. `block_in_place` is simply a wrapper that spawns a blocking task on the outer runtime context so that the inner `block_on` call doesn't starve (that is, if it is indeed called on the same runtime, which it isn't always).

joostjager: Discussed offline that even though `block_in_place` may work with caveats, we do prefer to implement sqlite async. And that for that to work, we need the versioning.

src/io/sqlite_store/mod.rs
```diff
@@ -54,7 +60,31 @@ impl SqliteStore {
...
-		Ok(Self { inner })
+		let next_write_version = AtomicU64::new(1);
+		Ok(Self { inner, next_write_version })
+	}
+
+	fn build_locking_key(
+		&self, primary_namespace: &str, secondary_namespace: &str, key: &str,
+	) -> String {
+		if primary_namespace.is_empty() {
+			key.to_owned()
```
joostjager: And secondary ns?

Is this actually used without a namespace?

src/io/sqlite_store/mod.rs
```diff
@@ -354,6 +436,7 @@ impl SqliteStoreInner {
 			io::Error::new(io::ErrorKind::Other, msg)
 		})?;
 		Ok(())
+        })
```
joostjager: Indent formatting issue?

src/io/sqlite_store/mod.rs
tnull: > And secondary ns?

The secondary namespace is allowed to be empty. 

> Is this actually used without a namespace?

Yes, keys like `manager` are residing in the top level for historical reasons (i.e., filesystem store compatibility). 
That said, for the locking key there is no need to follow this scheme exactly, I just copied what we did elsewhere. Now simplified the code a bit.

src/io/sqlite_store/mod.rs
tnull: Fixed, weird `rustfmt` didn't fix this automatically.

joostjager: Indeed. I also tried it locally, but didn't want to format it.

joostjager (approved):
I have to admit that I am a bit nervous about the versioning code, because it is involved in every persistence action. After landing this PR, do we need to do anything extra with regards to testing?

PR #634, opened 2025-09-16
Adapt channel balance reporting to use confirmed candidate

joostjager:
With splicing now implemented, a channel may have multiple holder commitment transactions and corresponding balance candidates. ldk-node now reports the confirmed balance candidate rather than a single static balance, ensuring the exposed value matches the channel's onchain state. Other candidate balances remain internal for now.

### DIFF
--- a/src/balance.rs
+++ b/src/balance.rs
@@ -73,7 +73,8 @@ pub struct BalanceDetails {
 pub enum LightningBalance {
 	/// The channel is not yet closed (or the commitment or closing transaction has not yet
 	/// appeared in a block). The given balance is claimable (less on-chain fees) if the channel is
-	/// force-closed now.
+	/// force-closed now. Values do not take into account any pending splices and are only based
+	/// on the confirmed state of the channel.
 	ClaimableOnChannelClose {
 		/// The identifier of the channel this balance belongs to.
 		channel_id: ChannelId,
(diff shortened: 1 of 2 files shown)


src/balance.rs
```diff
@@ -224,21 +224,25 @@ impl LightningBalance {
...
-				amount_satoshis,
-				transaction_fee_satoshis,
+				balance_candidates,
+				confirmed_balance_candidate_index,
 				outbound_payment_htlc_rounded_msat,
 				outbound_forwarded_htlc_rounded_msat,
 				inbound_claiming_htlc_rounded_msat,
 				inbound_htlc_rounded_msat,
+			} => {
+				let balance = balance_candidates.get(confirmed_balance_candidate_index).unwrap();
```
tnull, line 236: Can you add a `// unwrap safety: ` comment here to describe why this `unwrap` is always safe?

src/balance.rs
```diff
@@ -224,21 +224,25 @@ impl LightningBalance {
...
+
+				Self::ClaimableOnChannelClose {
```
tnull, line 238: Mind updating the docs on our `ClaimableOnChannelClose` to note that we're just exposing this balance in a splicing scenario?

joostjager: Added that values are based on the confirmed channel state, ignoring pending splices. I don't think it is what you suggest, that we are just exposing in a splicing scenario?

src/balance.rs
```diff
@@ -73,7 +73,8 @@ pub struct BalanceDetails {
 pub enum LightningBalance {
 	/// The channel is not yet closed (or the commitment or closing transaction has not yet
 	/// appeared in a block). The given balance is claimable (less on-chain fees) if the channel is
-	/// force-closed now.
+	/// force-closed now. Values do not take into account any pending splices and are only based
+	/// on the confirmed state of the channel.
```
jkczyz, line 77: I don't believe this is the case. It *will* take into account a pending splice if it is confirmed but not locked (i.e., doesn't have sufficient number of confirmations).

src/balance.rs
jkczyz: Noted above that this is not the case. Confirmed but not locked splices are taken into account.

Also, `LdkBalance::claimable_amount_satoshis` uses a different way of calculating this when the index is `0`. But maybe that is just to avoid the `unwrap`? (cc:: @wpaulino)

src/balance.rs
jkczyz: BTW, let us know if there is a way to update the docs to make this clear.

src/balance.rs
wpaulino: Yeah when the index is 0, either we don't have a pending splice, or we do but none of the splice transactions has confirmed. In the latter case, we still want to report some indication of the unconfirmed splice balance.

PR #637, opened 2025-09-16
Add LNURL-auth support

benthecarman:
Implements LNURL-auth (LUD-04) specification for secure, privacy-preserving authentication with Lightning services using domain-specific key derivation.

I used LUD-05 for deriving the keys to mimic what we do for VSS auth.

I was able to auth with stacker.news with this so it works!

### DIFF
--- a/bindings/ldk_node.udl
+++ b/bindings/ldk_node.udl
@@ -166,6 +166,8 @@ interface Node {
 	UnifiedPayment unified_payment();
 	LSPS1Liquidity lsps1_liquidity();
 	[Throws=NodeError]
+	void lnurl_auth(string lnurl);
+	[Throws=NodeError]
 	void connect(PublicKey node_id, SocketAddress address, boolean persist);
 	[Throws=NodeError]
 	void disconnect(PublicKey node_id);
@@ -364,6 +366,9 @@ enum NodeError {
 	"InvalidBlindedPaths",
 	"AsyncPaymentServicesDisabled",
 	"HrnParsingFailed",
+	"LnurlAuthFailed",
+	"LnurlAuthTimeout",
+	"InvalidLnurl",
 };
 
 dictionary NodeStatus {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -96,6 +96,7 @@ pub mod graph;
 mod hex_utils;
 pub mod io;
 pub mod liquidity;
+mod lnurl_auth;
 pub mod logger;
 mod message_handler;
 pub mod payment;
@@ -124,7 +125,8 @@ pub use builder::NodeBuilder as Builder;
 use chain::ChainSource;
 use config::{
 	default_user_config, may_announce_channel, AsyncPaymentsRole, ChannelConfig, Config,
-	NODE_ANN_BCAST_INTERVAL, PEER_RECONNECTION_INTERVAL, RGS_SYNC_INTERVAL,
+	LNURL_AUTH_TIMEOUT_SECS, NODE_ANN_BCAST_INTERVAL, PEER_RECONNECTION_INTERVAL,
+	RGS_SYNC_INTERVAL,
 };
 use connection::ConnectionManager;
 pub use error::Error as NodeError;
@@ -148,6 +150,7 @@ use lightning::util::persist::KVStoreSync;
 use lightning::util::wallet_utils::Wallet as LdkWallet;
 use lightning_background_processor::process_events_async;
 use liquidity::{LSPS1Liquidity, LiquiditySource};
+use lnurl_auth::LnurlAuth;
 use logger::{log_debug, log_error, log_info, log_trace, LdkLogger, Logger};
 use payment::asynchronous::om_mailbox::OnionMessageMailbox;
 use payment::asynchronous::static_invoice_store::StaticInvoiceStore;
@@ -220,6 +223,7 @@ pub struct Node {
 	scorer: Arc<Mutex<Scorer>>,
 	peer_store: Arc<PeerStore<Arc<Logger>>>,
 	payment_store: Arc<PaymentStore>,
+	lnurl_auth: Arc<LnurlAuth>,
 	is_running: Arc<RwLock<bool>>,
 	node_metrics: Arc<RwLock<NodeMetrics>>,
 	om_mailbox: Option<Arc<OnionMessageMailbox>>,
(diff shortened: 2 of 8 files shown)


bindings/ldk_node.udl
tnull, line 1: We probably also want to expose `Node::lnurl_auth` in bindings here?

benthecarman: added!

Camillarhi:
Thanks for this. I also tested this with https://stacker.news and https://satsback.com, and I was authenticated successfully on both.

src/lib.rs
```diff
@@ -187,7 +205,9 @@ pub struct Node {
 	scorer: Arc<Mutex<Scorer>>,
 	peer_store: Arc<PeerStore<Arc<Logger>>>,
 	payment_store: Arc<PaymentStore>,
+	lnurl_auth: LnurlAuth,
```
tnull: Let's just hold an `Arc`, as cloning that is at least cheaper than always cloning the whole struct below?

src/lib.rs
```diff
@@ -932,6 +939,14 @@ impl Node {
 		))
 	}
 
+	/// Authenticates the user via [LNURL-auth] for the given LNURL string.
+	///
+	/// [LNURL-auth]: https://github.com/lnurl/luds/blob/luds/04.md
+	pub fn lnurl_auth(&self, lnurl: String) -> Result<(), Error> {
+		let auth = self.lnurl_auth.clone();
+		self.runtime.block_on(async move { auth.authenticate(&lnurl).await })
```
tnull: We def. need to add a `tokio::timeout` to make sure we're not indefinitely keeping a blocking task around. Could be added here, or below around the actual calls to remote that could hang indefinitely if something goes wrong. Maybe here at the top level might even be safer.

benthecarman: added

src/lib.rs
benthecarman: done

benthecarman:
Sorry  forgot about this PR. Rebased and fixed for comments

src/lib.rs
```diff
@@ -1004,6 +1007,26 @@ impl Node {
 		))
 	}
 
+	/// Authenticates the user via [LNURL-auth] for the given LNURL string.
+	///
+	/// [LNURL-auth]: https://github.com/lnurl/luds/blob/luds/04.md
+	pub fn lnurl_auth(&self, lnurl: String) -> Result<(), Error> {
+		let auth = self.lnurl_auth.clone();
```
tnull: nit: 
```suggestion
		let auth = Arc::clone(&self.lnurl_auth);
```
is preferable.

src/lnurl_auth.rs
```diff
@@ -0,0 +1,224 @@
...
+use bitcoin::secp256k1::{Message, Secp256k1, SecretKey};
+use lightning::util::logger::Logger as LdkLogger;
+
+use bitcoin::bech32;
+use reqwest::Client;
+use serde::{Deserialize, Serialize};
+use std::sync::Arc;
+
+const LUD13_MESSAGE: &str = "DO NOT EVER SIGN THIS TEXT WITH YOUR PRIVATE KEYS! IT IS ONLY USED FOR DERIVATION OF LNURL-AUTH HASHING-KEY, DISCLOSING ITS SIGNATURE WILL COMPROMISE YOUR LNURL-AUTH IDENTITY AND MAY LEAD TO LOSS OF FUNDS!";
+pub(crate) const LNURL_AUTH_TIMEOUT_SECS: u64 = 15;
```
tnull: Let's move this next to the other `TIMEOUT` values in `config.rs`.

src/lnurl_auth.rs
```diff
@@ -0,0 +1,224 @@
...
+/// while allowing consistent authentication across sessions.
+#[derive(Clone)]
+pub struct LnurlAuth {
+	hashing_key: SecretKey,
+	client: Client,
+	logger: Arc<Logger>,
+}
+
+impl LnurlAuth {
+	pub(crate) fn new(hashing_key: SecretKey, logger: Arc<Logger>) -> Self {
```
tnull: Why do we need this constructor if it's only ever used through `from_keys_manager`? Can we merge them and just call it `new`, given that it takes a few different parameters?

src/lnurl_auth.rs
```diff
@@ -0,0 +1,224 @@
...
+
+		let lnurl_auth_url = String::from_utf8(bytes).map_err(|e| {
+			log_error!(self.logger, "Failed to convert LNURL bytes to string: {e}");
+			Error::InvalidLnurl
+		})?;
+
+		log_debug!(self.logger, "Starting LNURL-auth process for URL: {lnurl_auth_url}");
+
+		// Parse the URL to extract domain and parameters
+		let url = reqwest::Url::parse(&lnurl_auth_url).map_err(|e| {
```
tnull: We're about to drop our `reqwest` dependency soon. Can we avoid leaning on `reqwest` or `Url` and rather do parsing and some validation manually?

tnull: Actually, let's hold on that for a day or two. I now started a new crate for URL parsing with minimal dependencies so we don't have to reimplement the same logic over and over again. Will let you know once that's in a usable state.

benthecarman: addressed other comments, will leave this for now

tnull: @benthecarman Alright, excuse the delay here! By now we shipped `bitreq::Url`. Please bump the `bitreq` dependency in this PR to 0.3.4 and switch to use that `Url` type for parsing!

benthecarman: lnurl auth over bitreq :rocket:

src/lib.rs
```diff
@@ -1004,6 +1008,26 @@ impl Node {
 		))
 	}
 
+	/// Authenticates the user via [LNURL-auth] for the given LNURL string.
+	///
```
tnull, line 1013: Can you document here that we're using LUD-13 here, and **not** LUD-05? In particular, this might be important as the VSS-JWT-LNURL-auth is using LUD-05, i.e., uses incompatible signatures derived from the seed rather than `signmessage`.

benthecarman: Would it be better to switch to LUD-05 then?

tnull: Hmm, not sure? I don't see us dropping the implementation in VSS client itself any time soon. And LUD-13 is  a bit more expected for a lightning node. We mostly used LUD-5 in `vss-client` as there we didn't have access to `sign_message`.

benthecarman: switched it to lud-05, makes more sense that we use the same scheme as vss auth

PR #641, opened 2025-09-19
Fix wait_for_tx/outpoint exponential backoff

joostjager:
Backoff wasn't actually working and polling would happen without any delay at all.

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -437,32 +437,31 @@ pub(crate) fn wait_for_block<E: ElectrumApi>(electrs: &E, min_height: usize) {
 }
 
 pub(crate) fn wait_for_tx<E: ElectrumApi>(electrs: &E, txid: Txid) {
-	let mut tx_res = electrs.transaction_get(&txid);
-	loop {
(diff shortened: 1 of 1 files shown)


tests/common/mod.rs
```diff
@@ -437,16 +437,10 @@ pub(crate) fn wait_for_block<E: ElectrumApi>(electrs: &E, min_height: usize) {
 }
 
 pub(crate) fn wait_for_tx<E: ElectrumApi>(electrs: &E, txid: Txid) {
-	let mut tx_res = electrs.transaction_get(&txid);
```
tnull, line 440: Maybe it's worth to keep the pattern to avoid the unnecessary ping if the result is immediately available.

joostjager: Ok. What is the ping for actually?

tnull: See https://docs.rs/electrum-client/latest/electrum_client/trait.ElectrumApi.html#tymethod.ping

`ping` might trigger `electrs` to process new data before we actually attempt the lookup.

PR #644, opened 2025-09-22
Listen on all provided addresses 

### LINKED ISSUE #318: Failed to bind should not panic
If you run on a port that is already in use, the following will occur:

```
thread 'tokio-runtime-worker' panicked at src/lib.rs:513:45:
Failed to bind to listen address/port - is something else already listening on it?
signal: aborted (core dumped)
```
It would be nice instead to return an error message

joostjager:
Previously ldk-node would start binding after the first successful bind to an address.

Also fixes #318

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -195,7 +195,7 @@ pub(crate) fn random_storage_path() -> PathBuf {
 
 pub(crate) fn random_port() -> u16 {
 	let mut rng = thread_rng();
-	rng.gen_range(5000..65535)
+	rng.gen_range(5000..32768)
 }
 
 pub(crate) fn random_listening_addresses() -> Vec<SocketAddress> {


src/lib.rs
```diff
@@ -325,45 +321,61 @@ impl Node {
...
-
-				loop {
-					let peer_mgr = Arc::clone(&peer_manager_connection_handler);
-					tokio::select! {
-						_ = stop_listen.changed() => {
-							log_debug!(
-								listening_logger,
-								"Stopping listening to inbound connections."
+			let logger = Arc::clone(&listening_logger);
+			let listeners = self.runtime.block_on(async move {
```
joostjager, line 313: Need to block if we want to return an error.

src/lib.rs
```diff
@@ -203,7 +202,6 @@ pub struct Node {
 	peer_store: Arc<PeerStore<Arc<Logger>>>,
 	payment_store: Arc<PaymentStore>,
 	is_running: Arc<RwLock<bool>>,
-	is_listening: Arc<AtomicBool>,
```
tnull, line 192: Huh, please don't drop the `is_listening` field, we also support nodes that don't allow *any* inbound connections and want to indicate whether we're listening or not to a port or not here.

src/lib.rs
tnull: Hmm, I see the point. I guess we need to if we want to abort on error for now, but we'll probably make `start` async soonish anyways.

src/lib.rs
```diff
@@ -325,45 +321,61 @@ impl Node {
...
+								log_debug!(
+									logger,
+									"Stopping listening to inbound connections."
+								);
+								break;
+							}
+							res = listener.accept() => {
+								let tcp_stream = res.unwrap().0;
+								let peer_mgr = Arc::clone(&peer_mgr);
+								tokio::spawn(async move {
```
tnull: Mhh, looks pre-existing, but we should probably also use `spawn_cancellable_background_task` for this.

src/lib.rs
joostjager: Alternative is to use non-tokio binds here.

src/lib.rs
```diff
@@ -325,45 +321,61 @@ impl Node {
...
-								"Stopping listening to inbound connections."
+			let logger = Arc::clone(&listening_logger);
+			let listeners = self.runtime.block_on(async move {
+				let mut listeners = Vec::new();
+
+				// Try to bind to all addresses
+				for addr in &*bind_addrs {
+					match tokio::net::TcpListener::bind(addr).await {
+						Ok(listener) => {
+							log_trace!(logger, "Bound to {}", addr);
```
tnull: Can we make this message a bit more verbose?

src/lib.rs
joostjager: Discussed offline. This would just be a reflection of the config set.

src/lib.rs
joostjager: Quite vague indeed. Made it `Listener bound to {}`

src/lib.rs
joostjager: Kept it in, but dropped the polling for it which is no longer necessary.

src/lib.rs
```diff
@@ -675,7 +688,8 @@ impl Node {
 	/// Returns the status of the [`Node`].
 	pub fn status(&self) -> NodeStatus {
 		let is_running = *self.is_running.read().unwrap();
-		let is_listening = self.is_listening.load(Ordering::Acquire);
+		let is_listening =
```
tnull: Hmm, okay, I think I have to eat my words as this is odd indeed now. Feel free to drop the field if you still prefer it!

joostjager: Ok, dropping it again.

joostjager:
Now that we established that in any case we need to pick random ports from the lower half, I've also added that commit here.

src/lib.rs
joostjager: Done

PR #647, opened 2025-09-23
Fix Invalid Pending Payments and Strengthen RBF Tests

moisesPompilio:
Pending payments that are invalidated or removed from the mempool are now marked as failed. Currently, **“failed”** simply means the transaction is no longer in **BDK’s** canonical transaction list while still pending in our DB. It could later return to a pending or confirmed state if it reappears. More precise context for these **“failed”** states will be possible once **bdk_wallet**#[257](https://github.com/bitcoindevkit/bdk_wallet/pull/257) is merged, as that adds richer uncanonica transaction status details.

**Additional improvements:**
- Property-based tests simulating multiple RBF attempts, random confirmations, and chain reorgs to ensure correct handling when any RBF transaction confirms.
- `generate_block_and_insert_transactions` to mine blocks with chosen transactions, enabling tests where either the original or a replacement transaction is confirmed.
- Refactored node setup in tests with a `new_node` helper for clarity.

**Note:** This PR partially addresses issue #452. I believe it does not conflict with PR #628, as this change is more general. It handles any transaction that is no longer canonical in BDK, not just RBF replacements. This fixes the problem where such transactions would remain pending indefinitely due to lack of further updates, by marking them as failed. PR #628, on the other hand, specifically handles RBFs initiated by the node.

### DIFF
--- a/tests/reorg_test.rs
+++ b/tests/reorg_test.rs
@@ -1,20 +1,22 @@
 mod common;
-use std::collections::HashMap;
-
-use bitcoin::Amount;
+use bitcoin::{Amount, ScriptBuf};
 use ldk_node::payment::{PaymentDirection, PaymentKind};
 use ldk_node::{Event, LightningBalance, PendingSweepBalance};
-use proptest::prelude::prop;
-use proptest::proptest;
+use proptest::strategy::Strategy;
+use proptest::strategy::ValueTree;
+use proptest::{prelude::prop, proptest};
+use std::collections::{HashMap, HashSet};
 
 use crate::common::{
-	expect_event, generate_blocks_and_wait, invalidate_blocks, open_channel,
-	premine_and_distribute_funds, setup_bitcoind_and_electrsd, setup_node, wait_for_outpoint_spend,
-	TestChainSource,
+	bump_fee_and_broadcast, distribute_funds_unconfirmed, expect_event,
+	generate_block_and_insert_transactions, generate_blocks_and_wait, invalidate_blocks,
+	open_channel, premine_and_distribute_funds, premine_blocks, prepare_rbf,
+	setup_bitcoind_and_electrsd, setup_node, wait_for_outpoint_spend, TestChainSource,
 };
 
 proptest! {
 	#![proptest_config(proptest::test_runner::Config::with_cases(5))]
+
 	#[test]
 	fn reorg_test(reorg_depth in 1..=6usize, force_close in prop::bool::ANY) {
 		let (bitcoind, electrsd) = setup_bitcoind_and_electrsd();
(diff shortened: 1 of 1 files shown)


src/wallet/mod.rs
```diff
@@ -178,6 +179,9 @@ where
 					} else {
 						PaymentStatus::Pending
 					};
+					if payment_status == PaymentStatus::Pending {
```
Camillarhi: Thanks for the PR. Since `PaymentStatus::Pending` is only assigned in the else branch above, we can move the `unconfirmed_txid.push(id)` call directly into that block. This avoids the extra comparison afterward and makes the logic a bit more concise and self-contained.

moisesPompilio: Done!

Camillarhi: Great! Thanks

tests/common/mod.rs
```diff
@@ -301,6 +301,11 @@ pub(crate) fn setup_two_nodes(
 	(node_a, node_b)
 }
 
+pub(crate) fn new_node(chain_source: &TestChainSource, anchor_channels: bool) -> TestNode {
```
tnull: The change from the `config_node` macro to this method makes sense to me, but maybe we want to align names here?

Should this rather be renamed to `setup_node`, while renaming the current `setup_node` to `setup_node_from_config`?

tests/common/mod.rs
```diff
@@ -13,14 +13,15 @@ pub(crate) mod logging;
...
 use std::env;
 use std::path::PathBuf;
+use std::str::FromStr;
 use std::sync::{Arc, RwLock};
 use std::time::Duration;
 
-use bitcoin::hashes::hex::FromHex;
 use bitcoin::hashes::sha256::Hash as Sha256;
-use bitcoin::hashes::Hash;
+use bitcoin::hashes::{hex::FromHex, Hash};
```
tnull: nit: Please group imports at the module level.

tests/common/mod.rs
```diff
@@ -390,6 +391,136 @@ pub(crate) fn setup_node_for_async_payments(
...
+	bitcoind: &BitcoindClient, electrs: &E, txs: &[Transaction],
+) {
+	let _ = bitcoind.create_wallet("ldk_node_test");
+	let _ = bitcoind.load_wallet("ldk_node_test");
+	let blockchain_info = bitcoind.get_blockchain_info().expect("failed to get blockchain info");
+	let cur_height = blockchain_info.blocks;
+	let address = bitcoind.new_address().expect("failed to get new address");
+
+	let request_block_template =
+		corepc_node::TemplateRequest { rules: vec![electrsd::corepc_node::TemplateRules::Segwit] };
```
tnull: nit: Please avoid the full type qualifiers, rather import the types via `use` above.

tests/common/mod.rs
```diff
@@ -390,6 +391,136 @@ pub(crate) fn setup_node_for_async_payments(
 	node
 }
 
+pub(crate) fn generate_block_and_insert_transactions<E: ElectrumApi>(
```
tnull, line 405: I wonder if it would be simpler to

1. Put the RBF into the mempool
2. Sync the wallet so it registers
3. Mine a block
4. Call Bitcoin Core's `invalidateblock` to discard the tip and the RBF
5. Put the original TX into the mempool
6. Sync the wallet again

If I'm not mistaken that would save us all of this boilerplate?

src/wallet/mod.rs
```diff
@@ -219,6 +222,31 @@ impl Wallet {
...
+		});
+		if failed_payment.is_empty() {
+			return Ok(());
+		}
+
+		for failed_payment in &failed_payment {
+			let id = failed_payment.id;
+			let payment_update = PaymentDetailsUpdate {
+				id,
+				status: Some(PaymentStatus::Failed),
```
tnull: Hmm, not convinced we should do this, as on a semantic level there is no such thing as a 'failed' onchain payment. Valid transactions are forever pending as they could always be rebroadcasted even after being evicted from the local mempool. That is why we currently only track them as `Pending` or `Confirmed` (once they are irrevocably confirmed).

moisesPompilio: I understand your point. When a transaction leaves the mempool due to a low fee, it can still be valid and eventually confirmed. The issue arises with **RBF**, since only one of the conflicting transactions can be confirmed, making the others effectively **double-spent** and thus **“failed”**. The challenge is that **BDK** doesn’t currently provide a way to distinguish between these cases.

My main motivation is that keeping transactions evicted from the mempool marked as `Pending` can lead to **UX** confusion, especially with RBF. For example, if a user broadcasts a transaction and then later replaces it via RBF, they would see two onchain payments, one that gets confirmed and another that never will. Leaving the latter as Pending would incorrectly suggest it’s still valid. Maybe having an additional status like `Dropped` or `Evicted` could better represent transactions that are no longer **canonical** in BDK?

tests/common/mod.rs
moisesPompilio: That approach doesn’t seem to work. After the block is mined and then invalidated, the replaced transaction isn’t discarded. I tested this flow, and when the original transaction is reintroduced, it fails because the existing transaction has a higher fee.

tests/common/mod.rs
moisesPompilio: Done

tests/common/mod.rs
moisesPompilio: Done

tests/common/mod.rs
moisesPompilio: Yes, I agree with you. I’ve made the change.

tests/common/mod.rs
tnull: > That approach doesn’t seem to work. After the block is mined and then invalidated, the replaced transaction isn’t discarded. I tested this flow, and when the original transaction is reintroduced, it fails because the existing transaction has a higher fee.

Interesting, I was under the impression that `invalidateblock` doesn't put the transactions back into the mempool but would simply discard them. It's likely not working because you use the bitcoind wallet to generate and RBF the transaction, having it track as part of the wallet (rather than 'just' in the mempool).

In any case, the block mining boilerplate is probably fine here then, even though I had hoped it could be simplified.

src/wallet/mod.rs
tnull: > The challenge is that **BDK** doesn’t currently provide a way to distinguish between these cases.

I think the new `WalletEvent::TxReplaced` should allow us to distinguish the cases and track all 'failed'/RBF'd transactions as part of the same payment entry? That should be tackled as part of #658, IMO. 

> My main motivation is that keeping transactions evicted from the mempool marked as `Pending` can lead to **UX** confusion, especially with RBF. For example, if a user broadcasts a transaction and then later replaces it via RBF, they would see two onchain payments, one that gets confirmed and another that never will. Leaving the latter as Pending would incorrectly suggest it’s still valid. Maybe having an additional status like `Dropped` or `Evicted` could better represent transactions that are no longer **canonical** in BDK?

Right, that's basically #452 which should be addressed by #658. Apart from that, leaving non-RBF'd (but evicted) transactions as `Pending` is closer to reality and hence IMO preferable as long as we're not 100% they could eventually return and get mined.

tests/common/mod.rs
moisesPompilio: during a **reorg**, Bitcoin Core doesn’t drop the transactions but tries to reinsert them into the **mempool**. I think `invalidateblock` behaves similarly. So even if a different wallet were used instead of Bitcoin Core’s, the same behavior would likely occur.

src/wallet/mod.rs
moisesPompilio: I agree that this can be handled in PR #658, so I’ve removed the part that marked **non-canonical** transactions as failed and **kept only the additional tests**. I also tested the new event handling and confirmed it correctly differentiates between replaced and dropped transactions, except in one specific case when the RBF transaction **changes the outputs** and no longer points to the wallet address. In that case, the transaction is marked as dropped instead of replaced.

tnull:
Do you still intend to finish this PR? If so, this will need a rebase.

PR #650, opened 2025-09-25
Account for `LiquidityManager` persistence

tnull:
Based on #633.

We recently implemented persistence for the `lightning-liquidity` service
state. Here we make corresponding changes in LDK Node to have our
service state persisted.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1557,6 +1557,7 @@ fn build_with_store_internal(
 				Arc::clone(&channel_manager),
 				Arc::clone(&keys_manager),
 				Arc::clone(&chain_source),
+				Arc::clone(&kv_store),
 				Arc::clone(&config),
 				Arc::clone(&logger),
 			);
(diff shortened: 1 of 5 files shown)


src/liquidity.rs
```diff
@@ -230,17 +233,22 @@ where
...
+				Arc::clone(&self.keys_manager),
+				Arc::clone(&self.channel_manager),
+				Some(Arc::clone(&self.chain_source)),
+				None,
+				Arc::clone(&self.kv_store),
+				liquidity_service_config,
+				liquidity_client_config,
+			)
+			.await
+			.map_err(|_| BuildError::ReadFailed)?,
```
joostjager, line 248: Reading data in `new`, isn't that an undesirable pattern?

src/builder.rs
```diff
@@ -1590,7 +1591,8 @@ fn build_with_store_internal(
 				liquidity_source_builder.lsps2_service(promise_secret, config.clone())
 			});
 
-			let liquidity_source = Arc::new(liquidity_source_builder.build());
+			let liquidity_source = runtime
+				.block_on(async move { liquidity_source_builder.build().await.map(Arc::new) })?;
```
joostjager, line 1595: Assuming this is temporary until async building?

joostjager (approved):

src/builder.rs
tnull: Yes, we'll remove this when we move building to `Node::start` and make it `async`.

src/liquidity.rs
tnull: Maybe, yes. We did it over at https://github.com/lightningdevkit/rust-lightning/pull/4059 since the individual service states are persisted under different locations and the `LiquidityManager` is the singular interface on which you can decide which services to enable. So allowing an API that allows to enable some services but not reading all the service states seemed complicated and error-prone (i.e., worse than reading them all in `new`.

PR #652, opened 2025-09-28
Re-format all imports

tnull:
Nightly `rustfmt` allows to auto-group imports on the module level. While we're not quite convinced to switch to the nightly channel for this yet (mostly because not all contributors would have the right nightly version installed on their machines), we here make use of `cargo +nightly fmt` with some addtional import grouping options as a one-off. This cleans up our imports for the whole crate and gets us to a consistent state everywhere.

To reproduce, run `cargo +nightly fmt` with the following lines in `rustfmt.toml` enabled:

```
group_imports = "StdExternalCrate"
reorder_imports = true
imports_granularity = "Module"
```

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -10,3 +10,5 @@ match_block_trailing_comma = true
 # UNSTABLE: format_macro_matchers = true
 # UNSTABLE: format_strings = true
 # UNSTABLE: group_imports = "StdExternalCrate"
+# UNSTABLE: reorder_imports = true
+# UNSTABLE: imports_granularity = "Module"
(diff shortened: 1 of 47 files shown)


rustfmt.toml
```diff
@@ -10,3 +10,5 @@ match_block_trailing_comma = true
 # UNSTABLE: format_macro_matchers = true
 # UNSTABLE: format_strings = true
 # UNSTABLE: group_imports = "StdExternalCrate"
+# UNSTABLE: reorder_imports = true
+# UNSTABLE: imports_granularity = "Module"
```
joostjager, line 14: We put this in so that users who want to can still run the nightly version and get the imports fixed?

tnull: Yes, and this is the list of `rustfmt` features we want to enable once we can. Put them here also so we don't forget.

PR #653, opened 2025-10-06
Improve tests with external Lightning implementations

moisesPompilio:
Introduces the `ExternalLightningNode` trait to standardize how tests interact with external Lightning nodes and adds a generic test function that handles common test logic, requiring only an initialized node.

Adds new test targets: **external nodes** opening channels and performing simple transactions with **LDK**, and a **bolt12** target to cover BOLT12 offer/request flows.

Minor cleanup: update CLN version and remove an unused dependency.

**Nota:** Tests with LND opening channels to LDK were skipped due to dust exposure HTLC errors. I think recent changes in the rust-lightning channel logic might have resolved this issue.

### DIFF
--- a/tests/common/external_node.rs
+++ b/tests/common/external_node.rs
@@ -29,7 +29,8 @@ use std::sync::OnceLock;
 
 use crate::common::{
 	distribute_funds_unconfirmed, expect_channel_pending_event, expect_channel_ready_event,
-	expect_event, generate_blocks_and_wait, premine_blocks, random_config, wait_for_tx,
+	expect_event, expect_payment_successful_event, generate_blocks_and_wait, premine_blocks,
+	random_config, wait_for_tx,
 };
 
 pub trait ExternalLightningNode {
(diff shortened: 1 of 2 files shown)


moisesPompilio:
CI is failing due to the issue described in #654

tests/common/external_node.rs
```diff
@@ -0,0 +1,251 @@
...
+
+use std::str::FromStr;
+use std::sync::OnceLock;
+
+use crate::common::{
+	distribute_funds_unconfirmed, expect_channel_pending_event, expect_channel_ready_event,
+	expect_event, generate_blocks_and_wait, premine_blocks, random_config, wait_for_tx,
+};
+
+pub trait ExternalLightningNode {
```
tnull, line 36: Hmm, I'm not convinced DRYing up this code across implementations makes sense given their different feature sets. In particular, we'll want to add integration tests for splicing/BOLT12/etc going forward, i.e., features that LND doesn't implement. Wouldn't we then need to still workaround the `ExternalLightningNode`? At this point the trait would then just be an unnecessary complication?

moisesPompilio: I understand your point about different feature sets, but I still think using a trait for `ExternalLightningNode` makes sense. Most of the test logic is the same across implementations; the only difference is in the API calls. The trait allows us to unify these common tests while letting each implementation handle its own API calls.

Regarding features like **BOLT12**, I agree **LND** doesn’t support them, but **Core Lightning** and **Eclair** do. The `ExternalLightningNode` doesn’t force all implementations to implement **BOLT12** tests. **Core Lightning** can have the **BOLT12** tests now, and when **Eclair** support is added, it can reuse the same tests by just implementing the `ExternalLightningNode` API methods.

In short, the `ExternalLightningNode` isn’t meant to enforce uniformity for all features. It’s meant to reduce code duplication for shared logic (like channel open, sending/receiving payments, channel close). For feature-specific tests, if multiple implementations support a feature (for example, **BOLT12** in Core Lightning and Eclair), it makes sense to include it in the `ExternalLightningNode` so tests can be shared. If only one implementation supports a feature (for example, splicing only in Eclair), it should remain outside the `ExternalLightningNode`, as there’s no need for a shared test.

PR #655, opened 2025-10-07
Merge `develop` into `main`

tnull:
We recently introduced a `develop` branch that tracks more recent commits on LDK's `main` branch. This proved to be a good idea for more seamless development that require more recent LDK features. It however also increased complexity/confusion around what  branch contributors should base new PR off. 

Here, we simply merge the new `develop` onto main, and will delete it eventually. 

From now on, all PRs should be based off `main`, and we'll simply backport changes/cherry-pick commits for minor releases, just as we already do for patch releases.

### DIFF
--- a/bindings/ldk_node.udl
+++ b/bindings/ldk_node.udl
@@ -325,7 +325,6 @@ enum NodeError {
 
 dictionary NodeStatus {
 	boolean is_running;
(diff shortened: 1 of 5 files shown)


joostjager (approved):
Blind ack, everything has been reviewed already.

PR #658, opened 2025-10-09
Use BDK events in `update_payment_store`

Camillarhi:
This PR updates `update_payment_store` to use BDK 2.2’s `WalletEvent` stream during sync instead of iterating over the full list of wallet transactions every time. The new event-based approach reduces redundant work and ensures the payment store stays in sync with only the changes that actually occurred.

It also sets up the foundation for RBF support in #628 with `WalletEvent::TxReplaced`. Since #628 depends on this event handling, this PR should be merged first.

This PR will also address #452

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -55,12 +55,14 @@ use crate::gossip::GossipSource;
 use crate::io::sqlite_store::SqliteStore;
 use crate::io::utils::{
 	read_event_queue, read_external_pathfinding_scores_from_cache, read_network_graph,
-	read_node_metrics, read_output_sweeper, read_payments, read_peer_info, read_scorer,
-	write_node_metrics,
+	read_node_metrics, read_output_sweeper, read_payments, read_peer_info, read_pending_payments,
+	read_scorer, write_node_metrics,
 };
 use crate::io::vss_store::VssStoreBuilder;
 use crate::io::{
 	self, PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE, PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
+	PENDING_PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
+	PENDING_PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
 };
 use crate::liquidity::{
 	LSPS1ClientConfig, LSPS2ClientConfig, LSPS2ServiceConfig, LiquiditySourceBuilder,
@@ -73,8 +75,8 @@ use crate::runtime::{Runtime, RuntimeSpawner};
 use crate::tx_broadcaster::TransactionBroadcaster;
 use crate::types::{
 	AsyncPersister, ChainMonitor, ChannelManager, DynStore, DynStoreWrapper, GossipSync, Graph,
-	KeysManager, MessageRouter, OnionMessenger, PaymentStore, PeerManager, Persister,
-	SyncAndAsyncKVStore,
+	KeysManager, MessageRouter, OnionMessenger, PaymentStore, PeerManager, PendingPaymentStore,
+	Persister, SyncAndAsyncKVStore,
 };
 use crate::wallet::persist::KVStoreWalletPersister;
 use crate::wallet::Wallet;
@@ -1057,12 +1059,14 @@ fn build_with_store_internal(
 
 	let kv_store_ref = Arc::clone(&kv_store);
 	let logger_ref = Arc::clone(&logger);
-	let (payment_store_res, node_metris_res) = runtime.block_on(async move {
-		tokio::join!(
-			read_payments(&*kv_store_ref, Arc::clone(&logger_ref)),
-			read_node_metrics(&*kv_store_ref, Arc::clone(&logger_ref)),
-		)
-	});
+	let (payment_store_res, node_metris_res, pending_payment_store_res) =
+		runtime.block_on(async move {
+			tokio::join!(
+				read_payments(&*kv_store_ref, Arc::clone(&logger_ref)),
+				read_node_metrics(&*kv_store_ref, Arc::clone(&logger_ref)),
+				read_pending_payments(&*kv_store_ref, Arc::clone(&logger_ref))
+			)
+		});
 
 	// Initialize the status fields.
 	let node_metrics = match node_metris_res {
@@ -1243,6 +1247,20 @@ fn build_with_store_internal(
 		},
 	};
 
+	let pending_payment_store = match pending_payment_store_res {
+		Ok(pending_payments) => Arc::new(PendingPaymentStore::new(
+			pending_payments,
+			PENDING_PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE.to_string(),
+			PENDING_PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE.to_string(),
+			Arc::clone(&kv_store),
+			Arc::clone(&logger),
+		)),
+		Err(e) => {
+			log_error!(logger, "Failed to read pending payment data from store: {}", e);
+			return Err(BuildError::ReadFailed);
+		},
+	};
+
 	let wallet = Arc::new(Wallet::new(
 		bdk_wallet,
 		wallet_persister,
@@ -1251,6 +1269,7 @@ fn build_with_store_internal(
 		Arc::clone(&payment_store),
 		Arc::clone(&config),
 		Arc::clone(&logger),
+		Arc::clone(&pending_payment_store),
 	));
 
 	// Initialize the KeysManager
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -12,6 +12,7 @@ use std::sync::{Arc, Mutex};
 
 use bdk_chain::spk_client::{FullScanRequest, SyncRequest};
 use bdk_wallet::descriptor::ExtendedDescriptor;
+use bdk_wallet::event::WalletEvent;
 #[allow(deprecated)]
 use bdk_wallet::SignOptions;
 use bdk_wallet::{Balance, KeychainKind, PersistedWallet, Update};
@@ -49,8 +50,10 @@ use crate::config::Config;
 use crate::fee_estimator::{ConfirmationTarget, FeeEstimator, OnchainFeeEstimator};
 use crate::logger::{log_debug, log_error, log_info, log_trace, LdkLogger, Logger};
 use crate::payment::store::ConfirmationStatus;
-use crate::payment::{PaymentDetails, PaymentDirection, PaymentStatus};
-use crate::types::{Broadcaster, PaymentStore};
+use crate::payment::{
+	PaymentDetails, PaymentDirection, PaymentKind, PaymentStatus, PendingPaymentDetails,
+};
+use crate::types::{Broadcaster, PaymentStore, PendingPaymentStore};
 use crate::Error;
 
 pub(crate) enum OnchainSendAmount {
@@ -71,6 +74,7 @@ pub(crate) struct Wallet {
 	payment_store: Arc<PaymentStore>,
 	config: Arc<Config>,
 	logger: Arc<Logger>,
+	pending_payment_store: Arc<PendingPaymentStore>,
 }
 
 impl Wallet {
@@ -78,11 +82,20 @@ impl Wallet {
 		wallet: bdk_wallet::PersistedWallet<KVStoreWalletPersister>,
 		wallet_persister: KVStoreWalletPersister, broadcaster: Arc<Broadcaster>,
 		fee_estimator: Arc<OnchainFeeEstimator>, payment_store: Arc<PaymentStore>,
-		config: Arc<Config>, logger: Arc<Logger>,
+		config: Arc<Config>, logger: Arc<Logger>, pending_payment_store: Arc<PendingPaymentStore>,
 	) -> Self {
 		let inner = Mutex::new(wallet);
 		let persister = Mutex::new(wallet_persister);
-		Self { inner, persister, broadcaster, fee_estimator, payment_store, config, logger }
+		Self {
+			inner,
+			persister,
+			broadcaster,
+			fee_estimator,
+			payment_store,
+			config,
+			logger,
+			pending_payment_store,
+		}
 	}
 
 	pub(crate) fn get_full_scan_request(&self) -> FullScanRequest<KeychainKind> {
@@ -114,15 +127,15 @@ impl Wallet {
 
 	pub(crate) fn apply_update(&self, update: impl Into<Update>) -> Result<(), Error> {
 		let mut locked_wallet = self.inner.lock().unwrap();
-		match locked_wallet.apply_update(update) {
-			Ok(()) => {
+		match locked_wallet.apply_update_events(update) {
+			Ok(events) => {
 				let mut locked_persister = self.persister.lock().unwrap();
 				locked_wallet.persist(&mut locked_persister).map_err(|e| {
 					log_error!(self.logger, "Failed to persist wallet: {}", e);
 					Error::PersistenceFailed
 				})?;
 
-				self.update_payment_store(&mut *locked_wallet).map_err(|e| {
+				self.update_payment_store(&mut *locked_wallet, events).map_err(|e| {
 					log_error!(self.logger, "Failed to update payment store: {}", e);
 					Error::PersistenceFailed
 				})?;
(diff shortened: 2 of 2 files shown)


src/wallet/mod.rs
```diff
@@ -112,15 +113,16 @@ impl Wallet {
...
+		match locked_wallet.apply_update_events(update) {
+			Ok(events) => {
 				let mut locked_persister = self.persister.lock().unwrap();
 				locked_wallet.persist(&mut locked_persister).map_err(|e| {
 					log_error!(self.logger, "Failed to persist wallet: {}", e);
 					Error::PersistenceFailed
 				})?;
 
-				self.update_payment_store(&mut *locked_wallet).map_err(|e| {
+				let events_vec: Vec<WalletEvent> = events.into_iter().collect();
```
tnull: I don't think this re-allocation is necessary?

src/wallet/mod.rs
```diff
@@ -723,7 +943,7 @@ impl Listen for Wallet {
 
 		match locked_wallet.apply_block(block, height) {
```
tnull: Ugh, seems there is no corresponding `apply_block_events` method. I think we need that before actually moving forward here. Will raise it with the BDK folks.

src/wallet/mod.rs
```diff
@@ -693,6 +826,93 @@ impl Wallet {
...
+	}
+
+	fn find_payment_by_txid(&self, target_txid: Txid) -> Option<PaymentId> {
+		let direct_payment_id = PaymentId(target_txid.to_byte_array());
+		if self.payment_store.get(&direct_payment_id).is_some() {
+			return Some(direct_payment_id);
+		}
+
+		self.payment_store
+			.list_filter(|p| {
```
tnull: Hmm, ugh, that's already slow right now, but will be prohibitively slow when we don't keep our entire payment store in-memory. I think we can't get around adding another persisted lookup table that tracks RBF-`Txid` to original-`Txid`.

src/wallet/mod.rs
Camillarhi: Alright

src/wallet/mod.rs
Camillarhi: Yes, this is true, if we have another table, the lookup will be faster and the original `Txid` can be updated when the `RBF-Txid` for example, has a `confirmed` event from `BDK`

src/wallet/mod.rs
Camillarhi: Thanks! I have removed the re-allocation

src/wallet/mod.rs
Camillarhi: I saw that the PR on BDK wallet has been merged and added to the next release milestone. This will be updated as soon as there is a new release on BDK wallet

src/wallet/mod.rs
Camillarhi: @tnull, I created a new persisted `ReplacedTransactionStore` that maintains lookups from any txid in an `RBF` chain to its associated payment. Store entries are automatically cleaned up when any transaction in the chain confirms. This keeps lookups fast even with large payment histories.

tnull: > @tnull, I created a new persisted `ReplacedTransactionStore` that maintains lookups from any txid in an `RBF` chain to its associated payment. Store entries are automatically cleaned up when any transaction in the chain confirms. This keeps lookups fast even with large payment histories.

Huh, why do we need a whole other module/store for this? Let's just use a `HashMap<Txid, Txid>` and be done with it? Or do we need all that additionally tracked data somehow?

I guess we could use a `DataStore` implementation for this, but I don't quite see why we need to track `ConfirmationStatus` and `latest_update_timestamp`?

Camillarhi: > Huh, why do we need a whole other module/store for this? Let's just use a `HashMap<Txid, Txid>` and be done with it? Or do we need all that additionally tracked data somehow?
> 
Yes, we need to track this data for faster lookup of replaced transaction IDs instead of a full iteration

> I guess we could use a `DataStore` implementation for this, but I don't quite see why we need to track `ConfirmationStatus` and `latest_update_timestamp`?
>
You are right about the `ConfirmationStatus ` and `latest_update_timestamp`, they were part of my original design, and as I proceeded with my implementation, I decided to clean the store instead upon confirmation of any of the transactions. I will go ahead and remove them. I also used a `DataStore` implementation for this.

Cargo.toml
```diff
@@ -54,7 +54,7 @@ lightning-macros = { git = "https://github.com/lightningdevkit/rust-lightning",
 bdk_chain = { version = "0.23.0", default-features = false, features = ["std"] }
 bdk_esplora = { version = "0.22.0", default-features = false, features = ["async-https-rustls", "tokio"]}
 bdk_electrum = { version = "0.23.0", default-features = false, features = ["use-rustls-ring"]}
-bdk_wallet = { version = "2.2.0", default-features = false, features = ["std", "keys-bip39"]}
+bdk_wallet = { version = "2.3.0", default-features = false, features = ["std", "keys-bip39"]}
```
tnull, line 57: Please make the BDK bump a dedicated commit.

src/wallet/mod.rs
```diff
@@ -154,73 +155,165 @@ impl Wallet {
 
 	fn update_payment_store<'a>(
 		&self, locked_wallet: &'a mut PersistedWallet<KVStoreWalletPersister>,
+		events: &'a Vec<WalletEvent>,
```
tnull: If we use move semantics we can simplify the code quite a bit, and especially save on all the reallocations:

```diff
diff --git a/src/wallet/mod.rs b/src/wallet/mod.rs
index 5982eabc..f65b3772 100644
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -135,7 +135,7 @@ impl Wallet {
 					Error::PersistenceFailed
 				})?;
 
-				self.update_payment_store(&mut *locked_wallet, &events).map_err(|e| {
+				self.update_payment_store(&mut *locked_wallet, events).map_err(|e| {
 					log_error!(self.logger, "Failed to update payment store: {}", e);
 					Error::PersistenceFailed
 				})?;
@@ -167,15 +167,14 @@ impl Wallet {
 
 	fn update_payment_store<'a>(
 		&self, locked_wallet: &'a mut PersistedWallet<KVStoreWalletPersister>,
-		events: &'a Vec<WalletEvent>,
+		mut events: Vec<WalletEvent>,
 	) -> Result<(), Error> {
 		if events.is_empty() {
 			return Ok(());
 		}
 
-		let sorted_events: Vec<_> = if events.len() > 1 {
-			let mut events_vec: Vec<_> = events.iter().collect();
-			events_vec.sort_by_key(|e| match e {
+		if events.len() > 1 {
+			events.sort_by_key(|e| match e {
 				WalletEvent::TxReplaced { .. } => 0,
 				WalletEvent::TxUnconfirmed { .. } => 1,
 				WalletEvent::TxConfirmed { .. } => 2,
@@ -183,12 +182,9 @@ impl Wallet {
 				WalletEvent::TxDropped { .. } => 4,
 				_ => 5,
 			});
-			events_vec
-		} else {
-			events.iter().collect()
 		};
 
-		for event in sorted_events {
+		for event in events {
 			match event {
 				WalletEvent::TxConfirmed { txid, tx, block_time, .. } => {
 					let cur_height = locked_wallet.latest_checkpoint().height();
@@ -207,14 +203,14 @@ impl Wallet {
 					};
 
 					let payment_id = self
-						.find_payment_by_txid(*txid)
+						.find_payment_by_txid(txid)
 						.unwrap_or_else(|| PaymentId(txid.to_byte_array()));
 
 					let payment = self.create_payment_from_tx(
 						locked_wallet,
-						*txid,
+						txid,
 						payment_id,
-						tx,
+						&tx,
 						payment_status,
 						confirmation_status,
 						None,
@@ -261,14 +257,14 @@ impl Wallet {
 				},
 				WalletEvent::TxUnconfirmed { txid, tx, old_block_time: None } => {
 					let payment_id = self
-						.find_payment_by_txid(*txid)
+						.find_payment_by_txid(txid)
 						.unwrap_or_else(|| PaymentId(txid.to_byte_array()));
 
 					let payment = self.create_payment_from_tx(
 						locked_wallet,
-						*txid,
+						txid,
 						payment_id,
-						tx,
+						&tx,
 						PaymentStatus::Pending,
 						ConfirmationStatus::Unconfirmed,
 						None,
@@ -277,7 +273,7 @@ impl Wallet {
 				},
 				WalletEvent::TxReplaced { txid, conflicts, .. } => {
 					let payment_id = self
-						.find_payment_by_txid(*txid)
+						.find_payment_by_txid(txid)
 						.unwrap_or_else(|| PaymentId(txid.to_byte_array()));
 
 					// Collect all conflict txids
@@ -286,24 +282,21 @@ impl Wallet {
 
 					for conflict_txid in conflict_txids {
 						// Update the replaced transaction store
-						let replaced_tx_details = ReplacedOnchainTransactionDetails::new(
-							conflict_txid,
-							*txid,
-							payment_id,
-						);
+						let replaced_tx_details =
+							ReplacedOnchainTransactionDetails::new(conflict_txid, txid, payment_id);
 
 						self.replaced_tx_store.insert_or_update(replaced_tx_details)?;
 					}
 				},
 				WalletEvent::TxDropped { txid, tx } => {
 					let payment_id = self
-						.find_payment_by_txid(*txid)
+						.find_payment_by_txid(txid)
 						.unwrap_or_else(|| PaymentId(txid.to_byte_array()));
 					let payment = self.create_payment_from_tx(
 						locked_wallet,
-						*txid,
+						txid,
 						payment_id,
-						tx,
+						&tx,
 						PaymentStatus::Pending,
 						ConfirmationStatus::Unconfirmed,
 						None,
@@ -999,7 +992,7 @@ impl Listen for Wallet {
 
 		match locked_wallet.apply_block_events(block, height) {
 			Ok(events) => {
-				if let Err(e) = self.update_payment_store(&mut *locked_wallet, &events) {
+				if let Err(e) = self.update_payment_store(&mut *locked_wallet, events) {
 					log_error!(self.logger, "Failed to update payment store: {}", e);
 					return;
 				}
```

src/wallet/mod.rs
```diff
@@ -154,73 +155,165 @@ impl Wallet {
...
-			let (payment_status, confirmation_status) = match wtx.chain_position {
-				bdk_chain::ChainPosition::Confirmed { anchor, .. } => {
-					let confirmation_height = anchor.block_id.height;
+		if events.is_empty() {
+			return Ok(());
+		}
+
+		let sorted_events: Vec<_> = if events.len() > 1 {
+			let mut events_vec: Vec<_> = events.iter().collect();
+			events_vec.sort_by_key(|e| match e {
```
tnull: Why do we need to sort the events? Is it so we first process `TxConfirmed` before adjusting the height value in `ChainTipChanged`?

src/wallet/mod.rs
```diff
@@ -793,6 +886,93 @@ impl Wallet {
 
 		Ok(tx)
 	}
+
+	fn create_payment_from_tx(
+		&self, locked_wallet: &PersistedWallet<KVStoreWalletPersister>, txid: Txid,
+		payment_id: PaymentId, tx: &Transaction, payment_status: PaymentStatus,
+		confirmation_status: ConfirmationStatus, conflicting_txids: Option<Vec<Txid>>,
+	) -> PaymentDetails {
+		// TODO: It would be great to introduce additional variants for
```
tnull, line 914: Can we make this code move in a dedicated prefactor commit, so that `git diff --color-moved` clearly show the code move, before we then change anything in the following commits?

src/payment/replaced_transaction_store.rs
```diff
@@ -0,0 +1,104 @@
...
+	pub payment_id: PaymentId,
+}
+
+impl ReplacedOnchainTransactionDetails {
+	pub(crate) fn new(new_txid: Txid, original_txid: Txid, payment_id: PaymentId) -> Self {
+		Self { new_txid, original_txid, payment_id }
+	}
+}
+
+impl Writeable for ReplacedOnchainTransactionDetails {
```
tnull: Please use LDK's serialization macros (in this case `impl_writeable_tlv_based`) whenever you can.

src/wallet/mod.rs
```diff
@@ -793,6 +886,93 @@ impl Wallet {
...
+			amount_msat,
+			Some(fee_sat * 1000),
+			direction,
+			payment_status,
+		)
+	}
+
+	fn find_payment_by_txid(&self, target_txid: Txid) -> Option<PaymentId> {
+		let direct_payment_id = PaymentId(target_txid.to_byte_array());
+		if self.payment_store.get(&direct_payment_id).is_some() {
```
tnull: I think this would unnecessarily clone the entry. Let's just add a `contains` implementation to `DataStore` (preferably in a separate prefactor commit).

src/wallet/mod.rs
```diff
@@ -208,6 +220,17 @@ impl Wallet {
...
+
+					// Remove any replaced transactions associated with this payment
+					let replaced_txids = self
+						.replaced_tx_store
+						.list_filter(|r| r.payment_id == payment_id)
+						.iter()
+						.map(|p| p.new_txid)
+						.collect::<Vec<Txid>>();
+					for replaced_txid in replaced_txids {
+						self.replaced_tx_store.remove(&replaced_txid)?;
```
tnull: I think we can only remove the entry once we reach `ANTI_REORG_DELAY`, no? Otherwise we might see it confirmed, delete the entry, tip gets reorged, and we wouldn't track the original payment ID anymore?

src/wallet/mod.rs
Camillarhi: We process `TXReplaced` before `TxUnconfirmed` because when an `RBF` occurs, both a `TXReplaced` and a `TxUnconfirmed` event are triggered.  
If `TxUnconfirmed` were processed first, it would create a new payment record for the unconfirmed transaction, which we don’t want. Instead, we only want the original transaction to have a payment record, and link the replaced transaction to it.

By handling `TXReplaced` first, we store the replaced transaction in `ReplacedOnchainTransactionStore`. Then, when the `TxUnconfirmed` event arrives later, it checks and avoids creating a duplicate payment record for the now-conflicting transaction.


As for `TxConfirmed` and `ChainTipChanged`, the order matters there too. If we processed `ChainTipChanged` (which updates the block height) before `TxConfirmed`, the confirmation might reference an incorrect or outdated height. Sorting ensures `TxConfirmed` uses the correct chain state.

src/wallet/mod.rs
Camillarhi: Yes, thanks for pointing this out, this should be done when `ANTI_REORG_DELAY` is reached not on `TxConfirmed` event. I will make the update for it

Cargo.toml
Camillarhi: Thanks, this has been updated

src/wallet/mod.rs
Camillarhi: Thanks, I have added a `contains_key` implementation in a separate commit.

src/wallet/mod.rs
tnull: Makes sense, thanks!

src/wallet/mod.rs
```diff
@@ -154,42 +155,161 @@ impl Wallet {
...
+						payment_status,
+						confirmation_status,
+						None,
+					);
+					self.payment_store.insert_or_update(payment)?;
+				},
+				WalletEvent::ChainTipChanged { new_tip, .. } => {
+					// Get all payments that are Pending with Confirmed status
+					let pending_payments: Vec<PaymentDetails> =
+						self.payment_store.list_filter(|p| {
```
tnull: Hmm, this is unfortunate as it is basically back to scanning all transactions, though this time scanning even all *entries* in our payment store which might be worse. This makes me wonder if we should, rather than just track conflicting/replaced transactions, track 'pending payments' (with a field for conflicting/replaced txids) to also make this lookup reasonably fast. This might become especially crucial since we're looking to eventually only keep a cache of the payment store in memory, so scanning *all* entries might at that point mean a lot of slow, costly IO lookups from the `KVStore`.

src/wallet/mod.rs
```diff
@@ -780,27 +900,74 @@ impl Wallet {
 		// here to determine the `PaymentKind`, but that's not really satisfactory, so
 		// we're punting on it until we can come up with a better solution.
 
-		let kind = crate::payment::PaymentKind::Onchain { txid, status: confirmation_status };
+		let existing_payment = self.payment_store.get(&payment_id);
+		let final_conflicting_txids = if let Some(provided_conflicts) = conflicting_txids {
```
tnull: Hmm, tbh, this gets a bit confusing: why are we providing conflicts, and then *also* looking up conflicts from an existing payment here? Wouldn't it make sense to just have `create_payment_from_tx` a list of conflicts and populate that at the callsite?

src/io/utils.rs
```diff
@@ -576,6 +577,38 @@ pub(crate) fn read_bdk_wallet_change_set(
 	Ok(Some(change_set))
 }
 
+/// Read previously persisted replaced transaction information from the store.
+pub(crate) fn read_replaced_txs<L: Deref>(
```
tnull: Similar to what we did with the other `read_` methods this (or rather the pending payments counterpart) should be `async` now.

src/wallet/mod.rs
```diff
@@ -154,73 +167,147 @@ impl Wallet {
...
-			let txid = wtx.tx_node.txid;
-			let (payment_status, confirmation_status) = match wtx.chain_position {
-				bdk_chain::ChainPosition::Confirmed { anchor, .. } => {
-					let confirmation_height = anchor.block_id.height;
+		if events.is_empty() {
+			return Ok(());
+		}
+
+		if events.len() > 1 {
+			events.sort_by_key(|e| match e {
```
tnull, line 199: Mind adding the explanation you gave w.r.t. *why* we sort the event as a comment here in the code?

Camillarhi: Thanks for the suggestion, I'll add that explanation as a comment

src/wallet/mod.rs
Camillarhi: I've removed the `conflicting_txids` parameter from `create_payment_from_tx` and moved the logic to the callsite. In `WalletEvent::TxReplaced` we already have the conflicts from the event, so we just assign them after creating the payment.

src/builder.rs
```diff
@@ -1235,6 +1238,22 @@ fn build_with_store_internal(
 		},
 	};
 
+	let pending_payment_store = match runtime
+		.block_on(async { read_pending_payments(&*kv_store, Arc::clone(&logger)).await })
```
tnull: We should read these in parallel to reading the other payments, to avoid the extra roundtrip.

src/payment/store.rs
```diff
@@ -360,8 +351,6 @@ pub enum PaymentKind {
 		txid: Txid,
 		/// The confirmation status of this payment.
 		status: ConfirmationStatus,
-		/// Transaction IDs that have replaced or conflict with this payment.
-		conflicting_txids: Vec<Txid>,
```
tnull: We shouldn't first add this field and then remove it in a later commit. Can we maybe first add the pending payment store in a prefactor commit, and then directly make use of it when switching to the events?

src/payment/pending_payment_store.rs
```diff
@@ -0,0 +1,96 @@
...
+	payment::{store::PaymentDetailsUpdate, PaymentDetails},
+};
+
+/// Represents a pending payment
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct PendingPaymentDetails {
+	/// The full payment details
+	pub details: PaymentDetails,
+	/// Cached timestamp for efficient cleanup queries
+	pub created_at: u64,
```
tnull: AFAICT this field is never used? When/how would we use it? Do we intend this to happen in this PR?

src/payment/store.rs
Camillarhi: Thanks for the suggestion. I'll reorder the commits so the pending payment store comes before its usage in the wallet sync update.

src/builder.rs
Camillarhi: This has been updated to be read in parallel with other payments

src/payment/pending_payment_store.rs
Camillarhi: Yes, it's unused. I've removed the `created_at` field since the value is just derived from `latest_update_timestamp` on the payment detail, making it redundant.

PR #659, opened 2025-10-13
Bump to latest LDK `main`

tnull:
- We fix an oversight introduced in 9977903132e56afa78804b2e3de1bee29a9b410a where we switched `Cargo.toml` to use local dependencies instead of pointing to an appropriate upstream commit
- We account for Splicing-related events
- We account for `KeysManager` taking a new flag for v2-payment-key derivation.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -1601,6 +1601,18 @@ where
 			LdkEvent::FundingTransactionReadyForSigning { .. } => {
 				debug_assert!(false, "We currently don't support interactive-tx, so this event should never be emitted.");
 			},
+			LdkEvent::SplicePending { .. } => {
+				debug_assert!(
+					false,
+					"We currently don't support splicing, so this event should never be emitted."
+				);
+			},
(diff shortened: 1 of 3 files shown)


src/event.rs
```diff
@@ -1601,6 +1601,18 @@ where
 			LdkEvent::FundingTransactionReadyForSigning { .. } => {
 				debug_assert!(false, "We currently don't support interactive-tx, so this event should never be emitted.");
 			},
+			LdkEvent::SplicePending { .. } => {
+				debug_assert!(
+					false,
+					"We currently don't support splicing, so this event should never be emitted."
```
joostjager, line 1607: Just to double check - no splicing api is exposed in ldk-node and therefore this should never happen? I don't know the api, but is it possible for the counterparty to initiate this and get us to move along?

tnull: Yes, there is a new `reject_inbound_splices` on `UserConfig` that defaults to `true`. So we'll have to manually opt into splicing support as part of the integration PR.

PR #661, opened 2025-10-13
Use `MonitorUpdatingPersister`

### LINKED ISSUE #441: Switch to `MonitorUpdatingPersister`

tnull:
Closes #441 

We use LDK's `MonitorUpdatingPersister` to persist (and read) differential updates, instead of always repersisting the full monitors.

This is mostly a simple rebase of #456 (whose original author moved on) on top of current main.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -31,8 +31,8 @@ use lightning::routing::scoring::{
 };
 use lightning::sign::{EntropySource, NodeSigner};
 use lightning::util::persist::{
-	read_channel_monitors, KVStoreSync, CHANNEL_MANAGER_PERSISTENCE_KEY,
-	CHANNEL_MANAGER_PERSISTENCE_PRIMARY_NAMESPACE, CHANNEL_MANAGER_PERSISTENCE_SECONDARY_NAMESPACE,
+	KVStoreSync, CHANNEL_MANAGER_PERSISTENCE_KEY, CHANNEL_MANAGER_PERSISTENCE_PRIMARY_NAMESPACE,
+	CHANNEL_MANAGER_PERSISTENCE_SECONDARY_NAMESPACE,
 };
 use lightning::util::ser::ReadableArgs;
 use lightning::util::sweep::OutputSweeper;
@@ -66,7 +66,7 @@ use crate::runtime::Runtime;
 use crate::tx_broadcaster::TransactionBroadcaster;
 use crate::types::{
 	ChainMonitor, ChannelManager, DynStore, GossipSync, Graph, KeysManager, MessageRouter,
-	OnionMessenger, PaymentStore, PeerManager,
+	OnionMessenger, PaymentStore, PeerManager, Persister,
 };
 use crate::wallet::persist::KVStoreWalletPersister;
 use crate::wallet::Wallet;
@@ -75,6 +75,7 @@ use crate::{Node, NodeMetrics};
 const VSS_HARDENED_CHILD_INDEX: u32 = 877;
 const VSS_LNURL_AUTH_HARDENED_CHILD_INDEX: u32 = 138;
 const LSPS_HARDENED_CHILD_INDEX: u32 = 577;
+const PERSISTER_MAX_PENDING_UPDATES: u64 = 100;
 
 #[derive(Debug, Clone)]
 enum ChainDataSourceConfig {
@@ -1317,6 +1318,28 @@ fn build_with_store_internal(
 	));
 
 	let peer_storage_key = keys_manager.get_peer_storage_key();
+	let persister = Arc::new(Persister::new(
+		Arc::clone(&kv_store),
+		Arc::clone(&logger),
+		PERSISTER_MAX_PENDING_UPDATES,
+		Arc::clone(&keys_manager),
+		Arc::clone(&keys_manager),
+		Arc::clone(&tx_broadcaster),
+		Arc::clone(&fee_estimator),
+	));
+
+	// Read ChannelMonitor state from store
+	let channel_monitors = match persister.read_all_channel_monitors_with_updates() {
+		Ok(monitors) => monitors,
+		Err(e) => {
+			if e.kind() == lightning::io::ErrorKind::NotFound {
+				Vec::new()
+			} else {
+				log_error!(logger, "Failed to read channel monitors: {}", e.to_string());
+				return Err(BuildError::ReadFailed);
+			}
+		},
+	};
 
 	// Initialize the ChainMonitor
 	let chain_monitor: Arc<ChainMonitor> = Arc::new(chainmonitor::ChainMonitor::new(
@@ -1324,7 +1347,7 @@ fn build_with_store_internal(
 		Arc::clone(&tx_broadcaster),
 		Arc::clone(&logger),
 		Arc::clone(&fee_estimator),
-		Arc::clone(&kv_store),
+		Arc::clone(&persister),
 		Arc::clone(&keys_manager),
 		peer_storage_key,
 	));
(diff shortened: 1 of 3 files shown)


src/io/test_utils.rs
```diff
@@ -77,54 +91,73 @@ pub(crate) fn do_read_write_remove_list_persist<K: KVStoreSync + RefUnwindSafe>(
...
+	let persister: TestMonitorUpdatePersister<'a, K> = MonitorUpdatingPersister::new(
+		store,
+		&chanmon_cfg.logger,
+		max_pending_updates,
+		&chanmon_cfg.keys_manager,
+		&chanmon_cfg.keys_manager,
+		&chanmon_cfg.tx_broadcaster,
+		&chanmon_cfg.fee_estimator,
+	);
+	return persister;
```
joostjager: Return directly without intermediate var, here and below?

src/io/test_utils.rs
```diff
@@ -77,54 +91,73 @@ pub(crate) fn do_read_write_remove_list_persist<K: KVStoreSync + RefUnwindSafe>(
...
+	);
+	return chain_mon;
+}
+
 // Integration-test the given KVStore implementation. Test relaying a few payments and check that
 // the persisted data is updated the appropriate number of times.
 pub(crate) fn do_test_store<K: KVStoreSync + Sync>(store_0: &K, store_1: &K) {
+	// This value is used later to limit how many iterations we perform.
+	let persister_0_max_pending_updates = 7;
+	// Intentionally set this to a smaller value to test a different alignment.
```
joostjager, line 126: I've seen this also in rust-lightning, and also there I didn't understand what alignment there is to test because the nodes are independent?

src/io/test_utils.rs
```diff
@@ -137,10 +170,29 @@ pub(crate) fn do_test_store<K: KVStoreSync + Sync>(store_0: &K, store_1: &K) {
...
+	send_payment(&nodes[0], expected_route, 8_000_000);
+	check_persisted_data!(EXPECTED_UPDATES_PER_PAYMENT);
+	let expected_route = &[&nodes[0]][..];
+	send_payment(&nodes[1], expected_route, 4_000_000);
+	check_persisted_data!(2 * EXPECTED_UPDATES_PER_PAYMENT);
+
+	// Send a few more payments to try all the alignments of max pending updates with
+	// updates for a payment sent and received.
+	let mut sender = 0;
+	for i in 3..=persister_0_max_pending_updates * 2 {
```
joostjager, line 181: This also seems a bit overkill at the ldk node level?

src/io/test_utils.rs
```diff
@@ -163,27 +215,18 @@ pub(crate) fn do_test_store<K: KVStoreSync + Sync>(store_0: &K, store_1: &K) {
 	check_closed_broadcast!(nodes[0], true);
 	check_added_monitors!(nodes[0], 1);
 
-	let node_txn = nodes[0].tx_broadcaster.txn_broadcasted.lock().unwrap();
+	let node_txn = nodes[0].tx_broadcaster.txn_broadcast();
 	assert_eq!(node_txn.len(), 1);
+	let txn = vec![node_txn[0].clone(), node_txn[0].clone()];
+	let dummy_block = create_dummy_block(nodes[0].best_block_hash(), 42, txn);
+	connect_block(&nodes[1], &dummy_block);
```
joostjager, line 220: Some independent refactoring here?

src/builder.rs
```diff
@@ -1317,14 +1318,36 @@ fn build_with_store_internal(
...
+		Arc::clone(&logger),
+		PERSISTER_MAX_PENDING_UPDATES,
+		Arc::clone(&keys_manager),
+		Arc::clone(&keys_manager),
+		Arc::clone(&tx_broadcaster),
+		Arc::clone(&fee_estimator),
+	));
+
+	// Read ChannelMonitor state from store
+	let channel_monitors = match persister.read_all_channel_monitors_with_updates() {
```
joostjager, line 1332: Why the code move?

src/builder.rs
```diff
@@ -75,6 +75,7 @@ use crate::{Node, NodeMetrics};
 const VSS_HARDENED_CHILD_INDEX: u32 = 877;
 const VSS_LNURL_AUTH_HARDENED_CHILD_INDEX: u32 = 138;
 const LSPS_HARDENED_CHILD_INDEX: u32 = 577;
+const PERSISTER_MAX_PENDING_UPDATES: u64 = 100;
```
joostjager, line 78: Is there science to the number 100?

src/io/test_utils.rs
tnull: Done.

src/io/test_utils.rs
tnull: Yeah, you'd have to ask the OP regarding the comment, but it's also of course not wrong to set a different value here.

src/io/test_utils.rs
tnull: Mhh, a bit more test coverage doesn't hurt? FWIW, this could be made a proptest to have it be less deterministic, but apart from that there isn't much harm?

src/builder.rs
tnull: No, just handwaving currently. There has been some benchmarking over at https://github.com/lightningdevkit/rust-lightning/pull/3834 but it never got anywhere super conclusive. Let me know if you have a better guess.

src/builder.rs
tnull: You'd have to ask OP for the original reason, I left it in as grouping the channel monitor IO with setting up the persister and `ChainMonitor` made sense to me, but there might be different ways to think about it, probably. Let me know if you'd prefer to revert the change.

src/io/test_utils.rs
tnull: Yeah, doesn't hurt either though.

src/io/test_utils.rs
joostjager: It isn't wrong, but thought it might be worth knowing a bit more about it now that you are taking over.

src/io/test_utils.rs
joostjager: It doesn't hurt in terms over coverage, but it is probably duplicative, and code carries cost.

src/io/test_utils.rs
joostjager: I meant to say that it is better to isolate in a separate commit.

src/builder.rs
joostjager: It's ok. The origin of the comment was diff optimization.

src/builder.rs
joostjager: Does not seem unreasonable. Going higher is probably not gaining much. Maybe it could be a bit lower. But also fine as is.

src/io/test_utils.rs
domZippilli: @joostjager I think I wrote this sentence first in `rust-lightning` in this PR: https://github.com/lightningdevkit/rust-lightning/pull/2359/files#diff-0f1c2f21752b2bf468479020ff092a484d8b0a4a1c018241e3a127b62aa51de2R914

I think what we were concerned with was weird edge cases / off-by-one errors in the compaction algorithm in particular. I think there was a state of this PR where odd numbered `maximum_pending_updates` would drop one or something. Initially we just tested the kind of obvious/likely powers of 10. So deliberately using a few different primes became a thing.

PR #657, opened 2025-10-09
Switch to `nightly` `rustfmt`

tnull:
Just an experiment for now.

PR #669, opened 2025-10-17
Nightly `rustfmt` follow-up

tnull:
This is a follow-up to #657 which becomes necessary as `rust-toolchain.toml` unfortunately turned out not to work the way I thought it did.

Here, we revert our default `rustfmt` version to `stable` but also introduce a weekly cronjob running `cargo +nightly fmt` (which we stole from `rust-bitcoin`).

### DIFF
--- a/.github/workflows/cron-weekly-rustfmt.yml
+++ b/.github/workflows/cron-weekly-rustfmt.yml
@@ -1,0 +1,28 @@
+name: Nightly rustfmt
+on:
+  schedule:
+    - cron: "0 0 * * 0" # runs weekly on Sunday at 00:00
+  workflow_dispatch: # allows manual triggering
+jobs:
+  format:
+    name: Nightly rustfmt
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v5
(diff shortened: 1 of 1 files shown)


.github/workflows/cron-weekly-rustfmt.yml
```diff
@@ -0,0 +1,28 @@
+name: Nightly rustfmt
```
joostjager, line 1: Did you test-run this job in CI? (temp. schedule to trigger it?)

tnull: > Did you test-run this job in CI? (temp. schedule to trigger it?)

No, but it's a carbon copy of https://github.com/rust-bitcoin/rust-bitcoin/blob/master/.github/workflows/cron-weekly-rustfmt.yml, which is run weekly over at `rust-bitcoin`. See for example https://github.com/rust-bitcoin/rust-bitcoin/pull/5120

joostjager: I guess we'll find out if the weekly works or not...

joostjager (approved):

PR #670, opened 2025-10-17
Asyncify test suite and reintroduce VSS-internal runtime

tnull:
As we're looking to move to a more `async`-friendly API over time (which would allow us to drop the dependency on the `KVStoreSync` implementation entirely), we here make a few prefactors and minor improvements that move in that direction. 

Most notably, we switch to have our tests run in async contexts, which also should make them more efficient as not every node will spawn its own runtime, but rather reuse the one runtime spawned for each test case.

Moreover, we reintroduce a secondary runtime internal to `VssStore` in order to fix the blocking issue we recently hit in our CI. Additionally, we found that not all tests cfg-gated under `vss_test` were being run, so we fix these test cases and have them run in CI.

### DIFF
--- a/src/io/vss_store.rs
+++ b/src/io/vss_store.rs
@@ -606,38 +606,40 @@ mod tests {
 
 	use rand::distributions::Alphanumeric;
 	use rand::{thread_rng, Rng, RngCore};
-	use tokio::runtime;
 	use vss_client::headers::FixedHeaders;
 
 	use super::*;
 	use crate::io::test_utils::do_read_write_remove_list_persist;
+	use crate::logger::Logger;
 
 	#[test]
 	fn vss_read_write_remove_list_persist() {
-		let runtime = Arc::new(Runtime::new().unwrap());
 		let vss_base_url = std::env::var("TEST_VSS_BASE_URL").unwrap();
 		let mut rng = thread_rng();
 		let rand_store_id: String = (0..7).map(|_| rng.sample(Alphanumeric) as char).collect();
 		let mut vss_seed = [0u8; 32];
 		rng.fill_bytes(&mut vss_seed);
 		let header_provider = Arc::new(FixedHeaders::new(HashMap::new()));
+		let logger = Arc::new(Logger::new_log_facade());
+		let runtime = Arc::new(Runtime::new(logger).unwrap());
 		let vss_store =
-			VssStore::new(vss_base_url, rand_store_id, vss_seed, header_provider, runtime).unwrap();
+			VssStore::new(vss_base_url, rand_store_id, vss_seed, header_provider, runtime);
 
 		do_read_write_remove_list_persist(&vss_store);
 	}
 
 	#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
 	async fn vss_read_write_remove_list_persist_in_runtime_context() {
-		let runtime = Arc::new(Runtime::new().unwrap());
 		let vss_base_url = std::env::var("TEST_VSS_BASE_URL").unwrap();
 		let mut rng = thread_rng();
 		let rand_store_id: String = (0..7).map(|_| rng.sample(Alphanumeric) as char).collect();
 		let mut vss_seed = [0u8; 32];
 		rng.fill_bytes(&mut vss_seed);
 		let header_provider = Arc::new(FixedHeaders::new(HashMap::new()));
+		let logger = Arc::new(Logger::new_log_facade());
+		let runtime = Arc::new(Runtime::new(logger).unwrap());
 		let vss_store =
-			VssStore::new(vss_base_url, rand_store_id, vss_seed, header_provider, runtime).unwrap();
+			VssStore::new(vss_base_url, rand_store_id, vss_seed, header_provider, runtime);
 
 		do_read_write_remove_list_persist(&vss_store);
 		drop(vss_store)
--- a/.github/workflows/vss-integration.yml
+++ b/.github/workflows/vss-integration.yml
@@ -44,4 +44,5 @@ jobs:
         run: |
           cd ldk-node
           export TEST_VSS_BASE_URL="http://localhost:8080/vss"
+          RUSTFLAGS="--cfg vss_test" cargo test io::vss_store
           RUSTFLAGS="--cfg vss_test" cargo test --test integration_tests_vss


src/lib.rs
```diff
@@ -749,7 +749,8 @@ impl Node {
 	/// **Caution:** Users must handle events as quickly as possible to prevent a large event backlog,
 	/// which can increase the memory footprint of [`Node`].
 	pub fn wait_next_event(&self) -> Event {
-		self.event_queue.wait_next_event()
+		let fut = self.event_queue.next_event_async();
+		self.runtime.block_on(fut)
```
joostjager, line 755: It isn't immediately clear to me how this change makes it better? I've also heard `block_on` is bad... The hope is that it would just block the task rather than the thread?

src/event.rs
```diff
@@ -1625,7 +1625,6 @@ mod tests {
 
 		// Check we get the expected event and that it is returned until we mark it handled.
 		for _ in 0..5 {
-			assert_eq!(event_queue.wait_next_event(), expected_event);
```
joostjager, line 1640: The original idea was to assert that the event can be re-read also via the other method?

src/event.rs
```diff
@@ -1709,32 +1708,5 @@ mod tests {
...
-		let thread_event = expected_event.clone();
-		std::thread::spawn(move || {
-			// Sleep a bit before we enqueue the events everybody is waiting for.
-			std::thread::sleep(Duration::from_millis(20));
-			thread_queue.add_event(thread_event.clone()).unwrap();
-			thread_queue.add_event(thread_event.clone()).unwrap();
-		});
-
-		let e = event_queue.next_event_async().await;
-		assert_eq!(e, expected_event.clone());
```
joostjager, line 1746: Isn't this coverage that is still useful, or do we slowly start to deprioritize sync and testing of it?

tests/integration_tests_cln.rs
```diff
@@ -25,8 +25,8 @@ use lightning_invoice::{Bolt11Invoice, Bolt11InvoiceDescription, Description};
 use rand::distributions::Alphanumeric;
 use rand::{thread_rng, Rng};
 
-#[test]
-fn test_cln() {
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
```
joostjager, line 28: Is there a reason for specifically 1 thread?

tests/reorg_test.rs
```diff
@@ -17,179 +17,187 @@ proptest! {
...
-			($node:expr, $expected:expr) => {{
-				let mut user_channels = HashMap::new();
-				for _ in 0..$expected {
-					match $node.wait_next_event() {
-						Event::ChannelReady { user_channel_id, counterparty_node_id, .. } => {
-							$node.event_handled().unwrap();
-							user_channels.insert(counterparty_node_id, user_channel_id);
-						},
-						other => panic!("Unexpected event: {:?}", other),
+		let rt = tokio::runtime::Builder::new_multi_thread()
```
joostjager, line 20: Shouldn't this be a tokio::test?

src/lib.rs
tnull: It is not necessarily better, but our `Runtime::block_on` implementation would at least make sure to always call `tokio::task::block_in_place` if we happen to run in an outer runtime context, while `Condvar::wait` would just immediately park the thread. Yes, we should get rid of all 'internal' `block_on`s, but this might be a marginal improvement.

src/event.rs
tnull: But we now only have one method to test?

src/event.rs
tnull: Yes.

tests/integration_tests_cln.rs
tnull: Yes, efficiency, as otherwise each test would spawn `NUM_CORES` threads, which might make CI unnecessary slow.

tests/reorg_test.rs
tnull: It can't be as the `#![proptest_config(proptest::test_runner::Config::with_cases(5))]` macro conflicts with `#[tokio::test]`.

src/event.rs
joostjager: Just verifying the original idea.

src/event.rs
joostjager: But aren't there users still using the sync api, and doesn't that make it premature to remove coverage?

tests/integration_tests_cln.rs
joostjager: I do wonder if 1 thread simplifies concurrent behavior and might hide potential bugs?

tests/reorg_test.rs
joostjager: Ok

src/event.rs
tnull: > But aren't there users still using the sync api, and doesn't that make it premature to remove coverage?

I don't see an easy way to keep the coverage if we're reusing the same internal API for sync anyways?

tests/integration_tests_cln.rs
tnull: > I do wonder if 1 thread simplifies concurrent behavior and might hide potential bugs?

Usually it's the other way around: reducing the number of worker threads increases the chances of hitting any issues with blocking a thread.

src/lib.rs
joostjager: The scenario that you describe here is that we are in an async context, but for some reason ended in a sync fn that calls `wait_next_event`, and then we protect against issues by the extra `block_in_place`, right?

src/event.rs
joostjager: Indeed, it doesn't add much.

tests/integration_tests_cln.rs
joostjager: Okay, but also nothing is actually running in parallel? I suppose both options cover different but useful scenarios.

src/runtime.rs
```diff
@@ -67,7 +67,7 @@ impl Runtime {
 	{
 		let mut background_tasks = self.background_tasks.lock().unwrap();
 		let runtime_handle = self.handle();
-		background_tasks.spawn_on(future, runtime_handle);
+		background_tasks.spawn_on(async { future.await }, runtime_handle);
```
joostjager, line 73: Oh man, this really is starting to get a bit ridiculous. At least for a normal dev 😞 

I'd put the commit message here as a comment.

src/lib.rs
joostjager: Maybe add "our Runtime::block_on implementation would at least make sure to always call tokio::task::block_in_place if we happen to run in an outer runtime context" as a comment. It's all not that trivial if one isn't a tokio connaisseur.

src/io/vss_store.rs
```diff
@@ -60,7 +64,21 @@ impl VssStore {
 	) -> Self {
 		let inner = Arc::new(VssStoreInner::new(base_url, store_id, vss_seed, header_provider));
 		let next_version = AtomicU64::new(1);
-		Self { inner, next_version, runtime }
+		let internal_runtime = Some(
```
joostjager, line 75: Add the @TheBlueMatt analysis here as a comment to explain why we need this (for now)?

>there's a blocking task waiting on the VSS store to finish, but ultimately what's blocking the reactor is another task calling PeerManager::process_pending_events which calls ChannelManager::get_and_clear_pending_msg_events which then tries to take some locks in channelmanager that are held by the thread which is currently blocked waiting on the VSS store.

src/io/vss_store.rs
```diff
@@ -44,13 +44,17 @@ type CustomRetryPolicy = FilteredRetryPolicy<
 	Box<dyn Fn(&VssError) -> bool + 'static + Send + Sync>,
 >;
 
+const INTERNAL_RUNTIME_WORKERS: usize = 2;
```
joostjager, line 49: Document why 2?

src/io/vss_store.rs
```diff
@@ -94,46 +112,122 @@ impl KVStoreSync for VssStore {
...
+			async move { inner.read_internal(primary_namespace, secondary_namespace, key).await };
+		// TODO: We could drop the timeout here once we ensured vss-client's Retry logic always
+		// times out.
+		let spawned_fut = internal_runtime.spawn(async move {
+			tokio::time::timeout(VSS_IO_TIMEOUT, fut).await.map_err(|_| {
+				let msg = "VssStore::read timed out";
+				Error::new(ErrorKind::Other, msg)
+			})
+		});
+		self.runtime.block_on(spawned_fut).expect("We should always finish")?
```
joostjager, line 142: This is a sync method, so do we really need to worry about an outer runtime that we must call block_on on?

src/io/vss_store.rs
```diff
@@ -94,46 +112,122 @@ impl KVStoreSync for VssStore {
...
+		let internal_runtime = self.internal_runtime.as_ref().ok_or_else(|| {
+			debug_assert!(false, "Failed to access internal runtime");
+			let msg = format!("Failed to access internal runtime");
+			Error::new(ErrorKind::Other, msg)
+		})?;
+		let primary_namespace = primary_namespace.to_string();
+		let secondary_namespace = secondary_namespace.to_string();
+		let key = key.to_string();
+		let inner = Arc::clone(&self.inner);
+		let locking_key = self.build_locking_key(&primary_namespace, &secondary_namespace, &key);
```
joostjager, line 157: Is it possible to not duplicate this section?

src/lib.rs
tnull: Done.

tests/integration_tests_cln.rs
tnull: > Okay, but also nothing is actually running in parallel

Huh, why not? The background processor task and the test thread should introduce concurrency, no, etc? Or maybe I'm misunderstanding you?

src/io/vss_store.rs
tnull: Hmm, tbh. I don't have a super good rationale besides 'small number of threads that still allows progress if one would hit a blocking case'.

src/io/vss_store.rs
tnull: Done.

src/io/vss_store.rs
tnull: I'm not sure I'm following what you're asking here? We need to execute the future on our runtime and wait for the result, this is what `block_on` does.

src/io/vss_store.rs
tnull: Not quite sure what you mean? Which section?

src/runtime.rs
tnull: Done.

tests/integration_tests_cln.rs
joostjager: I mean if there is one worker thread, isn't it the case that no tasks actually run at the same time? Just that they interleave each other.

src/io/vss_store.rs
joostjager: Add that?

src/io/vss_store.rs
joostjager: This `read` is a sync function. I wondered whether we need to take into account the posssibility of an outer runtime really? I'd think that sync functions can be used without a runtime? Maybe that just how this is setup, but it feels a bit strange fundamentally.

src/io/vss_store.rs
joostjager: I mean that quite some code is duplicated between the various kv store methods. Maybe it is painful to reuse in rust though.

src/io/vss_store.rs
tnull: > This read is a sync function. I wondered whether we need to take into account the posssibility of an outer runtime really?

Yes, we do, as we have no clue what our users run our code in. They might not even have a direct dependency on LDK Node and unknowingly just run other code in a tokio context.

src/io/vss_store.rs
tnull: You mean the string allocations? FWIW, pulling out such small stuff into helper makes things harder to read, IMO, and would also only allow us to save ~3 lines per method or so?

src/io/vss_store.rs
tnull: Okay, if you insist. Not sure there is much benefit.

tnull: Done.

tests/integration_tests_cln.rs
tnull: > I mean if there is one worker thread, isn't it the case that no tasks actually run at the same time? Just that they interleave each other.

No, as at the very least we'll also have the test thread itself. Plus all of the threads spawned by `spawn_blocking`.

joostjager: Discussed offline. Both 1 and multiple have their own advantages in testing, but going for 1 thread to maximize deadlock potential if a deadlock would exist.

src/io/vss_store.rs
joostjager: Discussed offline. It should be possible to only use the internal runtime here, but in practice this created vague issues around timers, therefore the more complicated solution with two runtimes. It is temporary until the move to full async.

src/io/vss_store.rs
joostjager: Agreed that the savings are limited, and Rust isn't always that inviting to code reuse.

PR #674, opened 2025-10-22
Use `lightning::util::anchor_channel_reserves` 

### LINKED ISSUE #668: Make use of `lightning::util::anchor_channel_reserves` to improve anchor reserve estimation?
Rather than always requiring the user  a flat fee, we could consider to make use of the changes in 
https://github.com/lightningdevkit/rust-lightning/pull/3487 to allow for a more detailed estimation.

Camillarhi:
This update replaces the flat fee requirement for anchor channel reserves with a more accurate estimation using the new `lightning::util::anchor_channel_reserves` utility.

- Integrated anchor_channel_reserves for dynamic reserve estimation
- Removed flat static fee for anchors
- Updated relevant logic and documentation to reflect the new estimation approach

Closes #668

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -27,7 +27,6 @@ const DEFAULT_BDK_WALLET_SYNC_INTERVAL_SECS: u64 = 80;
 const DEFAULT_LDK_WALLET_SYNC_INTERVAL_SECS: u64 = 30;
 const DEFAULT_FEE_RATE_CACHE_UPDATE_INTERVAL_SECS: u64 = 60 * 10;
 const DEFAULT_PROBING_LIQUIDITY_LIMIT_MULTIPLIER: u64 = 3;
-const DEFAULT_ANCHOR_PER_CHANNEL_RESERVE_SATS: u64 = 25_000;
 
 /// The default log level.
 pub const DEFAULT_LOG_LEVEL: LogLevel = LogLevel::Debug;
@@ -233,7 +232,6 @@ impl Default for Config {
 /// | Parameter                  | Value  |
 /// |----------------------------|--------|
 /// | `trusted_peers_no_reserve` | []     |
-/// | `per_channel_reserve_sats` | 25000  |
 ///
 ///
 /// [BOLT 3]: https://github.com/lightning/bolts/blob/master/03-transactions.md#htlc-timeout-and-htlc-success-transactions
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -1593,6 +1594,12 @@ pub(crate) fn total_anchor_channels_reserve_sats(
 						.map_or(false, |t| t.requires_anchors_zero_fee_htlc_tx())
 			})
 			.count() as u64
-			* anchor_channels_config.per_channel_reserve_sats
+			* reserve_sat_per_channel
 	})
 }
+
+/// Returns the configured anchor channel reserve per channel in satoshis.
+pub fn get_anchor_reserve_per_channel() -> u64 {
```
tnull, line 1884: I don't think we should make this public, esp. given the only other call site is our tests.

tests/common/mod.rs
```diff
@@ -640,8 +641,9 @@ pub(crate) fn do_channel_full_cycle<E: ElectrumApi>(
 	let addr_a = node_a.onchain_payment().new_address().unwrap();
 	let addr_b = node_b.onchain_payment().new_address().unwrap();
 
-	let premine_amount_sat = if expect_anchor_channel { 2_125_000 } else { 2_100_000 };
+	let anchor_reserve = if expect_anchor_channel { get_anchor_reserve_per_channel() } else { 0 };
```
tnull, line 690: Rather than calling this to get the dynamic value, can we still keep a fixated value here that would also allow us to check if the method works as expected and detect if/when something changes.

src/config.rs
```diff
@@ -235,33 +233,11 @@ pub struct AnchorChannelsConfig {
...
-	/// spendable funds available in the on-chain wallet. Therefore, this per-channel multiplier is
-	/// really a emergency reserve that we maintain at all time to reduce reduce the risk of
-	/// insufficient funds at time of a channel closure. To this end, we will refuse to open
-	/// outbound or accept inbound channels if we don't have sufficient on-chain funds available to
-	/// cover the additional reserve requirement.
-	///
-	/// **Note:** Depending on the fee market at the time of closure, this reserve amount might or
-	/// might not suffice to successfully spend the Anchor output and have the HTLC transactions
-	/// confirmed on-chain, i.e., you may want to adjust this value accordingly.
-	pub per_channel_reserve_sats: u64,
```
tnull, line 270: I think we might want to keep this as an override and make it an `Option<u64>` so that user can still set a specific value if they want. But probably the default should be `None`, i.e., dynamic estimation.

src/lib.rs
```diff
@@ -1593,6 +1594,12 @@ pub(crate) fn total_anchor_channels_reserve_sats(
...
+	let context = AnchorChannelReserveContext::default();
+	get_reserve_per_channel(&context).to_sat()
```
tnull, line 1886: Hmm, it seems the current defaults would result in a channel reserve of 336513sats (~$300) per channel. This seems very conservative, and much too high for the typical use case we currently expect in LDK Node. This makes me wonder if we either need to use different defaults, or not opt for dynamic estimation right now. Will discuss this with a few other devs and get back to this.

Camillarhi: Just want to check if there is an update on this after discussing with the devs

tnull: Well, yes and no. I played around with the numbers here, and it seems we'll have to set really low `AnchorChannelReserveContext` parameters to arrive at reasonable (comaparable to our current default) numbers. So, just as before, it will be very hard to set sane default parameters that make sense for ~all or ~most users. It likely comes down to 'user profiles': if you run a forwarding node that is exposed to a lot of HTLC-related risks, you might want to bump the numbers or use `AnchorChannelReserveContext::default()` numbers. Most edge nodes however likely want to considerably reduce these numbers. 

For most users they'll have a very hard time to wrap their head around what risk they are taking and what numbers they should pick here, which is exactly what we hoped to improve upon by using the `anchor_channel_reserves` util. So, at some point it might make sense to introduce 'user profiles' as a general concept, but I'm not sure right now is the right time to do that. So I think we should just kick the can down the road a bit further, and revisit in a while. I'll make sure to update #668 accordingly, but for now I'm going ahead closing this exploratory PR - sorry!

PR #679, opened 2025-10-27
Pin-back `home` for MSRV

tnull:
Unfortunately, `home` bumped their MSRV to rustc 1.88 with their 0.5.12 patch release. Here, we pin it back to allow our CI to succeed under MSRV 1.85.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -41,10 +41,6 @@ jobs:
       - name: Check formatting on Rust ${{ matrix.toolchain }}
         if: matrix.check-fmt
         run: rustup component add rustfmt && cargo fmt --all -- --check
-      - name: Pin packages to allow for MSRV
-        if: matrix.msrv
-        run: |
-          cargo update -p home --precise "0.5.11" --verbose # home v0.5.12 requires rustc 1.88 or newer
       - name: Set RUSTFLAGS to deny warnings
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -41,6 +41,10 @@ jobs:
       - name: Check formatting on Rust ${{ matrix.toolchain }}
         if: matrix.check-fmt
         run: rustup component add rustfmt && cargo fmt --all -- --check
+      - name: Pin packages to allow for MSRV
+        if: matrix.msrv
+        run: |
+          cargo update -p home --precise "0.5.11" --verbose # home v0.5.12 requires rustc 1.88 or newer
```
joostjager: Isn't it a problem that (security?) patches are no longer coming in now?

tnull: > Isn't it a problem that (security?) patches are no longer coming in now?

No, a) this is a test-only dependency b) it would only be required for users building on MSRV c) `home` should be rather trivial in the first place.

However, we now opened and landed a PR upstream on the `electrsd` crate (https://github.com/RCasatta/electrsd/pull/114) that just removes the respective dependency entirely. Thankfully it was pushed out in a timely patch release, so we won't need the fix here after all. Pushed a revert commit to verify CI passes this step again, will close afterwards.

PR #682, opened 2025-10-28
Fix bitcoind shutdown hang

benthecarman:
Previously, the shutdown process could hang when using the bitcoind
    backend because the syncing process was not always checking if we had
    received the shutdown signal. Now with any future we call during the
    sync process we will use a `tokio::select!` to make sure we abort early
    if we receive a stop signal.

### DIFF
--- a/src/chain/bitcoind.rs
+++ b/src/chain/bitcoind.rs
@@ -147,6 +147,12 @@ impl BitcoindChainSource {
 		const MAX_BACKOFF_SECS: u64 = 300;
 
 		loop {
+			// if the stop_sync_sender has been dropped, we should just exit
+			if stop_sync_receiver.has_changed().unwrap_or(true) {
+				log_trace!(self.logger, "Stopping initial chain sync.");
+				return;
+			}
+
 			let channel_manager_best_block_hash = channel_manager.current_best_block().block_hash;
 			let sweeper_best_block_hash = output_sweeper.current_best_block().block_hash;
 			let onchain_wallet_best_block_hash =
@@ -226,7 +232,18 @@ impl BitcoindChainSource {
 									e,
 									backoff
 								);
-						tokio::time::sleep(Duration::from_secs(backoff)).await;
+						// Sleep with stop signal check to allow immediate shutdown
+						tokio::select! {
+							biased;
+							_ = stop_sync_receiver.changed() => {
+								log_trace!(
+									self.logger,
+									"Stopping initial chain sync.",
+								);
+								return;
+							}
+							_ = tokio::time::sleep(Duration::from_secs(backoff)) => {}
+						}
 						backoff = std::cmp::min(backoff * 2, MAX_BACKOFF_SECS);
 					} else {
 						log_error!(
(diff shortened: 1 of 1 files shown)


src/chain/bitcoind.rs
```diff
@@ -147,6 +147,19 @@ impl BitcoindChainSource {
 		const MAX_BACKOFF_SECS: u64 = 300;
 
 		loop {
+			// Check for stop signal at the beginning of each iteration
+			tokio::select! {
```
tnull: Why do we use a `select` here instead of just polling for `has_changed` once?

src/chain/bitcoind.rs
```diff
@@ -147,6 +147,19 @@ impl BitcoindChainSource {
 		const MAX_BACKOFF_SECS: u64 = 300;
 
 		loop {
+			// Check for stop signal at the beginning of each iteration
```
tnull: Please drop these redundant comments on all the inserted codeblocks.

src/chain/bitcoind.rs
```diff
@@ -226,7 +239,18 @@ impl BitcoindChainSource {
...
 									backoff
 								);
-						tokio::time::sleep(Duration::from_secs(backoff)).await;
+						// Sleep with stop signal check to allow immediate shutdown
+						tokio::select! {
+							biased;
+							_ = stop_sync_receiver.changed() => {
+								log_trace!(
+									self.logger,
+									"Stopping initial chain sync during backoff.",
```
tnull: I don't think it's interesting we do it 'during backoff'.

src/chain/bitcoind.rs
```diff
@@ -260,6 +295,8 @@ impl BitcoindChainSource {
 		let mut last_best_block_hash = None;
 		loop {
 			tokio::select! {
+				biased;
+
```
tnull: Please drop spurious newline.

src/chain/bitcoind.rs
```diff
@@ -268,17 +305,43 @@ impl BitcoindChainSource {
...
+						biased;
+						_ = stop_sync_receiver.changed() => {
+							log_trace!(
+								self.logger,
+								"Stopping polling for new chain data.",
+							);
+							return;
+						}
+						res = async {
+							if last_best_block_hash != Some(channel_manager.current_best_block().block_hash) {
```
tnull: This if should still be outside of the select.

src/chain/bitcoind.rs
```diff
@@ -268,17 +305,43 @@ impl BitcoindChainSource {
...
+								let update_res = self.update_fee_rate_estimates().await;
+								if update_res.is_ok() {
+									last_best_block_hash = Some(channel_manager.current_best_block().block_hash);
+								}
+							}
+						} => {
+							let _ = res;
```
tnull: Please avoid this, just don't bind `res` in the first place here.

src/chain/bitcoind.rs
```diff
@@ -268,17 +305,43 @@ impl BitcoindChainSource {
...
+								"Stopping polling for new chain data.",
+							);
+							return;
+						}
+						res = self.poll_and_update_listeners(
+							Arc::clone(&channel_manager),
+							Arc::clone(&chain_monitor),
+							Arc::clone(&output_sweeper)
+						) => {
+							let _ = res;
```
tnull: Sam here, don't bind `res`.

src/chain/bitcoind.rs
benthecarman: fixed

PR #683, opened 2025-10-29
Upgrade `rand` dependency, use `os_rng` for seed generation

tnull:
We bump our `rand` dependency to the latest stable version.

Additionally, we now use `os_rng` for seed/mnemonic generation. While the previously-used `thread_rng` should be totally fine, but `os_rng` is guaranteed to block until there is sufficient entropy available (e.g., after startup), which might slightly improve security here.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -1137,7 +1137,7 @@ where
 					}
 				}
 
-				let user_channel_id: u128 = rng().random::<u128>();
+				let user_channel_id: u128 = rng().random();
 				let allow_0conf = self.config.trusted_peers_0conf.contains(&counterparty_node_id);
 				let mut channel_override_config = None;
 				if let Some((lsp_node_id, _)) = self
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1117,7 +1117,7 @@ impl Node {
 		}
 
 		let push_msat = push_to_counterparty_msat.unwrap_or(0);
-		let user_channel_id: u128 = rand::rng().random::<u128>();
+		let user_channel_id: u128 = rand::rng().random();
 
 		match self.channel_manager.create_channel(
 			peer_info.node_id,
--- a/src/liquidity.rs
+++ b/src/liquidity.rs
@@ -562,7 +562,7 @@ where
 						return;
 					};
 
-					let user_channel_id: u128 = rand::rng().random::<u128>();
+					let user_channel_id: u128 = rand::rng().random();
 					let intercept_scid = self.channel_manager.get_intercept_scid();
 
 					if let Some(payment_size_msat) = payment_size_msat {


Cargo.toml
```diff
@@ -90,7 +90,7 @@ bip39 = "2.0.0"
 bip21 = { version = "0.5", features = ["std"], default-features = false }
 
 base64 = { version = "0.22.1", default-features = false, features = ["std"] }
-rand = "0.8.5"
+rand = { version = "0.9.2", default-features = false, features = ["std", "thread_rng", "os_rng"] }
```
tankyleo, line 93: Looks like `thread_rng` enables `std` and `os_rng`, do we still want to list them all out explicitly ?

src/event.rs
```diff
@@ -1137,7 +1137,7 @@ where
 					}
 				}
 
-				let user_channel_id: u128 = thread_rng().gen::<u128>();
+				let user_channel_id: u128 = rng().random::<u128>();
```
tankyleo: nit nit: we specify the type in two places here but we only need to specify it once. I am myself not a fan of the aesthetics of turbofish, but your call :)

It's like this in other diff'ed lines too.

src/io/vss_store.rs
```diff
@@ -592,7 +592,7 @@ pub(crate) struct RandEntropySource;
 
 impl EntropySource for RandEntropySource {
 	fn fill_bytes(&self, buffer: &mut [u8]) {
-		rand::thread_rng().fill_bytes(buffer);
+		rand::rng().fill_bytes(buffer);
```
tankyleo, line 595: In the next commit, I would have used `OsRng` here too since it's used for cryptographic nonces.

src/io/utils.rs
```diff
@@ -96,7 +97,10 @@ where
 		Ok(key)
 	} else {
 		let mut key = [0; WALLET_KEYS_SEED_LEN];
-		rng().fill_bytes(&mut key);
+		OsRng.try_fill_bytes(&mut key).map_err(|e| {
+			log_error!(logger, "Failed to generate entropy: {}", e);
+			std::io::Error::new(std::io::ErrorKind::Other, "Failed to generate seed bytes")
+		})?;
```
tankyleo, line 103: If we propagate the `Err` from the `OsRng` here should we also propagate the `Err` in `generate_entropy_mnemonic` above ?

src/io/vss_store.rs
tankyleo: I am conscious of the performance tradeoff here as this is a frequent operation, maybe `OsRng` would be overkill ? Cryptographic nonces have stronger requirements than the identifiers we use `ThreadRng` for elsewhere

src/io/utils.rs
tankyleo: I guess I am looking for consistency here, but maybe the difference is a public API function vs a function internal to the crate ?

tankyleo (commented):
Thank you !

Cargo.toml
tnull: I think so, yes. They are also listed like that in the default features.

src/io/utils.rs
tnull: > If we propagate the `Err` from the `OsRng` here should we also propagate the `Err` in `generate_entropy_mnemonic` above ?

Well, we would need to introduce an `enum` error type just for this case, which seems like overkill. It could even be argued that we *should* to panic in this case, as according to the docs of `OsRng` failing is only expected in case of system misconfiguration and we fail to actually provide sufficient entropy, i.e., every operation after this could be considered unsafe.

src/io/vss_store.rs
tnull: Yes, that's why I kept it with the thread RNG, which should be perfectly fine for cryptographic operations. I guess if we'd want to be super cautious we could look into periodic re-seeding via `ReseedingRng`, but not even sure if that is warranted. WDYT?

src/event.rs
tnull: Sure, removed the turbofish.

src/io/utils.rs
tankyleo: hmm ok would you consider panicking here ? sounds to me like `OsRng` returning an `Error` is very rare, but if it ever does, user should probably stop what they are doing and consider running `ldk-node` on a different machine.

src/io/vss_store.rs
tankyleo: Looked into the current `reseed` behaviour, it's already automatically called every 64kB, I think we are good there.

They do recommend to explicitly reseed on any fork, but I don't think we ever fork right ? If so I think I am good with the current choice.

tnull: > They do recommend to explicitly reseed on any fork, but I don't think we ever fork right ? If so I think I am good with the current choice.

Yeah, was also tripping over that and considered reseeding on any new thread spawned by tokio. But, IIUC, that should happen automatically in the `rng()` call.

src/io/utils.rs
tnull: > hmm ok would you consider panicking here ? sounds to me like `OsRng` returning an `Error` is very rare, but if it ever does, user should probably stop what they are doing and consider running `ldk-node` on a different machine.

Well, the other path will happen on `build` only, so instead of panicking we'd just be erroring out before we did anything really.

PR #687, opened 2025-10-29
Support client_trusts_lsp=true on ldk-node

martinsaposnic:
implement changes introduced on https://github.com/lightningdevkit/rust-lightning/pull/3838 for rust-lightning.

as discussed, client_trusts_lsp is a flag set at startup.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -542,8 +542,12 @@ where
 						match result {
 							Ok(()) => {},
 							Err(APIError::APIMisuseError { err }) => {
-								log_error!(self.logger, "Panicking due to APIMisuseError: {}", err);
-								panic!("APIMisuseError: {}", err);
+								log_error!(
+									self.logger,
+									"Encountered APIMisuseError, this should never happen: {}",
+									err
+								);
+								debug_assert!(false, "APIMisuseError: {}", err);
 							},
 							Err(APIError::ChannelUnavailable { err }) => {
 								log_error!(
@@ -571,7 +575,7 @@ where
 							)
 							.unwrap_or_else(|e| {
 								log_error!(self.logger, "Failed to force close channel after funding generation failed: {:?}", e);
-								panic!(
+								debug_assert!(false,
 									"Failed to force close channel after funding generation failed"
 								);
 							});


toneloc:
With LDK Node including this PR, I got a compilation error and had to add these imports

```
use crate::lightning::util::errors::APIError;
use crate::bitcoin::Transaction;
```
It compiled after that, then I tried to do LSPS2 JIT onboarding, but it failed ("route not found" from the client side).

Note sure if that's an issue with this PR or me.

martinsaposnic:
@tnull this should be ready now. let me know if you have comments

src/event.rs
```diff
@@ -506,12 +506,43 @@ where
 					locktime,
 				) {
 					Ok(final_tx) => {
-						// Give the funding transaction back to LDK for opening the channel.
-						match self.channel_manager.funding_transaction_generated(
-							temporary_channel_id,
-							counterparty_node_id,
-							final_tx,
-						) {
+						let needs_manual_broadcast =
```
tnull, line 513: nit: Would be cleaner if we did:

```diff
diff --git a/src/event.rs b/src/event.rs
index 7f467977..9a5516d8 100644
--- a/src/event.rs
+++ b/src/event.rs
@@ -506,20 +506,17 @@ where
                                        locktime,
                                ) {
                                        Ok(final_tx) => {
-                                               let needs_manual_broadcast =
-                                                       match self.liquidity_source.as_ref().map(|ls| {
-                                                               ls.as_ref().lsps2_channel_needs_manual_broadcast(
+                                               let needs_manual_broadcast = self
+                                                       .liquidity_source
+                                                       .as_ref()
+                                                       .and_then(|ls| {
+                                                               ls.lsps2_channel_needs_manual_broadcast(
                                                                        counterparty_node_id,
                                                                        user_channel_id,
                                                                )
-                                                       }) {
-                                                               Some(Ok(v)) => v,
-                                                               Some(Err(e)) => {
-                                                                       log_error!(self.logger, "Failed to determine if channel needs manual broadcast: {:?}", e);
-                                                                       false
-                                                               },
-                                                               None => false,
-                                                       };
+                                                               .ok()
+                                                       })
+                                                       .unwrap_or(false);

                                                let result = if needs_manual_broadcast {
                                                        self.liquidity_source.as_ref().map(|ls| {
```

src/liquidity.rs
```diff
@@ -147,6 +150,7 @@ where
 	kv_store: Arc<DynStore>,
 	config: Arc<Config>,
 	logger: L,
+	broadcaster: Arc<Broadcaster>,
```
tnull: No need to add this field if we already have `tx_broadcaster: Arc<Broadcaster>` above.

src/liquidity.rs
```diff
@@ -305,6 +310,79 @@ where
 		self.lsps2_client.as_ref().map(|s| (s.lsp_node_id, s.lsp_address.clone()))
 	}
 
+	pub(crate) fn lsps2_channel_needs_manual_broadcast(
+		&self, counterparty_node_id: PublicKey, user_channel_id: u128,
+	) -> Result<bool, APIError> {
+		// if we are not in a client_trusts_lsp model, we don't check and just return false
+		if !self.is_client_trusts_lsp() {
+			log_debug!(self.logger, "Skipping funding transaction broadcast as client trusts LSP.");
```
tnull: Huh, this is very confusing/wrong. We're actually auto-broadcasting, not manually/**witholding** the broadcast because we're in LSP-trusts-client, right?

Same below

src/liquidity.rs
```diff
@@ -305,6 +310,79 @@ where
 		self.lsps2_client.as_ref().map(|s| (s.lsp_node_id, s.lsp_address.clone()))
 	}
 
+	pub(crate) fn lsps2_channel_needs_manual_broadcast(
+		&self, counterparty_node_id: PublicKey, user_channel_id: u128,
+	) -> Result<bool, APIError> {
+		// if we are not in a client_trusts_lsp model, we don't check and just return false
```
tnull: Please drop this redundant comment, it doesn't explain anything beyond the code below, just potentially adds to the confusion.

src/liquidity.rs
```diff
@@ -305,6 +310,79 @@ where
 		self.lsps2_client.as_ref().map(|s| (s.lsp_node_id, s.lsp_address.clone()))
 	}
 
+	pub(crate) fn lsps2_channel_needs_manual_broadcast(
+		&self, counterparty_node_id: PublicKey, user_channel_id: u128,
+	) -> Result<bool, APIError> {
```
tnull: The `Result<.., APIError>` on `lightning-liquidity` pattern isn't great to begin with, but we should def. not have any methods in LDK Node using `APIError`. Please define an error variant if you really need one, but I suspect we don't actually need to.

Please clean this up, for example something like this
```diff
diff --git a/src/liquidity.rs b/src/liquidity.rs
index a293311d..ddd517b1 100644
--- a/src/liquidity.rs
+++ b/src/liquidity.rs
@@ -312,22 +312,18 @@ where

        pub(crate) fn lsps2_channel_needs_manual_broadcast(
                &self, counterparty_node_id: PublicKey, user_channel_id: u128,
-       ) -> Result<bool, APIError> {
-               // if we are not in a client_trusts_lsp model, we don't check and just return false
-               if !self.is_client_trusts_lsp() {
-                       log_debug!(self.logger, "Skipping funding transaction broadcast as client trusts LSP.");
-                       return Ok(false);
-               }
-
-               // if we are in a client_trusts_lsp model, then we check if the LSP has an LSPS2 operation in progress
-               self.lsps2_service.as_ref().map_or(Ok(false), |_| {
-                       let lsps2_service_handler = self.liquidity_manager.lsps2_service_handler();
-                       if let Some(handler) = lsps2_service_handler {
-                               handler.channel_needs_manual_broadcast(user_channel_id, &counterparty_node_id)
-                       } else {
-                               log_error!(self.logger, "LSPS2 service handler is not available.");
-                               Ok(false)
-                       }
+       ) -> bool {
+               self.lsps2_service.as_ref().map_or(false, |lsps2_service| {
+                       lsps2_service.service_config.client_trusts_lsp
+                               || self
+                                       .liquidity_manager()
+                                       .lsps2_service_handler()
+                                       .and_then(|handler| {
+                                               handler
+                                                       .channel_needs_manual_broadcast(user_channel_id, &counterparty_node_id)
+                                                       .ok()
+                                       })
+                                       .unwrap_or(false)
                })
        }
```
.. would be much more concise and readable, IMO. Also no need for 3-line util methods like `is_client_trusts_lsp` that just complicate things in this case.
 
(likewise below)

src/liquidity.rs
```diff
@@ -305,6 +310,79 @@ where
...
+			if let Some(handler) = lsps2_service_handler {
+				handler.channel_needs_manual_broadcast(user_channel_id, &counterparty_node_id)
+			} else {
+				log_error!(self.logger, "LSPS2 service handler is not available.");
+				Ok(false)
+			}
+		})
+	}
+
+	pub(crate) fn lsps2_store_funding_transaction(
```
tnull, line 327: Given that we use them both at the same callsite and for the same purpose, should we just merge `lsps2_channel_needs_manual_broadcast` / `lsps2_store_funding_transaction` to `broadcast_or_store_funding_transaction` and call that in the event handler? Then most of this redundant boilerplate goes away, we just check once what we should do and call the respective method on channelmanager here?

tests/integration_tests_rust.rs
```diff
@@ -1860,3 +1861,260 @@ async fn drop_in_async_context() {
 	let node = setup_node(&chain_source, config, Some(seed_bytes));
 	node.stop().unwrap();
 }
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn lsps2_client_trusts_lsp() {
```
tnull, line 1877: Can we simply DRY this up with `lsps2_client_server_integration` by renaming it `do_lsps2_client_server_integration` taking a `client_trusts_lsp: bool` that is set. Then we can simply add `lsps2_client_server_integration_client_trusts_lsp` that calls `do_lsps2_client_server_integration(true)`.

tests/integration_tests_rust.rs
```diff
@@ -1860,3 +1861,260 @@ async fn drop_in_async_context() {
...
+	println!("Paying JIT invoice!");
+	let _payment_id = payer_node.bolt11_payment().send(&res, None).unwrap();
+	expect_channel_pending_event!(service_node, client_node.node_id());
+	expect_channel_ready_event!(service_node, client_node.node_id());
+	expect_channel_pending_event!(client_node, service_node.node_id());
+	expect_channel_ready_event!(client_node, service_node.node_id());
+	println!("Waiting for funding transaction to be broadcast... It will be there because LSP trusts the client, even though the client has not claimed it yet.");
+	let mut funding_tx_found = false;
+	for _ in 0..500 {
+		std::thread::sleep(std::time::Duration::from_millis(100));
```
tnull: Ugh, let's avoid too many sleeps in tests please, as over time they will considerably slow down our CI for no reason. Please use `wait_for_tx` for stuff like this, which at least does some polling with back-off.

tests/integration_tests_rust.rs
```diff
@@ -1860,3 +1861,260 @@ async fn drop_in_async_context() {
...
+			funding_tx_found = true;
+			break;
+		}
+	}
+
+	assert!(funding_tx_found, "Funding transaction should be broadcast after the client claims it");
+}
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn lsps2_lsp_trusts_client_but_client_does_not_claim() {
```
tnull, line 2050: Hmm, as mentioned above, I'd prefer if we could add the mempool check based on the flag in `do_lsps2_client_server_integration(false)`. I don't think this needs yet another full test case, as `lsps2_client_service_integration` provides the same (actually more) coverage here already.

src/event.rs
```diff
@@ -511,19 +511,12 @@ where
 				) {
 					Ok(final_tx) => {
 						let needs_manual_broadcast =
-							match self.liquidity_source.as_ref().map(|ls| {
+							self.liquidity_source.as_ref().map_or(false, |ls| {
```
joostjager, line 514: Don't you want to surface the error in the log?

src/liquidity.rs
```diff
@@ -310,22 +310,18 @@ where
...
-		self.lsps2_service.as_ref().map_or(Ok(false), |_| {
-			let lsps2_service_handler = self.liquidity_manager.lsps2_service_handler();
-			if let Some(handler) = lsps2_service_handler {
-				handler.channel_needs_manual_broadcast(user_channel_id, &counterparty_node_id)
-			} else {
-				log_error!(self.logger, "LSPS2 service handler is not available.");
-				Ok(false)
-			}
+	) -> bool {
+		self.lsps2_service.as_ref().map_or(false, |lsps2_service| {
```
joostjager, line 313: Same q about error logging.

tests/integration_tests_rust.rs
```diff
@@ -1969,21 +1968,41 @@ async fn lsps2_client_trusts_lsp() {
...
-		let current_mempool = bitcoind.client.get_raw_mempool().unwrap();
-		if current_mempool.0.len() > initial_mempool_size {
-			funding_tx_found = true;
-			break;
-		}
+	for _ in 0..30 {
+		tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+		let mempool = bitcoind.client.get_raw_mempool().unwrap().into_model().unwrap();
+		let funding_tx_found = mempool.0.iter().any(|txid| *txid == funding_txo.txid);
+		assert!(!funding_tx_found, "Funding transaction should NOT be broadcast yet");
```
joostjager: NIcer mempool presence test 👍

tests/integration_tests_rust.rs
```diff
@@ -1969,21 +1968,41 @@ async fn lsps2_client_trusts_lsp() {
...
-	let mut funding_tx_found = false;
-	for _ in 0..50 {
-		std::thread::sleep(std::time::Duration::from_millis(100));
-		let current_mempool = bitcoind.client.get_raw_mempool().unwrap();
-		if current_mempool.0.len() > initial_mempool_size {
-			funding_tx_found = true;
-			break;
-		}
+	for _ in 0..30 {
+		tokio::time::sleep(std::time::Duration::from_millis(100)).await;
```
joostjager: In the happy flow, we are always waiting 3s I think. Is a loop necessary then? Maybe just delay 3s and check.

tests/integration_tests_rust.rs
```diff
@@ -1969,21 +1968,41 @@ async fn lsps2_client_trusts_lsp() {
...
+		service_node
+			.list_channels()
+			.iter()
+			.find(|c| c.counterparty_node_id == client_node_id)
+			.unwrap()
+			.confirmations,
+		Some(0)
+	);
+
+	// No claim the JIT payment, which should release the funding transaction
```
joostjager: No -> Now?

tests/integration_tests_rust.rs
tnull: Done.

tests/integration_tests_rust.rs
tnull: Fixed, plus another typo

src/event.rs
tnull: Hmm, no, we'd get an `APIError` here if we call this for any counterparty that isn't tracked by the `LSPS2ServiceHandler`. So we shouldn't log an error at this step at all, as it would otherwise surface for any non-JIT-related channel.

src/liquidity.rs
tnull: See above, I think continuing and just logging on the manual-broadcast step is preferable.

toneloc:
> Awesome guys! Thanks. Could I test this from main? Any config / upgrade considerations to run? Or just the "client_trusts_lsp" flag to "true"?

I'm curious as to how this onboarding flow should work. Would love some docs helping users understand the flow (happy path, unhappy path), and ready help in that regard.

toneloc:
Thank you. Will continue conversation at #707.

PR #688, opened 2025-10-30
Add simple workflow to generate Swift bindings in CI

tnull:
We add a simple workflows to generate the Swift bindings in CI, ensuring the build is succeeding.

### DIFF
--- a/bindings/ldk_node.udl
+++ b/bindings/ldk_node.udl
@@ -771,7 +771,7 @@ interface Offer {
 	constructor([ByRef] string offer_str);
 	OfferId id();
 	boolean is_expired();
-	string? description();
+	string? offer_description();
 	string? issuer();
 	OfferAmount? amount();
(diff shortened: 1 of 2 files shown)


bindings/ldk_node.udl
```diff
@@ -771,7 +771,7 @@ interface Offer {
 	constructor([ByRef] string offer_str);
 	OfferId id();
 	boolean is_expired();
-	string? description();
+	string? offer_description();
```
jkczyz, line 774: CI is failing. Looks like some tests need to be updated, e.g. https://github.com/lightningdevkit/ldk-node/blob/134b2f7fc88a205538d744e785d3127d448089a8/src/ffi/types.rs#L1364

Same for the other structs.

tnull: Ah, right. Force-pushed fixes.

jkczyz (approved):

PR #689, opened 2025-10-31
Implement `lazy` deletes for `VssStore`

tnull:
Now based on #691 

We bump the LDK depedencies to the just-released 2.0 release candidate and account for minor last-minute API changes.

Furthermore, we implement `lazy` deletion for `VssStore` by tracking pending lazy deletes and supplying them as `delete_items` on the next `put` operation.

### DIFF
--- a/src/io/vss_store.rs
+++ b/src/io/vss_store.rs
@@ -666,4 +666,87 @@ mod tests {
 		do_read_write_remove_list_persist(&vss_store);
 		drop(vss_store)
 	}
+
+	#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+	async fn vss_lazy_delete() {
+		let vss_base_url = std::env::var("TEST_VSS_BASE_URL").unwrap();
+		let mut rng = rng();
+		let rand_store_id: String = (0..7).map(|_| rng.sample(Alphanumeric) as char).collect();
+		let mut vss_seed = [0u8; 32];
+		rng.fill_bytes(&mut vss_seed);
+		let header_provider = Arc::new(FixedHeaders::new(HashMap::new()));
+		let logger = Arc::new(Logger::new_log_facade());
+		let runtime = Arc::new(Runtime::new(logger).unwrap());
+		let vss_store =
+			VssStore::new(vss_base_url, rand_store_id, vss_seed, header_provider, runtime);
+
+		let primary_namespace = "test_namespace";
+		let secondary_namespace = "";
+		let key_to_delete = "key_to_delete";
+		let key_for_trigger = "key_for_trigger";
+		let data_to_delete = b"data_to_delete".to_vec();
+		let trigger_data = b"trigger_data".to_vec();
+
+		// Write the key that we'll later lazily delete
+		KVStore::write(
+			&vss_store,
+			primary_namespace,
+			secondary_namespace,
+			key_to_delete,
+			data_to_delete.clone(),
+		)
+		.await
+		.unwrap();
+
+		// Verify the key exists
+		let read_data =
+			KVStore::read(&vss_store, primary_namespace, secondary_namespace, key_to_delete)
+				.await
+				.unwrap();
+		assert_eq!(read_data, data_to_delete);
+
+		// Perform a lazy delete
+		KVStore::remove(&vss_store, primary_namespace, secondary_namespace, key_to_delete, true)
+			.await
+			.unwrap();
+
+		// Verify the key still exists (lazy delete doesn't immediately remove it)
+		let read_data =
+			KVStore::read(&vss_store, primary_namespace, secondary_namespace, key_to_delete)
+				.await
+				.unwrap();
+		assert_eq!(read_data, data_to_delete);
+
+		// Verify the key is still in the list
+		let keys = KVStore::list(&vss_store, primary_namespace, secondary_namespace).await.unwrap();
+		assert!(keys.contains(&key_to_delete.to_string()));
+
+		// Trigger the actual deletion by performing a write operation
+		KVStore::write(
+			&vss_store,
+			primary_namespace,
+			secondary_namespace,
+			key_for_trigger,
+			trigger_data.clone(),
+		)
+		.await
+		.unwrap();
+
+		// Now verify the key is actually deleted
+		let read_result =
+			KVStore::read(&vss_store, primary_namespace, secondary_namespace, key_to_delete).await;
+		assert!(read_result.is_err());
+		assert_eq!(read_result.unwrap_err().kind(), ErrorKind::NotFound);
+
+		// Verify the key is no longer in the list
+		let keys = KVStore::list(&vss_store, primary_namespace, secondary_namespace).await.unwrap();
+		assert!(!keys.contains(&key_to_delete.to_string()));
+
+		// Verify the trigger key still exists
+		let read_data =
+			KVStore::read(&vss_store, primary_namespace, secondary_namespace, key_for_trigger)
+				.await
+				.unwrap();
+		assert_eq!(read_data, trigger_data);
+	}
 }


src/io/vss_store.rs
```diff
@@ -451,6 +456,12 @@ impl VssStoreInner {
 			"write",
 		)?;
 
+		let delete_items = self
+			.pending_lazy_deletes
+			.try_lock()
+			.ok()
+			.and_then(|mut guard| guard.take())
```
tnull, line 463: Hmm, we might lose some `lazy` deletes if the write below would fail. I do wonder if we should go out of our way to restore the pending items in such a case, or if we're fine just leaning into the 'may or may not succeed' API contract here.

@joostjager Any opinion?

Similiarly, I do wonder if we should `spawn`-and-forget some tasks on `Drop` to attempt cleaning up the pending deletes on shutdown?

joostjager (commented):
I wonder if lazy delete is needed for VSS? Maybe we can just ignore the flag and delete sync always.

tnull:
> I wonder if lazy delete is needed for VSS? Maybe we can just ignore the flag and delete sync always.

Well, esp. now that we're on `MonitorUpdatingPersister`, it will have performance benefits for cleanup.

joostjager:
I just don't know if it is worth it for end-user nodes to worry about this. Even persisting full monitors always apparently was good enough performance? 

In this PR the delete semantics are again different, because a write is needed before deletes are executed. It doesn't make it easier to understand.

src/io/vss_store.rs
joostjager: Good question how loose we can get away with. Those keys are then never deleted anymore, which isn't great? I am not sure.

src/io/vss_store.rs
```diff
@@ -451,6 +456,12 @@ impl VssStoreInner {
 			"write",
 		)?;
 
+		let delete_items = self
+			.pending_lazy_deletes
+			.try_lock()
```
joostjager, line 461: I briefly thought "oh why can't we do this inside the lock that we already obtain", but that doesn't work because here we are also processing deletes on other keys.

src/io/vss_store.rs
```diff
@@ -491,13 +502,19 @@ impl VssStoreInner {
 			"remove",
 		)?;
 
+		let obfuscated_key =
+			self.build_obfuscated_key(&primary_namespace, &secondary_namespace, &key);
+
+		let key_value = KeyValue { key: obfuscated_key, version: -1, value: vec![] };
```
joostjager, line 513: Can't we store just keys here?

joostjager (commented):
Code looks good. Just not sure about the implications of lingering data if something goes wrong / a write never happens.

In particular for the incremental channel updates, it would be good if that is always followed by a write somehow without much delay. Idk if that's indeed the case?

src/io/vss_store.rs
enigbe: What do you both think about re-adding the `delete_items` back to `pending_lazy_deletes` if write fails? We get to retry deleting them on a subsequent write attempt.

src/io/vss_store.rs
tnull: Currently we might get away with it, but going forward we might utilize the `version` field there, so I'd prefer to do the future-safer thing here.

src/io/vss_store.rs
tnull: > What do you both think about re-adding the `delete_items` back to `pending_lazy_deletes` if write fails? We get to retry deleting them on a subsequent write attempt.

Yeah, that's the straightforward approach. Technically, we wouldn't even need to do this according to the API contract, but it's probably best to keep it best-effort. Blocking on re-acquiring the `pending_lazy_deletes` mutex in `write` isn't too great as it suddenly creates yet another interdependency between `remove` and `write`, but it should be safe here.

Went ahead and did that now.

src/io/vss_store.rs
```diff
@@ -666,4 +666,87 @@ mod tests {
...
+				.await
+				.unwrap();
+		assert_eq!(read_data, data_to_delete);
+
+		// Verify the key is still in the list
+		let keys = KVStore::list(&vss_store, primary_namespace, secondary_namespace).await.unwrap();
+		assert!(keys.contains(&key_to_delete.to_string()));
+
+		// Trigger the actual deletion by performing a write operation
+		KVStore::write(
```
joostjager, line 725: You could do a failed write first perhaps, to verify that the fixup commit works.

src/io/vss_store.rs
joostjager: You say 'might'. Do you mean that just using keys can already be a problem without the version field?

Even if we use the version field in the future, `value` still isn't needed?

In general, I would avoid dead code (or fields in this case). It just raises questions with devs. At least it did with me. Or otherwise clearly document.

Final question - is a version needed for lazy deletes?

src/io/vss_store.rs
```diff
@@ -475,10 +475,15 @@ impl VssStoreInner {
 					version: vss_version,
 					value: storable.encode_to_vec(),
 				}],
-				delete_items,
+				delete_items: delete_items.clone(),
 			};
 
 			self.client.put_object(&request).await.map_err(|e| {
+				// Restore delete items so they'll be retried on next write.
```
joostjager, line 482: I don't think a 'best effort' clean up is great, but probably acceptable.

tnull: > I don't think a 'best effort' clean up is great, but probably acceptable.

Hmm, we could discuss that at length, but that is what the `lazy` flag fundamentally is about no?

src/io/vss_store.rs
tnull: > You say 'might'. Do you mean that just using keys can already be a problem without the version field?

No, but it's something we could easily overlook when making use of the version going forward.

> Even if we use the version field in the future, `value` still isn't needed?

Well, whether the API design of `KeyValue` is good or not is debatable (I didn't invent it), but it is the type that we use to reference entries in the store. I.e., it is the thing used in `remove`/`DeleteObjectRequest`.

> Final question - is a version needed for lazy deletes?

Well, if we issue a delete on a certain version, the service could in the future ignore a delete if it already has a higher version, no? I.e., it would thereby detect that the value has been updated since the delete was issued.

src/io/vss_store.rs
joostjager: I don't think lazy means that it may or may not happen? I thought it means that it is going to happen, but unclear when?

src/io/vss_store.rs
tnull: Discussed offline, concluded that testing this particular behavior is currently not worth adding a lot of test boilerplate code (which this would likely require).

src/io/vss_store.rs
tnull: > I don't think lazy means that it may or may not happen? I thought it means that it is going to happen, but unclear when?

See the docs on `KVStore::remove`:

> Note that while setting the lazy flag reduces the I/O burden of multiple subsequent remove calls, it also influences the atomicity guarantees as lazy removes could potentially get lost on crash after the method returns. Therefore, this flag should only be set for remove operations that can be safely replayed at a later time.

joostjager: Ah ok. Maybe `lazy` isn't the most descriptive term. 

Is replay implemented for all lazily deleted data?

tnull: > Ah ok. Maybe `lazy` isn't the most descriptive term.
> 
> Is replay implemented for all lazily deleted data?

I think so. Currently we only use it for cleanup in `MonitorUpdatingPersister` and `LSPS2ServiceHandler`, and in both cases we'd worst case re-read the data on startup and prune/remove it again if its stale AFAIU.

joostjager: Perfect

src/io/vss_store.rs
joostjager: We could map the key to `KeyValue` at a later point, and not use it internally? Not a blocker.

>Well, if we issue a delete on a certain version, the service could in the future ignore a delete if it already has a higher version, no? I.e., it would thereby detect that the value has been updated since the delete was issued.

I thought we didn't want rewrites of the same key after it has been deleted lazily, but not 100% sure anymore. There's no version currently anyway, so a worry for later.

PR #693, opened 2025-11-03
Only use internal runtime in `VssStore`

tnull:
We previously attempted to drop the internal runtime from `VssStore`, resulting into blocking behavior. While we recently made changes that improved our situation (having VSS CI pass again pretty reliably), we just ran into yet another case where the VSS CI hung (cf. https://github.com/lightningdevkit/vss-server/actions/runs/19023212819/job/54322173817?pr=59).

Here we attempt to restore even more of the original pre- ab3d78d1ecd05a755c836915284e5ca60c65692a / #623 behavior to get rid of the reappearing blocking behavior, i.e., only use the internal runtime in `VssStore`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,7 +101,9 @@ serde = { version = "1.0.210", default-features = false, features = ["std", "der
 serde_json = { version = "1.0.128", default-features = false, features = ["std"] }
 log = { version = "0.4.22", default-features = false, features = ["std"]}
 
-vss-client = "0.3"
+#vss-client = "0.3"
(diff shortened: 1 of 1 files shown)


tnull:
> IIRC the previous hang was reproducible locally. Is that still the case with the hang that this PR fixes?

Unfortunately it's not. And after kicking CI on https://github.com/lightningdevkit/vss-server/pull/59, the flake went away (just as CI in generally now seemed to 'reliably' pass). I am worried though that it might reappear randomly in production.

PR #694, opened 2025-11-05
Use `[patch]` instead of switching all the dependencies

tnull:
We previously added a bunch of commented-out `rust-lightning` dependencies in our `Cargo.toml` to be able to easily switch between `rust-lightning` locations. However, this is exactly what the `[patch]` command is for, which in particular also allows to patch a dependency for the whole tree, not only this one project.

Therefore, we move the examples to a commented-out `patch` section.

### DIFF
--- a/.github/workflows/cron-weekly-rustfmt.yml
+++ b/.github/workflows/cron-weekly-rustfmt.yml
@@ -1,4 +1,9 @@
 name: Nightly rustfmt
+
+permissions:
+  contents: write
+  pull-requests: write
+
 on:
   schedule:
     - cron: "0 0 * * 0" # runs weekly on Sunday at 00:00


.github/workflows/cron-weekly-rustfmt.yml
```diff
@@ -1,4 +1,9 @@
 name: Nightly rustfmt
+
+permissions:
+  contents: write
+  pull-requests: write
```
joostjager, line 5: Maybe run it hourly temporarily to see if this fixes it?

tnull: Well, no, after this is merged I can just manually run it in Github UI.

PR #697, opened 2025-11-10
Make `EventQueue` persistence `async`

tnull:
Previously, we'd still use `KVStoreSync` for persistence of our event queue, which also meant calling the sync persistence through our otherwise-async background processor/event handling flow.

Here we switch our `EventQueue` persistence to be async, which gets us one step further towards async-everything.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -26,7 +26,7 @@ use lightning::util::config::{
 	ChannelConfigOverrides, ChannelConfigUpdate, ChannelHandshakeConfigUpdate,
 };
 use lightning::util::errors::APIError;
-use lightning::util::persist::KVStoreSync;
+use lightning::util::persist::KVStore;
 use lightning::util::ser::{Readable, ReadableArgs, Writeable, Writer};
 use lightning_liquidity::lsps2::utils::compute_opening_fee;
 use lightning_types::payment::{PaymentHash, PaymentPreimage};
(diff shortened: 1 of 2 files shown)


src/event.rs
```diff
@@ -301,12 +301,14 @@ where
 		Self { queue, waker, kv_store, logger }
 	}
 
-	pub(crate) fn add_event(&self, event: Event) -> Result<(), Error> {
-		{
+	pub(crate) async fn add_event(&self, event: Event) -> Result<(), Error> {
```
joostjager, line 304: I was looking at what's left in EventQueue that is sync. Just `next_event`. Unrelated to this PR, but in every test where that method is used, the test is already async. So it could use the async version?

src/lib.rs
```diff
@@ -777,7 +777,10 @@ impl Node {
 	///
 	/// **Note:** This **MUST** be called after each event has been handled.
 	pub fn event_handled(&self) -> Result<(), Error> {
-		self.event_queue.event_handled().map_err(|e| {
+		// We use our runtime for the sync variant to ensure `tokio::task::block_in_place` is
+		// always called if we'd ever hit this in an outer runtime context.
```
joostjager, line 781: Every time I read this, the same question marks pop up. It feels so strange to have regular sync code and worry about an outer runtime. And then we also make it more complicated because our own `block_on` variant isn't using the internal handle if there is already a context.

joostjager (approved):
Straight forward change

src/event.rs
tnull: > I was looking at what's left in EventQueue that is sync. Just `next_event`. Unrelated to this PR, but in every test where that method is used, the test is already async. So it could use the async version?

But `next_event` doesn't do anything that would require async operation? It doesn't repersist or anything?

joostjager: Shouldn't we set the example and use the async api of eventqueue?

joostjager (approved):

src/event.rs
joostjager: Ah maybe the naming isn't optimally clear. `next_event_async` waits, and `next_event` does not?

tnull: The former returns a future, the second one could be named `try_next_event`.

joostjager: Yes that would be an improvement indeed.

PR #699, opened 2025-11-10
Add support for configurable BIP39 mnemonic word counts

coreyphillips:
## Summary

This PR introduces support for generating BIP39 mnemonics with configurable word counts - **12, 15, 18, 21, or 24 words**.  
The `word_count` parameter is optional and defaults to **24 words (256-bit entropy)**.

## Motivation

Different use cases benefit from different mnemonic lengths:
- **12 words:** Simplified UX and easier memorization for users *(default in Bitkit)*  
- Supporting configurable lengths increases flexibility for downstream applications integrating with this library.

## Changes

- Added `WordCount` enum (`Words12`, `Words15`, `Words18`, `Words21`, `Words24`)
- Updated `generate_entropy_mnemonic()` to accept an **optional** `word_count` parameter
- Default: **24 words** when no parameter is provided
- Updated UDL bindings to expose the new enum and function signature
- Updated tests to cover all variants and validate entropy mapping

## Breaking Changes

⚠️ **This is a breaking change** - `generate_entropy_mnemonic()` now requires a `word_count` parameter.

Users will need to update their code from:
```rust
let mnemonic = generate_entropy_mnemonic();
```

To:
```rust
let mnemonic = generate_entropy_mnemonic(None);  // Uses default 24 words
// Or specify a word count:
let mnemonic = generate_entropy_mnemonic(Some(WordCount::Words12));
```

## Testing

- Verified default behavior (24 words)
- Added tests for all word count variants
- All tests pass (`cargo test`)

_Summary was co-created with AI_

### DIFF
--- a/src/io/utils.rs
+++ b/src/io/utils.rs
@@ -47,13 +47,15 @@ use crate::io::{
 };
 use crate::logger::{log_error, LdkLogger, Logger};
 use crate::peer_store::PeerStore;
-use crate::types::{Broadcaster, DynStore, KeysManager, Sweeper};
+use crate::types::{Broadcaster, DynStore, KeysManager, Sweeper, WordCount};
 use crate::wallet::ser::{ChangeSetDeserWrapper, ChangeSetSerWrapper};
 use crate::{Error, EventQueue, NodeMetrics, PaymentDetails};
 
 pub const EXTERNAL_PATHFINDING_SCORES_CACHE_KEY: &str = "external_pathfinding_scores_cache";
 
-/// Generates a random [BIP 39] mnemonic.
+/// Generates a random [BIP 39] mnemonic with the specified word count.
+///
+/// If no word count is specified, defaults to 24 words (256-bit entropy).
 ///
 /// The result may be used to initialize the [`Node`] entropy, i.e., can be given to
 /// [`Builder::set_entropy_bip39_mnemonic`].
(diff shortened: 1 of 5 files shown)


src/io/utils.rs
```diff
@@ -47,23 +47,26 @@ use crate::io::{
...
 ///
 /// [BIP 39]: https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki
 /// [`Node`]: crate::Node
 /// [`Builder::set_entropy_bip39_mnemonic`]: crate::Builder::set_entropy_bip39_mnemonic
-pub fn generate_entropy_mnemonic() -> Mnemonic {
-	// bip39::Mnemonic supports 256 bit entropy max
-	let mut entropy = [0; 32];
+pub fn generate_entropy_mnemonic(word_count: Option<MnemonicWordCount>) -> Mnemonic {
+	let word_count = word_count.unwrap_or(MnemonicWordCount::Words24);
+	let entropy_bytes = word_count.entropy_bytes();
```
tnull: Hmm, rather than dealing with different entropy byte lengths here, can we just use `Mnemonic::generate` which allows us to specify the word count? Having an `enum` for it still makes sense (actually we might want to switch to that in the `bip39` API https://github.com/rust-bitcoin/rust-bip39/issues/97).

tnull (commented):
Thanks, I think it makes sense to add this option.

Note that we appreciate human-written descriptions and when AI agent co-authorship is disclosed in the commit message.

src/io/utils.rs
coreyphillips: The latest commit [here](https://github.com/lightningdevkit/ldk-node/pull/699/commits/0bf87217e5ab4be59ca81388b36150e35ffc3e07) passes the `WordCount` directly to `Mnemonic::generate()` as suggested. However, I'm not sure how you may feel about the function name `generate_entropy_mnemonic` after this change. I've left it as is for now.

I've also created a PR adding the `enum WordCount` to `rust-bip39` [here](https://github.com/rust-bitcoin/rust-bip39/pull/98).

coreyphillips:
Squashed as requested.

PR #702, opened 2025-11-13
Allow to set optional `RouteParametersConfig` in BOLT12 API

### LINKED ISSUE #347: Allow to set `SendingParameters` in BOLT12 send methods
In #336 we added support for specifying `SendingParameters` for BOLT11 and spontaneous payments.

Unfortunately, the BOLT12 API currently doesn't allow to specify these fields, but we should also add them to BOLT12 sending methods once it becomes available in LDK (see tracking issue https://github.com/lightningdevkit/rust-lightning/issues/3262)

tnull:
Closes #347.

Previously, LDK only allowed to set this for BOLT11 payments. Since we now can, we allow to specify the `RouteParametersConfig` in BOLT12 and `UnifiedQrPayment` APIs.

### DIFF
--- a/src/payment/bolt12.rs
+++ b/src/payment/bolt12.rs
@@ -23,7 +23,7 @@ use lightning::util::ser::{Readable, Writeable};
 use lightning_types::string::UntrustedString;
 use rand::RngCore;
 
-use crate::config::{AsyncPaymentsRole, LDK_PAYMENT_RETRY_TIMEOUT};
+use crate::config::{AsyncPaymentsRole, Config, LDK_PAYMENT_RETRY_TIMEOUT};
 use crate::error::Error;
 use crate::ffi::{maybe_deref, maybe_wrap};
 use crate::logger::{log_error, log_info, LdkLogger, Logger};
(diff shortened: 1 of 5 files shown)


src/payment/bolt12.rs
```diff
@@ -74,8 +75,12 @@ impl Bolt12Payment {
 	/// response.
 	///
 	/// If `quantity` is `Some` it represents the number of items requested.
+	///
+	/// If `route_parameters` are provided they will override the default as well as the
+	/// node-wide parameters configured via [`Config::route_parameters`] on a per-field basis.
 	pub fn send(
 		&self, offer: &Offer, quantity: Option<u64>, payer_note: Option<String>,
+		route_parameters: Option<RouteParametersConfig>,
```
jkczyz, line 83: Not sure your feeling on this, but could avoid an additional `Option` parameter by chaining a call when creating `Bolt12Payment`:

```rust
node.bolt12_payment().with_route_params(route_params).send()
```

Though maybe an argument against is that it is specific to sending so not relevant to other calls. At very least, these aren't specific to a given `Offer`.

tnull: Yeah, I guess that would make sense if we did the builder-style refactor eventually. Not necessarily opposed to it, but IMO if we would go in that direction we should probably consider refactor the payment APIs to be fully in builder -pattern style.

jkczyz (approved):

PR #703, opened 2025-11-13
Add explicit type definition for ChannelDetails

chuksys:
This change uses an alias (`LdkChannelDetails`) and an explicit `Vec<LdkChannelDetails>` type annotation for `open_channels` in `close_channel_internal` and `update_channel_config`. This resolves type ambiguity caused by a name collision with the local `ChannelDetails` struct, which prevents rust-analyzer from correctly inferring the type as `Vec`, leading to an incorrect _len() is private_ error.

This is a minor fix that would improve the development experience.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -133,7 +133,7 @@ use io::utils::write_node_metrics;
 use lightning::chain::BestBlock;
 use lightning::events::bump_transaction::Wallet as LdkWallet;
 use lightning::impl_writeable_tlv_based;
-use lightning::ln::channel_state::ChannelShutdownState;
+use lightning::ln::channel_state::{ChannelDetails as LdkChannelDetails, ChannelShutdownState};
 use lightning::ln::channelmanager::PaymentId;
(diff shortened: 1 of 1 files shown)


tnull (commented):
Hmm, can't this be fixed on the `rust-analyzer` side of things? Not a big fan of making changes just for tooling, especially given that every contributor might run slightly different setups.

chuksys:
I totally get not wanting to fix things just for tooling.

I think `rust-analyzer` struggling to infer the correct type likely means we might want to make the code clearer here by using an explicit alias so that it's easy for anyone to know that what is returned here is a `Vec` of `ChannelDetails` from LDK and not the local wrapper.

Also I think that since `rust-analyzer` is a relatively popular tool (over 5 million installs), this small change gives us better readability in addition to the significant developer experience gain! 

Let me know if that makes sense - would be happy to hold-off on this if otherwise.

chuksys:
Thank you!

PR #706, opened 2025-11-18
`bitcoind` chain source: Poll tip before intialization

### LINKED ISSUE #415: Consider introducing checkpoints to shorten initial header validation
Currently, we initialize all chain listeners with the genesis hash, which results in a complete validation of the header chain upon first startup.

We should consider a way to cut this short. Unfortunately it's not as easy as polling the best tip and initializing the chain listeners with the most-recent tip as we don't have the chain source available during initialization in the builder (cf. https://github.com/lightningdevkit/ldk-node/pull/414#issuecomment-2511801852). 

However, one way to improve the initial startup time could be to introduce a hard-coded, more recent, checkpoint with which we initialize all the chain listeners.

tnull:
Closes #415.

Previously, we couldn't poll the chain tip in `Builder::build` as we wouldn't have a runtime available. Since we now do, we can at least attempt to poll for the chain tip before initializing objects, avoiding that fresh nodes need to re-validate everything from genesis.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1195,10 +1195,10 @@ fn build_with_store_internal(
 		},
 	};
 
-	let chain_source = match chain_data_source_config {
+	let (chain_source, chain_tip_opt) = match chain_data_source_config {
 		Some(ChainDataSourceConfig::Esplora { server_url, headers, sync_config }) => {
 			let sync_config = sync_config.unwrap_or(EsploraSyncConfig::default());
-			Arc::new(ChainSource::new_esplora(
+			ChainSource::new_esplora(
 				server_url.clone(),
 				headers.clone(),
 				sync_config,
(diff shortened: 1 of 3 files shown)


src/builder.rs
```diff
@@ -1499,13 +1524,13 @@ fn build_with_store_internal(
 			channel_manager
 		} else {
 			// We're starting a fresh node.
-			let genesis_block_hash =
-				bitcoin::blockdata::constants::genesis_block(config.network).block_hash();
+			let best_block = chain_tip_opt.unwrap_or_else(|| {
+				let genesis_block_hash =
+					bitcoin::blockdata::constants::genesis_block(config.network).block_hash();
+				BestBlock::new(genesis_block_hash, 0)
```
benthecarman: you can just do `BestBlock::from_network` instead of grabbing the genesis block yourself

tnull: Done

benthecarman (approved):

PR #707, opened 2025-11-19
Expand docs on `LSPS2ServiceConfig::client_trusts_lsp` field

tnull:
Previously the docs have been a bit sparse. Now that we actually implement the client-trusts-LSP flow, we should expand a bit on what the bool actually does.

### DIFF
--- a/src/liquidity.rs
+++ b/src/liquidity.rs
@@ -130,7 +130,18 @@ pub struct LSPS2ServiceConfig {
 	pub min_payment_size_msat: u64,
 	/// The maximum payment size that we will accept when opening a channel.
 	pub max_payment_size_msat: u64,
-	/// Use the client trusts lsp model
+	/// Use the 'client-trusts-LSP' trust model.
+	///
+	/// When set, the service will delay *broadcasting* the JIT channel's funding transaction until
+	/// the client claimed sufficient HTLC parts to pay for the channel open.
+	///
+	/// Note this will render the flow incompatible with clients utilizing the 'LSP-trust-client'
+	/// trust model, i.e., in turn delay *claiming* any HTLCs until they see the funding
+	/// transaction in the mempool.
+	///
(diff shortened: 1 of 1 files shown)


src/liquidity.rs
```diff
@@ -130,7 +130,18 @@ pub struct LSPS2ServiceConfig {
 	pub min_payment_size_msat: u64,
 	/// The maximum payment size that we will accept when opening a channel.
 	pub max_payment_size_msat: u64,
-	/// Use the client trusts lsp model
+	/// Use the 'client-trusts-LSP' trust model.
+	///
+	/// When set, the service will delay *broadcasting* the JIT channel's funding transaction until
+	/// the client claimed sufficient HTLC parts to pay for the channel open.
+	///
+	/// Note this will render the flow incompatible with clients utilizing the 'LSP-trust-client'
```
joostjager, line 138: Perhaps make it more clear that this bool switches between the two models, and that it should be false for LSP-trust-client. Maybe an enum would have been better here?

tnull: Well, the thing is that if we set this to `false` the client can do whatever (wait or immediately claim). While technically there are two modes, no client that I'm aware of will actually run `LSP-trusts-client` model. I even attempted to get it removed from the spec, but there are .. reasons .. why we can't at this point. Really, any client should just claim the HTLCs ASAP. So this flag is really only concerning the LSP, allowing to determine how strict/safe it wants to be.

joostjager: Why offer the boolean then if there is no demand yet and you think it shouldn't even be in the spec?

tnull: Because the LSP might opt to be less strict. Some require to support 'LSP-trusts-client' from their (legal) side, but none actually expects the client to be strict and implement mempool-checking-before-claiming.

PR #710, opened 2025-11-19
Replace docs.rs build `doc_auto_cfg` feature with `doc_cfg`

tnull:
Unfortunately, `doc_auto_cfg` was removed, breaking doc builds for v0.7.0-rc.0. Here we replace it with the `doc_cfg` attribute.


tnull:
> Looked up `docs.rs` docs, thought you might want to add this to CI somewhere instead of the script we've got in LDK: https://docs.rs/about/builds#testing-builds-ci https://github.com/dtolnay/cargo-docs-rs

Yeah, maybe that would be worth. For now going ahead landing this.

PR #711, opened 2025-11-19
Replace docs.rs build `doc_auto_cfg` feature with `doc_cfg` (main)

tnull:
Unfortunately, `doc_auto_cfg` was removed, breaking doc builds for v0.7.0-rc.0. Here we replace it with the `doc_cfg` attribute.


tankyleo (approved):
As mentioned in the other PR, thought you might find these interesting :)

https://docs.rs/about/builds#testing-builds-ci
https://github.com/dtolnay/cargo-docs-rs

PR #715, opened 2025-11-24
Add a `docs.rs` CI job checking documentation builds

tnull:
We previouly ran into a quiet error that lead to `docs.rs` not rendering our docs properly, which unfortunately didn't surface until after we pushed out a releas (thankfully only an RC in this case).

Here, we add a CI job that tests our docs build with exactly the settings `docs.rs` uses. Additionally, the change also has the benefit that we now only build docs once rather than for every combiantion in our workflow matrix, which was a bit overkill.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,8 +12,6 @@ readme = "README.md"
 keywords = ["bitcoin", "lightning", "ldk", "bdk"]
 categories = ["cryptography::cryptocurrencies"]
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
 [lib]
 crate-type = ["lib", "staticlib", "cdylib"]
 name = "ldk_node"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -12,7 +12,8 @@ readme = "README.md"
 keywords = ["bitcoin", "lightning", "ldk", "bdk"]
 categories = ["cryptography::cryptocurrencies"]
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[package.metadata.docs.rs]
+features = ["uniffi"]
```
tankyleo: Hmm so on `docs.rs`, I'll see for example `pub fn bolt11_payment(&self) -> Arc<Bolt11Payment>` instead of `pub fn bolt11_payment(&self) -> Bolt11Payment`. Since `uniffi` is off by default, I think I would have also turned it off for `docs.rs` builds, your call.

tnull: Ah, yeah, no, good point. Reverted that change.

tankyleo (approved):

PR #716, opened 2025-11-24
Introduce new mandatory `NodeEntropy` object

tnull:
Second attempt at https://github.com/lightningdevkit/ldk-node/pull/240.

This is a prefactor for a following PR then using `EntropySource` in a `VssStoreBuilder` to replace #713. Will open that once this one lands.

```
Previously, the `Builder` allowed setting different entropy sources via
its `set_entropy...` methods, defaulting to sourcing from an
auto-generated seed file in the storage path. While this allowed for
really easy setup, it spared the user to actually think about where to
store their node secret.

Here, we therefore introduce a mandatory `NodeEntropy` object that, as
before, allows the user to source entropy from BIP39 Mnemonic, seed
bytes, or a seed file. However, it doesn't implement any default and
hence intentionally requires manually setup by the user. Moreover, this
API refactor also allows to reuse the same object outside of the
`Node`'s `Builder` in a future commit.
```

### DIFF
--- a/src/io/utils.rs
+++ b/src/io/utils.rs
@@ -575,3 +575,18 @@ pub(crate) fn read_bdk_wallet_change_set(
 		.map(|indexer| change_set.indexer = indexer);
 	Ok(Some(change_set))
 }
+
+#[cfg(test)]
+mod tests {
+	use super::read_or_generate_seed_file;
+	use super::test_utils::random_storage_path;
+
+	#[test]
+	fn generated_seed_is_readable() {
+		let mut rand_path = random_storage_path();
+		rand_path.push("test_keys_seed");
+		let expected_seed_bytes = read_or_generate_seed_file(&rand_path.to_str().unwrap()).unwrap();
+		let read_seed_bytes = read_or_generate_seed_file(&rand_path.to_str().unwrap()).unwrap();
+		assert_eq!(expected_seed_bytes, read_seed_bytes);
+	}
+}


src/io/utils.rs
```diff
@@ -47,45 +46,19 @@ use crate::io::{
...
-	Mnemonic::generate(word_count).expect("Failed to generate mnemonic")
-}
-
-pub(crate) fn read_or_generate_seed_file<L: Deref>(
-	keys_seed_path: &str, logger: L,
-) -> std::io::Result<[u8; WALLET_KEYS_SEED_LEN]>
-where
-	L::Target: LdkLogger,
-{
+pub(crate) fn read_or_generate_seed_file(
```
benthecarman, line 55: would be good to have a test for this function, that it can read the seeds it writes

tests/integration_tests_rust.rs
```diff
@@ -2030,7 +2027,7 @@ async fn lsps2_client_trusts_lsp() {
 	setup_builder!(service_builder, service_config.node_config);
 	service_builder.set_chain_source_esplora(esplora_url.clone(), Some(sync_config));
 	service_builder.set_liquidity_provider_lsps2(lsps2_service_config);
-	let service_node = service_builder.build().unwrap();
+	let service_node = service_builder.build(service_config.node_entropy.into()).unwrap();
```
benthecarman, line 2030: you don't need the `into()` here and everywhere below

tnull: Yes we do, as under the `uniffi` feature `build` expects an `Arc<NodeEntropy>`. So yes, unnecessary without `uniffi`, but necessary with `uniffi`, and using `into()` is less annoying than a bunch of `cfg` flags everywhere.

src/io/utils.rs
tnull: Done.

PR #718, opened 2025-11-25
Avoid `TcpListener` port collisions

tnull:
Second, more thorough attempt:

Previously, we could in tests potentially run into listening port collisions resulting into `InvalidSocketAddress` errors. These errors could surface if we rolled port numbers that either collided with other concurrent tests *or* with other unrelated services running on localhost.

Here, we simply let the OS assign us a free port number when setting up the testing nodes, which avoids such collisions altoghether (mod the potential TOCTOU race here, which we ignore for now).

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -215,22 +215,17 @@ pub(crate) fn random_storage_path() -> PathBuf {
 	temp_path
 }
 
-pub(crate) fn random_port() -> u16 {
-	let mut rng = rng();
-	rng.random_range(5000..32768)
-}
-
 pub(crate) fn random_listening_addresses() -> Vec<SocketAddress> {
 	let num_addresses = 2;
-	let mut listening_addresses = Vec::with_capacity(num_addresses);
+	let mut listening_addresses = HashSet::new();
 
-	for _ in 0..num_addresses {
-		let rand_port = random_port();
-		let address: SocketAddress = format!("127.0.0.1:{}", rand_port).parse().unwrap();
-		listening_addresses.push(address);
+	while listening_addresses.len() < num_addresses {
+		let socket = std::net::TcpListener::bind("127.0.0.1:0").unwrap();
+		let address: SocketAddress = socket.local_addr().unwrap().into();
+		listening_addresses.insert(address);
 	}
 
-	listening_addresses
+	listening_addresses.into_iter().collect()
 }
 
 pub(crate) fn random_node_alias() -> Option<NodeAlias> {


tests/common/mod.rs
```diff
@@ -222,15 +222,14 @@ pub(crate) fn random_port() -> u16 {
...
 	let num_addresses = 2;
-	let mut listening_addresses = Vec::with_capacity(num_addresses);
+	let mut listening_ports = HashSet::with_capacity(num_addresses);
 
-	for _ in 0..num_addresses {
+	while listening_ports.len() < num_addresses {
 		let rand_port = random_port();
-		let address: SocketAddress = format!("127.0.0.1:{}", rand_port).parse().unwrap();
-		listening_addresses.push(address);
+		listening_ports.insert(rand_port);
```
joostjager: Was this collision actually hit? ~1/27000 probability seems pretty unlikely.

Another option is to start random, and increase by 1 for every port. Also no collisions, and there could be a small benefit in having it deterministic.

src/lib.rs
```diff
@@ -1835,3 +1834,18 @@ pub(crate) fn total_anchor_channels_reserve_sats(
...
 }
+
+fn setup_tcp_listener(addr: SocketAddr) -> Result<TcpListener, tokio::io::Error> {
+	let socket = match addr {
+		SocketAddr::V4(_) => TcpSocket::new_v4()?,
+		SocketAddr::V6(_) => TcpSocket::new_v6()?,
+	};
+	socket.set_reuseaddr(true)?;
+	#[cfg(not(target_os = "windows"))]
+	socket.set_reuseport(true)?;
```
joostjager: Isn't this a load balancing option, where the kernel distributes load to any of the process that bind to the same port? Not sure if that is desirable here?

tests/common/mod.rs
tnull: Yes, in a test setting where I had reduced the possible ranges of ports (for other reasons). FWIW, IMO it's just cleaner to avoid the potential collision altogether.

src/lib.rs
tnull: Hmm, not exactly, but on second look it might not be necessary as `SO_REUSEADDR` should at least suffice for lingering connections.

tests/common/mod.rs
```diff
@@ -215,22 +215,18 @@ pub(crate) fn random_storage_path() -> PathBuf {
...
-}
-
 pub(crate) fn random_listening_addresses() -> Vec<SocketAddress> {
 	let num_addresses = 2;
-	let mut listening_addresses = Vec::with_capacity(num_addresses);
+	let mut listening_addresses = HashSet::new();
 
-	for _ in 0..num_addresses {
-		let rand_port = random_port();
+	while listening_addresses.len() < num_addresses {
```
joostjager, line 222: I think the former for loop can remain?

tests/common/mod.rs
```diff
@@ -215,22 +215,18 @@ pub(crate) fn random_storage_path() -> PathBuf {
...
+		let socket = std::net::TcpListener::bind("127.0.0.1:0").unwrap();
+		let rand_port = socket.local_addr().unwrap().port();
```
joostjager: Isn't there a way to get the full address out of socket without repeating the IP below?

tests/common/mod.rs
tnull: Well, I'm not sure that there is a guarantee that we don't get the same port back if we drop `socket` each iteration. I guess we could push them to a `Vec` to keep them alive while we're still iterating, but this seemed easier.

tests/common/mod.rs
tnull: Ah, indeed. I somehow remembered that we only have a `From` implementation for the other way, but a simple `into()` works. Now force-pushed:

```diff
> git diff-tree -U2 1b7d919d c0b4933a
diff --git a/tests/common/mod.rs b/tests/common/mod.rs
index 0795946a..0cdd5e12 100644
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -222,6 +222,5 @@ pub(crate) fn random_listening_addresses() -> Vec<SocketAddress> {
        while listening_addresses.len() < num_addresses {
                let socket = std::net::TcpListener::bind("127.0.0.1:0").unwrap();
-               let rand_port = socket.local_addr().unwrap().port();
-               let address: SocketAddress = format!("127.0.0.1:{}", rand_port).parse().unwrap();
+               let address: SocketAddress = socket.local_addr().unwrap().into();
                listening_addresses.insert(address);
        }
```

tests/common/mod.rs
joostjager: Hmm although obvious, I didn't realize that we don't actually use the socket. In that case, it definitely seems better to keep the sockets alive?

tnull: > Hmm although obvious, I didn't realize that we don't actually use the socket. In that case, it definitely seems better to keep the sockets alive?

No, why? We def. don't want to keep it alive to allow the `Node` setup a new one?

joostjager: I mean during the iteration, for the reason you brought up?

tnull: Right, but the OS likely gives back random available ports, and we dedup via the `HashSet`, so we're good?

joostjager: Idk if that is always the case for all OSes? Seems there is endless loop potential here. It's just test, fine to leave as is, but it doesn't feel 100% solid.

joostjager (approved):

PR #721, opened 2025-12-01
Update `ChannelDetails` docs for splicing

tnull:
This simply adopts the changes of rust-lightning#4250.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -228,12 +228,15 @@ impl fmt::Display for UserChannelId {
 
 /// Details of a channel as returned by [`Node::list_channels`].
 ///
+/// When a channel is spliced, most fields continue to refer to the original pre-splice channel
+/// state until the splice transaction reaches sufficient confirmations to be locked (and we
+/// exchange `splice_locked` messages with our peer). See individual fields for details.
+///
 /// [`Node::list_channels`]: crate::Node::list_channels
 #[derive(Debug, Clone)]
 pub struct ChannelDetails {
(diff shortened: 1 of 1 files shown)


src/types.rs
```diff
@@ -228,12 +228,15 @@ impl fmt::Display for UserChannelId {
 
 /// Details of a channel as returned by [`Node::list_channels`].
 ///
+/// When a channel is spliced, most fields continue to refer to the original pre-splice channel
+/// state until the splice transaction reaches sufficient confirmations to be locked (and we
+/// exchange `splice_locked` messages with our peer). See individual fields for details.
+///
 /// [`Node::list_channels`]: crate::Node::list_channels
 #[derive(Debug, Clone)]
 pub struct ChannelDetails {
```
valentinewallace, line 237: For my understanding of LDK Node, why do we duplicate this struct here vs using the `rust-lightning` type?

tnull: The reason is mostly bindings, as we don't have a good way to expose all fields on the LDK variant (e.g., `InitFeatures`, but we also have our own version of `config: ChannelConfig` as the LDK `enum MaxDustHTLCExposure` has tuple enum variants which our bindings generator `uniffi` doesn't support). It's also due to the docs on `ChannelDetails` referring to LDK APIs that aren't exposed in LDK Node. And, lastly, in LDK Node we also made the decision to not expose LDK's forked `OutPoint` variant but rather consistently only ever expose `bitcoin::OutPoint`.

valentinewallace: Ah, got it. Thanks for that explanation!

PR #726, opened 2025-12-03
Insert channel funding outputs into Wallet

benthecarman:
When doing a splice, to properly calculate fees we need the channel's funding utxo in our wallet, otherwise our wallet won't know the channel's original size. This adds the channel funding txo to the wallet before a splice as we only really need this information for splices and not for all channels.

### DIFF
--- a/src/ffi/types.rs
+++ b/src/ffi/types.rs
@@ -20,7 +20,7 @@ pub use bip39::Mnemonic;
 use bitcoin::hashes::sha256::Hash as Sha256;
 use bitcoin::hashes::Hash;
 use bitcoin::secp256k1::PublicKey;
-pub use bitcoin::{Address, BlockHash, FeeRate, Network, OutPoint, Txid};
+pub use bitcoin::{Address, BlockHash, FeeRate, Network, OutPoint, ScriptBuf, Txid};
 pub use lightning::chain::channelmonitor::BalanceSource;
 pub use lightning::events::{ClosureReason, PaymentFailureReason};
 use lightning::ln::channelmanager::PaymentId;
@@ -99,6 +99,22 @@ impl UniffiCustomTypeConverter for Address {
 		}
 
 		Err(Error::InvalidAddress.into())
+	}
+
+	fn from_custom(obj: Self) -> Self::Builtin {
+		obj.to_string()
+	}
+}
+
+impl UniffiCustomTypeConverter for ScriptBuf {
+	type Builtin = String;
+
+	fn into_custom(val: Self::Builtin) -> uniffi::Result<Self> {
+		if let Ok(key) = ScriptBuf::from_hex(&val) {
+			return Ok(key);
+		}
+
+		Err(Error::InvalidScriptPubKey.into())
 	}
 
 	fn from_custom(obj: Self) -> Self::Builtin {
--- a/src/types.rs
+++ b/src/types.rs
@@ -9,7 +9,7 @@ use std::fmt;
 use std::sync::{Arc, Mutex};
 
 use bitcoin::secp256k1::PublicKey;
-use bitcoin::OutPoint;
+use bitcoin::{OutPoint, ScriptBuf};
 use lightning::chain::chainmonitor;
 use lightning::impl_writeable_tlv_based;
 use lightning::ln::channel_state::ChannelDetails as LdkChannelDetails;
@@ -222,6 +222,15 @@ pub struct ChannelDetails {
 	/// state until the splice transaction reaches sufficient confirmations to be locked (and we
 	/// exchange `splice_locked` messages with our peer).
 	pub funding_txo: Option<OutPoint>,
+	/// The witness script that is used to lock the channel's funding output to commitment transactions.
+	///
+	/// This field will be `None` if we have not negotiated the funding transaction with our
+	/// counterparty already.
+	///
+	/// When a channel is spliced, this continues to refer to the original pre-splice channel
+	/// state until the splice transaction reaches sufficient confirmations to be locked (and we
+	/// exchange `splice_locked` messages with our peer).
+	pub funding_redeem_script: Option<ScriptBuf>,
 	/// The position of the funding transaction in the chain. None if the funding transaction has
 	/// not yet been confirmed and the channel fully opened.
 	///
@@ -378,6 +387,7 @@ impl From<LdkChannelDetails> for ChannelDetails {
 			channel_id: value.channel_id,
 			counterparty_node_id: value.counterparty.node_id,
 			funding_txo: value.funding_txo.map(|o| o.into_bitcoin_outpoint()),
+			funding_redeem_script: value.funding_redeem_script,
 			short_channel_id: value.short_channel_id,
 			outbound_scid_alias: value.outbound_scid_alias,
 			inbound_scid_alias: value.inbound_scid_alias,
--- a/bindings/ldk_node.udl
+++ b/bindings/ldk_node.udl
@@ -339,6 +339,7 @@ enum NodeError {
 	"InvalidNodeAlias",
 	"InvalidDateTime",
 	"InvalidFeeRate",
+	"InvalidScriptPubKey",
 	"DuplicatePayment",
 	"UnsupportedCurrency",
 	"InsufficientFunds",
@@ -575,6 +576,7 @@ dictionary ChannelDetails {
 	ChannelId channel_id;
 	PublicKey counterparty_node_id;
 	OutPoint? funding_txo;
+	ScriptBuf? funding_redeem_script;
 	u64? short_channel_id;
 	u64? outbound_scid_alias;
 	u64? inbound_scid_alias;
@@ -901,3 +903,6 @@ typedef string LSPS1OrderId;
 
 [Custom]
 typedef string LSPSDateTime;
+
+[Custom]
+typedef string ScriptBuf;
--- a/src/error.rs
+++ b/src/error.rs
@@ -113,6 +113,8 @@ pub enum Error {
 	InvalidDateTime,
 	/// The given fee rate is invalid.
 	InvalidFeeRate,
+	/// The given script public key is invalid.
+	InvalidScriptPubKey,
 	/// A payment with the given hash has already been initiated.
 	DuplicatePayment,
 	/// The provided offer was denonminated in an unsupported currency.
@@ -186,6 +188,7 @@ impl fmt::Display for Error {
 			Self::InvalidNodeAlias => write!(f, "The given node alias is invalid."),
 			Self::InvalidDateTime => write!(f, "The given date time is invalid."),
 			Self::InvalidFeeRate => write!(f, "The given fee rate is invalid."),
+			Self::InvalidScriptPubKey => write!(f, "The given script pubkey is invalid."),
 			Self::DuplicatePayment => {
 				write!(f, "A payment with the given hash has already been initiated.")
 			},


src/event.rs
```diff
@@ -1458,6 +1458,31 @@ where
...
+						.iter()
+						.find(|c| c.user_channel_id == user_channel_id)
+						.and_then(|c| c.get_funding_output());
+					match chan_output {
+						None => {
+							log_error!(
+								self.logger,
+								"Failed to find channel info for pending channel {channel_id} with counterparty {counterparty_node_id}"
+							);
+							return Err(ReplayEvent());
```
tnull: This would induce an infinite loop in event handling, no? Rather than doing this, we probably need to just log the error for now, and abort the flow when we can with 0.3? Also, would it make sense to add a `debug_assert` here, given that we expect this always to be avaialble? (same below)

src/event.rs
```diff
@@ -1458,6 +1458,31 @@ where
...
+								log_error!(
+									self.logger,
+									"Failed to insert funding TXO into wallet: {e}"
+								);
+								return Err(ReplayEvent());
+							}
+						},
+					}
 				} else {
 					log_info!(
```
tnull, line 1462: In what case would we end up in this `else` branch, and would we need to do something related to the redeemscript, too?

tests/integration_tests_rust.rs
```diff
@@ -995,7 +995,7 @@ async fn splice_channel() {
 	// Splice-in funds for Node B so that it has outbound liquidity to make a payment
 	node_b.splice_in(&user_channel_id_b, node_a.node_id(), 4_000_000).unwrap();
 
-	expect_splice_pending_event!(node_a, node_b.node_id());
+	let txo = expect_splice_pending_event!(node_a, node_b.node_id());
```
tnull, line 1001: These test changes pass on `main`. Can you add coverage for the cases where the previous approach would lead to inaccurate fee estimations?

src/ffi/types.rs
```diff
@@ -106,6 +106,22 @@ impl UniffiCustomTypeConverter for Address {
 	}
 }
 
+impl UniffiCustomTypeConverter for ScriptBuf {
```
tnull, line 109: nit: While this should be fine for now, we generally try to get away from the `UniffiCustomTypeConverter` approach, and instead convert more and more objects to proper interfaces.

tnull (commented):
Makes sense to me. Do we think this is worth a backport for 0.7 or are we fine waiting for 0.8 with shipping this?

The different cases in the event handler could use a comment or two to explain when we'd land on which side of the `if`/`else` clauses there. Also, we really need to be careful with just replaying events in cases where we don't expect immediate resolution of the error on retry, as we'd otherwise just spin and block any other event processing. This could also use better test coverage to show why we need this change in the first place.

Besides that, also pinging @jkczyz as a second reviewer on this.

tests/integration_tests_rust.rs
benthecarman: The bug/issue only exists with bitcoind syncing so to do so, we'd have to change the test to sync with bitcoind rpc. Is that wanted?

src/event.rs
benthecarman: According to the docs, the `funding_txo` "Will be None if the channel's funding transaction reached an acceptable depth prior to version 0.2". Since this was also added in 0.2, the redeem script would also be `None` so there shouldn't be anything to add. Either way, we need the funding txo to be able to insert so we can't add it.

src/event.rs
benthecarman: The error below can only happen on a persistence failure, elsewhere in the event handling we replay on those errors so I think it makes sense to keep as is.

For this one, it should always be available so I will just change to a `debug_assert`

src/event.rs
```diff
@@ -1458,6 +1458,33 @@ where
...
+							log_error!(
+								self.logger,
+								"Failed to find channel info for pending channel {channel_id} with counterparty {counterparty_node_id}"
+							);
+							debug_assert!(false,
+								"Failed to find channel info for pending channel {channel_id} with counterparty {counterparty_node_id}"
+							);
+						},
+						Some(output) => {
+							if let Err(e) = self.wallet.insert_txo(funding_txo, output) {
```
jkczyz: Hmmm... I guess we didn't really need the redeem script in `ChannelPending` and `SplicePending` since we only need the previous funding output in the wallet for the next splice. Might have been better in `ChannelReady`. We'd be able to avoid the look-up, although it wouldn't remove the `Option` check.

src/event.rs
```diff
@@ -1458,6 +1458,33 @@ where
 						counterparty_node_id,
 						funding_txo,
 					);
+
+					let chans =
+						self.channel_manager.list_channels_with_counterparty(&counterparty_node_id);
+					let chan_output = chans
+						.iter()
+						.find(|c| c.user_channel_id == user_channel_id)
+						.and_then(|c| c.get_funding_output());
```
jkczyz: Doing this in `ChannelReady` wouldn't work for splicing preexisting channels. Wonder if we should instead do this when initiating a splice? There we already look up the channel. It would also let us use the real `shared_input` when selecting UTXOs.

benthecarman: > Doing this in ChannelReady wouldn't work for splicing preexisting channels.

I dont think this is right, in the docs it says we get a ChannelReady for splices, also my test seems to confirm

src/event.rs
benthecarman: Yeah sadly I  messed up here, i think we needed to add the redeem script along with the new channel size so we had the full utxo information

src/event.rs
jkczyz: Correct, we do get it for spliced channels, but we need the initial funding output because that is the one being spent. And for channels existing before this change, we would have already processed the `ChannelReady` event for the initial funding.

src/event.rs
jkczyz: Yeah, though if we insert the previous funding txo when initiating the splice as mentioned in the other comment, then I think we'll have all the necessary information in `ChannelDetails`? Essentially, we'd be lazily adding the previous funding output to the wallet when initiating the splice.

src/event.rs
benthecarman: ah i see for backward compat sake, added!

src/event.rs
benthecarman: Made so we add it before a splice-in. I still think this one is good to keep so if you're doing any other wallet related action, it will have a more complete set of information and you don't need to initiate a splice for it to have the channel information

src/lib.rs
```diff
@@ -1267,29 +1267,25 @@ impl Node {
...
-						.to_p2wsh(),
-				},
+				previous_utxo: funding_output.clone(),
 				satisfaction_weight: EMPTY_SCRIPT_SIG_WEIGHT + FUNDING_TRANSACTION_WITNESS_WEIGHT,
 			};
 
 			let shared_output = bitcoin::TxOut {
 				value: shared_input.previous_utxo.value + Amount::from_sat(splice_amount_sats),
-				script_pubkey: make_funding_redeemscript(&dummy_pubkey, &dummy_pubkey).to_p2wsh(),
+				script_pubkey: funding_output.script_pubkey,
```
jkczyz: This should still be the dummy one or note that it is not really the shared output's `script_pubkey`. I guess the amount is already fake since we don't know what the counterparty will contribute.

src/lib.rs
```diff
@@ -1317,6 +1317,15 @@ impl Node {
...
 			};
 
+			// insert channel's funding utxo into the wallet so we can later calculate fees
+			// correctly when viewing this splice-in.
+			self.wallet.insert_txo(funding_txo.into_bitcoin_outpoint(), funding_output).map_err(
+				|e| {
+					log_error!(self.logger, "Failed to splice channel: {:?}", e);
+					Error::ChannelSplicingFailed
+				},
+			)?;
```
jkczyz: @tnull Don't know what the convention is here. Should we replace the `Error::PersistenceFailed` here with `Error::ChannelSplicingFailed`?

src/lib.rs
benthecarman: added a comment

src/lib.rs
tnull: Yeah, I think it's preferable to bubble up the `PersistenceFailed` error.

src/event.rs
tnull: > I still think this one is good to keep so if you're doing any other wallet related action, it will have a more complete set of information and you don't need to initiate a splice for it to have the channel information

Could you provide an example for what else we would use this information? If we don't need it I'd prefer to drop this change to declutter the logic here and avoid the extra persistence roundtrip.

tests/integration_tests_rust.rs
tnull: > The bug/issue only exists with bitcoind syncing so to do so, we'd have to change the test to sync with bitcoind rpc. Is that wanted?

Sure, feel free to do so. Alternatively, you could also add a `_bitcoind`  variant test case (ofc reusing the logic) so we have two test cases for bitcoind and esplora - that's if we think there would be more chain source specific edge cases.

src/types.rs
```diff
@@ -222,6 +222,15 @@ pub struct ChannelDetails {
...
 	pub funding_txo: Option<OutPoint>,
+	/// The witness script that is used to lock the channel's funding output to commitment transactions.
+	///
+	/// This field will be `None` if we have not negotiated the funding transaction with our
+	/// counterparty already.
+	///
+	/// When a channel is spliced, this continues to refer to the original pre-splice channel
+	/// state until the splice transaction reaches sufficient confirmations to be locked (and we
+	/// exchange `splice_locked` messages with our peer).
+	pub funding_redeem_script: Option<bitcoin::ScriptBuf>,
```
tnull: nit: Please import `ScriptBuf` instead of using the `bitcoin::` prefix.

src/event.rs
benthecarman: Okay yeah on further reflection, no real use case, especially with the api that ldk-node exposes to the wallet.

tests/integration_tests_rust.rs
benthecarman: added

benthecarman:
Addressed review, ended up removing it on events and just doing it before a splice. Made sure test coverage hits on both backends and for splice in and out

src/lib.rs
```diff
@@ -1319,6 +1319,10 @@ impl Node {
 				},
 			};
 
+			// insert channel's funding utxo into the wallet so we can later calculate fees
+			// correctly when viewing this splice-in.
+			self.wallet.insert_txo(funding_txo.into_bitcoin_outpoint(), funding_output)?;
```
jkczyz: We'll want to do this before we get the change address. Otherwise, if this fails the `cancel_tx` clean-up below will not be called.

benthecarman: done

tests/integration_tests_rust.rs
```diff
@@ -1044,18 +1052,29 @@ async fn splice_channel() {
...
 	assert_eq!(
 		node_a.list_balances().total_lightning_balance_sats,
-		4_000_000 - closing_transaction_fee_sat - anchor_output_sat - splice_out_fee_sat
+		4_000_000 - closing_transaction_fee_sat - anchor_output_sat - expected_splice_out_fee_sat
 	);
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn splice_channel() {
+	run_splice_channel_test(false).await;
```
tnull, line 1074: nit: Not that it matters much, but elsewhere in LDK Node / LDK tests the usual pattern would be to name the 'inner' test methods `do_splice_channel`. Feel free to make that rename if you touch the code again.

tnull (approved):
LGTM, one nit. Otherwise should be ready pending @jkczyz's review.

jkczyz (approved):

PR #727, opened 2025-12-04
Expose a `VssStoreBuilder` allowing to build `VssStore` independently

tnull:
Previously, `VssStore` was a private object that could only be constructed internally via the `Builder` object. However, some users might want to use `VssStore` independently of/before the `Node` is constructed.

To this end, we here expose a new `VssStoreBuilder` that allows to independently construct a `VssStore` instance that then can be handed to `Builder::build_with_store`.

For now we keep the previous VSS-related methods on `Builder` around, but have them reuse the `VssStoreBuilder` internally. This also allows us to not change anything related to bindings. For bindings, we still have to explore if/how we can expose `VssStore` in the future.

FWIW, after this PR lands it might also be a good time to finally upstream `VssStore` to the `lightning-persister` crate, which was the original plan ever since we added `VssStore`.

### DIFF
--- a/src/io/vss_store.rs
+++ b/src/io/vss_store.rs
@@ -5,8 +5,11 @@
 // http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
 // accordance with one or both of these licenses.
 
+//! Objects related to [`VssStore`] live here.
+
 use std::boxed::Box;
 use std::collections::HashMap;
+use std::fmt;
 use std::future::Future;
 #[cfg(test)]
 use std::panic::RefUnwindSafe;
@@ -15,7 +18,10 @@ use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
+use bitcoin::bip32::{ChildNumber, Xpriv};
 use bitcoin::hashes::{sha256, Hash, HashEngine, Hmac, HmacEngine};
+use bitcoin::key::Secp256k1;
+use bitcoin::Network;
 use lightning::impl_writeable_tlv_based_enum;
 use lightning::io::{self, Error, ErrorKind};
 use lightning::util::persist::{KVStore, KVStoreSync};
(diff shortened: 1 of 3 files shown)


src/io/vss_store.rs
```diff
@@ -800,6 +817,149 @@ impl EntropySource for RandEntropySource {
...
+					.derive_priv(
+						&Secp256k1::new(),
+						&[ChildNumber::Hardened { index: VSS_HARDENED_CHILD_INDEX }],
+					)
+					.map_err(|_| VssStoreBuildError::KeyDerivationFailed)
+			})?;
+
+		let lnurl_auth_xprv = vss_xprv
+			.derive_priv(
+				&Secp256k1::new(),
```
benthecarman: nit: you're initializing 2 secp contexts here, could just use 1

src/io/vss_store.rs
```diff
@@ -800,6 +817,149 @@ impl EntropySource for RandEntropySource {
...
+	/// server.
+	///
+	/// **Caution**: VSS support is in **alpha** and is considered experimental. Using VSS (or any
+	/// remote persistence) may cause LDK to panic if persistence failures are unrecoverable, i.e.,
+	/// if they remain unresolved after internal retries are exhausted.
+	///
+	/// [VSS]: https://github.com/lightningdevkit/vss-server/blob/main/README.md
+	/// [LNURL-auth]: https://github.com/lnurl/luds/blob/luds/04.md
+	pub fn build(
+		&self, lnurl_auth_server_url: String, fixed_headers: HashMap<String, String>,
```
benthecarman, line 876: Is there a reason for these we using `&self` instead of consuming it with `self`? I see we do the same for the node builder but typically builders consume themselves when done, it would allow you to remove some clones

tnull: Yes, for the `Builder` we do that to be able to reuse it in the bindings variant `ArcedNodeBuilder` (as bindings don't support move semantics, only reference-counted `Arc`s). Here, we go the same route as we will at least attempt to expose `VssStore`/`VssStoreBuilder` in bidnings, though with the new async `KVStore` trait that could prove difficult, we'll see.

src/io/vss_store.rs
tnull: Ah, preexisting, but good point. Added a fixup.

Issue #529, opened 2025-04-28
Multi-LSP support


tnull:
Currently, we only allow to set a single LSP for each protocol (i.e., bLIP-51/bLIP-52). We however got users requests for the ability to configure multiple LSPs at the same time. 

While the technical change to do this is trivial, we'll need to figure out a good API design to allow for this. In case of bLIP-52 / LSPS2 we could actually request fee params from all LSPs in parallel and choose the cheapest one. For bLIP-51/LSPS1 it's however not quite so clear how we can handle multiple LSPs in parallel in the API.

MegalithicBTC:
I can see this has been implemented here: 
https://github.com/lightningdevkit/ldk-node/blob/ddd1768d33c91b69fd990b188f47b1b677099718/src/liquidity.rs#L1031

One thing to think about -- to my knowledge, in any instance where multiple LSP offers are actually being used currently on the Lightning Network -- for example, Amboss Magma, or Alby Hub, if a user were to always choose the "cheapest" liquidity offer, that user would end up with a channel with a LSP that one or all of the following characteristics:

1) The LSP would lack inbound liquidity and would often fail payments
2) The LSP would have intermittent connectivity to the internet, so payments would often fail (A major seller on Amboss Magma actually just went down due to a power outage for several days last week.)
3) The LSP would practice aggressive "liquidity reuse", whereby the LSP would often rebalance "outbound heavy" channels, turning them into "inbound heavy", in order to re-use the outbound liquidity for other LSP channel sales

All to say: I'm not sure that "automatically choose the cheapest LSP" will be a great user experience, in practice.

MegalithicBTC:
Ah, I see, the cheapest offer from the SINGLE configured LSP. Got it.

tnull:
Done in #792.

PR #729, opened 2025-12-10
Add LSPS5 webhook notification support

Camillarhi:
Integrates LSPS5 (bLIP-0055) from lightning-liquidity, enabling webhook-based push notifications so clients can be alerted to events while their app is offline. Built on the refactored multi-LSP liquidity module `(src/liquidity/{client,service})`.

- set_webhook - register/update a webhook.
- list_webhooks - query registered webhooks.
- remove_webhook - remove a registration.

When `node_id` is `None`, `set_webhook` and `remove_webhook` fan out to every LSPS5-capable LSP, so a webhook can be configured once across all configured LSPs.

Lets a node act as an LSPS5 server:
- Delivers outgoing webhook notifications over HTTP in response to `LSPS5ServiceEvent::SendWebhookNotification`.
- Auto-triggers an `onion_message_incoming` notification when an intercepted onion message targets a client that is currently offline (wired from `LdkEvent::OnionMessageIntercepted`, gated on peer connectivity via the peer manager).

### DIFF
--- a/tests/probing_tests.rs
+++ b/tests/probing_tests.rs
@@ -296,8 +296,9 @@ async fn locked_msat_accounts_for_routing_fees() {
 /// HTLC pending across the restart we observe `locked_msat > 0` and then *immediately*
 /// call `node_a.disconnect(node_b)`, which closes A's socket to B in-process — much
 /// faster than `node_b.stop()` — so any failure message from B is dropped before A
-/// processes it. If the race is lost on a given probe (locked_msat drops back to 0
-/// after the disconnect), we reconnect and let the next probe tick try again.
+/// processes it. Before rechecking, we allow monitor persistence and queued failure
+/// handling to settle. If the race is lost on a given probe (locked_msat drops back to
+/// 0 after the disconnect), we reconnect and let the next probe tick try again.
 /// The pending Probe entry persists in `node_a`'s channel manager and must be
 /// rebuilt by the prober's `locked_msat` on restart via `list_recent_payments()`.
 #[tokio::test(flavor = "multi_thread")]
@@ -363,11 +364,14 @@ async fn probing_budget_restored_after_node_restart() {
 	let isolated = tokio::time::timeout(Duration::from_secs(30), async {
 		loop {
 			if node_a.prober().unwrap().locked_msat() > 0 {
+				strategy.stop_probing();
 				node_a.disconnect(node_b_id).ok();
+				tokio::time::sleep(Duration::from_secs(1)).await;
 				if node_a.prober().unwrap().locked_msat() > 0 {
 					return true;
 				}
 				node_a.connect(node_b_id, node_b_addr.clone(), false).ok();
+				strategy.start_probing();
 			}
 			tokio::time::sleep(Duration::from_millis(1)).await;
 		}


Cargo.toml
```diff
@@ -56,7 +56,7 @@ bdk_esplora = { version = "0.22.0", default-features = false, features = ["async
 bdk_electrum = { version = "0.23.0", default-features = false, features = ["use-rustls-ring"]}
 bdk_wallet = { version = "2.2.0", default-features = false, features = ["std", "keys-bip39"]}
 
-reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls"] }
+reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls", "blocking"] }
```
tnull: We shouldn't ever add the `blocking` feature for `reqwest` as it's fundamentally incompatible with the async variant.

src/liquidity.rs
```diff
@@ -54,6 +60,77 @@ const LIQUIDITY_REQUEST_TIMEOUT_SECS: u64 = 5;
...
+			HttpClientError::Status(code) => write!(f, "HTTP error: {}", code),
+			HttpClientError::Other(msg) => write!(f, "HTTP client error: {}", msg),
+		}
+	}
+}
+
+impl std::error::Error for HttpClientError {}
+
+/// Trait for making HTTP requests.
+pub trait HttpClient: Send + Sync + std::fmt::Debug {
```
tnull: All of this code seems like an outdated copy of what was there on the upstream LSPS5 PR at one point in time. Please remove anything related to `HttpClient` etc. Rather, we should just ... make the call via spawned async task using `reqwest` when we handle the `SendWebhookNotification` event.

src/liquidity.rs
```diff
@@ -145,13 +222,71 @@ pub struct LSPS2ServiceConfig {
...
+	ldk_service_config: LdkLSPS5ServiceConfig,
+}
+
+/// Represents the configuration of the LSPS5 service.
+///
+/// See [bLIP-55 / LSPS5] for more information.
+///
+/// [bLIP-55 / LSPS5]: https://github.com/lightning/blips/blob/master/blip-0055.md
+#[derive(Debug, Clone)]
+pub struct LSPS5ServiceConfig {
```
tnull: Not sure why we need to replicate this struct?

tnull (commented):
Thanks for taking this over!

Need to look into this PR closer, here are some early comments that I noted when briefly scrolling through.

src/liquidity.rs
Camillarhi: Thanks for the feedback! I've updated the PR to use `reqwest` and removed the` HttpClient` dependencies. This is still a work in progress, but that should address your concern about the outdated approach.

src/liquidity.rs
```diff
@@ -1540,3 +2181,213 @@ impl LSPS1Liquidity {
...
+		self.set_webhook_impl(app_name, webhook_url)
+	}
+
+	/// Connects to the configured LSP and registers a webhook URL for receiving LSPS5 notifications.
+	///
+	/// The webhook will receive signed push notifications for Lightning events such as incoming
+	/// payments when the client is offline.
+	#[cfg(feature = "uniffi")]
+	pub fn set_webhook(
+		&self, app_name: String, webhook_url: String,
```
TheBlueMatt: These appear to be the same?

src/liquidity.rs
```diff
@@ -54,6 +60,29 @@ const LIQUIDITY_REQUEST_TIMEOUT_SECS: u64 = 5;
 const LSPS2_GETINFO_REQUEST_EXPIRY: Duration = Duration::from_secs(60 * 60 * 24);
 const LSPS2_CHANNEL_CLTV_EXPIRY_DELTA: u32 = 72;
 
+/// Error type for HTTP client operations.
+#[derive(Debug)]
+pub enum HttpClientError {
```
tnull: Seems this isn't used anywhere. Let's drop it!

src/builder.rs
```diff
@@ -444,6 +452,44 @@ impl NodeBuilder {
...
+		&mut self, service_config: LSPS5ServiceConfig,
+	) -> &mut Self {
+		let liquidity_source_config =
+			self.liquidity_source_config.get_or_insert(LiquiditySourceConfig::default());
+		liquidity_source_config.lsps5_service = Some(service_config);
+		self
+	}
+
+	/// Sets a custom HTTP client to be used by the LSPS5 service.
+	pub fn set_liquidity_http_client(&mut self, http_client: reqwest::Client) -> &mut Self {
```
tnull: No need to set a 'custom' client. Let's just use our usual client and drop this builder method.

src/builder.rs
```diff
@@ -444,6 +452,44 @@ impl NodeBuilder {
...
 	}
 
+	/// Configures the [`Node`] instance to source webhook notifications from the given
+	/// [bLIP-55 / LSPS5] service.
+	///
+	/// This allows the client to register webhook endpoints with the LSP to receive
+	/// push notifications for Lightning events when the client is offline.
+	///
+	/// [bLIP-55 / LSPS5]: https://github.com/lightning/blips/blob/master/blip-0055.md
+	pub fn set_liquidity_source_lsps5(
```
tnull: I think there is ~no immediate scenario where we'd like to set a *different* LSP for LSPS5 compared to LSPS1/LSPS2. So rather than having the user configure this explicitly, it should be automatically enabled if we detect that our LSPS1 or LSPS2 counterparty also supports LSPS5 (you can check that via `LSPS0ClientHandler::list_protocols`). Note that this however poses an API design challenge as suddenly we'll have to support getting notifications from *multiple* LSPs.

Note that for a while some users have requested multi-LSP support (cf. https://github.com/lightningdevkit/ldk-node/issues/529) and for the sake of simplicity we so far have punted on it. But I do wonder if it would now make sense to first make this refactor to a model where users just configure a number of LSP nodes and we lean on `list_protocols` to discover what the configured nodes are supporting. I think I'll have to think about it a bit more - what are your thoughts on this?

tnull (commented):
Thanks for looking into this, before I go into too much details, here are some comments regarding the chosen approach and/or the API design. Let me know what you think!

src/builder.rs
Camillarhi: I did this because I wanted to mock the client in the test in order to test the `SendWebhookNotification` event. I will try another approach

Camillarhi:
> @Camillarhi Do you still intend to move forward with this PR?

Yes, I do, apologies for the delayed response

src/builder.rs
Camillarhi: Yeah, what if Instead of auto-enabling LSPS5 on top of the current model, we do the multi-LSP refactor first. Users would just configure LSP nodes (node_id + address + token), and we use `list_protocols` to discover what each one supports. LSPS5 then just works for any LSP that advertises it.

PR #733, opened 2025-12-29
[RFC] event: expose BOLT12 invoice in PaymentSuccessful for proof of payment

vincenzopalazzo:
This patch adds the `bolt12_invoice` field to the `PaymentSuccessful` event, enabling users to obtain proof of payment for BOLT12 transactions.

Problem:
Previously, after a successful BOLT12 payment, users had no way to access the paid invoice data. This made it impossible to provide proof of payment to third parties, who need both the payment preimage and the original invoice to verify that sha256(preimage) matches the invoice's payment_hash.

Solution:
Add a `bolt12_invoice: Option<Vec<u8>>` field to `PaymentSuccessful` that contains the serialized BOLT12 invoice bytes. The invoice is serialized using LDK's standard encoding, which can be parsed back using `Bolt12Invoice::try_from(bytes)` in native Rust, or by hex-encoding the bytes and using `Bolt12Invoice.from_str()` in FFI bindings.

Design decisions:
- Store as `Vec<u8>` rather than the complex `PaidBolt12Invoice` type to avoid UniFFI limitations with objects in enum variants
- Return `None` for `StaticInvoice` (async payments) since proof of payment is not possible for those payment types anyway
- Use TLV tag 7 for serialization, maintaining backward compatibility with existing persisted events

This implementation follows the maintainer guidance from PR #563 to expose the invoice via the event rather than storing it in the payment store.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -15,6 +15,8 @@ use bitcoin::blockdata::locktime::absolute::LockTime;
 use bitcoin::secp256k1::PublicKey;
 use bitcoin::{Amount, OutPoint};
 use lightning::events::bump_transaction::BumpTransactionEvent;
+#[cfg(not(feature = "uniffi"))]
+use lightning::events::PaidBolt12Invoice;
 use lightning::events::{
 	ClosureReason, Event as LdkEvent, FundingInfo, PaymentFailureReason, PaymentPurpose,
 	ReplayEvent,
@@ -37,6 +39,8 @@ use crate::config::{may_announce_channel, Config};
 use crate::connection::ConnectionManager;
 use crate::data_store::DataStoreUpdateResult;
 use crate::fee_estimator::ConfirmationTarget;
+#[cfg(feature = "uniffi")]
+use crate::ffi::PaidBolt12Invoice;
 use crate::io::{
 	EVENT_QUEUE_PERSISTENCE_KEY, EVENT_QUEUE_PERSISTENCE_PRIMARY_NAMESPACE,
 	EVENT_QUEUE_PERSISTENCE_SECONDARY_NAMESPACE,
@@ -79,6 +83,17 @@ pub enum Event {
 		payment_preimage: Option<PaymentPreimage>,
 		/// The total fee which was spent at intermediate hops in this payment.
 		fee_paid_msat: Option<u64>,
+		/// The BOLT12 invoice that was paid.
+		///
+		/// This is useful for proof of payment. A third party can verify that the payment was made
+		/// by checking that the `payment_hash` in the invoice matches `sha256(payment_preimage)`.
+		///
+		/// Will be `None` for non-BOLT12 payments.
+		///
+		/// Note that static invoices (indicated by [`PaidBolt12Invoice::StaticInvoice`], used for
+		/// async payments) do not support proof of payment as the payment hash is not derived
+		/// from a preimage known only to the recipient.
+		bolt12_invoice: Option<PaidBolt12Invoice>,
 	},
 	/// A sent payment has failed.
 	PaymentFailed {
@@ -268,6 +283,7 @@ impl_writeable_tlv_based_enum!(Event,
 		(1, fee_paid_msat, option),
 		(3, payment_id, option),
 		(5, payment_preimage, option),
+		(7, bolt12_invoice, option),
 	},
 	(1, PaymentFailed) => {
 		(0, payment_hash, option),
@@ -1028,6 +1044,7 @@ where
 				payment_preimage,
 				payment_hash,
 				fee_paid_msat,
+				bolt12_invoice,
 				..
 			} => {
 				let payment_id = if let Some(id) = payment_id {
@@ -1073,6 +1090,7 @@ where
 					payment_hash,
 					payment_preimage: Some(payment_preimage),
 					fee_paid_msat,
+					bolt12_invoice: bolt12_invoice.map(Into::into),
 				};
 
 				match self.event_queue.add_event(event).await {
--- a/src/ffi/types.rs
+++ b/src/ffi/types.rs
@@ -10,6 +10,7 @@
 //
 // Make sure to add any re-exported items that need to be used in uniffi below.
 
+use std::collections::HashMap;
 use std::convert::TryInto;
 use std::ops::Deref;
 use std::str::FromStr;
@@ -22,17 +23,20 @@ use bitcoin::hashes::Hash;
 use bitcoin::secp256k1::PublicKey;
 pub use bitcoin::{Address, BlockHash, FeeRate, Network, OutPoint, ScriptBuf, Txid};
 pub use lightning::chain::channelmonitor::BalanceSource;
+use lightning::events::PaidBolt12Invoice as LdkPaidBolt12Invoice;
 pub use lightning::events::{ClosureReason, PaymentFailureReason};
 use lightning::ln::channelmanager::PaymentId;
+use lightning::ln::msgs::DecodeError;
 pub use lightning::ln::types::ChannelId;
 use lightning::offers::invoice::Bolt12Invoice as LdkBolt12Invoice;
 pub use lightning::offers::offer::OfferId;
 use lightning::offers::offer::{Amount as LdkAmount, Offer as LdkOffer};
 use lightning::offers::refund::Refund as LdkRefund;
+use lightning::offers::static_invoice::StaticInvoice as LdkStaticInvoice;
 use lightning::onion_message::dns_resolution::HumanReadableName as LdkHumanReadableName;
 pub use lightning::routing::gossip::{NodeAlias, NodeId, RoutingFees};
 pub use lightning::routing::router::RouteParametersConfig;
-use lightning::util::ser::Writeable;
+use lightning::util::ser::{Readable, Writeable, Writer};
 use lightning_invoice::{Bolt11Invoice as LdkBolt11Invoice, Bolt11InvoiceDescriptionRef};
 pub use lightning_invoice::{Description, SignedRawBolt11Invoice};
 pub use lightning_liquidity::lsps0::ser::LSPSDateTime;
@@ -41,10 +45,10 @@ pub use lightning_liquidity::lsps1::msgs::{
 };
 pub use lightning_types::payment::{PaymentHash, PaymentPreimage, PaymentSecret};
 pub use lightning_types::string::UntrustedString;
-use std::collections::HashMap;
-
-use vss_client::headers::VssHeaderProvider as VssClientHeaderProvider;
-use vss_client::headers::VssHeaderProviderError as VssClientHeaderProviderError;
+use vss_client::headers::{
+	VssHeaderProvider as VssClientHeaderProvider,
+	VssHeaderProviderError as VssClientHeaderProviderError,
+};
 
 /// Errors around providing headers for each VSS request.
 #[derive(Debug, uniffi::Error)]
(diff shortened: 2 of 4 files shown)


src/event.rs
```diff
@@ -75,6 +76,17 @@ pub enum Event {
...
+		///
+		/// This is useful for proof of payment. A third party can verify that the payment was made
+		/// by checking that the `payment_hash` in the invoice matches `sha256(payment_preimage)`.
+		///
+		/// Will be `None` for non-BOLT12 payments, or for async payments (`StaticInvoice`)
+		/// where proof of payment is not possible.
+		///
+		/// To parse the invoice in native Rust, use `Bolt12Invoice::try_from(bytes)`.
+		/// In FFI bindings, hex-encode the bytes and use `Bolt12Invoice.from_str(hex_string)`.
+		bolt12_invoice: Option<Vec<u8>>,
```
tnull: I don't think we should expose the raw bytes here, although I have to say that I also find the `PaidBolt12Invoice` type pretty awkward tbh. But I guess we'll have to bite that bullet now. 

Note that we'll probably need to add our own version of `PaidBolt12Invoice` though as `Uniffi` doesn't support tuple enum variants, so we'll need to add a local version with named enum variants, and then add an `impl From<LdkPaidBolt12Invoice> for PaidBolt12Invoice`, as well as expose the `StaticInvoice` APIs in the `ffi/types.rs`/UDL.

tnull (commented):
Exposing the BOLT12 invoice makes sense to me, though we should do it properly instead of just exposing the bytes.

src/event.rs
vincenzopalazzo: What do you think about the commit https://github.com/lightningdevkit/ldk-node/pull/733/changes/4c6b18ed6b9ac258d4f89d98370b21588b98c751

bindings/ldk_node.udl
```diff
@@ -850,6 +850,33 @@ interface Bolt12Invoice {
...
+	u64? absolute_expiry_seconds();
+	sequence<u8> encode();
+};
+
+enum PaidBolt12InvoiceKind {
+	"Bolt12Invoice",
+	"StaticInvoice",
+};
+
+dictionary PaidBolt12Invoice {
```
tnull: Can we make this an

```udl
[Enum]
interface PaidBolt12Invoice {
```
instead?

src/types.rs
```diff
@@ -606,3 +616,199 @@ impl From<&(u64, Vec<u8>)> for CustomTlvRecord {
 		CustomTlvRecord { type_num: tlv.0, value: tlv.1.clone() }
 	}
 }
+
+// Re-export the invoice types. When UniFFI is enabled, we use the types from ffi::types
+// which have additional UniFFI-specific implementations. Otherwise, we use simpler wrappers.
+#[cfg(feature = "uniffi")]
+pub use crate::ffi::{Bolt12Invoice, PaidBolt12Invoice, PaidBolt12InvoiceKind, StaticInvoice};
```
tnull: No, please ensure anything `uniffi`-related is kept in the `ffi` module and doesn't leak elsewhere.

src/types.rs
```diff
@@ -606,3 +616,199 @@ impl From<&(u64, Vec<u8>)> for CustomTlvRecord {
...
+	}
+
+	/// A wrapper around a [`lightning::offers::static_invoice::StaticInvoice`].
+	///
+	/// A `StaticInvoice` is used for async payments where the recipient may be offline.
+	///
+	/// Unlike [`Bolt12Invoice`], a `StaticInvoice` does not support proof of payment
+	/// because the payment hash is not derived from a preimage known only to the recipient.
+	#[derive(Debug, Clone, PartialEq, Eq)]
+	pub struct StaticInvoice {
```
tnull: No need to wrap `StaticInvoice` in the non-`uniffi` case. Please follow the same approach as we did for `Offer`, `Refund`, etc. and reuse the existing `maybe_wrap` macros etc.

src/types.rs
```diff
@@ -606,3 +616,199 @@ impl From<&(u64, Vec<u8>)> for CustomTlvRecord {
...
+				LdkPaidBolt12Invoice::StaticInvoice(invoice) => PaidBolt12Invoice {
+					kind: PaidBolt12InvoiceKind::StaticInvoice,
+					bolt12_invoice: None,
+					static_invoice: Some(Arc::new(StaticInvoice { inner: invoice })),
+				},
+			}
+		}
+	}
+
+	impl Writeable for PaidBolt12Invoice {
```
tnull: We def. shouldn't need to reimplement anything related to serialization.

tnull (commented):
The changes are getting closer, but please make sure to avoid unnecessary boilerplate and stick to the approach we took for the other types (like `Offer`, `Refund`, etc).

This also needs a rebase by now, sorry!

Btw, please re-request review when you made updates, otherwise I might not always see it immediately.

bindings/ldk_node.udl
vincenzopalazzo: Mh I was hacking round this this afternoon and this is what I got

```
  error: failed to run custom build command for `ldk-node v0.8.0+git`

  Caused by:
    process didn't exit successfully (exit status: 101)
    --- stdout
    cargo:rerun-if-changed=bindings/ldk_node.udl
    cargo:rerun-if-env-changed=UNIFFI_TESTS_DISABLE_EXTENSIONS

    --- stderr
    thread 'main' panicked at build.rs:10:59:
    called `Result::unwrap()` on an `Err` value: Objects cannot currently be used in enum variant data
```

The key part is:

Objects cannot currently be used in enum variant data

```rust
  This occurs when trying to use:
  [Enum]
  interface PaidBolt12Invoice {
      Bolt12Invoice(Bolt12Invoice invoice);
      StaticInvoice(StaticInvoice invoice);
  };
```

It looks like that it is happening because `Bolt12Invoice` and `StaticInvoice` are defined as interface (Objects) in UniFFI, they cannot be used as parameters in enum variants.

Am I missing something?

src/types.rs
vincenzopalazzo: Good catch now I am  delegating it to the inner type

src/types.rs
vincenzopalazzo: Ok let me know if now this is solving your concern, thanks!

bindings/ldk_node.udl
tnull: No, at least on our current `uniffi` version you are indeed correct I'm afraid. 

So we are at an impasse here: I really dislike making this a `struct` with optional fields (and also adding serialization on bindings wrapper, but we could solve for this otherwise). Unfortunately, Uniffi only added support for enum fields that are objects in https://github.com/mozilla/uniffi-rs/pull/2412, which however shipped in v0.29. We however can't update to v0.29 as we have users that use `NordSecurity/uniffi-bindgen-go` which notoriously lags behind and doesn't get updated (see https://github.com/NordSecurity/uniffi-bindgen-go/issues/40).

So I think this is a rare occasion where I'd be in favor of just not keeping the bindings in sync with the Rust interface for now and coming back to it either when a bindings users shout or once we finally can upgrade.

Could you:

1) Revert back to the Rust-only changes, i.e., only using LDK types and putting the `Event` enum field behind `#[cfg(not(feature = "uniffi"))]`
2) Add a TODO there in the code noting that we still need to expose this in bindings when we can post uniffi v0.29
3) Open a corresponding tracking issue to also keep track of this on github
4) Rebase and squash any fixup commits so we can land this soon

Sorry for the churn, hope you didn't spend too much time trying to make it work!

vincenzopalazzo: No problem, thanks for helping navigate UniFFI. So I did what you suggested in this commit: [e6200d6](https://github.com/lightningdevkit/ldk-node/pull/733/commits/e6200d6e17669f7a15cd59d96b94806739f7af90). But to me, this solution still feels a little bit incomplete because I think we can do better and remove the branching inside UniFFI.

Do you think it's worth reconsidering my first approach to pass the raw bytes of the invoice? That way, people using the bindings can still perform proof of payment, and then when the new version of UniFFI lands, we can deprecate the raw bytes and switch to the proper type?

let me know what do you preferer, the PR to be ready need just to squash the last commit

tnull: > No problem, thanks for helping navigate UniFFI. So I did what you suggested in this commit: [e6200d6](https://github.com/lightningdevkit/ldk-node/pull/733/commits/e6200d6e17669f7a15cd59d96b94806739f7af90). But to me, this solution still feels a little bit incomplete because I think we can do better and remove the branching inside UniFFI.

Yeah, I agree it's unfortunate, but probably best to just hold off on it until we can do it the right way.

> Do you think it's worth reconsidering my first approach to pass the raw bytes of the invoice? That way, people using the bindings can still perform proof of payment, and then when the new version of UniFFI lands, we can deprecate the raw bytes and switch to the proper type?

No, I don't think returning bytes is an option as AFAIU `Bolt12Invoice` never had a blessed byte *or* string, so even if we'd expose it the user would have no idea how to deserialize it again on the other end. Hence we should really just wait until we can expose the proper `Bolt12Invoice` type.

All of this makes me wonder if we could prioritize the upgrade first, will explore that in the next few days and come back here.

bindings/ldk_node.udl
```diff
@@ -860,6 +861,8 @@ interface Bolt12Invoice {
 	sequence<u8> encode();
 };
 
+// TODO: Add StaticInvoice and PaidBolt12Invoice once we upgrade to UniFFI 0.29+. See #757.
```
tnull: nit: Let's drop this TODO, one above on `Event` is sufficient.

src/payment/mod.rs
```diff
@@ -23,3 +23,6 @@ pub use store::{
 	ConfirmationStatus, LSPFeeLimits, PaymentDetails, PaymentDirection, PaymentKind, PaymentStatus,
 };
 pub use unified::{UnifiedPayment, UnifiedPaymentResult};
+
+#[cfg(not(feature = "uniffi"))]
+pub use crate::types::PaidBolt12Invoice;
```
tnull: No need to re-export the type here, let's just drop it and use the LDK type directly.

src/event.rs
```diff
@@ -48,6 +48,8 @@ use crate::payment::store::{
 	PaymentDetails, PaymentDetailsUpdate, PaymentDirection, PaymentKind, PaymentStatus,
 };
 use crate::runtime::Runtime;
+#[cfg(not(feature = "uniffi"))]
+use crate::types::PaidBolt12Invoice;
```
tnull: Let's just import the type from `lightning` here.

src/event.rs
```diff
@@ -258,6 +276,90 @@ pub enum Event {
...
+// only exists in non-UniFFI builds, and the macro generates code that references struct fields
+// by name, so we can't use a single macro for both. The duplication can be removed once we
+// upgrade to UniFFI 0.29+, which supports Object types in enum variants.
+// See https://github.com/lightningdevkit/ldk-node/issues/757
+//
+// Note: The serialization formats are compatible - TLV tag 7 (bolt12_invoice) written by
+// non-UniFFI builds is silently skipped by UniFFI builds (unknown optional TLV), and UniFFI
+// builds write without tag 7, which non-UniFFI builds read as `bolt12_invoice: None`.
+#[cfg(not(feature = "uniffi"))]
+impl_writeable_tlv_based_enum!(Event,
```
tnull, line 280: Ugh, that's really ugly. Makes me wonder if it wouldn't be worth prioritizing the `uniffi` upgrade even more and then still going for the full thing right away. A few days ago I put up the dependency bump on `uniffi-bindgen-go`, but let me see how much work the actual upgrade is on our end.

vincenzopalazzo: I agree @tnull to hold off on e this and finish the unifi ASAP, this functionality is not crucial because the people who have this in prod probably do not care about this functionality.

This code duplication, I do not think, deserves to be merged, even if we are blocked by the unifi.

bindings/ldk_node.udl
vincenzopalazzo: I would love if we move with the upgrade first and then we have a way to proper implement this

src/payment/mod.rs
tnull: We should be good to drop these changes now?

src/types.rs
```diff
@@ -626,3 +626,6 @@ impl From<&(u64, Vec<u8>)> for CustomTlvRecord {
 }
 
 pub(crate) type PendingPaymentStore = DataStore<PendingPaymentDetails, Arc<Logger>>;
+
+#[cfg(not(feature = "uniffi"))]
+pub use lightning::events::PaidBolt12Invoice;
```
tnull: I don't think we need this?

tests/integration_tests_rust.rs
```diff
@@ -1273,6 +1274,95 @@ async fn simple_bolt12_send_receive() {
 	assert_eq!(node_a_payments.first().unwrap().amount_msat, Some(overpaid_amount));
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn bolt12_proof_of_payment() {
+	let (bitcoind, electrsd) = setup_bitcoind_and_electrsd();
+	let chain_source = TestChainSource::Esplora(&electrsd);
```
tnull: Please use `random_chain_source` here so this test is run on all chain sources.

tests/integration_tests_rust.rs
```diff
@@ -1273,6 +1274,95 @@ async fn simple_bolt12_send_receive() {
 	assert_eq!(node_a_payments.first().unwrap().amount_msat, Some(overpaid_amount));
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn bolt12_proof_of_payment() {
```
tnull: AFAICT this is essentially the same as `simple_bolt12_send_receive`? Should we just add a check for the proof-of-payment/invoice there?

tests/integration_tests_rust.rs
```diff
@@ -1273,6 +1274,95 @@ async fn simple_bolt12_send_receive() {
...
+				PaidBolt12Invoice::Bolt12Invoice(invoice) => {
+					assert_eq!(invoice.payment_hash(), *payment_hash);
+					assert_eq!(invoice.amount_msats(), expected_amount_msat);
+				},
+				PaidBolt12Invoice::StaticInvoice(_) => {
+					panic!("Expected Bolt12Invoice, got StaticInvoice");
+				},
+			}
+
+			// Verify serialization round-trip (tests TLV tag 7 backward compatibility)
```
tnull: Hmm, that seems very out-of-place here? If anything, a serialization roundtrip test should be a unit test in `event.rs`, but I don't think we need it.

tnull (commented):
Cool, mostly looks good, just some comments regarding cleanups we should now be able to do?

src/payment/mod.rs
vincenzopalazzo: Done — removed the non-uniffi re-export and the comment block. Only the uniffi re-export from crate::ffi remains.

src/types.rs
vincenzopalazzo: Dropped — event.rs now imports PaidBolt12Invoice directly from lightning::events for non-uniffi builds.

tests/integration_tests_rust.rs
vincenzopalazzo: Resolved by removing the separate test entirely and merging the check into simple_bolt12_send_receive, which already uses random_chain_source.

tests/integration_tests_rust.rs
vincenzopalazzo: Done — deleted bolt12_proof_of_payment and added the bolt12_invoice.is_some() assertion directly in simple_bolt12_send_receive.

tests/integration_tests_rust.rs
vincenzopalazzo: Agreed, removed along with the rest of the deleted test. The unused Readable/Writeable imports are cleaned up too.

src/ffi/types.rs
```diff
@@ -775,6 +778,100 @@ impl AsRef<LdkBolt12Invoice> for Bolt12Invoice {
...
+	fn as_ref(&self) -> &LdkStaticInvoice {
+		self.deref()
+	}
+}
+
+/// The BOLT12 invoice that was paid, surfaced in [`Event::PaymentSuccessful`].
+///
+/// [`Event::PaymentSuccessful`]: crate::Event::PaymentSuccessful
+///
+/// Note: The variant names intentionally differ from their contained types to avoid
```
tnull: Let's drop this note please. Such a comment would make sense int he commit message, but not in user-facing docs.

src/ffi/types.rs
```diff
@@ -775,6 +778,100 @@ impl AsRef<LdkBolt12Invoice> for Bolt12Invoice {
...
+				LdkPaidBolt12Invoice::StaticInvoice(invoice.inner.clone())
+			},
+		}
+	}
+}
+
+impl Writeable for PaidBolt12Invoice {
+	fn write<W: Writer>(&self, w: &mut W) -> Result<(), lightning::io::Error> {
+		// We clone `self` to convert it into the LDK native type for serialization.
+		// This only runs during event persistence, so the overhead is acceptable.
```
tnull: Ugh, this is still pretty bad. We should really avoid this, though I played with the different options here and I'm still not sure what else I find preferable. Seems every approach that avoids cloning the data (in particular struct wrapping for `uniffi`) will revert back to a lot of complications. 

Let's make this comment a `TODO: Find way to avoid cloning invoice data` so I don't totally forget to revisit this, and then we should be able to move forward here to unblock you.

src/payment/mod.rs
```diff
@@ -25,3 +25,6 @@ pub use store::{
 	ConfirmationStatus, LSPFeeLimits, PaymentDetails, PaymentDirection, PaymentKind, PaymentStatus,
 };
 pub use unified::{UnifiedPayment, UnifiedPaymentResult};
+
+#[cfg(feature = "uniffi")]
```
tnull: Let's drop this please.

vincenzopalazzo: ops good catch thanks!

src/ffi/types.rs
vincenzopalazzo: Yeah, I was not able to avoid it because the `into()` will close the type anyway, and I do not think it is a good idea to duplicate the code here, but probably I am underestimating how many times this method is called?

vincenzopalazzo: BTW, I updated the comment! Thanks for the feedback

Issue #730, opened 2025-12-22
[good first issue] Add `PaymentKind::payment_hash` as an accessor method


TheBlueMatt-macbook:
nice to not have to match on several variants to extra the hash, probably a few other common fields could use it too.

jgmcalpine:
I've implemented the accessor methods and opened a PR for review here: #734

PR #734, opened 2025-12-29
feat: add accessor methods for common PaymentKind fields

jgmcalpine:
## Description
Addresses issue #730.

Added `payment_hash()` as requested, but also identified that `preimage` and `secret` are common fields shared across most `PaymentKind` variants. Added accessor methods for all three to unify the API.

## Changes
- Added `payment_hash()`, `preimage()`, and `secret()` to `PaymentKind`.
- Refactored `PaymentDetailsUpdate::from` to use these new methods, removing boilerplate matching logic.
- Added unit test `payment_kind_accessor_methods` to verify behavior across all variants (Bolt11, Bolt12, Spontaneous, Onchain).

## PR Tasks
- [x] Added tests
- [x] Ran `cargo fmt`
- [x] Ran `cargo test`

### DIFF
--- a/bindings/ldk_node.udl
+++ b/bindings/ldk_node.udl
@@ -1,6 +1,11 @@
 namespace ldk_node {
 	Mnemonic generate_entropy_mnemonic(WordCount? word_count);
 	Config default_config();
+
+	// PaymentKind Accessors
+	PaymentHash? payment_kind_get_payment_hash(PaymentKind kind);
+	PaymentPreimage? payment_kind_get_preimage(PaymentKind kind);
+	PaymentSecret? payment_kind_get_secret(PaymentKind kind);
 };
 
 dictionary Config {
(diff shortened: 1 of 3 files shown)


tnull (commented):
> I implemented the accessors as top-level helpers in the UDL, but I gated the Rust logic behind #[cfg(feature = "uniffi")] and #[doc(hidden)] to make sure they are only compiled for the bindings and don't clutter the standard Rust API or docs.

Ah, right, we can't even expose the actual methods given that it's an enum `interface`. Well, tbh, I don't find exposing this at the top-level acceptable, and also dislike if we'd resorted to just do it for the Rust API. 

Given that I wasn't convinced we should do this from an API-design perspective in the to begin with, I now more and more think it's not worth the interface complication/incongruence across languages.

I now closed #730. Thank you for looking into it though!

PR #735, opened 2025-12-30
Update peer socket address when peer already existsRefresh the stored address if an existing peer is added again

### LINKED ISSUE #700: [BUG] Peer node socketAddress is not updated when passed a new socketAddress
if the LSP node’s IP address changes, calling set_liquidity_source_lsps2 with the new IP doesn’t update the existing peer’s socket address. The node keeps trying to connect using the old one.

if self.peers contains node_id then it returns early:
```rust

		if locked_peers.contains_key(&peer_info.node_id) {
			return Ok(());
		}

		locked_peers.insert(peer_info.node_id, peer_info);
		self.persist_peers(&*locked_peers)
	}
```

chahat-101:
Fixes #700.

Previously, re-adding an existing peer would return early and ignore an
updated socket address, causing stale peer addresses to be persisted.
This change updates and persists the peer info when the address differs,
and adds a regression test covering the behavior.

### DIFF
--- a/src/peer_store.rs
+++ b/src/peer_store.rs
@@ -43,12 +43,21 @@ where
 	pub(crate) fn add_peer(&self, peer_info: PeerInfo) -> Result<(), Error> {
 		let mut locked_peers = self.peers.write().unwrap();
 
-		if locked_peers.contains_key(&peer_info.node_id) {
-			return Ok(());
+		match locked_peers.get_mut(&peer_info.node_id) {
+			Some(existing_peer) => {
+				if existing_peer.address != peer_info.address {
+					*existing_peer = peer_info;
+					self.persist_peers(&*locked_peers)?;
+				}
+			},
+
+			None => {
+				locked_peers.insert(peer_info.node_id, peer_info);
+				self.persist_peers(&*locked_peers)?;
+			},
 		}
 
-		locked_peers.insert(peer_info.node_id, peer_info);
-		self.persist_peers(&*locked_peers)
+		Ok(())
 	}
 
 	pub(crate) fn remove_peer(&self, node_id: &PublicKey) -> Result<(), Error> {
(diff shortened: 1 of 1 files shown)


src/peer_store.rs
```diff
@@ -43,12 +43,21 @@ where
 	pub(crate) fn add_peer(&self, peer_info: PeerInfo) -> Result<(), Error> {
 		let mut locked_peers = self.peers.write().unwrap();
 
-		if locked_peers.contains_key(&peer_info.node_id) {
-			return Ok(());
+		match locked_peers.get_mut(&peer_info.node_id) {
```
tnull, line 46: I'm not sure why we bother to first look it up if the following behavior is identical to `insert` anyways? Why then not just always `insert`?

src/peer_store.rs
```diff
@@ -43,12 +43,21 @@ where
 	pub(crate) fn add_peer(&self, peer_info: PeerInfo) -> Result<(), Error> {
 		let mut locked_peers = self.peers.write().unwrap();
 
-		if locked_peers.contains_key(&peer_info.node_id) {
```
tnull, line 46: We can't just drop this, as there are instances where we'd `add_peer` but *don't* want to override the peer address if we already have one (e.g., when adding peers for inbound channels in `event.rs`). To solves this, we should probably just introduce an `override: bool` flag that lets us not return early. Alternatively, we could have two different methods for updating or inserting a peer.

chahat-101: To make sure I implement this the way you expect: would you prefer:

(a) keeping add_peer but adding an explicit override_existing: bool, or

(b) splitting this into two methods (e.g. add_peer_if_missing and upsert_peer)?

I’m happy to update call sites accordingly once I know which behavior you want to be the default.

src/peer_store.rs
chahat-101: Makes sense — I missed the inbound-channel case where we shouldn’t overwrite an existing peer address.

src/peer_store.rs
enigbe: > We can't just drop this, as there are instances where we'd add_peer but don't want to override the peer address if we already have one (e.g., when adding peers for inbound channels in event.rs).

@tnull, when adding peers for inbound channels, we only persist to the store if:
- the peer doesn't already exist, and
- we can extract an address from the counterparty's announcement info,
Given that, there shouldn't be anything to override, right?

@chahat-101 ACK for introducing an override flag

tnull: > @tnull, when adding peers for inbound channels, we only persist to the store if:
> 
> * the peer doesn't already exist, and
> * we can extract an address from the counterparty's announcement info,
>   Given that, there shouldn't be anything to override, right?

Hmm, good point, you're right. Could be argued that behavior is also odd as we might add a peer at the time of channel creation but then never update its address if they change it. However, if the user already provided some address for this peer (e.g., for an outbound channel), we shouldn't just override it arbitrarily, so this is probably fine for now.

I think then I have to eat my words (sorry for that!) and we can simply drop the `contains_key` and be done with it? We'll then always override with whatever the last-provided address is for that peer, but I guess that's expected?

PR #736, opened 2025-12-31
Fix two circular `Arc` references

TheBlueMatt:
Based on https://github.com/lightningdevkit/rust-lightning/pull/4294, basically, but the first fix should probably get backported to 0.7 as it fixes the issue for RGS ldk-node instances even though we'll have to wait for an LDK update to fix the issue for P2P gossip instances.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -110,6 +110,8 @@ use std::default::Default;
 use std::net::ToSocketAddrs;
 use std::sync::{Arc, Mutex, RwLock};
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
+#[cfg(cycle_tests)]
+use std::{any::Any, sync::Weak};
 
 pub use balance::{BalanceDetails, LightningBalance, PendingSweepBalance};
 use bitcoin::secp256k1::PublicKey;
@@ -173,6 +175,23 @@ use crate::scoring::setup_background_pathfinding_scores_sync;
 #[cfg(feature = "uniffi")]
 uniffi::include_scaffolding!("ldk_node");
 
+#[cfg(cycle_tests)]
+/// A list of [`Weak`]s which can be used to check that a [`Node`]'s inner fields are being
+/// properly released after the [`Node`] is dropped.
+pub struct LeakChecker(Vec<Weak<dyn Any + Send + Sync>>);
+
+#[cfg(cycle_tests)]
+impl LeakChecker {
+	/// Asserts that all the stored [`Weak`]s point to contents which have been freed.
+	///
+	/// This will (obviously) panic if the [`Node`] has not yet been dropped.
+	pub fn assert_no_leaks(&self) {
+		for weak in self.0.iter() {
+			assert_eq!(weak.strong_count(), 0);
+		}
+	}
+}
+
 /// The main interface object of LDK Node, wrapping the necessary LDK and BDK functionalities.
 ///
 /// Needs to be initialized and instantiated through [`Builder::build`].
@@ -208,6 +227,8 @@ pub struct Node {
 	om_mailbox: Option<Arc<OnionMessageMailbox>>,
 	async_payments_role: Option<AsyncPaymentsRole>,
 	hrn_resolver: Arc<HRNResolver>,
+	#[cfg(cycle_tests)]
+	_leak_checker: LeakChecker,
 }
 
 impl Node {
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -80,11 +80,11 @@ jobs:
       - name: Test on Rust ${{ matrix.toolchain }}
         if: "matrix.platform != 'windows-latest'"
         run: |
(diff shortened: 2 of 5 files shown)


TheBlueMatt:
Note that this is (obviously) blocked on landing the upstream LDK PR.

.github/workflows/cln-integration.yml
```diff
@@ -27,4 +27,4 @@ jobs:
           socat -d -d UNIX-LISTEN:/tmp/lightning-rpc,reuseaddr,fork TCP:127.0.0.1:9937&
 
       - name: Run CLN integration tests
-        run: RUSTFLAGS="--cfg cln_test" cargo test --test integration_tests_cln
+        run: RUSTFLAGS="--cfg cln_test --cfg cycle_tests" cargo test --test integration_tests_cln
```
tnull: Hmm, any reason we need to add this `cfg` to all integration tests? Why not only on the Rust tests where the test lives?

tests/common/mod.rs
```diff
@@ -261,10 +262,51 @@ pub(crate) fn random_config(anchor_channels: bool) -> TestConfig {
...
+		}
+		#[cfg(not(feature = "uniffi"))]
+		{
+			&self.inner.as_ref().unwrap()
+		}
+	}
+}
+
+#[cfg(cycle_tests)]
+impl Drop for TestNode {
```
tnull: I don't quite understand why we need all this `TestNode` boilerplate. If this is `cfg`-gated anways, why not just add the `assert_eq` behind the `cfg`-gate (or even as a `debug_assert`) in the actual `Node::drop` implementation?

src/lib.rs
```diff
@@ -1761,6 +1761,12 @@ impl Node {
 			Error::PersistenceFailed
 		})
 	}
+
+	#[cfg(cycle_tests)]
```
tnull: Rather than leaking internals and adding all that boilerplate which only makes sense in the `cycle_tests` context, could we just add a `LeakChecker` object (probably living in `src/leak_checker.rs`) taking a number of weak references and holding it as the last field in `Node`. This `LeakChecker` could then check all counts are 0 in its drop implementation? Seems like that would a) encapsulate the checks more cleanly b) could also be easily used to check that more (most?) of the main object `Arc`s aren't leaking on `Node::drop`?

tnull (commented):
Mostly looks good, two questions regarding the cycle tests commit.

src/lib.rs
TheBlueMatt: Sure, yea, that's cleaner. I didn't bother putting it in a separate file cause its literally 10 lines of code without comments and that seemed kinda absurd. let me know if you prefer a file.

tests/common/mod.rs
```diff
@@ -261,10 +262,55 @@ pub(crate) fn random_config(anchor_channels: bool) -> TestConfig {
...
+	fn drop(&mut self) {
+		if !std::thread::panicking() {
+			let leak_checks = (**self).leak_checker();
+			match self.inner.take().unwrap().stop() {
+				Ok(()) => {},
+				Err(e) if e == NodeError::NotRunning => {},
+				Err(e) => panic!("{e:?}"),
+			}
+			leak_checks.assert_no_leaks();
```
tnull: I'm confused: why can we not do this in `Node::drop` and avoid all the `TestNode`-related changes now? If we add a `_leak_checker` field as the last field in the `struct Node`, Rust's drop order should ensure it's always dropped last, i.e., when all other references should be gone, no?

src/lib.rs
tnull: Yeah, keeping it in the same file is fine for if you so prefer, but I'm confused why we need the `TestNode` still then? See https://github.com/lightningdevkit/ldk-node/pull/736#discussion_r2682255345

tests/common/mod.rs
TheBlueMatt: Mmm, I hadn't considered that, nifty idea.

TheBlueMatt: Made this change, though sadly note that we still need/want `TestNode` cause we have to make sure that we `stop` the node before `drop`ping it, which we don't always do in tests.

tnull: > cause we have to make sure that we stop the node before dropping it

Not sure I'm following. I guess explicitly `stop`ping in tests might be preferable, but why do we *have* "to make sure we `stop` before `drop`?

In particular, why did you remove the `Drop` implementation from `Node`? I don't think we want to do that, as we need to try to do everything to cleanly shutdown, even if users don't explicitly call stop.

TheBlueMatt: Ooops, uhhhh, cause I'm blind :). I reverted the pile of `TestNode` changes to just leave the checker.

tnull: > Ooops, uhhhh, cause I'm blind :). I reverted the pile of `TestNode` changes to just leave the checker.

Cool, seems you might still need to push these changes though :)

TheBlueMatt: lol oops. done

PR #737, opened 2026-01-01
Switch from `reqwest` to `bitreq`

TheBlueMatt:
`reqwest` is one of the largest contributors of code size and dependencies (including single-author dependencies) to much of the rust-bitcoin ecosystem, including ldk-node.

Thus, Tobin took the time to (ask an LLM to) fork `minreq` and add async support to it, including async `rustls` support. As its now a functional HTTP(s) client, its time for the ecosystem to start switching over. Luckily, its ~trivial to do.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ bdk_esplora = { version = "0.22.0", default-features = false, features = ["async
 bdk_electrum = { version = "0.23.0", default-features = false, features = ["use-rustls-ring"]}
 bdk_wallet = { version = "2.2.0", default-features = false, features = ["std", "keys-bip39"]}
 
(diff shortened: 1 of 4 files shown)


tnull (commented):
This needs a formatter run, otherwise generally LGTM.

Should be okay for RGS/scorer downloads, but for other applications (syncing, mostly) I think we want to wait switching over until `bitreq` is properly async and they have a way to reuse a single connection instead of reopening one per request (see https://github.com/rust-bitcoin/corepc/issues/441 and https://github.com/rust-bitcoin/corepc/issues/442).

tnull (commented):
This needs a bump to `bitreq` 0.3.

tnull (approved):
Thanks!

PR #739, opened 2026-01-05
Parallelize `read_payments`

tnull:
Previously, we would read entries of our payment store sequentially.
This is more or less fine when we read from a local store, but when we
read from a remote (e.g., VSS) store, all the latency could result in
considerable slowdown during startup. Here, we opt to read store entries
in batches.

### DIFF
--- a/tests/integration_tests_rust.rs
+++ b/tests/integration_tests_rust.rs
@@ -23,7 +23,8 @@ use common::{
 	expect_splice_pending_event, generate_blocks_and_wait, open_channel, open_channel_push_amt,
 	premine_and_distribute_funds, premine_blocks, prepare_rbf, random_config,
 	random_listening_addresses, setup_bitcoind_and_electrsd, setup_builder, setup_node,
-	setup_node_for_async_payments, setup_two_nodes, wait_for_tx, TestChainSource, TestSyncStore,
+	setup_node_for_async_payments, setup_two_nodes, wait_for_tx, TestChainSource, TestStoreType,
+	TestSyncStore,
 };
 use ldk_node::config::{AsyncPaymentsRole, EsploraSyncConfig};
 use ldk_node::liquidity::LSPS2ServiceConfig;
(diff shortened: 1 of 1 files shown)


src/io/utils.rs
```diff
@@ -88,38 +88,44 @@ pub(crate) fn read_or_generate_seed_file(
...
 where
 	L::Target: LdkLogger,
 {
-	let mut reader = Cursor::new(KVStoreSync::read(
-		&*kv_store,
-		NETWORK_GRAPH_PERSISTENCE_PRIMARY_NAMESPACE,
-		NETWORK_GRAPH_PERSISTENCE_SECONDARY_NAMESPACE,
-		NETWORK_GRAPH_PERSISTENCE_KEY,
-	)?);
+	let mut reader = Cursor::new(
```
TheBlueMatt: nit: might be a good chance to drop the `Cursor` use - `&mut &[u8]` also implements a read stream so you can just use that, not that its critical.

src/io/utils.rs
```diff
@@ -88,38 +88,44 @@ pub(crate) fn read_or_generate_seed_file(
 }
 
 /// Read a previously persisted [`NetworkGraph`] from the store.
-pub(crate) fn read_network_graph<L: Deref + Clone>(
+pub(crate) async fn read_network_graph<L: Deref + Clone>(
 	kv_store: Arc<DynStore>, logger: L,
```
TheBlueMatt: I don't see any reason for these methods to take the `Arc`s by ownership rather than by reference, which would avoid a bunch of `Arc::clone` calls in the builder.

src/io/utils.rs
```diff
@@ -175,63 +184,101 @@ where
...
-		let mut reader = Cursor::new(KVStoreSync::read(
-			&*kv_store,
-			PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
-			PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
-			&stored_key,
-		)?);
+	)
+	.await?;
+
+	const BATCH_SIZE: usize = 20;
```
TheBlueMatt: seems low tbh. maybe 100?

tnull: > seems low tbh. maybe 100?

Hmm, but firing off 100 requests in parallel seems also bit much, to the point where it actually hit some rate-limit once we get around to add proper per-user rate-limits to VSS? Now bumped it to 50 in a fixup.

src/io/utils.rs
tnull: Done, across the codebase.

src/io/utils.rs
tnull: Yeah, we mostly kept it consistent for convenience and to avoid constant refactoring churn (as we previously often ended up discovering we might need some downstream object to hold an `Arc` overall, requiring to change everything back from `Deref`, etc). In any case, probably a good point to start reducing the unnecessary allocations. Now added a commit that drops unnecessary `Arc` usage from a bunch of places and opts to go for plain references instead.

tnull:
> I assume the next step will be to parallelize every read operation that we can in the builder?

Not quite sure if next-next, but generally yes. Generally the plan is still to move most/all setup logic from `build` to `start`, and make the latter `async`.

For now addressed the feedback, let me know if I can squash the fixup.

src/io/utils.rs
TheBlueMatt: > to the point where it actually hit some rate-limit once we get around to add proper per-user rate-limits to VSS?

Maybe? We should definitely set VSS rate limits to be appropriate for whatever the `ldk-node` code does :). Generally, IMO, rate-limits shouldn't be (and usually aren't) "max parallel requests" but rather a token bucket over a number of seconds/minutes, so batch size here won't actually impact our rate-limit chance at all.

> firing off 100 requests in parallel seems also bit much

I was thinking of it in terms of RTTs to load the `Node` - if we have 1000 payments (pretty reasonable "high but common" use case?) a batch size of 100 is still 10 round-trips, which is pretty high. Batch size of 50 is 20 round-trips, and we start getting into 2-3 seconds to load, which is unacceptably bad IMO. Now, because there may be rate-limit concerns we can certainly just leave this at 50 and fix it properly later by doing actual batch requests to VSS, but I think we'll still get complaints at 50.

src/io/utils.rs
TheBlueMatt: nit: `Arc::clone` doesn't allocate, it only manipulates an atomic counter.

tnull:
> Fixups look fine, feel free to squash, I didn't check that no other changes slipped in.

src/io/utils.rs
tnull: Ah, true, still.

tnull:
> CI is quite sad

Ah, force-pushed a fix:

```diff
> git diff-tree -U2 dec6f220 c724a893
diff --git a/tests/integration_tests_rust.rs b/tests/integration_tests_rust.rs
index 8771c51b..4d2a1742 100644
--- a/tests/integration_tests_rust.rs
+++ b/tests/integration_tests_rust.rs
@@ -24,5 +24,6 @@ use common::{
        premine_and_distribute_funds, premine_blocks, prepare_rbf, random_config,
        random_listening_addresses, setup_bitcoind_and_electrsd, setup_builder, setup_node,
-       setup_node_for_async_payments, setup_two_nodes, wait_for_tx, TestChainSource, TestSyncStore,
+       setup_node_for_async_payments, setup_two_nodes, wait_for_tx, TestChainSource, TestStoreType,
+       TestSyncStore,
 };
 use ldk_node::config::{AsyncPaymentsRole, EsploraSyncConfig};
@@ -2326,5 +2327,6 @@ async fn payment_persistence_after_restart() {
        // Setup nodes manually so we can restart node_a with the same config
        println!("== Node A ==");
-       let config_a = random_config(true);
+       let mut config_a = random_config(true);
+       config_a.store_type = TestStoreType::Sqlite;

        let num_payments = 200;
```

PR #741, opened 2026-01-07
Set restrictive file permissions for seed file

tnull:
Previously, seed files were created using `fs::File::create()` which inherits the default umask, potentially making the sensitive seed material world-readable on Unix systems.

This change:
- Creates seed files with mode 0o600 (owner read/write only) on Unix
- Uses `create_new` instead of `create` to atomically fail if the file already exists, providing defense-in-depth against TOCTOU race conditions

Co-Authored-By: Claude AI

### DIFF
--- a/src/io/utils.rs
+++ b/src/io/utils.rs
@@ -5,11 +5,14 @@
 // http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
 // accordance with one or both of these licenses.
 
-use std::fs;
+use std::fs::{self, OpenOptions};
 use std::io::Write;
 use std::ops::Deref;
 use std::path::Path;
 use std::sync::Arc;
+
+#[cfg(unix)]
+use std::os::unix::fs::OpenOptionsExt;
 
 use bdk_chain::indexer::keychain_txout::ChangeSet as BdkIndexerChangeSet;
 use bdk_chain::local_chain::ChangeSet as BdkLocalChainChangeSet;
@@ -78,7 +81,11 @@ pub(crate) fn read_or_generate_seed_file(
 			fs::create_dir_all(parent_dir)?;
 		}
 
-		let mut f = fs::File::create(keys_seed_path)?;
+		#[cfg(unix)]
+		let mut f = OpenOptions::new().write(true).create_new(true).mode(0o400).open(keys_seed_path)?;
+
+		#[cfg(not(unix))]
+		let mut f = OpenOptions::new().write(true).create_new(true).open(keys_seed_path)?;
 
 		f.write_all(&key)?;
 


src/io/utils.rs
```diff
@@ -78,7 +81,14 @@ pub(crate) fn read_or_generate_seed_file(
 			fs::create_dir_all(parent_dir)?;
 		}
 
-		let mut f = fs::File::create(keys_seed_path)?;
+		// Create file with restrictive permissions (owner read/write only on Unix)
+		// to protect the sensitive seed material. We use `create_new` to fail if the
+		// file already exists, providing defense-in-depth against race conditions.
+		#[cfg(unix)]
+		let mut f = { OpenOptions::new().write(true).create_new(true).mode(0o600).open(keys_seed_path)? };
```
tankyleo: I'd prefer we set permissions here to 0400 to protect against accidental writes. Once the seed file is created there really shouldn't be a reason to write anything to the file.

tnull: Fair:

```diff
> git diff-tree -U2 af77366c 7e2f1fa9
diff --git a/src/io/utils.rs b/src/io/utils.rs
index 47b408e2..965831aa 100644
--- a/src/io/utils.rs
+++ b/src/io/utils.rs
@@ -86,5 +86,5 @@ pub(crate) fn read_or_generate_seed_file(
                // file already exists, providing defense-in-depth against race conditions.
                #[cfg(unix)]
-               let mut f = { OpenOptions::new().write(true).create_new(true).mode(0o600).open(keys_seed_path)? };
+               let mut f = { OpenOptions::new().write(true).create_new(true).mode(0o400).open(keys_seed_path)? };

                #[cfg(not(unix))]
```

src/io/utils.rs
```diff
@@ -78,7 +81,14 @@ pub(crate) fn read_or_generate_seed_file(
 			fs::create_dir_all(parent_dir)?;
 		}
 
-		let mut f = fs::File::create(keys_seed_path)?;
+		// Create file with restrictive permissions (owner read/write only on Unix)
```
tankyleo: Let's update the comment, and the commit message too

src/io/utils.rs
```diff
@@ -78,7 +81,14 @@ pub(crate) fn read_or_generate_seed_file(
...
+		// to protect the sensitive seed material. We use `create_new` to fail if the
+		// file already exists, providing defense-in-depth against race conditions.
+		#[cfg(unix)]
+		let mut f = { OpenOptions::new().write(true).create_new(true).mode(0o400).open(keys_seed_path)? };
```
tankyleo: nit: Don't think we need the block expression here

src/io/utils.rs
tnull: Sure, dropped the comment as it was redundant anyways and updated the commit message.

src/io/utils.rs
tnull: Done.

PR #746, opened 2026-01-08
Add Payjoin Receiver Support (BIP 77)

### LINKED ISSUE #177: Explore integration of Payjoin/PDK

### LINKED ISSUE #1019: Payjoin V2 (BIP 77) integration
## Context
Tracking issue for Payjoin V2 ([BIP 77](https://github.com/bitcoin/bips/blob/master/bip-0077.md)) onchain integration, covering receive and send. Builds on #177.

## Receiver (#746)
- [x] OHTTP relay configuration
- [x] Session persistence
- [x] PSBT handling
- [x] Mempool monitoring
- [x] Node restart recovery

## Sender
- [ ] Session initiation
- [ ] PSBT construction
- [ ] Payment tracking
- [ ] Integration tests

Camillarhi:
This PR adds support for receiving payjoin payments in LDK Node. This is currently a work in progress and implements the receiver side of the payjoin protocol.

- Add session store for persisting payjoin receiver events across restarts
- Implement `KVStorePayjoinReceiverPersister` to handle session persistence
- Add `Payjoin` as a `PaymentKind` to the payment store
- Add event polling mechanism for active payjoin sessions
- Wire up payjoin payment request and receive flows

Note on persistence: The payjoin library currently only supports synchronous persistence, but they're working on adding async support(https://github.com/payjoin/rust-payjoin/pull/1235). This PR sets up the persistence structure (`KVStorePayjoinReceiverPersister`), which will be updated to use async operations once the upstream PR is merged.

This PR partially fixes #177 and fixes #1019

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,6 +87,8 @@ prost = { version = "0.11.6", default-features = false}
 #bitcoin-payment-instructions = { version = "0.6" }
 bitcoin-payment-instructions = { git = "https://github.com/tnull/bitcoin-payment-instructions", rev = "0e430be98c09540624a68a68022ee0551e86d1be" }
 
+payjoin = { version = "0.25.0", default-features = false, features = ["v2", "io"] }
+
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3", features = ["winbase"] }
 
--- a/bindings/ldk_node.udl
+++ b/bindings/ldk_node.udl
@@ -112,6 +112,8 @@ interface Node {
 	SpontaneousPayment spontaneous_payment();
 	OnchainPayment onchain_payment();
 	UnifiedPayment unified_payment();
+	[Throws=NodeError]
+	PayjoinPayment payjoin_payment();
 	Liquidity liquidity();
 	[Throws=NodeError]
 	void lnurl_auth(string lnurl);
@@ -182,6 +184,8 @@ interface FeeRate {
 
 typedef interface UnifiedPayment;
 
+typedef interface PayjoinPayment;
+
 typedef interface Liquidity;
 
 [Error]
@@ -209,6 +213,8 @@ enum NodeError {
 	"OnchainTxSigningFailed",
 	"TxSyncFailed",
 	"TxSyncTimeout",
+	"TxLookupFailed",
+	"TxLookupTimeout",
 	"GossipUpdateFailed",
 	"GossipUpdateTimeout",
 	"LiquidityRequestFailed",
@@ -247,6 +253,9 @@ enum NodeError {
 	"LnurlAuthTimeout",
 	"InvalidLnurl",
 	"ChainSourceNotSupported",
+	"PayjoinNotConfigured",
+	"PayjoinSessionCreationFailed",
+	"PayjoinSessionFailed",
 };
 
 typedef dictionary NodeStatus;
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -49,7 +49,7 @@ use crate::chain::ChainSource;
 use crate::config::{
 	default_user_config, may_announce_channel, AnnounceError, AsyncPaymentsRole,
 	BitcoindRestClientConfig, Config, ElectrumSyncConfig, EsploraSyncConfig, HRNResolverConfig,
-	TorConfig, DEFAULT_ESPLORA_SERVER_URL, DEFAULT_LOG_FILENAME, DEFAULT_LOG_LEVEL,
+	PayjoinConfig, TorConfig, DEFAULT_ESPLORA_SERVER_URL, DEFAULT_LOG_FILENAME, DEFAULT_LOG_LEVEL,
 	DEFAULT_MAX_PROBE_AMOUNT_MSAT, DEFAULT_MIN_PROBE_AMOUNT_MSAT,
 };
 use crate::connection::ConnectionManager;
@@ -65,7 +65,8 @@ use crate::io::utils::{
 };
 use crate::io::vss_store::VssStoreBuilder;
 use crate::io::{
-	self, PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE, PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
+	self, PAYJOIN_SESSION_STORE_PRIMARY_NAMESPACE, PAYJOIN_SESSION_STORE_SECONDARY_NAMESPACE,
+	PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE, PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
 	PENDING_PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
 	PENDING_PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
 };
@@ -74,6 +75,7 @@ use crate::lnurl_auth::LnurlAuth;
 use crate::logger::{log_error, LdkLogger, LogLevel, LogWriter, Logger};
 use crate::message_handler::NodeCustomMessageHandler;
 use crate::payment::asynchronous::om_mailbox::OnionMessageMailbox;
+use crate::payment::payjoin::manager::PayjoinManager;
 use crate::peer_store::PeerStore;
 use crate::probing::{
 	HighDegreeStrategy, Prober, ProbingConfig, ProbingStrategy, ProbingStrategyKind,
@@ -83,8 +85,8 @@ use crate::runtime::{Runtime, RuntimeSpawner};
 use crate::tx_broadcaster::TransactionBroadcaster;
 use crate::types::{
 	AsyncPersister, ChainMonitor, ChannelManager, DynStore, DynStoreRef, DynStoreWrapper,
-	GossipSync, Graph, HRNResolver, KeysManager, MessageRouter, OnionMessenger, PaymentStore,
-	PeerManager, PendingPaymentStore,
+	GossipSync, Graph, HRNResolver, KeysManager, MessageRouter, OnionMessenger,
+	PayjoinSessionStore, PaymentStore, PeerManager, PendingPaymentStore,
 };
 use crate::wallet::persist::KVStoreWalletPersister;
 use crate::wallet::Wallet;
@@ -211,6 +213,8 @@ pub enum BuildError {
 	ChainTipFetchFailed,
 	/// The configured wallet rescan height is above the current chain tip.
 	WalletRescanHeightTooHigh,
+	/// The payjoin configuration requires a Bitcoin Core backend, but a different chain source was configured.
+	PayjoinConfigMismatch,
 }
 
 impl fmt::Display for BuildError {
@@ -256,6 +260,9 @@ impl fmt::Display for BuildError {
 			},
 			Self::WalletRescanHeightTooHigh => {
 				write!(f, "Wallet rescan height is above the current chain tip.")
+			},
+			Self::PayjoinConfigMismatch => {
+				write!(f, "Payjoin requires a Bitcoin Core chain source, but a different one was configured.")
 			},
 		}
 	}
@@ -635,6 +642,15 @@ impl NodeBuilder {
 
 		self.async_payments_role = role;
 		Ok(self)
+	}
+
+	/// Configures the [`Node`] instance to enable payjoin payments.
+	///
+	/// The `payjoin_config` specifies the PayJoin directory and OHTTP relay URLs required
+	/// for payjoin V2 protocol.
+	pub fn set_payjoin_config(&mut self, payjoin_config: PayjoinConfig) -> &mut Self {
+		self.config.payjoin_config = Some(payjoin_config);
+		self
 	}
 
 	/// Sets background probing config.
@@ -1199,6 +1215,14 @@ impl ArcedNodeBuilder {
 		self.inner.write().expect("lock").set_async_payments_role(role).map(|_| ())
 	}
 
+	/// Configures the [`Node`] instance to enable payjoin payments.
+	///
+	/// The `payjoin_config` specifies the PayJoin directory and OHTTP relay URLs required
+	/// for payjoin V2 protocol.
+	pub fn set_payjoin_config(&self, payjoin_config: PayjoinConfig) {
+		self.inner.write().expect("lock").set_payjoin_config(payjoin_config);
+	}
+
 	/// Configures background probing.
 	///
 	/// Use [`ProbingConfigBuilder`] to build the configuration.
@@ -1439,7 +1463,7 @@ fn build_with_store_internal(
 
 	let kv_store_ref = Arc::clone(&kv_store);
 	let logger_ref = Arc::clone(&logger);
-	let (payment_store_res, node_metris_res, pending_payment_store_res) =
+	let (payment_store_res, node_metris_res, pending_payment_store_res, payjoin_session_store_res) =
 		runtime.block_on(async move {
 			tokio::join!(
 				read_all_objects(
@@ -1453,6 +1477,12 @@ fn build_with_store_internal(
 					&*kv_store_ref,
 					PENDING_PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
 					PENDING_PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
+					Arc::clone(&logger_ref),
+				),
+				read_all_objects(
+					&*kv_store_ref,
+					PAYJOIN_SESSION_STORE_PRIMARY_NAMESPACE,
+					PAYJOIN_SESSION_STORE_SECONDARY_NAMESPACE,
 					Arc::clone(&logger_ref),
 				)
 			)
@@ -2249,6 +2279,42 @@ fn build_with_store_internal(
 
 	let pathfinding_scores_sync_url = pathfinding_scores_sync_config.map(|c| c.url.clone());
 
+	let payjoin_manager = if config.payjoin_config.is_some() {
+		if !matches!(chain_data_source_config, Some(ChainDataSourceConfig::Bitcoind { .. })) {
+			return Err(BuildError::PayjoinConfigMismatch);
+		}
+
+		let payjoin_session_store = match payjoin_session_store_res {
+			Ok(payjoin_sessions) => Arc::new(PayjoinSessionStore::new(
+				payjoin_sessions,
+				PAYJOIN_SESSION_STORE_PRIMARY_NAMESPACE.to_string(),
+				PAYJOIN_SESSION_STORE_SECONDARY_NAMESPACE.to_string(),
+				Arc::clone(&kv_store),
+				Arc::clone(&logger),
+			)),
+			Err(e) => {
+				log_error!(logger, "Failed to read payjoin session data from store: {}", e);
+				return Err(BuildError::ReadFailed);
+			},
+		};
+
+		Some(Arc::new(PayjoinManager::new(
+			Arc::clone(&payjoin_session_store),
+			Arc::clone(&logger),
+			Arc::clone(&config),
+			Arc::clone(&wallet),
+			Arc::clone(&fee_estimator),
+			Arc::clone(&chain_source),
+			Arc::clone(&channel_manager),
+			stop_sender.subscribe(),
+			Arc::clone(&payment_store),
+			Arc::clone(&pending_payment_store),
+			Arc::clone(&tx_broadcaster),
+		)))
+	} else {
+		None
+	};
+
 	#[cfg(cycle_tests)]
 	let mut _leak_checker = crate::LeakChecker(Vec::new());
 	#[cfg(cycle_tests)]
@@ -2342,6 +2408,7 @@ fn build_with_store_internal(
 		prober,
 		#[cfg(cycle_tests)]
 		_leak_checker,
+		payjoin_manager,
 	})
 }
 
--- a/src/chain/bitcoind.rs
+++ b/src/chain/bitcoind.rs
@@ -34,7 +34,7 @@ use serde::Serialize;
 use super::WalletSyncStatus;
 use crate::config::{
 	BitcoindRestClientConfig, Config, DEFAULT_FEE_RATE_CACHE_UPDATE_TIMEOUT_SECS,
-	DEFAULT_TX_BROADCAST_TIMEOUT_SECS,
+	DEFAULT_TX_BROADCAST_TIMEOUT_SECS, DEFAULT_TX_LOOKUP_TIMEOUT_SECS,
 };
 use crate::fee_estimator::{
 	apply_post_estimation_adjustments, get_all_conf_targets, get_num_block_defaults_for_target,
@@ -650,6 +650,57 @@ impl BitcoindChainSource {
 						Err(e) => self.log_broadcast_error(e, &[txid], &txs),
 					}
 				}
+			},
+		}
+	}
+
+	pub(crate) async fn can_broadcast_transaction(&self, tx: &Transaction) -> Result<bool, Error> {
+		let timeout_fut = tokio::time::timeout(
+			Duration::from_secs(DEFAULT_TX_LOOKUP_TIMEOUT_SECS),
+			self.api_client.test_mempool_accept(tx),
+		);
+
+		match timeout_fut.await {
+			Ok(res) => res.map_err(|e| {
+				log_error!(
+					self.logger,
+					"Failed to test mempool accept for transaction {}: {}",
+					tx.compute_txid(),
+					e
+				);
+				Error::WalletOperationFailed
+			}),
+			Err(e) => {
+				log_error!(
+					self.logger,
+					"Failed to test mempool accept for transaction {} due to timeout: {}",
+					tx.compute_txid(),
+					e
+				);
+				log_trace!(
+					self.logger,
+					"Failed test mempool accept transaction bytes: {}",
+					log_bytes!(tx.encode())
+				);
+				Err(Error::WalletOperationTimeout)
+			},
+		}
+	}
+
+	pub(crate) async fn get_transaction(&self, txid: &Txid) -> Result<Option<Transaction>, Error> {
+		let timeout_fut = tokio::time::timeout(
+			Duration::from_secs(DEFAULT_TX_LOOKUP_TIMEOUT_SECS),
+			self.api_client.get_raw_transaction(txid),
+		);
+
+		match timeout_fut.await {
+			Ok(res) => res.map_err(|e| {
+				log_error!(self.logger, "Failed to get transaction {}: {}", txid, e);
+				Error::TxLookupFailed
+			}),
+			Err(e) => {
+				log_error!(self.logger, "Failed to get transaction {} due to timeout: {}", txid, e);
+				Err(Error::TxLookupTimeout)
 			},
 		}
 	}
@@ -1269,6 +1320,34 @@ impl BitcoindClient {
 			.collect();
 		Ok(evicted_txids)
 	}
+
+	/// Tests whether the provided transaction would be accepted by the mempool.
+	pub(crate) async fn test_mempool_accept(
+		&self, tx: &Transaction,
+	) -> Result<bool, RpcClientError> {
+		match self {
+			BitcoindClient::Rpc { rpc_client, .. } => {
+				Self::test_mempool_accept_inner(Arc::clone(rpc_client), tx).await
+			},
+			BitcoindClient::Rest { rpc_client, .. } => {
+				// We rely on the internal RPC client to make this call, as this
+				// operation is not supported by Bitcoin Core's REST interface.
+				Self::test_mempool_accept_inner(Arc::clone(rpc_client), tx).await
+			},
+		}
+	}
+
+	async fn test_mempool_accept_inner(
+		rpc_client: Arc<RpcClient>, tx: &Transaction,
+	) -> Result<bool, RpcClientError> {
+		let tx_serialized = bitcoin::consensus::encode::serialize_hex(tx);
+		let tx_array = serde_json::json!([tx_serialized]);
+
+		rpc_client
+			.call_method::<TestMempoolAcceptResponse>("testmempoolaccept", &[tx_array])
+			.await
+			.map(|resp| resp.0)
+	}
 }
 
 impl BlockSource for BitcoindClient {
@@ -1438,6 +1517,23 @@ impl TryInto<SubmitPackageResponse> for JsonResponse {
 				return Err(response);
 			},
 		}
+	}
+}
+
+pub(crate) struct TestMempoolAcceptResponse(pub bool);
+
+impl TryInto<TestMempoolAcceptResponse> for JsonResponse {
+	type Error = String;
+	fn try_into(self) -> Result<TestMempoolAcceptResponse, String> {
+		let array =
+			self.0.as_array().ok_or("Failed to parse testmempoolaccept response".to_string())?;
+		let first =
+			array.first().ok_or("Empty array response from testmempoolaccept".to_string())?;
+		let allowed = first
+			.get("allowed")
+			.and_then(|v| v.as_bool())
+			.ok_or("Missing 'allowed' field in testmempoolaccept response".to_string())?;
+		Ok(TestMempoolAcceptResponse(allowed))
 	}
 }
 
--- a/src/chain/electrum.rs
+++ b/src/chain/electrum.rs
@@ -20,6 +20,7 @@ use bitcoin::transaction::Version;
 use bitcoin::{FeeRate, Network, Script, ScriptBuf, Transaction, Txid};
 use electrum_client::{
 	Batch, Client as ElectrumClient, ConfigBuilder as ElectrumConfigBuilder, ElectrumApi,
+	Error as ElectrumError,
 };
 use lightning::chain::{Confirm, Filter, WatchedOutput};
 use lightning::util::ser::Writeable;
@@ -27,8 +28,8 @@ use lightning_transaction_sync::ElectrumSyncClient;
 
 use super::WalletSyncStatus;
 use crate::config::{
-	clamp_full_scan_stop_gap, Config, ElectrumSyncConfig, MAX_FULL_SCAN_STOP_GAP,
-	MIN_FULL_SCAN_STOP_GAP,
+	clamp_full_scan_stop_gap, Config, ElectrumSyncConfig, DEFAULT_TX_LOOKUP_TIMEOUT_SECS,
+	MAX_FULL_SCAN_STOP_GAP, MIN_FULL_SCAN_STOP_GAP,
 };
 use crate::error::Error;
 use crate::fee_estimator::{
@@ -374,6 +375,22 @@ impl ElectrumChainSource {
 				}
 			},
 		}
+	}
+
+	pub(crate) async fn get_transaction(&self, txid: &Txid) -> Result<Option<Transaction>, Error> {
+		let electrum_client: Arc<ElectrumRuntimeClient> = if let Some(client) =
+			self.electrum_runtime_status.read().expect("lock").client().as_ref()
+		{
+			Arc::clone(client)
+		} else {
+			debug_assert!(
+				false,
+				"We should have started the chain source before getting transactions"
+			);
+			return Err(Error::TxLookupFailed);
+		};
+
+		electrum_client.get_transaction(txid).await
 	}
 }
 
@@ -792,6 +809,37 @@ impl ElectrumRuntimeClient {
 		}
 
 		Ok(new_fee_rate_cache)
+	}
+
+	async fn get_transaction(&self, txid: &Txid) -> Result<Option<Transaction>, Error> {
+		let electrum_client = Arc::clone(&self.electrum_client);
+		let txid_copy = *txid;
+
+		let spawn_fut =
+			self.runtime.spawn_blocking(move || electrum_client.transaction_get(&txid_copy));
+		let timeout_fut =
+			tokio::time::timeout(Duration::from_secs(DEFAULT_TX_LOOKUP_TIMEOUT_SECS), spawn_fut);
+
+		match timeout_fut.await {
+			Ok(res) => match res {
+				Ok(inner_res) => match inner_res {
+					Ok(tx) => Ok(Some(tx)),
+					Err(ElectrumError::Protocol(_)) => Ok(None),
+					Err(e) => {
+						log_error!(self.logger, "Failed to get transaction {}: {}", txid, e);
+						Err(Error::TxLookupFailed)
+					},
+				},
+				Err(e) => {
+					log_error!(self.logger, "Failed to get transaction {}: {}", txid, e);
+					Err(Error::TxLookupFailed)
+				},
+			},
+			Err(e) => {
+				log_error!(self.logger, "Failed to get transaction {} due to timeout: {}", txid, e);
+				Err(Error::TxLookupTimeout)
+			},
+		}
 	}
 }
 
--- a/src/chain/mod.rs
+++ b/src/chain/mod.rs
@@ -13,7 +13,7 @@ use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
-use bitcoin::{Script, Txid};
+use bitcoin::{Script, Transaction, Txid};
 use lightning::chain::{BlockLocator, Filter};
 
 use crate::chain::bitcoind::{BitcoindChainSource, UtxoSourceClient};
@@ -533,6 +533,42 @@ impl ChainSource {
 				}
 			}
 		}
+	}
+
+	pub(crate) fn can_broadcast_transaction(&self, tx: &Transaction) -> Result<bool, Error> {
+		tokio::task::block_in_place(|| {
+			tokio::runtime::Handle::current().block_on(async {
+				match &self.kind {
+					ChainSourceKind::Bitcoind(bitcoind_chain_source) => {
+						bitcoind_chain_source.can_broadcast_transaction(tx).await
+					},
+					ChainSourceKind::Esplora { .. } => {
+						// Esplora does not support a testmempoolaccept equivalent
+						unreachable!(
+							"Broadcast suitability check not supported by Esplora backend."
+						)
+					},
+					ChainSourceKind::Electrum { .. } => {
+						// Electrum does not support a testmempoolaccept equivalent.
+						unreachable!(
+							"Broadcast suitability check not supported by Electrum backend."
+						)
+					},
+				}
+			})
+		})
+	}
+
+	pub(crate) fn get_transaction(&self, txid: &Txid) -> Result<Option<Transaction>, Error> {
+		tokio::task::block_in_place(|| {
+			tokio::runtime::Handle::current().block_on(async {
+				match &self.kind {
+					ChainSourceKind::Bitcoind(bitcoind) => bitcoind.get_transaction(txid).await,
+					ChainSourceKind::Esplora(esplora) => esplora.get_transaction(txid).await,
+					ChainSourceKind::Electrum(electrum) => electrum.get_transaction(txid).await,
+				}
+			})
+		})
 	}
 }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -13,6 +13,7 @@ use std::time::Duration;
 
 use bitcoin::secp256k1::PublicKey;
 use bitcoin::Network;
+use bitreq::URL;
 use lightning::ln::msgs::SocketAddress;
 use lightning::routing::gossip::NodeAlias;
 use lightning::routing::router::RouteParametersConfig;
@@ -136,6 +137,18 @@ pub(crate) const LIQUIDITY_DISCOVERY_RETRY_INITIAL_DELAY: Duration = Duration::f
 // thereafter until every configured LSP has been discovered.
 pub(crate) const LIQUIDITY_DISCOVERY_RETRY_MAX_DELAY: Duration = Duration::from_secs(60 * 60);
 
+// The time interval at which we resume persisted payjoin sessions.
+pub(crate) const PAYJOIN_RESUME_INTERVAL: Duration = Duration::from_secs(15);
+
+// The duration after which completed or failed payjoin sessions are cleaned up (24 hours).
+pub(crate) const PAYJOIN_SESSION_CLEANUP_AGE_SECS: u64 = 24 * 60 * 60;
+
+// The interval at which we check for old payjoin sessions to clean up (1 hour).
+pub(crate) const PAYJOIN_SESSION_CLEANUP_INTERVAL: Duration = Duration::from_secs(60 * 60);
+
+// The default timeout after which we abort a transaction lookup operation.
+pub(crate) const DEFAULT_TX_LOOKUP_TIMEOUT_SECS: u64 = 10;
+
 #[derive(Debug, Clone)]
 #[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
 /// Represents the configuration of an [`Node`] instance.
@@ -153,8 +166,9 @@ pub(crate) const LIQUIDITY_DISCOVERY_RETRY_MAX_DELAY: Duration = Duration::from_
 /// | `probing_liquidity_limit_multiplier`   | 3                                    |
 /// | `anchor_channels_config`               | AnchorChannelsConfig::default()      |
 /// | `route_parameters`                     | None                                 |
-/// | `tor_config`                           | None                                 |
+/// | `tor_config`                           | None               				|
 /// | `hrn_config`                           | HumanReadableNamesConfig::default()  |
+/// | `payjoin_config`                     	 | None                                 |
 ///
 /// See [`AnchorChannelsConfig`] and [`RouteParametersConfig`] for more information regarding their
 /// respective default values.
@@ -219,6 +233,8 @@ pub struct Config {
 	///
 	/// [BIP 353]: https://github.com/bitcoin/bips/blob/master/bip-0353.mediawiki
 	pub hrn_config: HumanReadableNamesConfig,
+	/// Configuration options for PayJoin payments.
+	pub payjoin_config: Option<PayjoinConfig>,
 }
 
 impl Default for Config {
@@ -235,6 +251,7 @@ impl Default for Config {
 			route_parameters: None,
 			node_alias: None,
 			hrn_config: HumanReadableNamesConfig::default(),
+			payjoin_config: None,
 		}
 	}
 }
@@ -776,6 +793,16 @@ pub enum AsyncPaymentsRole {
 	/// Node acts as a server in an async payments context. This means that it will hold async payments HTLCs and onion
 	/// messages for its peers.
 	Server,
+}
+
+/// Configuration options for PayJoin payments.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct PayjoinConfig {
+	/// The URL of the PayJoin directory
+	pub payjoin_directory: URL,
+	/// The URLs of the OHTTP relays to use for sending OHTTP requests to PayJoin receivers.
+	pub ohttp_relays: Vec<URL>,
 }
 
 #[cfg(test)]
--- a/src/error.rs
+++ b/src/error.rs
@@ -63,6 +63,10 @@ pub enum Error {
 	TxSyncFailed,
 	/// A transaction sync operation timed out.
 	TxSyncTimeout,
+	/// A transaction lookup operation failed.
+	TxLookupFailed,
+	/// A transaction lookup operation timed out.
+	TxLookupTimeout,
 	/// A gossip updating operation failed.
 	GossipUpdateFailed,
 	/// A gossip updating operation timed out.
@@ -139,6 +143,12 @@ pub enum Error {
 	InvalidLnurl,
 	/// The configured chain source is not supported.
 	ChainSourceNotSupported,
+	/// Payjoin is not configured.
+	PayjoinNotConfigured,
+	/// Payjoin session creation failed.
+	PayjoinSessionCreationFailed,
+	/// Payjoin session failed.
+	PayjoinSessionFailed,
 }
 
 impl fmt::Display for Error {
@@ -173,6 +183,8 @@ impl fmt::Display for Error {
 			Self::OnchainTxSigningFailed => write!(f, "Failed to sign given transaction."),
 			Self::TxSyncFailed => write!(f, "Failed to sync transactions."),
 			Self::TxSyncTimeout => write!(f, "Syncing transactions timed out."),
+			Self::TxLookupFailed => write!(f, "Failed to look up transaction."),
+			Self::TxLookupTimeout => write!(f, "Transaction lookup timed out."),
 			Self::GossipUpdateFailed => write!(f, "Failed to update gossip data."),
 			Self::GossipUpdateTimeout => write!(f, "Updating gossip data timed out."),
 			Self::LiquidityRequestFailed => write!(f, "Failed to request inbound liquidity."),
@@ -227,6 +239,9 @@ impl fmt::Display for Error {
 			Self::ChainSourceNotSupported => {
 				write!(f, "The configured chain source is not supported.")
 			},
+			Self::PayjoinNotConfigured => write!(f, "Payjoin is not configured."),
+			Self::PayjoinSessionCreationFailed => write!(f, "Payjoin session creation failed."),
+			Self::PayjoinSessionFailed => write!(f, "Payjoin session failed."),
 		}
 	}
 }
--- a/src/payment/payjoin/manager.rs
+++ b/src/payment/payjoin/manager.rs
@@ -1,0 +1,893 @@
+// This file is Copyright its original authors, visible in version control history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
+// accordance with one or both of these licenses.
+
+use bitcoin::{Amount, FeeRate, TxIn};
+use lightning::ln::channelmanager::PaymentId;
+use lightning::log_warn;
+use payjoin::persist::OptionalTransitionOutcome;
+use payjoin::receive::InputPair;
+use payjoin::ImplementationError;
+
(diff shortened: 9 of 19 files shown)


DanGould:
Are you stuck? Did something in our library break CI @Camillarhi

bindings/ldk_node.udl
```diff
@@ -221,6 +224,9 @@ enum NodeError {
 	"LnurlAuthFailed",
 	"LnurlAuthTimeout",
 	"InvalidLnurl",
+	"PayjoinNotConfigured",
+	"PayjoinSessionCreationFailed",
+	"PayjoinSessionFailed"
```
bc1cindy: `PayjoinSessionFailed` is used for infrastructure failures (coin selection, `InputPair` construction) and network errors (poll/post requests). The upstream lib distinguishes these via `SelectionError`, `InputContributionError`, and `ReceiverError` , would it be worth preserving that granularity so callers have more info and know how to act?

src/chain/bitcoind.rs
```diff
@@ -620,6 +620,57 @@ impl BitcoindChainSource {
 			}
 		}
 	}
+
+	pub(crate) async fn can_broadcast_transaction(&self, tx: &Transaction) -> Result<bool, Error> {
+		let timeout_fut = tokio::time::timeout(
+			Duration::from_secs(DEFAULT_TX_BROADCAST_TIMEOUT_SECS),
```
bc1cindy: since `testmempoolaccept` is local validation only, shouldn't this use `DEFAULT_TX_LOOKUP_TIMEOUT_SECS`?

src/chain/electrum.rs
```diff
@@ -652,6 +667,48 @@ impl ElectrumRuntimeClient {
...
+		);
+
+		match timeout_fut.await {
+			Ok(res) => match res {
+				Ok(inner_res) => match inner_res {
+					Ok(tx) => Ok(Some(tx)),
+					Err(e) => {
+						// Check if it's a "not found" error
+						let error_str = e.to_string();
+						if error_str.contains("No such mempool or blockchain transaction")
```
bc1cindy: matching errors by string seems fragile, as the message varies by server implementation. 
e.g. tested: `ssl://electrum.blockstream.info:60002: Protocol(String("missing transaction"))` , which doesn't match either string here and `Ok(None)` would never be returned. the caller sees a sync failure instead

src/chain/electrum.rs
```diff
@@ -652,6 +667,48 @@ impl ElectrumRuntimeClient {
...
+
+	async fn get_transaction(&self, txid: &Txid) -> Result<Option<Transaction>, Error> {
+		let electrum_client = Arc::clone(&self.electrum_client);
+		let txid_copy = *txid;
+
+		let spawn_fut =
+			self.runtime.spawn_blocking(move || electrum_client.transaction_get(&txid_copy));
+		let timeout_fut = tokio::time::timeout(
+			Duration::from_secs(
+				self.sync_config.timeouts_config.lightning_wallet_sync_timeout_secs,
```
bc1cindy: this is a single onchain tx lookup, should this use `tx_broadcast_timeout_secs` (10s)  instead of `lightning_wallet_sync_timeout_secs` (30s)? or would a dedicated `tx_lookup_timeout_secs` field make more sense?

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,779 @@
...
+			},
+		}
+		Ok(())
+	}
+
+	async fn monitor_payjoin_proposal(
+		&self, proposal: Receiver<Monitor>, persister: &KVStorePayjoinReceiverPersister,
+	) -> Result<(), Error> {
+		// On a session resumption, the receiver will resume again in this state.
+		let poll_interval = tokio::time::Duration::from_millis(200);
```
bc1cindy: this seems aggressive for a daemon (`payjoin-cli` uses 200ms because its a one-shot CLI with manual retry) but for ldk node maybe 1-5 seconds would be more appropriate?

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,779 @@
...
+						"Failed to create persister for session {:?}: {:?}",
+						session_id,
+						e
+					);
+					continue;
+				},
+			};
+
+			match replay_receiver_event_log_async(&recv_persister).await {
+				Ok((receiver_state, _)) => {
```
bc1cindy, line 857: `SessionHistory` is discarded with _ here.

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,779 @@
...
+
+		Ok(())
+	}
+
+	async fn handle_closed_session(
+		&self, outcome: SessionOutcome, persister: &KVStorePayjoinReceiverPersister,
+	) -> Result<(), Error> {
+		let session_id = persister.session_id();
+		let mut session =
+			self.payjoin_session_store.get(&session_id).ok_or(Error::InvalidPaymentId)?;
```
bc1cindy: `Error::InvalidPaymentId` is used here and in lines 537, 574, and 585, but the session was found successfully, the ID is valid. The actual error is a missing `txid`/`fallback_tx`, which makes failures hard to debug in prod

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,779 @@
...
+			},
+			SessionOutcome::FallbackBroadcasted => {
+				log_info!(self.logger, "Payjoin failed. Fallback transaction was broadcasted.");
+				session.status = PayjoinStatus::Failed;
+				self.payjoin_session_store.insert_or_update(session)?;
+			},
+			SessionOutcome::Cancel => {
+				log_info!(self.logger, "Payjoin session was cancelled.");
+
+				let fallback_tx = session.fallback_tx.as_ref().ok_or(Error::InvalidPaymentId)?;
```
bc1cindy: `Error::InvalidPaymentId` is used here but the session was found successfully, the ID is valid. The actual error is a missing `txid`/`fallback_tx`, which makes failures hard to debug in prod

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,779 @@
...
+	) -> Result<(), Error> {
+		let proposal = proposal
+			.finalize_proposal(|psbt| {
+				self.wallet
+					.process_psbt(psbt.clone())
+					.map_err(|e| ImplementationError::from(e.to_string().as_str()))
+			})
+			.save_async(persister)
+			.await
+			.map_err(|_| Error::PersistenceFailed)?;
```
bc1cindy: `finalize_proposal` returns `MaybeTransientTransition` to distinguish transient from fatal errors, but `.map_err(|_| Error::PersistenceFailed)` collapses both into the same error. Transient failures won't be retried and will surface as `PersistenceFailed`, which is misleading

src/payment/store.rs
```diff
@@ -493,7 +506,11 @@ impl_writeable_tlv_based_enum!(PaymentKind,
 		(2, preimage, option),
 		(3, quantity, option),
 		(4, secret, option),
-	}
+	},
+	(11, Payjoin)=>{
```
bc1cindy: all other `PaymentKind` variants use even TLV tags (0, 2, 4, 6, 8, 10), was 11 intentional, or should this be 12?

src/builder.rs
```diff
@@ -1841,6 +1861,34 @@ fn build_with_store_internal(
...
 
+	let payjoin_session_store = match payjoin_session_store_res {
+		Ok(payjoin_sessions) => Arc::new(PayjoinSessionStore::new(
+			payjoin_sessions,
+			PAYJOIN_SESSION_STORE_PRIMARY_NAMESPACE.to_string(),
+			PAYJOIN_SESSION_STORE_SECONDARY_NAMESPACE.to_string(),
+			Arc::clone(&kv_store),
+			Arc::clone(&logger),
+		)),
+		Err(e) => {
```
bc1cindy: `payjoin_session_store_res` doesn't handle `ErrorKind::NotFound`, which will occur on any node that has never used payjoin. fix suggestion in `read_payjoin_sessions`

src/builder.rs
```diff
@@ -1841,6 +1861,34 @@ fn build_with_store_internal(
...
+			log_error!(logger, "Failed to read payjoin session data from store: {}", e);
+			return Err(BuildError::ReadFailed);
+		},
+	};
+
+	let payjoin_manager = Arc::new(PayjoinManager::new(
```
bc1cindy: `payjoin_manager` is unconditionally initialized even when `set_payjoin_config` was never called.
This causes two background tasks to always run (`resume_payjoin_sessions` and `cleanup_old_sessions`). I understand the established pattern for optional subsystems at ldk node is `Option<Arc<T>>` (see `liquidity_source: Option<Arc<LiquiditySource<...>>>`)

Cargo.toml
```diff
@@ -81,6 +81,8 @@ prost = { version = "0.11.6", default-features = false}
 #bitcoin-payment-instructions = { version = "0.6" }
 bitcoin-payment-instructions = { git = "https://github.com/jkczyz/bitcoin-payment-instructions", rev = "869fd348c3ca0c78f439d2f31181f4d798c6b20e" }
 
+payjoin = { git = "https://github.com/payjoin/rust-payjoin.git", package = "payjoin", default-features = false, features = ["v2", "io"] }
```
bc1cindy: should this pin a `rev`?

src/io/utils.rs
```diff
@@ -702,6 +703,83 @@ where
...
+	L::Target: LdkLogger,
+{
+	let mut res = Vec::new();
+
+	let mut stored_keys = KVStore::list(
+		&*kv_store,
+		PAYJOIN_SESSION_STORE_PRIMARY_NAMESPACE,
+		PAYJOIN_SESSION_STORE_SECONDARY_NAMESPACE,
+	)
+	.await?;
```
bc1cindy: `list()` returns `NotFound` when the namespace has never been written, maybe consider something like this handling it?

```
rust.or_else(|e| if e.kind() == lightning::io::ErrorKind::NotFound {
    Ok(vec![])
} else {
    Err(e)
})?;
```

bc1cindy (commented):
hi there!

congrats and thanks for your work! 

I talked to @spacebear21 and we agreed that I would review this PR, so I took a first look and here are some points:

needs to run `cargo fmt --all`

bindings/ldk_node.udl
Camillarhi: These are the errors exposed at the ldk-node API level. More granular details from the upstream errors are logged for debugging purposes.

src/chain/bitcoind.rs
Camillarhi: Yes, will update to use `DEFAULT_TX_LOOKUP_TIMEOUT_SECS`

src/payment/payjoin/manager.rs
Camillarhi: The `SessionHistory` is intentionally discarded here. The current implementation doesn't have a use for it. Is there a specific use case you have in mind that we should handle?

src/payment/store.rs
Camillarhi: 11 is intentional here. Since this is newly added, we're following the "it's okay to be odd" rule, which allows odd TLV types for optional fields that can be safely ignored by older readers. So 11 is correct rather than 12.

src/io/utils.rs
Camillarhi: Thanks for the suggestion, but the `list()` implementations across all return `Ok(vec![])` when the namespace is empty or hasn't been written to yet.

src/builder.rs
Camillarhi: Thanks for the suggestion, but the `list()` implementations across all return `Ok(vec![])` when the namespace is empty or hasn't been written to yet.

src/payment/payjoin/manager.rs
Camillarhi: `PersistenceFailed` here doesn't cancel the session. The session remains active in its current state and will be retried on the next poll cycle. Sessions are only marked as failed when we receive a `Closed` event. Retrying transient failures is handled naturally by the polling loop.

src/payment/payjoin/manager.rs
Camillarhi: Thanks for pointing this out. These are actually defensive checks for fields that should always be held by the time we reach these states. I will return a more descriptive error with a log line that makes failures clearer in prod.

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,779 @@
...
+					fallback_tx,
+					lightning::chain::chaininterface::TransactionType::Sweep { channels: vec![] },
+				)]);
+				session.status = PayjoinStatus::Failed;
+				self.payjoin_session_store.insert_or_update(session)?;
+			},
+			SessionOutcome::Failure => {
+				log_error!(self.logger, "Payjoin session failed due to a protocol error.");
+
+				let fallback_tx = session.fallback_tx.as_ref().ok_or(Error::InvalidPaymentId)?;
```
bc1cindy: `test_session_fatal_error` covers `Failure` with `fallback_tx: None`, when `check_broadcast_suitability` rejects the PSBT. In that case, the `?` returns before failed status is persisted

src/chain/mod.rs
```diff
@@ -468,6 +468,42 @@ impl ChainSource {
...
+						// Esplora does not support a testmempoolaccept equivalent, so we skip
+						// the broadcast suitability check and assume the transaction is broadcastable.
+						log_debug!(self.logger, "Skipping broadcast suitability check: not supported by Esplora backend.");
+						Ok(true)
+					},
+					ChainSourceKind::Electrum{..} => {
+						// Electrum does not support a testmempoolaccept equivalent, so we skip
+						// the broadcast suitability check and assume the transaction is broadcastable.
+						log_debug!(self.logger, "Skipping broadcast suitability check: not supported by Electrum backend.");
+						Ok(true)
```
spacebear21: The broadcast suitability check is important for non-interactive receivers (e.g. payment processors) to protect against probing attacks, where a malicious sender can repeatedly send proposals to have the non-interactive receiver reveal the UTXOs it owns with the proposals it modifies. The "broadcastable" check ensures the receiver can broadcast the fallback transaction in case the payjoin fails, effectively incurring a cost to the attacker.

Electrum's lack of support for this kind of check is a [longstanding issue](https://github.com/orgs/payjoin/discussions/900) unfortunately.

I'm not familiar enough with LDK-node usage to know if it's typically used in an interactive way (receiver takes manual action to receive payment) or non-interactive (e.g. bolt12 or similar static payment with no receiver action per payment)?

In interactive conditions, the receiver may use the `assume_interactive_receiver` transition method to explicitly skip the broadcast check altogether. In non-interactive conditions though, we should find an alternative way to perform this check. Or at least return an error instead of silently skipping the check, so it can be handled by the implementer.

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,779 @@
...
+		let session_id = persister.session_id();
+		let mut session =
+			self.payjoin_session_store.get(&session_id).ok_or(Error::InvalidPaymentId)?;
+
+		// Note: broadcast suitability can only be fully verified when using the BitcoindRpc
+		// backend. For Esplora and Electrum backends this check is skipped and the transaction
+		// is assumed to be broadcastable.
+		let proposal = proposal
+			.check_broadcast_suitability(None, |tx| {
+				self.chain_source
```
spacebear21: So here for example we could call `proposal.assume_interactive_receiver()` if applicable.

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,779 @@
...
+
+		let selected_input = proposal.try_preserving_privacy(candidate_inputs).map_err(|e| {
+			log_error!(self.logger, "Failed to select input for payjoin contribution: {}", e);
+			Error::PayjoinSessionFailed
+		})?;
+		let proposal = proposal
+			.contribute_inputs(vec![selected_input])
+			.map_err(|e| {
+				log_error!(self.logger, "Failed to contribute inputs to payjoin: {}", e);
+				Error::PayjoinSessionFailed
```
spacebear21, line 474: Are these failure modes handled by broadcasting the fallback transaction? I'm trying to find where these errors are handled but not sure I'm tracing it correctly.

src/config.rs
```diff
@@ -619,6 +636,16 @@ pub enum AsyncPaymentsRole {
...
 }
 
+/// Configuration options for PayJoin payments.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct PayjoinConfig {
+	/// The URL of the PayJoin directory to use for discovering PayJoin receivers.
+	pub payjoin_directory: URL,
+	/// The URL of the OHTTP relay to use for sending OHTTP requests to PayJoin receivers.
+	pub ohttp_relay: URL,
```
spacebear21: We can make this a list of OHTTP relays and randomize the relay per-request. This also improves reliability by enabling retries to other relays in case one is offline for any reason.

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,779 @@
...
+		let now = std::time::SystemTime::now()
+			.duration_since(std::time::UNIX_EPOCH)
+			.unwrap_or(std::time::Duration::from_secs(0))
+			.as_secs();
+
+		let sessions_to_remove: Vec<PaymentId> = self
+			.payjoin_session_store
+			.list_filter(|s| {
+				let is_terminal =
+					s.status == PayjoinStatus::Completed || s.status == PayjoinStatus::Failed;
```
spacebear21, line 1072: It may be desirable to keep "Completed" sessions ~forever since it makes it possible to reconstruct past Payjoins for e.g. displaying transaction history.

spacebear21 (commented):
This is looking great! I have high-level feedback from the payjoin side, but overall agree with the approach and implementation.

src/chain/electrum.rs
Camillarhi: You're right, the string matching is fragile.  I'll fix it by matching on `Error::Protocol` directly.

src/payment/payjoin/manager.rs
Camillarhi: Good point, I'll update it to 2 seconds to be less aggressive while still being reasonably responsive.

Cargo.toml
Camillarhi: Thanks for pointing that out. I'll pin it to a specific rev

src/chain/mod.rs
zealsham: > I'm not familiar enough with LDK-node usage to know if it's typically used in an interactive way (receiver takes manual action to receive payment) or non-interactive 

it's mostly non-interactive

Camillarhi: It is non-interactive. I'll update both Esplora and Electrum to return an error instead so the limitation is explicit. Thanks for the context on the probing attack.

src/payment/payjoin/manager.rs
Camillarhi: When the session propagates to a `Closed` state which is handled in `handle_closed_session`, On `Failure` or `Cancel`, the fallback transaction (extracted and stored at the start of `check_proposal`) is broadcast to ensure the receiver still gets paid.

Camillarhi: I also have a question, Should failures at `create_poll_request`, `try_preserving_privacy`, `contribute_inputs`, and `create_post_request` trigger fallback broadcast immediately, or are these considered retryable until the session expires or propagates a `Closed` state? Wanted to align on the intended behavior before handling these inline.

src/payment/payjoin/manager.rs
Camillarhi: Completed payjoin sessions are already recorded as `Payjoin` payments in the `PaymentStore` with full details on success, so the session itself doesn't need to be kept long-term for history purposes. @tnull, what are your thoughts on this as well.

src/config.rs
Camillarhi: I changed `ohttp_relay` to `ohttp_relays: Vec<URL>`. Each request now picks a random starting relay and iterates through the rest as fallbacks, so if one relay is offline the next is tried immediately without waiting for a retry cycle.

src/chain/mod.rs
```diff
@@ -468,6 +468,46 @@ impl ChainSource {
...
+			tokio::runtime::Handle::current().block_on(async {
+				match &self.kind {
+					ChainSourceKind::Bitcoind(bitcoind_chain_source) => {
+						bitcoind_chain_source.can_broadcast_transaction(tx).await
+					},
+					ChainSourceKind::Esplora { .. } => {
+						// Esplora does not support a testmempoolaccept equivalent
+						log_error!(
+							self.logger,
+							"Broadcast suitability check not supported by Esplora backend."
```
bc1cindy: I think it's important to be explicit about probing attack, like: "Broadcast suitability check not supported by Esplora backend. Probing attack risk not mitigated."

src/chain/mod.rs
```diff
@@ -468,6 +468,46 @@ impl ChainSource {
...
+						);
+						Err(Error::UnsupportedChainSource)
+					},
+					ChainSourceKind::Electrum { .. } => {
+						// Electrum does not support a testmempoolaccept equivalent.
+						log_error!(
+							self.logger,
+							"Broadcast suitability check not supported by Electrum backend."
```
bc1cindy: I think it's important to be explicit about probing attack, like: "Broadcast suitability check not supported by Electrum backend. Probing attack risk not mitigated."

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,872 @@
...
+		log_info!(self.logger, "Request Payjoin by sharing this Payjoin Uri: {}", pj_uri);
+
+		Ok(pj_uri.to_string())
+	}
+
+	fn relay_order<'a>(payjoin_config: &'a PayjoinConfig) -> Vec<&'a str> {
+		let count = payjoin_config.ohttp_relays.len();
+		let start = if count > 0 {
+			let mut bytes = [0u8; 8];
+			getrandom::fill(&mut bytes).unwrap_or(());
```
bc1cindy: If `getrandom::fill` fails silently  via `unwrap_or(())` the selection it's going to be deterministically and not random

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,872 @@
...
+				)]);
+			},
+			SessionOutcome::Failure => {
+				log_error!(self.logger, "Payjoin session failed due to a protocol error.");
+
+				session.status = PayjoinStatus::Failed;
+				self.payjoin_session_store.insert_or_update(session.clone())?;
+
+				let fallback_tx = session.fallback_tx.as_ref().ok_or_else(|| {
+					log_error!(self.logger, "Payjoin session {} missing txid.", session_id);
```
bc1cindy: I believe it will be more clear using  `"Payjoin session {} missing fallback transaction", session_id)` here

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,872 @@
...
+				self.payjoin_session_store.insert_or_update(session)?;
+			},
+			SessionOutcome::Cancel => {
+				log_info!(self.logger, "Payjoin session was cancelled.");
+
+				session.status = PayjoinStatus::Failed;
+				self.payjoin_session_store.insert_or_update(session.clone())?;
+
+				let fallback_tx = session.fallback_tx.as_ref().ok_or_else(|| {
+					log_error!(self.logger, "Payjoin session {} missing txid.", session_id);
```
bc1cindy: I believe it will be more clear using  `"Payjoin session {} missing fallback transaction", session_id)` here

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,872 @@
...
+		&self, proposal: Receiver<UncheckedOriginalPayload>,
+		persister: &KVStorePayjoinReceiverPersister,
+	) -> Result<(), Error> {
+		let session_id = persister.session_id();
+		let mut session =
+			self.payjoin_session_store.get(&session_id).ok_or(Error::InvalidPaymentId)?;
+
+		// Note: broadcast suitability can only be fully verified when using the BitcoindRpc
+		// backend. For Esplora and Electrum backends this check is skipped and the transaction
+		// is assumed to be broadcastable.
```
bc1cindy: it might also be good to explain probing attacks risk

src/payment/payjoin/manager.rs
spacebear21: Makes sense, I see now PaymentDetails already contains all necessary information regarding payment amount and direction that would be ambiguous from the txid alone.

bc1cindy (commented):
this is really looking good! all fixes applied.

here are some minor points:

src/payment/payjoin/persist.rs
```diff
@@ -0,0 +1,151 @@
...
+				.iter()
+				.map(|e| serde_json::from_str(&e.event_json))
+				.collect::<Result<Vec<_>, _>>()
+				.map_err(|_| Error::PersistenceFailed)?;
+
+			Ok(Box::new(events.into_iter()) as Box<dyn Iterator<Item = Self::SessionEvent> + Send>)
+		}
+	}
+
+	fn close(
```
bc1cindy, line 136: `close()` sets `Completed` status, but its called from `close_failed_session()` (manager.rs:820), with log message "Closed failed {} session", which is semantically inconsistent. Should this be addressed?

src/payment/payjoin/manager.rs
spacebear21: Good question, I had to go back and think through each case which means our API should be clearer about this... Anyway these are my thoughts:
- `create_poll_request` and `create_post_request` are fatal, they only error due to crypto or config errors that won't resolve. In these cases we should trigger fallback immediately.
- `try_preserving_privacy` and `contribute_inputs` depend on the provided candidate inputs, so they are retryable in the sense that they could succeed with a different input set. Since `list_input_pairs()` is returning all spendable utxos, the "right" answer here depends on whether you want the user to maybe wait until new inputs become available, or just assume they won't change within the expiration window and broadcast the fallback immediately. I would guess there aren't many on-chain state changes to a typical ldk-node within a 24h window, so the latter choice seems appropriate.

src/chain/mod.rs
Camillarhi: This method could be called from other contexts down the line where probing attacks aren't relevant, so I'd rather keep the message generic and let the caller log any threat-specific context.

src/chain/mod.rs
Camillarhi: This method could be called from other contexts down the line where probing attacks aren't relevant, so I'd rather keep the message generic and let the caller log any threat-specific context.

src/payment/payjoin/manager.rs
Camillarhi: Thanks for pointing this out, I will also update the comment here, as the broadcast check will no longer be skipped

src/config.rs
```diff
@@ -619,6 +636,16 @@ pub enum AsyncPaymentsRole {
 	Server,
 }
 
+/// Configuration options for PayJoin payments.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct PayjoinConfig {
+	/// The URL of the PayJoin directory to use for discovering PayJoin receivers.
```
zealsham: Nit:  the sender does not discover  receivers using the directory. 

i think just "The url of the payjoin directory" is fine

src/payment/payjoin/manager.rs
Camillarhi: Thanks. I will update the failure handling for these cases accordingly

Camillarhi: > * `try_preserving_privacy` and `contribute_inputs` depend on the provided candidate inputs, so they are retryable in the sense that they could succeed with a different input set. Since `list_input_pairs()` is returning all spendable utxos, the "right" answer here depends on whether you want the user to maybe wait until new inputs become available, or just assume they won't change within the expiration window and broadcast the fallback immediately. I would guess there aren't many on-chain state changes to a typical ldk-node within a 24h window, so the latter choice seems appropriate.

Since these aren't fatal, I'd lean toward letting the session expire naturally rather than broadcasting immediately. The receiver is still covered on expiry.

src/payment/payjoin/manager.rs
Camillarhi: The randomisation here is just for load balancing. It's not security-critical. Even if getrandom fails, we still try all relays in order, so nothing breaks. The worst case is `relay[0]` always goes first. Failing hard on this would be overkill.

src/payment/payjoin/persist.rs
Camillarhi: Yes, it should be addressed. Thanks

src/chain/mod.rs
tnull: > Electrum's lack of support for this kind of check is a [longstanding issue](https://github.com/orgs/payjoin/discussions/900) unfortunately.

Hmm, did anybody ever raise this explicitly with the Electrum folks? I think I briefly brought this up (though mostly for completeness sake) here when reviewing the recent v1.6: https://github.com/spesmilo/electrum-protocol/pull/6#discussion_r2489302775

If I had known the history I would have pushed harder on this, but I assume it might make sense to let them know about this and try to get it into the next release at least?

For this PR, I think we had discussed on the initial call that we'd skip the validation for non-bitcoind chain sources, as they are considered semi-trusted already? Any reason why we can't/shouldn't do that?

Cargo.toml
```diff
@@ -81,6 +81,8 @@ prost = { version = "0.11.6", default-features = false}
 #bitcoin-payment-instructions = { version = "0.6" }
 bitcoin-payment-instructions = { git = "https://github.com/jkczyz/bitcoin-payment-instructions", rev = "0138feb7acefb1e49102a6fb46d7b776bf43265e" }
 
+payjoin = { git = "https://github.com/payjoin/rust-payjoin.git", rev= "dd21835e55116c681fddae29bf3cd363dd177745", package = "payjoin", default-features = false, features = ["v2", "io"] }
```
tnull: Given we're about to finally drop the `reqwest`/`url` dependencies from our dependency tree soon, it's a bit unfortunate that payjoin might take some more time to do so. I guess that's however expected.

I do however wonder about the other dependencies, e.g., are there any further efforts to reduce the dependency tree? (cc @spacebear21 @bc1cindy)

src/error.rs
```diff
@@ -137,6 +141,14 @@ pub enum Error {
...
 	/// The provided lnurl is invalid.
 	InvalidLnurl,
+	/// Payjoin is not configured.
+	PayjoinNotConfigured,
+	/// Payjoin session creation failed.
+	PayjoinSessionCreationFailed,
+	/// Payjoin session failed.
+	PayjoinSessionFailed,
+	/// The operation is not supported by the current chain source.
+	UnsupportedChainSource,
```
tnull: I don't think we want to expose this as a user-facing runtime error ~ever. Anything we add as a feature should work with any of the chain sources. If we really find that some configuration can't work at all, we should catch that early the the `build` step and return a `BuildError`.

src/chain/electrum.rs
```diff
@@ -288,6 +291,21 @@ impl ElectrumChainSource {
...
+	pub(crate) async fn get_transaction(&self, txid: &Txid) -> Result<Option<Transaction>, Error> {
+		let electrum_client: Arc<ElectrumRuntimeClient> =
+			if let Some(client) = self.electrum_runtime_status.read().unwrap().client().as_ref() {
+				Arc::clone(client)
+			} else {
+				debug_assert!(
+					false,
+					"We should have started the chain source before getting transactions"
+				);
+				return Err(Error::TxLookupTimeout);
```
tnull: Why are we returning a `Timeout` error type here?

src/error.rs
```diff
@@ -63,6 +63,10 @@ pub enum Error {
 	TxSyncFailed,
 	/// A transaction sync operation timed out.
 	TxSyncTimeout,
+	/// A transaction lookup operation failed.
+	TxLookupFailed,
```
tnull, line 69: Hmm, 'tx lookup' is very generic and not very informative. It would be much better to return error types that reflect what payjoin operation failed exactly?

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,899 @@
...
+
+		let confirmation_target = ConfirmationTarget::OnchainPayment;
+		let fee_rate =
+			fee_rate.unwrap_or_else(|| self.fee_estimator.estimate_fee_rate(confirmation_target));
+
+		// Create a new persister for this session
+		let persister = KVStorePayjoinReceiverPersister::new(
+			session_id,
+			Some(amount_sats * 1000),
+			self.payjoin_session_store.clone(),
```
tnull: nit: Please use `Arc::clone` here and elsewhere.

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,899 @@
...
+		let session_id = PaymentId(random_bytes);
+
+		let confirmation_target = ConfirmationTarget::OnchainPayment;
+		let fee_rate =
+			fee_rate.unwrap_or_else(|| self.fee_estimator.estimate_fee_rate(confirmation_target));
+
+		// Create a new persister for this session
+		let persister = KVStorePayjoinReceiverPersister::new(
+			session_id,
+			Some(amount_sats * 1000),
```
tnull: Seems odd this is hardcoded?

src/payment/payjoin/payjoin_session.rs
```diff
@@ -0,0 +1,247 @@
...
+	/// The session has completed successfully
+	Completed,
+	/// The session has failed
+	Failed,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct SerializedSessionEvent {
+	/// JSON representation of the event
+	pub event_json: String,
```
tnull: Ugh, let's avoid mixing serialization formats please. Rather than using a JSON string, please create a serialization wrapper (similar to what we did for the BDK objects) and have the event serialized as TLV.

Speaking of, does `rust-payjoin` offer any backwards/forwards compatibility guarantees for the data types we need to persist? (cc @bc1cindy @spacebear21)

src/payment/store.rs
```diff
@@ -456,6 +457,18 @@ pub enum PaymentKind {
 		/// The pre-image used by the payment.
 		preimage: Option<PaymentPreimage>,
 	},
+	/// An on-chain Payjoin payment.
+	///
+	/// Payments of this kind will be considered pending until the respective transaction has
+	/// reached [`ANTI_REORG_DELAY`] confirmations on-chain.
+	///
+	/// [`ANTI_REORG_DELAY`]: lightning::chain::channelmonitor::ANTI_REORG_DELAY
+	Payjoin {
```
tnull: I do wonder if we need a new variant here, or if it would make more sense to list it as `Onchain` with sub-classification `Payjoin`? (cf. open questions on #791)

tnull (commented):
Thanks, and excuse the delay here!

Given this is a huge PR I'm still in the process of going through anything, but here's an initial rounds of comments that stood out on the first pass.

Cargo.toml
DanGould: You don't need reqwest from payjoin for LDK-Node. The `io` feature pulls it in but you can drop that and write an inline HTTP CONNECT client fetch function with a client other than reqwest to replace it.

DanGould: Regarding `url` @BenAllenG spent significant time removing that. I thought it was removed except for the HTTP client in `io`

benalleng: For clarity as it stands in 0.25 we do still depend on url. while the public interface for url has been removed we are working on removing the url dep [entirely](https://github.com/payjoin/rust-payjoin/pull/1377) _with expection to the `io` feature_ with our own internal Url but we need to merge that first.

DanGould: I do think bitreq may still be missing HTTP CONNECT support as well https://github.com/rust-bitcoin/corepc/issues/472 but there's a PR that adds it https://github.com/rust-bitcoin/corepc/pull/533

tnull: > For clarity as it stands in 0.25 we do still depend on url.

Def. no huge blocker if there is a path to dropping it!

> I do think bitreq may still be missing HTTP CONNECT support as well [rust-bitcoin/corepc#472](https://github.com/rust-bitcoin/corepc/issues/472) but there's a PR that adds it [rust-bitcoin/corepc#533](https://github.com/rust-bitcoin/corepc/pull/533)

I think the author in that PR is under the impression that HTTP CONNECT is already implemented and you guys don't need SOCKS support: https://github.com/rust-bitcoin/corepc/pull/533#discussion_r2968209906

Might be easy to resolve, but maybe it would be good clarifying which one it is you need exactly? IIUC, you indeed need HTTP CONNECT support, but bitreq just needs to make sure it adds HTTPS support for it?

DanGould: Ah, I see we just need to
test that the relay can't see and that the connection from client to directory is indeed e2ee with https. That could be done in this PR dropping io or in a follow-up to remove the reqwest dep.

src/chain/electrum.rs
Camillarhi: You're right, this should be `Error::TxLookupFailed` instead, as the chain source wasn't started. I'll fix this.

src/error.rs
Camillarhi: These errors are only used internally in the chain layer and get converted before reaching the user (the payjoin manager maps them to `ImplementationError` in the `check_payment` closure). They don't actually surface as user-facing errors. Would you prefer I keep them out of the public Error enum entirely, or is it fine to leave them as internal errors given how the enum is structured?

src/payment/payjoin/manager.rs
Camillarhi: This is converting sats to millisats for the `amount_msat` parameter.

src/payment/payjoin/payjoin_session.rs
Camillarhi: I'll switch to storing the event as `Vec<u8>` instead of a JSON string.

src/payment/store.rs
Camillarhi: Agreed, I think it makes more sense to add it as a subtype of `Onchain` rather than a separate variant. I'll follow the approach in #791 once it lands.

src/payment/payjoin/payjoin_session.rs
spacebear21: > does rust-payjoin offer any backwards/forwards compatibility guarantees for the data types we need to persist?

Not until we hit 1.0. Since payjoin sessions are ephemeral (24h default by default), you could simply discard those sessions that fail to replay due to unrecognized data types, and optionally broadcast the fallback transaction where applicable.

tnull: Hmm, okay. So we'd def. need some robust logic for that then, otherwise it will be a blocker for merging this as we need to provide these guarantees (at the very least backwards, but hopefully forwards, too, at some poitn) to our users.

spacebear21: @nothingmuch elaborates on this here https://github.com/payjoin/rust-payjoin/issues/1456#issuecomment-4157925453 

TLDR; the append-only log w/ explicit types design was specifically chosen to make that as easy as possible to extend without versioning. If new event types or data get introduced we can have the replay logic handle old sessions without much difficulty in most cases and no impact on implementers.

spacebear21:
@Camillarhi I keep getting reminders from the review bot, are you actually waiting for my input here on anything? I think I already answered the outstanding questions but please let me know if there's anything I can do to keep this moving forward.

tnull:
> @Camillarhi I keep getting reminders from the review bot, are you actually waiting for my input here on anything? I think I already answered the outstanding questions but please let me know if there's anything I can do to keep this moving forward.

Ah, I was also under the impression that we're still waiting on more conceptual review here.

Went ahead and removed the review requests for now. @Camillarhi let us know if you require further input here.

I think we still have a few open questions, around https://github.com/lightningdevkit/ldk-node/pull/746#discussion_r3014386989 and also whether it's fine to skip the checks when used in conjunction with Electrum/Esplora chain backends (cf. https://github.com/lightningdevkit/ldk-node/pull/746#discussion_r2941670442), as we had discussed in the initial call with @nothingmuch.

src/chain/mod.rs
DanGould: Since our implementation here is the non-automated case, the spec can already be implemented completely without this check intended for automated integrations (like payjoin on a BTCPayServer where a receiver could be probed many times through an open server). There's no massive reason *to* check, but if we could it would be nice to have.

I don't think we've spoken to the electrum folks specifically but we have [proposed a fix to electrs](https://github.com/romanz/electrs/pull/1212). It's been brouht up [in Fulcrum](https://github.com/cculianu/Fulcrum/issues/44) as well. Do you know the team? Would be great to get that final suggestion in Electurm and then all (all?) implementations are on the same page about the demand at least.

tnull: @DanGould Excuse the delay here, fell off my radar for a bit. Yeah, I talked to them and they seem open to add it to v1.7, opened an issue now here: https://github.com/spesmilo/electrum-protocol/issues/11

Cargo.toml
DanGould: I left a comment here because I don't think this one is going to be closed out with this issue / merge. Looks like you're also depending on `reqwest` not just through us but also through [`esplora-client`](https://github.com/bitcoindevkit/rust-esplora-client/blob/master/Cargo.toml#L29)

DanGould: URL did just get removed  outside of `reqwest` tho https://github.com/payjoin/rust-payjoin/pull/1377

tnull: Yes, esplora-client still needs to switch too, although they are blocked on requiring a blocking client also. (pun intended)

src/error.rs
Camillarhi: Bringing your attention to this, any preference on this?

src/error.rs
Camillarhi: Payjoin now requires the chain source to be Bitcoin Core at build time.

src/payment/payjoin/payjoin_session.rs
Camillarhi: On replay, if loading the persisted events fails (e.g. due to a mismatched data type), the session is closed as `Failed` and the fallback transaction is broadcast if one was persisted.

Camillarhi:
Apologies for the delayed response. I've addressed the reviews above.

Camillarhi:
Tested this end-to-end on regtest with payjoin-cli as the v2 sender and bitcoind 29.0 in Polar. The session establishes, the sender's PSBT gets picked up from the directory, all checks are run, proposal posted back, tx hits mempool, confirms, payment lands in the store as `PaymentKind::Payjoin`.
Hit three bugs while testing, all fixed:
- `list_input_pairs` was passing `Some(satisfaction_weight)` to `InputPair::new`, which the payjoin crate rejects when the input has a `witness_utxo` since the weight is derivable. Now passes `None`.
- The persister's `close()` was unconditionally setting status to `Failed`, but the v2 crate calls `close()` on success too, so the success-handler never ran. Switched to just stamping `completed_at`, like payjoin-cli does.
- `check_proposal` and `send_payjoin_proposal` were retrieving the session from the store before calling `save_async`, then writing the stale snapshot back to update `fallback_tx`, `fee_paid_msat`, and `txid`, wiping the event `save_async` had just appended. That broke replay on the next resume tick. The session is now fetched after `save_async`.
 
Right now, the wallet's `TxConfirmed` handler also writes an `Onchain` record for the same txid, so the payjoin tx
ends up in the store twice (once Payjoin/Pending, once Onchain/Confirmed). Once #791 lands and Payjoin becomes an Onchain subtype, I'll update the method for updating the payment store for Payjoin payments

src/payment/payjoin/manager.rs
bc1cindy: this is a privacy concern, its exactly the convergence the threat model wants to avoid

https://github.com/payjoin/rust-payjoin/issues/919#issuecomment-4411012631 - shared relay for handover request (worst case scenario) 

> "the main leak with two clients using the same relay is that this changes the threat model for traffic 
> analysis so that even a local passive adversary can link the sender and the receiver by observing > that the post of one client interrupts the polling request of the other." 

so, if `getrandom::fill` fails and `start = 0` consistently, every `ldk-node` client converges on `relay[0]` for the first request

its the worst case scenario the threat model identifies, affected ldk-node clients converging on relay[0] increases the probability of sender/receiver sharing a relay

makes sense?

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,886 @@
...
+		Ok(())
+	}
+
+	async fn handle_closed_session(
+		&self, outcome: SessionOutcome, persister: &KVStorePayjoinReceiverPersister,
+	) -> Result<(), Error> {
+		let session_id = persister.session_id();
+		let mut session =
+			self.payjoin_session_store.get(&session_id).ok_or(Error::InvalidPaymentId)?;
+		println!("Session outcome: {:?}", outcome);
```
bc1cindy: looks like a debug leftover?

src/payment/payjoin/manager.rs
bc1cindy: endorsing this

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,886 @@
...
+		}?;
+		let payjoin_psbt = proposal.psbt().clone();
+		let session = proposal
+			.process_response(&res.as_bytes(), ohttp_ctx)
+			.save_async(persister)
+			.await
+			.map_err(|_| Error::PersistenceFailed)?;
+
+		// at this point we will persist the fee and txid to the session store
+		let payjoin_psbt_clone = payjoin_psbt.clone();
```
bc1cindy: is this `clone()` needed?

src/payment/payjoin/manager.rs
bc1cindy: I would like to highlight this comment

this makes the selection deterministic and non-random

bc1cindy (commented):
https://github.com/lightningdevkit/ldk-node/pull/746/changes#r3214853417

I would like to highlight this comment

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,884 @@
...
+				);
+				let kind = PaymentKind::Payjoin {
+					txid: session.txid.ok_or_else(|| {
+						log_error!(self.logger, "Payjoin session {} missing txid.", session_id);
+						Error::PayjoinSessionFailed
+					})?,
+					status: ConfirmationStatus::Unconfirmed,
+				};
+
+				let payment = PaymentDetails::new(
```
bc1cindy, line 644: `find_payment_by_txid` and the `Pending→Succeeded` promotion are onchain-only 

since the payjoin tx spends a wallet UTXO, the sync records it as its own onchain payment keyed by txid

won't this payjoin entry (keyed by `session_id`) end up stuck `Pending` and duplicated alongside it?

src/payment/payjoin/manager.rs
Camillarhi: Thanks for highlighting this comment, this will no longer silently fail and will now propagate the error matching the pattern already used by the session ID generation.

src/payment/payjoin/manager.rs
Camillarhi: Once https://github.com/lightningdevkit/ldk-node/pull/791 lands and Payjoin becomes an Onchain subtype, I'll update `find_payment_by_txid` so the payment store can be updated for Payjoin payments.  Already flagged this in https://github.com/lightningdevkit/ldk-node/pull/746#issuecomment-4408358938

src/payment/payjoin/manager.rs
```diff
@@ -0,0 +1,887 @@
...
+			})?
+			.commit_inputs()
+			.save_async(persister)
+			.await
+			.map_err(|_| Error::PersistenceFailed)?;
+		self.apply_fee_range(proposal, persister).await
+	}
+
+	fn list_input_pairs(&self) -> Result<Vec<InputPair>, Error> {
+		let unspent = self.wallet.list_unspent_utxos()?;
```
bc1cindy: `list_unspent_utxos()` includes unconfirmed, it iterates BDK's `list_unspent()` and pushes each segwit UTXO with no confirmation filter ([`wallet/mod.rs:1449-1519`](https://github.com/Camillarhi/ldk-node/blob/5ff4dc92b4bc7bfc7bf73435fcc89ed5f3d2297c/src/wallet/mod.rs#L1449-L1519)), unlike the crate's other coin-selection which excludes it ([`exclude_unconfirmed()`](https://github.com/Camillarhi/ldk-node/blob/5ff4dc92b4bc7bfc7bf73435fcc89ed5f3d2297c/src/wallet/mod.rs#L927), [`is_confirmed()`](https://github.com/Camillarhi/ldk-node/blob/5ff4dc92b4bc7bfc7bf73435fcc89ed5f3d2297c/src/wallet/mod.rs#L977)). So a payjoin can contribute a 0-conf input, built on RBF-able funds, and a temporal fingerprint that rules out confirmed-only wallets. 

reuse `list_confirmed_utxos_inner()` instead?

ref: https://github.com/payjoin/rust-payjoin/issues/1597#issuecomment-4649767163

spacebear21 (commented):
I gave @Camillarhi minor feedback out-of-band, overall looks good to me but seems like this is blocked by https://github.com/lightningdevkit/ldk-node/pull/791 for the time being. Can we turn off the noisy PR review reminder bot for this PR?

Camillarhi:
> I gave @Camillarhi minor feedback out-of-band, overall looks good to me but seems like this is blocked by #791 for the time being.


Thanks. I’m keeping an eye on [spesmilo/electrum-protocol#2](https://github.com/spesmilo/electrum-protocol/pull/2) which adds `testmempoolaccept` support, that would allow broadcast suitability checks so a node supporting payjoin payments won’t be limited to just bitcoind as a chain source.

tnull:
As mentioned elsewhere, we'll defer this to the 0.9 milestone. For now removed the review requests to silence the 5-fold notifications every day. Still ofc. intend to get back to this soon though.

src/wallet/mod.rs
```diff
@@ -1423,6 +1423,100 @@ impl Wallet {
...
+		// Return the signed PSBT (not extracted transaction)
+		Ok(psbt)
+	}
+
+	pub fn list_unspent_utxos(&self) -> Result<Vec<Utxo>, Error> {
+		let locked_wallet = self.inner.lock().expect("lock");
+
+		let mut utxos = Vec::new();
+
+		for u in locked_wallet.list_unspent() {
```
bc1cindy: this path contributes `0-conf UTXOs`. `list_unspent()` is filtered only by witness version, never by confirmation depth. 

two downsides: it leaks a temporal fingerprint (an unconfirmed-ancestor spend is observable and partitions the anonymity set by elimination), and a 0-conf parent can be RBF'd out, invalidating the payjoin.

`ldk-node` already excludes unconfirmed everywhere else, `exclude_unconfirmed()`, `list_confirmed_utxos_inner()`

suggestion

```
              for u in locked_wallet.list_unspent() {
                      // Skip 0-conf UTXOs: temporal-fingerprint leak, and the parent can be RBF'd out, invalidating the payjoin.
                      if !u.chain_position.is_confirmed() {
                              continue;
                      }
```

ref: https://github.com/payjoin/rust-payjoin/issues/1597#issuecomment-4649767163

Camillarhi:
@spacebear21 notified me out of band that version 1.0 has been released. This PR has been updated and now builds on the released 1.0.0. The 1.0 API moved a fair bit from the version the PR was originally based on, and I have handled all the changes it introduced. I also tested end-to-end on regtest against a Polar bitcoind, with payjoin-cli as the sender. 


@bc1cindy @spacebear21 @DanGould @zealsham, ready for another look since some changes have been made since the last review due to the version bump.

xstoicunicornx:
I did a cursory review using Opus and Sol and found a couple larger design considerations that probably should be reworked before doing a low level review. There were also some straightforward implementation issues that I think are secondary to the design considerations. For now I think we should focus on resolving the design considerations before proceeding with cleaning up the implementation issues.

With that being said, please let me know if there is anything here is inaccurate.

### Design considerations

#### Payjoin is not feature gated

I guess whether Payjoin should be feature gated is something that would require input from LDK team, however I would imagine that it probably should be.

Even if the LDK team determines that Payjoin should not be feature gated, the current implementation breaks other currently supported features like `uniffi-default`. When you run:

```bash
cargo check --no-default-features --features uniffi-default
```
it fails with errors.

#### Bitcoin core is mandatory dependency (which is not explicitly enforced at build time)

Currently Payjoin requires the Bitcoin core backend during the LDK node construction and throws an error if not found. This basically makes Bitcoin core a silent dependency of Payjoin that is only found at run time instead of build time when the Payjoin configuration exists. This feels like a bad anti-pattern.

The main part of this concern is that nodes that rely on Esplora or Electrum backends cannot use Payjoin. Additionally, this also makes Payjoin unusable by the uniffi bindings (as mentioned in previous section).

#### My suggestion

The best suggestion I can think of is to enable the `bitccoinconcensus` feature within `bitcoin` library and do a best effort check for Esplora and Electrum:
 1. Fetch every previous output.
 2. Check that those outputs appear unspent.
 3. Run Transaction::verify.
 4. Check input and output totals and fee rate.
 5. Check lock-time constraints.
 6. Apply locally implemented standardness limits.

What do you think of this approach? I think the 2 main requirements that implementing Payjoin in LDK node should have (other than not breaking anything) are:
1. Can run with all backends
2. Can be exported in uniffi bindings

As for whether Payjoin should be feature gated I think I lean towards not feature gating it, or at least making it a default feature, however I would defer to LDK maintainers on that.

### Implementation details

#### Anchor reserve not protected from being used in Payjoin tx

Anchor reserve UTXO can potentially be selected as an input to the Payjoin tx. `contribute_inputs` reads `get_spendable_amount_sats(cur_anchor_reserve_sats)` and bails only if it's zero, then hands `try_preserving_privacy` the output of `list_unspent_confirmed_utxos`, which is every confirmed UTXO with no exclusions. Nothing stops the selected input from being the UTXO backing the anchor reserve.

#### Lack of UTXO locking

Contributed UTXOs are never locked or reserved. A concurrent `send_to_address`, a channel open, or a second payjoin session can select the same UTXO. Selection and reservation should be one wallet operation. The selected outpoint should be persisted in the session, excluded from other spends, and released when the session becomes terminal.

#### Mid-session errors strand the fallback for ~24h

Non-transient `Err` out of `process_receiver_session` is caught in the `JoinSet` drain and only logged. Session stays Active, replays and fails every 15s, fallback never broadcasts, sender waits out directory expiry. A temporary UTXO shortage becomes a day-long hang.

#### Relay requests have no timeout and allow very large responses

post_request() uses a bare bitreq request:

```rust
bitreq::post(req.url)
    .with_header(...)
    .with_body(...)
    .send_async()
```

See src/payment/payjoin/manager.rs:282-291.

bitreq has no timeout by default and permits bodies up to 1 GiB by default. A stalled or hostile relay can hold a session task forever. Because resume_payjoin_sessions() waits for the whole JoinSet, one stuck session also prevents later 15-second resume rounds.

These requests need a timeout, a small protocol-specific body limit, and explicit status validation.

#### The scheduler does unbounded work every 15 seconds

Every cycle:

- Scans all active sessions.
- Replays every complete event log.
- Spawns one task per session.
- Polls monitoring sessions every two seconds.
- Potentially issues concurrent Bitcoin Core RPC requests.

There is no session limit, concurrency limit, per-session backoff, or next-attempt timestamp. Applications can create unused URIs for 24 hours, making this an easy local resource exhaustion path.

A per-session worker or persisted next-action schedule with bounded concurrency would fit better.

#### The public API hides the session lifecycle

receive() returns only a URI:

```rust
pub async fn receive(...) -> Result<String, Error>
```

There is no public session ID, status query, cancellation method, or user-facing event. Background failures are only logged. The application cannot reliably correlate a URI with the resulting payment or distinguish an active session from a failed one.

Returning a session handle containing the URI and ID would leave room for status() and cancel().

#### Disabling Payjoin strands existing sessions

The manager is constructed only when config.payjoin_config is present. If a node issued Payjoin URIs and restarts without that config, its active persisted sessions are ignored. They will not be polled, expired, or closed with fallback.

Existing active sessions should at least cause a warning or build error when Payjoin is disabled.

Camillarhi:
Thanks for the review @xstoicunicornx. I've gone through each point below. I agree with most of it and will address those, but there are a few I'd push back on.

> Payjoin is not feature gated

Agreed, I'll add the gate.

> Bitcoin Core is a mandatory dependency

It isn't a compile-time constraint, but it is enforced at node construction. `Builder::build()` returns `BuildError::PayjoinConfigMismatch` when a payjoin config is present alongside a non-Bitcoind chain source, so an unsupported combination can't produce a running node.

On requiring Bitcoin Core at all, that was settled earlier in this review. @tnull opened an [issue](https://github.com/spesmilo/electrum-protocol/issues/11) on the `electrum-protocol` repo off the back of that discussion, and was merged into the 1.7 spec [PR](https://github.com/spesmilo/electrum-protocol/pull/2) that adds `testmempoolaccept`. Once it's exposed by a released `electrum-client`, I'll wire it up. It's one method on `ElectrumChainSource` plus loosening the check above.

> On the `bitcoinconsensus` suggestion

I don't think a best-effort check gets us what the real one does. `Transaction::verify` validates scripts, not mempool acceptance, so a transaction can pass every item on that list and still be rejected for low feerate, non-standard scripts, RBF rules or dust.

> Anchor reserve not protected

It's the same shape as `send_to_address`, which builds with BDK's normal selection and then validates the reserve-adjusted balance afterwards rather than excluding UTXOs up front. Is there a specific reason it needs to be handled differently for payjoin?

> Lack of UTXO locking

There's a PR in progress for this: [#1037](https://github.com/lightningdevkit/ldk-node/pull/1037). If it lands first I'll rebase onto it.

> Mid-session errors strand the fallback for ~24h

The fallback isn't stranded. At expiry, the replay returns `Expired`, which routes through `cancel_receiver_session` and broadcasts it.

I also don't think closing the session and broadcasting on a non-transient error is the right fix. The errors that escape here mostly aren't permanent. For example, `InsufficientFunds` just means the wallet isn't funded yet. Broadcasting the fallback on any of those abandons a payjoin that would have succeeded on the next tick, and the sender ends up paying with their original transaction for nothing.

The real problem is that the retry is hot and unbounded, every 15 seconds for up to 24 hours with no backoff. I'll add per-session backoff instead.

> Relay requests have no timeout and allow very large responses

I'll set a timeout, a protocol-sized body limit, and status validation.

> The scheduler does unbounded work every 15 seconds

Agreed. I'll add bounded concurrency alongside the per-session backoff above.

> The public API hides the session lifecycle

On cancellation, that's a fair point, and I'll work on adding it here. I will update the `receive()` API to return the session ID alongside the URI, and add a `cancel` method taking that ID. 

> Disabling Payjoin strands existing sessions

I will look into a good approach for this.

PR #747, opened 2026-01-08
Parallelize init further

TheBlueMatt:
Based on #736, this uses the freshly-merged https://github.com/lightningdevkit/rust-lightning/pull/4147 to parallelize `ChannelMonitor` loading, and while we're here because I couldn't avoid taking the fun part (sorry @tnull) I went ahead and `tokio::join`ed some additional reading during init.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -33,7 +33,7 @@ use lightning::routing::scoring::{
 };
 use lightning::sign::{EntropySource, NodeSigner};
 use lightning::util::persist::{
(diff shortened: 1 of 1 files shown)


src/builder.rs
```diff
@@ -51,7 +51,7 @@ use crate::connection::ConnectionManager;
 use crate::entropy::NodeEntropy;
 use crate::event::EventQueue;
 use crate::fee_estimator::OnchainFeeEstimator;
-use crate::gossip::GossipSource;
+use crate::gossip::{GossipSource, RuntimeSpawner};
```
tnull: If we use it more generally now, let's move `RuntimeSpawner` to `runtime.rs` (in a prefactor commit, if possible).

tnull (approved):

PR #748, opened 2026-01-08
Fix backwards compatibility of `NodeMetrics` reads

tnull:
In commit https://github.com/lightningdevkit/ldk-node/commit/dd55d4758d8ea0b71312de574fd42fa25fc5f5d0 we started ignoring the legacy `latest_channel_monitor_archival_height` field of `NodeMetrics`. However, we erroneously started reading it as `Option<u32>`, though, given it's an optional field, it should have been read as a plain `u32` that might or might not be present. Here we fix this error.

We also add a test case asserting serialization backwards compatibility for non-VSS backends.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -20,7 +20,7 @@ use bitcoin::key::Secp256k1;
 use bitcoin::secp256k1::PublicKey;
 use bitcoin::{BlockHash, Network};
 use bitcoin_payment_instructions::onion_message_resolver::LDKOnionMessageDNSSECHrnResolver;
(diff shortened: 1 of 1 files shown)


TheBlueMatt (dismissed):
Oops thanks

TheBlueMatt (approved):

tnull:
> CI is still sad though.

Yeah, that seems really unrelated this time:

```
error: failed to build archive at `/home/runner/work/ldk-node/ldk-node/target/debug/deps/libldk_node.rlib`: No space left on device (os error 28)
```

PR #750, opened 2026-01-11
Fix NodeEntropy type mismatch in UniFFI builds

chuksys:
### Context 
In https://github.com/lightningdevkit/ldk-node/commit/06a8108aff12196210728b0078453641a9aa85aa, `persistence_backwards_compatibility` was introduced. Currently, the `from_seed_bytes` constructor and the subsequent build method have diverging signatures depending on whether the `uniffi` feature is enabled.

### Changes 
- A helper function is added to unify entropy creation.
-  `into()` is used to conditionally wrap `node_entropy` with an Arc or not.

### Why this is necessary 
UniFFI requires Arc<T> wrappers for cross-language memory management. Without this change, the integration tests fail to compile when the `uniffi` feature is active because the builder receives a raw struct instead of the expected shared pointer.

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -1288,6 +1288,18 @@ pub(crate) async fn do_channel_full_cycle<E: ElectrumApi>(
 	println!("\nB stopped");
 }
 
+// Helper to unify entropy creation
+pub(crate) fn create_test_entropy(seed: [u8; 64]) -> NodeEntropy {
+	#[cfg(feature = "uniffi")]
+	{
+		NodeEntropy::from_seed_bytes(seed.to_vec()).unwrap()
+	}
+	#[cfg(not(feature = "uniffi"))]
+	{
+		NodeEntropy::from_seed_bytes(seed)
+	}
+}
+
 // A `KVStore` impl for testing purposes that wraps all our `KVStore`s and asserts their synchronicity.
 #[derive(Clone)]
 pub(crate) struct TestSyncStore {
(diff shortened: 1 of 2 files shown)


tests/integration_tests_rust.rs
tnull: Ah, rather than adding extra `cfg` flags above, can we just follow the pattern we did elswhere and add an `into()` here?
```suggestion
	let node_new = builder_new.build(node_entropy.into()).unwrap();
```

chuksys: Added a helper function to unify entropy creation. Also used `into()` as suggested.

tests/common/mod.rs
```diff
@@ -1288,6 +1288,18 @@ pub(crate) async fn do_channel_full_cycle<E: ElectrumApi>(
 	println!("\nB stopped");
 }
 
+// Helper to unify entropy creation
```
tnull, line 1291: Ah, actually, if need this let's rather just avoid the issue altogether by using what we do elsewhere (i.e., in `TestConfig::default`):

```rust
		let mnemonic = generate_entropy_mnemonic(None);
		let node_entropy = NodeEntropy::from_bip39_mnemonic(mnemonic, None);
```

rather than seed bytes in `fn persistence_backwards_compatibility`.

tnull (commented):
Landing this to unbreak CI, might make minor changes in a follow-up.

tnull (approved):

Issue #712, opened 2025-11-20
Update logging to include structured fields


joostjager:
Since https://github.com/lightningdevkit/rust-lightning/pull/4217, LDK log messages no longer contain data that is also present in the structured log fields for peer id, channel id and payment hash. This means that `Logger` implementations, including `ldk-node`'s, now need to print the structured fields for the logs to remain comprehensible.

tnull:
> Hi! I’d like to work on this issue if it’s still open. I plan to update the LDK-Node logger to include structured fields introduced in rust-lightning. Please let me know if that sounds good or if there’s anything I should be aware of.

Ah, AFAIR @joostjager wanted to do that eventually himself. But if not, you should be good to give it a shot.

PR #751, opened 2026-01-12
Add structured logging context fields to LogRecord

joostjager:
Extend LogRecord with peer_id, channel_id, and payment_hash fields from
LDK's Record struct. These structured fields are now available to custom
LogWriter implementations and are automatically appended to log messages
by the built-in FileWriter and LogFacadeWriter.

- Add peer_id, channel_id, payment_hash fields to LogRecord (both uniffi
  and non-uniffi versions)
- Add LogContext struct with Display impl to format fields with truncated
  hex values, avoiding intermediate heap allocations
- Update FileWriter and LogFacadeWriter to append context to messages
- Update UDL bindings with new LogRecord fields
- Add unit tests for LogContext and LogFacadeWriter

Closes https://github.com/lightningdevkit/ldk-node/issues/712

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -7,14 +7,16 @@
 
 //! Logging-related objects.
 
-#[cfg(not(feature = "uniffi"))]
 use core::fmt;
 use std::fs;
 use std::io::Write;
 use std::path::Path;
 use std::sync::Arc;
 
+use bitcoin::secp256k1::PublicKey;
 use chrono::Utc;
+use lightning::ln::types::ChannelId;
+use lightning::types::payment::PaymentHash;
 pub use lightning::util::logger::Level as LogLevel;
 pub(crate) use lightning::util::logger::{Logger as LdkLogger, Record as LdkRecord};
 pub(crate) use lightning::{log_bytes, log_debug, log_error, log_info, log_trace};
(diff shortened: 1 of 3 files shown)


src/logger.rs
```diff
@@ -113,19 +159,22 @@ pub(crate) enum Writer {
...
+
 		match self {
 			Writer::FileWriter { file_path, max_log_level } => {
 				if record.level < *max_log_level {
 					return;
 				}
 
 				let log = format!(
-					"{} {:<5} [{}:{}] {}\n",
+					"{} {:<5} [{}:{}] {}{}\n",
```
tnull, line 201: Just appending it plain looks a bit odd for some messages:

```
Persistence of ChannelMonitorUpdate id 13 completed ch:d6487d p:02d87e
```

Can we maybe wrap the context in parentheses or brackets?

joostjager: Parentheses added. Also made the implementation a bit more efficient by avoiding some heap allocs.

src/logger.rs
```diff
@@ -32,6 +34,72 @@ pub struct LogRecord<'a> {
...
+	/// The payment hash pertaining to the logged record.
+	pub payment_hash: Option<PaymentHash>,
+}
+
+/// Structured context fields for log messages.
+///
+/// Implements `Display` to format context fields (channel_id, peer_id, payment_hash) directly
+/// into a formatter, avoiding intermediate heap allocations when used with `format_args!` or
+/// `write!` macros.
+pub struct LogContext<'a> {
```
tnull, line 55: Why does this need to be `pub`? Seems it's an internal utility only? When you drop it, please also drop the redundant doc comments here and on the `Display` impl.

joostjager: For ldk-node users that have their own logger implementation, but also want to reuse the context fmting?

tnull: > For ldk-node users that have their own logger implementation, but also want to reuse the context fmting?

Hmm, so do you plan to use it for LDK Server? If not, I'd say we can make it private as other users implementation might look differently. If we want to keep it `pub` for reuse in Server, let's still drop the docs on `impl Display` and also the entire `pub fn new()` which is just unnecessary (and to a degree an anti-pattern) if you have only `pub` struct fields anyways.

joostjager: Yes, plan to use in server. Dropped new and moved Display docs.

joostjager: Turns out ldk-server is using log facade, so the context should already be logged.

tnull: > Turns out ldk-server is using log facade, so the context should already be logged.

Yes, though I wondered if there are some log lines there that are related to specific channels that could also make use of the context? Same goes for here in LDK Node? But true, that's somewhat orthogonal I guess.

joostjager: Yes, that would indeed be a continuation of the log conversion that was for the most part done in ldk too.

PR #752, opened 2026-01-13
Drop entropy creation helper, cleanup remaining warnings

tnull:
Just a tiny follow-up to #750.

Previously we introduced a `create_test_entropy` helper method in `common.rs` that was only used in one specific test. We here opt to drop this again and also clean up the remaining warnings when run under `--features uniffi`.


joostjager (approved):
Should uniffi warnings fail CI?

PR #755, opened 2026-01-15
Add support for the simple "sigs-based auth" VSS scheme

TheBlueMatt:
At https://github.com/lightningdevkit/vss-server/pull/79 we added
    a new, trivial, VSS authentication scheme that ensures client
    isolation without much else. This is great for testing, and we
    expect some to do new-account-rate-limiting via other means, so
    might well become a common default.

    Here we add support to it in ldk-node.

### DIFF
--- a/tests/integration_tests_vss.rs
+++ b/tests/integration_tests_vss.rs
@@ -28,7 +28,7 @@ async fn channel_full_cycle_with_vss_store() {
 		.build_with_vss_store(
 			config_a.node_entropy,
 			vss_base_url.clone(),
-			"".to_owned(),
+			"store_id".to_owned(),
 			HashMap::new(),
 		)
 		.unwrap();
@@ -39,7 +39,12 @@ async fn channel_full_cycle_with_vss_store() {
 	let mut builder_b = Builder::from_config(config_b.node_config);
 	builder_b.set_chain_source_esplora(esplora_url.clone(), None);
 	let node_b = builder_b
-		.build_with_vss_store(config_b.node_entropy, vss_base_url, "".to_owned(), HashMap::new())
+		.build_with_vss_store(
+			config_b.node_entropy,
+			vss_base_url,
+			"store".to_owned(),
+			HashMap::new(),
+		)
 		.unwrap();
 	node_b.start().unwrap();
 
@@ -73,7 +78,12 @@ async fn vss_node_restart() {
 		builder.set_storage_dir_path(storage_path.clone());
 		builder.set_chain_source_esplora(esplora_url.clone(), None);
 		let node = builder
-			.build_with_vss_store(node_entropy, vss_base_url.clone(), "".to_owned(), HashMap::new())
+			.build_with_vss_store(
+				node_entropy,
+				vss_base_url.clone(),
+				"store_id".to_owned(),
+				HashMap::new(),
+			)
 			.unwrap();
 
 		node.start().unwrap();
@@ -102,7 +112,7 @@ async fn vss_node_restart() {
 	builder.set_chain_source_esplora(esplora_url, None);
 
 	let node = builder
-		.build_with_vss_store(node_entropy, vss_base_url, "".to_owned(), HashMap::new())
+		.build_with_vss_store(node_entropy, vss_base_url, "store_id".to_owned(), HashMap::new())
 		.unwrap();
 
 	node.start().unwrap();


tests/integration_tests_vss.rs
```diff
@@ -25,10 +25,10 @@ async fn channel_full_cycle_with_vss_store() {
 	builder_a.set_chain_source_esplora(esplora_url.clone(), None);
 	let vss_base_url = std::env::var("TEST_VSS_BASE_URL").unwrap();
 	let node_a = builder_a
-		.build_with_vss_store_and_fixed_headers(
+		.build_with_vss_store(
 			config_a.node_entropy,
 			vss_base_url.clone(),
-			"node_1_store".to_string(),
```
tnull, line 31: Why are we changing the `store_id` here and below?

tests/integration_tests_vss.rs
```diff
@@ -66,9 +66,6 @@ async fn vss_v0_schema_backwards_compatibility() {
 	let esplora_url = format!("http://{}", electrsd.esplora_url.as_ref().unwrap());
 	let vss_base_url = std::env::var("TEST_VSS_BASE_URL").unwrap();
 
-	let rand_suffix: String =
-		(0..7).map(|_| rng().sample(rand::distr::Alphanumeric) as char).collect();
-	let store_id = format!("v0_compat_test_{}", rand_suffix);
```
tnull, line 71: No, please leave this in place, otherwise running this test repeatedly against the same backend won't start from scratch every time.

tests/integration_tests_vss.rs
TheBlueMatt: To demonstrate/test client isolation in VSS by default.

tests/integration_tests_vss.rs
TheBlueMatt: Oops, sorry, missed that they're using a static seed. I randomized the seed.

tests/integration_tests_vss.rs
tnull: Hmm, not sure it's worth the changes, but if you think it somehow preferable, sure..

src/builder.rs
```diff
@@ -581,10 +582,10 @@ impl NodeBuilder {
 	///
 	/// [VSS]: https://github.com/lightningdevkit/vss-server/blob/main/README.md
 	pub fn build_with_vss_store(
-		&self, node_entropy: NodeEntropy, vss_url: String, store_id: String,
-		fixed_headers: HashMap<String, String>,
+		&self, node_entropy: NodeEntropy, vss_url: String, fixed_headers: HashMap<String, String>,
 	) -> Result<Node, BuildError> {
 		let logger = setup_logger(&self.log_writer_config, &self.config)?;
+		let store_id = "ldk-node".to_owned();
```
tnull: I don't think we should do this. Adding pubkey-auth is just another authentication mechanism, which is otherwise unrelated to the VSS API contract. Note that we had users pick `store_id`s freely already and they are running in production. Suddenly fixating the `store_id` breaks the VSS API contract but also disallows them to *switch* to pubkey-auth from whatever they are running right now.

tnull (commented):
Generally looks good, would still be good to see a `vss-client` dependency bump (preferably *before* the next `vss-client` release so we get a chance to amend bugs).

I don't mind the test changes, but leaning NACK on the fixating the `store_id` suddenly, see below.

tests/integration_tests_vss.rs
TheBlueMatt: Entirely up to you. I did it cause coverage is coverage, even if the coverage is in totally the wrong repo. If you don't want coverage in the wrong repo, that's also totally fine :).

src/builder.rs
TheBlueMatt: I'm not quite sure I understand the argument around the API contract - there's an API contract defined in vss-client and between the VSS client and server, but I'm not really sure that applies to new instances of ldk-node using VSS. Do we really think they're going to re-derive the VSS auth key from their entropy in order to instantiate another instance of the VSS client with the same auth to store other stuff there and need to put it in the same `store_id`? If not, making them pass another argument just seems like yet more cognitive burden on devs to figure out what a `store_id` is.

> disallows them to switch to pubkey-auth from whatever they are running right now.

Fwiw they can manually instantiate the vss-client `VssHeaderProvider` instance and call `build_with_vss_store_and_header_provider`, though we could also add a separate method if we anticipate users switching authentication mechanisms but keeping their existing VSS instance. It wasn't clear that we anticipate that kind of migration.

tests/integration_tests_vss.rs
tnull: Yeah, feel free to leave as-is.

src/builder.rs
tnull: > I'm not quite sure I understand the argument around the API contract - there's an API contract defined in vss-client and between the VSS client and server, but I'm not really sure that applies to new instances of ldk-node using VSS.

Well, but we're re-exposing that API in LDK Node. In particular, we recently introduced `VssStoreBuilder` explicitly so that *other* projects (such as Orange) can build a `VssStore` instance outside of/prior to LDK Node initialization to be able to store some extra data. We also considered that a pre-factor to then finally upstream `VssStore` and `VssStoreBuilder` to `lightning-persister` soon, which has been the goal since the beginning, but we never go around to do it so far. That's all to say that I'm not quite seeing the need to fixate the `store_id` at all, and in fact we'll need to revert that (and then re-apply for backwards  compat if we'd go that way) once we upstream the code.

> Fwiw they can manually instantiate the vss-client VssHeaderProvider instance and call build_with_vss_store_and_header_provider, though we could also add a separate method if we anticipate users switching authentication mechanisms but keeping their existing VSS instance. It wasn't clear that we anticipate that kind of migration.

I mean, we should anticipate such a migration, as we have a bunch of users on JWT right now, that might opt to choose the new auth mechanism which seems to be the new default over time? At least I don't think we should make it impossible or have them some custom route just to do that? FWIW, we also have `MigratableKVStore`, why not allow users to build two independent `VssStore`s with different store_id (but both using same auth) and then migrate between them, e.g., for backup purposes.

tnull: Now opened a draft for this https://github.com/lightningdevkit/rust-lightning/pull/4323 (i.e., draft until this lands so we upstream the latest version).

TheBlueMatt: > Well, but we're re-exposing that API in LDK Node. In particular, we recently introduced VssStoreBuilder explicitly so that other projects (such as Orange) can build a VssStore instance outside of/prior to LDK Node initialization to be able to store some extra data. We also considered that a pre-factor to then finally upstream VssStore and VssStoreBuilder to lightning-persister soon, which has been the goal since the beginning, but we never go around to do it so far. That's all to say that I'm not quite seeing the need to fixate the store_id at all, and in fact we'll need to revert that (and then re-apply for backwards compat if we'd go that way) once we upstream the code.

Right, the change here did not remove the store-id parameter from the `VssStoreBuilder`, but only from the `NodeBuilder::build_with_vss_store` method that uses an underlying `VssStoreBuilder`. My assumption here was that we wanted as simple an API as possible under the new sigs-based auth scheme, and for those using `NodeBuilder::build_with_vss_store` we assume that they are just using LDK Node, as they would otherwise use a `VssStoreBuilder` and pass a reference to the VSS Store to `NodeBuilder::build_with_store` so that they can also reuse the VSS Store elsewhere in their code.

I guess a related question here is whether something like orange can/should/will use a separate store-id from ldk node. I assume that it should, given we don't want it storing stuff somewhere LDK Node might want to in the future? In that case it would need multiple VSS Store objects. Sadly I believe the store-ids aren't obfuscated so the service would trivially see which objects are for which submodule which does kinda suck.

> I mean, we should anticipate such a migration, as we have a bunch of users on JWT right now, that might opt to choose the new auth mechanism which seems to be the new default over time? At least I don't think we should make it impossible or have them some custom route just to do that? FWIW, we also have MigratableKVStore, why not allow users to build two independent VssStores with different store_id (but both using same auth) and then migrate between them, e.g., for backup purposes.

Right, my question was only how prevalent it would be. Of course the current API supports it just fine (either with a manual header provider or via the `VssStoreBuilder`), just a question of if we expect it to be universal so much that we should make the API more complicated for new uses.

tankyleo: @tnull As matt mentions, what do you think of the `VssStoreBuilder` -> `VssStore` -> `Node::build_with_store` route that people can take today to build a sig-auth based VSS backend, with a custom store_id ? Seems reasonable enough to ask people to take this route if they want to customize the `store_id` IMHO.

Tangentially: as to the goal of minimizing confusion / cognitive load, I would delete all the existing `Builder::build_with_vss_store*` methods, and make all VSS users go through `VssStoreBuilder`. We currently have two ways of achieving the same thing, which is confusing to me (see for example `Builder::build_with_vss_store_and_lnurl_auth` and `VssStoreBuilder::build_with_lnurl`).

Then with `store_id`, I think it would be more helpful to add one or two documentation sentences explaining to people how they should think about `store_id` at a high-level, rather than the current choice of either 1) fixing the store_id to `ldk-node` under the hood, or 2) letting people customize it, with no guidance on how they should think about selecting the `store_id` string, and no explanation on the purpose of `store_id` in the VSS design.

tnull: > Right, the change here did not remove the store-id parameter from the `VssStoreBuilder`, but only from the `NodeBuilder::build_with_vss_store` method that uses an underlying `VssStoreBuilder`. My assumption here was that we wanted as simple an API as possible under the new sigs-based auth scheme, and for those using `NodeBuilder::build_with_vss_store` we assume that they are just using LDK Node, as they would otherwise use a `VssStoreBuilder` and pass a reference to the VSS Store to `NodeBuilder::build_with_store` so that they can also reuse the VSS Store elsewhere in their code.

No, note that `VssStoreBuilder` is not (and given we're going to upstream it, never will) be exposed in bindings, so we still need a way for bindings users to set their store id.

> I guess a related question here is whether something like orange can/should/will use a separate store-id from ldk node. I assume that it should, given we don't want it storing stuff somewhere LDK Node might want to in the future? In that case it would need multiple VSS Store objects. Sadly I believe the store-ids aren't obfuscated so the service would trivially see which objects are for which submodule which does kinda suck.

Yeah, maybe they should build two separate stores, though not sure what's @benthecarman's thoughts here. Any case, choosing store ids is left to the user, if they obfuscate it the service won't learn anything.

tnull: > @tnull As matt mentions, what do you think of the `VssStoreBuilder` -> `VssStore` -> `Node::build_with_store` route that people can take today to build a sig-auth based VSS backend, with a custom store_id ? Seems reasonable enough to ask people to take this route if they want to customize the `store_id` IMHO.

Unfortunately, I'm still not the biggest fan of throwing in that semantic API change with just adding another auth mechanism. It's really an entirely orthogonal choice, that now unnecessarily complicates the discussion on this PR, IMO. As noted above, bindings users won't have access to the `VssStoreBuilder`, so we need to provide a way to set the (potentially legacy) store id via `Builder`.

> Tangentially: as to the goal of minimizing confusion / cognitive load, I would delete all the existing `Builder::build_with_vss_store*` methods, and make all VSS users go through `VssStoreBuilder`. We currently have two ways of achieving the same thing, which is confusing to me (see for example `Builder::build_with_vss_store_and_lnurl_auth` and `VssStoreBuilder::build_with_lnurl`).

See above, we can't do that as we don't expose `VssStoreBuilder` in bindings, and  likely won't ever given we're going to upstream it to `lightning-persister`, where we don't have the notion of `uniffi` bindings.

TheBlueMatt: Mmm, that's a good point about bindings. I'll re-add the argument but try to add more context in the documentation.

tnull (commented):
Alright, we now shipped `vss-client-ng` v0.5.0 and bumped the dependency in #796. Please rebase!

TheBlueMatt:
Done!

src/builder.rs
```diff
@@ -584,6 +584,37 @@ impl NodeBuilder {
...
+	/// [`VssStoreBuilder`] using the same [`NodeEntropy`] (as storage with different keys is
+	/// obviously segmented to prevent wallets from reading data for unrelated wallets). It can be
+	/// any value.
+	///
+	/// **Caution**: VSS support is in **alpha** and is considered experimental.
+	/// Using VSS (or any remote persistence) may cause LDK to panic if persistence failures are
+	/// unrecoverable, i.e., if they remain unresolved after internal retries are exhausted.
+	///
+	/// [VSS]: https://github.com/lightningdevkit/vss-server/blob/main/README.md
+	pub fn build_with_vss_store(
```
tnull, line 606: You'll also need to make these changes to the `ArcedNodeBuilder` variant to have `uniffi` builds succeed.
Note that for the new method, you'll want to follow our usual pattern to simply calling through to `inner`, add the new method to `bindings/ldk_node.udl`, and ofc also rename the prior method signature in both `ArcedNodeBuilder` and `ldk_node.udl`.

tests/integration_tests_vss.rs
```diff
@@ -60,96 +55,15 @@ async fn channel_full_cycle_with_vss_store() {
 	.await;
 }
 
-#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
-async fn vss_v0_schema_backwards_compatibility() {
```
tnull, line 64: Why was this removed?

src/io/vss_store.rs
```diff
@@ -972,11 +1011,13 @@ mod tests {
 		let vss_base_url = std::env::var("TEST_VSS_BASE_URL").unwrap();
 		let mut rng = rng();
 		let rand_store_id: String = (0..7).map(|_| rng.sample(Alphanumeric) as char).collect();
-		let mut vss_seed = [0u8; 32];
+		let mut vss_seed = [0u8; 64];
```
tnull: No, the `vss_seed` is 32 bytes, this is **not** the node entropy seed. Renaming and mixing the concepts here is confusing. Please revert.

tests/integration_tests_vss.rs
```diff
@@ -159,12 +73,7 @@ async fn vss_node_restart() {
 		builder.set_storage_dir_path(storage_path.clone());
 		builder.set_chain_source_esplora(esplora_url.clone(), None);
 		let node = builder
-			.build_with_vss_store_and_fixed_headers(
-				node_entropy,
-				vss_base_url.clone(),
-				store_id.clone(),
-				HashMap::new(),
-			)
+			.build_with_vss_store(node_entropy, vss_base_url.clone(), "".to_owned(), HashMap::new())
```
tnull: The empty store ID here is invalid, we expect it to be `character varying(120) NOT NULL` (see https://github.com/lightningdevkit/vss-server/blob/main/rust/impls/src/postgres/sql/v0_create_vss_db.sql#L3). Please revert this to use the alphanumeric store ids we were using before.

tests/integration_tests_vss.rs
TheBlueMatt: I updated the commit message but its not really practical to support anymore - we don't have LNURL-auth support in the old version of LDK Node being tested here and VSS wants auth now, so getting the data in the table isn't trivial. We'd have to use a custom VSS server authenticator or disable the auth...given VSS is labeled "alpha" still it didn't seem worth the complexity.

src/io/vss_store.rs
TheBlueMatt: The code had to change to use a `NodeEntropy` rather than the previous version because VSS server no longer accepts fixed empty headers so we have to use the builder. It can't be reverted, but I renamed the field.

tests/integration_tests_vss.rs
TheBlueMatt: Hmm, shouldn't we update vss-server to accept an empty store-id? I imagine the most common use will in fact by an empty store-id (because there's no reason to set a store id) and it doesn't seem worth forcing people to set it.

tests/integration_tests_vss.rs
tnull: > we don't have LNURL-auth support in the old version of LDK Node being tested here

Yes we did? https://docs.rs/ldk-node/0.6.2/ldk_node/struct.Builder.html#method.build_with_vss_store

> and VSS wants auth now

No, test setups like this are exactly the reason why we kept `NoopAuthorizer` around behind the `cfg(noop_authorizer)` flag: https://github.com/lightningdevkit/vss-server/commit/52fd9193478686e0d02bdfbef3261fe5fb46841b

> so getting the data in the table isn't trivial

It should be relatively easy, given one of the two paths mentioned above?

Note that we did the schema upgrade just last release and existing users aren't actually migrated to the new schema. So IMO it would be good to keep the backwards compatibility around to ensure we don't break anything for existing users.

src/io/vss_store.rs
tnull: Right, I mostly meant the renaming, though we will need to revisit this as part of https://github.com/lightningdevkit/rust-lightning/pull/4323 anyways, as `NodeEntropy` is ofc an LDK Node concept, so the builder can't take it as an argument then.

tests/integration_tests_vss.rs
TheBlueMatt: Err, sorry, I meant that we don't have LNURL-auth tests setup because we don't have an LNURL-auth server (sample) anywhere AFAIK. Using `NoopAuthorizer` would require having a different VSS server for some tests than others, which we currently don't do. Would you prefer this test be pulled out into its own integration test file/CI job so that it gets a different VSS?

tnull: > Would you prefer this test be pulled out into its own integration test file/CI job so that it gets a different VSS?

Yeah, maybe that would then make the most sense. I def. think we should keep some variant of the schema backward compat test around to make sure legacy users don't run into issues.

tests/integration_tests_vss.rs
tnull: > Hmm, shouldn't we update vss-server to accept an empty store-id? 

No strong opinion, but would prefer to just revert back to the old behavior in this PR so we can get it landed soon.

TheBlueMatt:
Okay this should pass once `vss-server` allows empty `store_id`s. We could remove it but I figured the test is nice and https://github.com/lightningdevkit/vss-server/pull/95 is basically ready anyway so not much harm in waiting a day.

src/builder.rs
G8XSU: ~~Let's not conflate/ hack together `store_id` for a different purpose altogether.~~

Ah.. ignore the above comment, I thought we were using store_id for some authScheme but I think we are using user_token now.

tnull:
> Okay this should pass once `vss-server` allows empty `store_id`s. We could remove it but I figured the test is nice and [lightningdevkit/vss-server#95](https://github.com/lightningdevkit/vss-server/pull/95) is basically ready anyway so not much harm in waiting a day.

Can we just revert here for now to finally fix CI and get this landed, and then I'm happy to bikeshed further on whether or not we want to allow empty `store_id`s? If we conclude we should, happy to switch back to use empty ones in the tests here then.

PR #764, opened 2026-01-21
Add commented-out `patch` section for LDK Git dependencies

### LINKED ISSUE #758: `[patch.crates-io]` does not patch git dependencies in Cargo.toml
When trying to use a local version of `rust-lightning` for development, the `[patch.crates-io]` section in `Cargo.toml` does not work because the lightning dependencies are specified as git dependencies, not crates.io dependencies.

tnull:
Fixes #758.

For convenience we added ready-to-go `patch` setions to `Cargo.toml`, which however only allowed us to quickly override official LDK releases.

Here we also add the corrsponding counterparts for LDK `git` dependencies.

### DIFF
--- a/.github/workflows/semver.yml
+++ b/.github/workflows/semver.yml
@@ -7,9 +7,5 @@ jobs:
     steps:
       - name: Checkout source code
         uses: actions/checkout@v4
-      - name: Install Rust stable toolchain
-        run: |
-          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --profile=minimal --default-toolchain stable
-          rustup override set stable
       - name: Check SemVer
         uses: obi1kenobi/cargo-semver-checks-action@v2


Cargo.toml
```diff
@@ -168,3 +168,16 @@ harness = false
 
 #vss-client-ng = { path = "../vss-client" }
 #vss-client-ng = { git = "https://github.com/lightningdevkit/vss-client", branch = "main" }
+#
+#[patch."https://github.com/lightningdevkit/rust-lightning"]
```
joostjager, line 172: With multiple patch sections, idk if it is still better than just having a commented-out section in the regular location (no patch)?

tnull: It is. Note that patch is fundamentally different in that it will transitively patch the dependency for *all* sub-dependencies. So if you `patch` in LDK Node, you won't need to have it point to a changed version of `bitcoin-payments-instruction` also pointing to the same LDK dependency, etc.

joostjager: Ah, good to know. I guess it's arguable how useful the `crates-io` patch still is. Because very soon after an LDK release, we'll probably bump to an `ldk` latest commit again.

PR #769, opened 2026-01-23
Randomize chain source selection in tests

tnull:
.. all of our tests should be robust against switching chain sources. We
here opt to pick a random one each time to considerably extend our test
coverage, instead of just running some cases against non-Esplora chain
sources.

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -132,14 +132,14 @@ impl Wallet {
 		let mut locked_wallet = self.inner.lock().unwrap();
 		match locked_wallet.apply_update_events(update) {
 			Ok(events) => {
-				let mut locked_persister = self.persister.lock().unwrap();
-				locked_wallet.persist(&mut locked_persister).map_err(|e| {
-					log_error!(self.logger, "Failed to persist wallet: {}", e);
+				self.update_payment_store(&mut *locked_wallet, events).map_err(|e| {
+					log_error!(self.logger, "Failed to update payment store: {}", e);
 					Error::PersistenceFailed
 				})?;
 
-				self.update_payment_store(&mut *locked_wallet, events).map_err(|e| {
-					log_error!(self.logger, "Failed to update payment store: {}", e);
+				let mut locked_persister = self.persister.lock().unwrap();
+				locked_wallet.persist(&mut locked_persister).map_err(|e| {
+					log_error!(self.logger, "Failed to persist wallet: {}", e);
 					Error::PersistenceFailed
 				})?;
 


src/wallet/mod.rs
```diff
@@ -1215,3 +1246,105 @@ impl ChangeDestinationSource for WalletKeysManager {
 		}
 	}
 }
+
+// FIXME/TODO: This is copied-over from bdk_wallet and only used to generate `WalletEvent`s after
+// applying mempool transactions. We should drop this when BDK offers to generate events for
+// mempool transactions natively.
```
TheBlueMatt, line 1261: Hmm, if BDK doesn't support this natively, do they have logic to correctly handle conflicts and build balance knowledge appropriately? ie if there's an RBF that conflicts do we remove the 0conf tx from our balance?

src/builder.rs
```diff
@@ -544,6 +547,16 @@ impl NodeBuilder {
 		Ok(self)
 	}
 
+	/// Configures the [`Node`] to resync chain data from genesis on first startup, recovering any
+	/// historical wallet funds.
+	///
+	/// This should only be set on first startup when importing an older wallet from a previously
+	/// used [`NodeEntropy`].
+	pub fn set_wallet_recovery_mode(&mut self) -> &mut Self {
```
TheBlueMatt, line 555: Rather than a coarse "recovery mode" can we just enable setting the "wallet birthday"? In some setups that might be available and would avoid a lot of effort.

src/wallet/mod.rs
tnull: Yes, they support all of the canonical transaction handling, it's just when they recently introduced the `WalletEvent`, they didn't update all APIs we require to return them. Arguably, they should just expose the helper we here copied over which would avoid them having to provide `_event` variants for all API endpoints individually. See also the discussion on https://github.com/bitcoindevkit/bdk_wallet/issues/374

src/builder.rs
tnull: Hmm, wallet birthday support is still inflight on the BDK side (see https://github.com/bitcoindevkit/bdk_wallet/pull/368) and I'd like to punt on it until we can actually make use of that and remove our hacky birthday logic. 

That said, if you prefer we could of course rename the current "recovery_mode" API to "disable_wallet_birthday" for now which ~does the same thing?

src/wallet/mod.rs
```diff
@@ -152,16 +152,47 @@ impl Wallet {
...
+		let events =
+			wallet_events(&mut *locked_wallet, chain_tip1, chain_tip2, wallet_txs1, wallet_txs2);
+
 		let mut locked_persister = self.persister.lock().unwrap();
 		locked_wallet.persist(&mut locked_persister).map_err(|e| {
 			log_error!(self.logger, "Failed to persist wallet: {}", e);
 			Error::PersistenceFailed
 		})?;
 
+		self.update_payment_store(&mut *locked_wallet, events).map_err(|e| {
```
TheBlueMatt, line 188: Should this happen before `locked_wallet.persist()`? What happens if we crash between the wallet persist and the payment store update (presumably the second is idempotent so we should do it first to reply on restart if needed?)?

src/wallet/mod.rs
```diff
@@ -152,16 +152,47 @@ impl Wallet {
 	pub(crate) fn apply_mempool_txs(
 		&self, unconfirmed_txs: Vec<(Transaction, u64)>, evicted_txids: Vec<(Txid, u64)>,
 	) -> Result<(), Error> {
+		if unconfirmed_txs.is_empty() {
+			return Ok(());
```
TheBlueMatt, line 159: This doesn't appear to be tested. Also not clear to me why we'd want to skip this if we don't have any unconfirmed txs in the mempool but *do* have some txs that got evicted (iiuc).

tnull: Hmm, good point, we should only skip if there's nothing to update. Added a fixup

src/wallet/mod.rs
tnull: Also a good point. I think the current approach worked fine when we still iterated on the wallet txs, but now with the events we should consistently update first to ensure events are replayed. Now added a fixup and another commit fixing it for `apply_update`.

PR #771, opened 2026-01-27
Add basic `CLAUDE.md` file

tnull:


CLAUDE.md
```diff
@@ -0,0 +1,11 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Development Rules
+
+- Always ensure tests pass before committing.
+- Run `cargo fmt --all` after every code change
```
joostjager, line 8: Does it need a version specified?

CLAUDE.md
```diff
@@ -0,0 +1,11 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
```
benthecarman, line 4: having some basic description of the project architecture can go a long way, otherwise it will initially read a bunch of files trying to find things and can waste a lot of the context window

CLAUDE.md
tnull: Don't think so, default `stable` works here.

CLAUDE.md
tnull: Alright, asked Claude to add an architecture overview ^^

PR #772, opened 2026-01-27
Add storage for forwarded payments 

benthecarman:
Routing nodes and LSPs want to track forwarded payments so they can run accounting on fees earned and track profitability across time. We now store these to make it easier to track and allows for future accounting utils in the future.

This shouldn't effect edge user nodes as they should never be forwarding payments.

Implementation is mostly just copied how we currently handle normal payments and adapted for forwarded payments.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -65,7 +65,13 @@ use crate::io::utils::{
 };
 use crate::io::vss_store::VssStoreBuilder;
 use crate::io::{
-	self, PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE, PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
+	self, CHANNEL_FORWARDING_STATS_PERSISTENCE_PRIMARY_NAMESPACE,
+	CHANNEL_FORWARDING_STATS_PERSISTENCE_SECONDARY_NAMESPACE,
+	CHANNEL_PAIR_FORWARDING_STATS_PERSISTENCE_PRIMARY_NAMESPACE,
+	CHANNEL_PAIR_FORWARDING_STATS_PERSISTENCE_SECONDARY_NAMESPACE,
+	FORWARDED_PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
+	FORWARDED_PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
+	PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE, PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
 	PENDING_PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
 	PENDING_PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
 };
@@ -82,7 +88,8 @@ use crate::probing::{
 use crate::runtime::{Runtime, RuntimeSpawner};
 use crate::tx_broadcaster::TransactionBroadcaster;
 use crate::types::{
-	AsyncPersister, ChainMonitor, ChannelManager, DynStore, DynStoreRef, DynStoreWrapper,
+	AsyncPersister, ChainMonitor, ChannelForwardingStatsStore, ChannelManager,
+	ChannelPairForwardingStatsStore, DynStore, DynStoreRef, DynStoreWrapper, ForwardedPaymentStore,
 	GossipSync, Graph, HRNResolver, KeysManager, MessageRouter, OnionMessenger, PaymentStore,
 	PeerManager, PendingPaymentStore,
 };
@@ -1439,24 +1446,34 @@ fn build_with_store_internal(
 
 	let kv_store_ref = Arc::clone(&kv_store);
 	let logger_ref = Arc::clone(&logger);
-	let (payment_store_res, node_metris_res, pending_payment_store_res) =
-		runtime.block_on(async move {
-			tokio::join!(
-				read_all_objects(
-					&*kv_store_ref,
-					PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
-					PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
-					Arc::clone(&logger_ref),
-				),
-				read_node_metrics(&*kv_store_ref, Arc::clone(&logger_ref)),
-				read_all_objects(
-					&*kv_store_ref,
-					PENDING_PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
-					PENDING_PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
-					Arc::clone(&logger_ref),
-				)
+	let (
+		payment_store_res,
+		channel_forwarding_stats_res,
+		node_metris_res,
+		pending_payment_store_res,
+	) = runtime.block_on(async move {
+		tokio::join!(
+			read_all_objects(
+				&*kv_store_ref,
+				PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
+				PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
+				Arc::clone(&logger_ref),
+			),
+			read_all_objects(
+				&*kv_store_ref,
+				CHANNEL_FORWARDING_STATS_PERSISTENCE_PRIMARY_NAMESPACE,
+				CHANNEL_FORWARDING_STATS_PERSISTENCE_SECONDARY_NAMESPACE,
+				Arc::clone(&logger_ref),
+			),
+			read_node_metrics(&*kv_store_ref, Arc::clone(&logger_ref)),
+			read_all_objects(
+				&*kv_store_ref,
+				PENDING_PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
+				PENDING_PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
+				Arc::clone(&logger_ref),
 			)
-		});
+		)
+	});
 
 	// Initialize the status fields.
 	let node_metrics = match node_metris_res {
@@ -1484,6 +1501,34 @@ fn build_with_store_internal(
 			return Err(BuildError::ReadFailed);
 		},
 	};
+
+	let forwarded_payment_store = Arc::new(ForwardedPaymentStore::new(
+		FORWARDED_PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE.to_string(),
+		FORWARDED_PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE.to_string(),
+		Arc::clone(&kv_store),
+		Arc::clone(&logger),
+	));
+
+	let channel_forwarding_stats_store = match channel_forwarding_stats_res {
+		Ok(stats) => Arc::new(ChannelForwardingStatsStore::new(
+			stats,
+			CHANNEL_FORWARDING_STATS_PERSISTENCE_PRIMARY_NAMESPACE.to_string(),
+			CHANNEL_FORWARDING_STATS_PERSISTENCE_SECONDARY_NAMESPACE.to_string(),
+			Arc::clone(&kv_store),
+			Arc::clone(&logger),
+		)),
+		Err(e) => {
+			log_error!(logger, "Failed to read channel forwarding stats from store: {}", e);
+			return Err(BuildError::ReadFailed);
+		},
+	};
+
+	let channel_pair_forwarding_stats_store = Arc::new(ChannelPairForwardingStatsStore::new(
+		CHANNEL_PAIR_FORWARDING_STATS_PERSISTENCE_PRIMARY_NAMESPACE.to_string(),
+		CHANNEL_PAIR_FORWARDING_STATS_PERSISTENCE_SECONDARY_NAMESPACE.to_string(),
+		Arc::clone(&kv_store),
+		Arc::clone(&logger),
+	));
 
 	let (chain_source, chain_tip_opt) = match chain_data_source_config {
 		Some(ChainDataSourceConfig::Esplora { server_url, headers, sync_config }) => {
@@ -2306,6 +2351,14 @@ fn build_with_store_internal(
 		_leak_checker.0.push(Arc::downgrade(&wallet) as Weak<dyn Any + Send + Sync>);
 	}
 
+	let forwarded_payment_aggregation_retention_secs = match config.forwarded_payment_tracking_mode
+	{
+		crate::config::ForwardedPaymentTrackingMode::Detailed => {
+			Some(crate::payment::forwarding_store::FORWARDED_PAYMENT_AGGREGATION_BUCKET_SIZE_SECS)
+		},
+		crate::config::ForwardedPaymentTrackingMode::Stats => Some(0),
+	};
+
 	Ok(Node {
 		runtime,
 		stop_sender,
@@ -2333,6 +2386,10 @@ fn build_with_store_internal(
 		scorer,
 		peer_store,
 		payment_store,
+		forwarded_payment_store,
+		channel_forwarding_stats_store,
+		channel_pair_forwarding_stats_store,
+		forwarded_payment_aggregation_retention_secs,
 		lnurl_auth,
 		is_running,
 		node_metrics,
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -80,7 +80,9 @@ use crate::io::utils::{
 #[cfg(feature = "storage-vss")]
 use crate::io::vss_store::VssStoreBuilder;
 use crate::io::{
-	self, PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE, PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
+	self, CHANNEL_FORWARDING_STATS_PERSISTENCE_SECONDARY_NAMESPACE,
+	FORWARDED_PAYMENT_PERSISTENCE_PRIMARY_NAMESPACE, PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
+	PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
 	PENDING_PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
 	PENDING_PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
 };
@@ -89,6 +91,7 @@ use crate::lnurl_auth::LnurlAuth;
 use crate::logger::{log_error, LdkLogger, LogLevel, LogWriter, Logger};
 use crate::message_handler::NodeCustomMessageHandler;
 use crate::payment::asynchronous::om_mailbox::OnionMessageMailbox;
+use crate::payment::forwarding_store::ForwardingStore;
 #[cfg(feature = "unified-payments")]
 use crate::payment::HRNResolver;
 use crate::peer_store::PeerStore;
@@ -1524,26 +1527,37 @@ fn build_with_store_internal(
 
 	let kv_store_ref = Arc::clone(&kv_store);
 	let logger_ref = Arc::clone(&logger);
-	let (payment_store_res, node_metris_res, pending_payment_store_res, address_pool_res) = runtime
-		.block_on(async move {
-			tokio::join!(
-				read_n_objects(
-					&*kv_store_ref,
-					PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
-					PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
-					PAYMENT_CACHE_WARMUP_COUNT,
-					Arc::clone(&logger_ref),
-				),
-				read_node_metrics(&*kv_store_ref, Arc::clone(&logger_ref)),
-				read_all_objects(
-					&*kv_store_ref,
-					PENDING_PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
-					PENDING_PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
-					Arc::clone(&logger_ref),
-				),
-				read_address_pool(&*kv_store_ref, &*logger_ref)
-			)
-		});
+	let (
+		payment_store_res,
+		channel_forwarding_stats_res,
+		node_metris_res,
+		pending_payment_store_res,
+		address_pool_res,
+	) = runtime.block_on(async move {
+		tokio::join!(
+			read_n_objects(
+				&*kv_store_ref,
+				PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
+				PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
+				PAYMENT_CACHE_WARMUP_COUNT,
+				Arc::clone(&logger_ref),
+			),
+			read_all_objects(
+				&*kv_store_ref,
+				FORWARDED_PAYMENT_PERSISTENCE_PRIMARY_NAMESPACE,
+				CHANNEL_FORWARDING_STATS_PERSISTENCE_SECONDARY_NAMESPACE,
+				Arc::clone(&logger_ref),
+			),
+			read_node_metrics(&*kv_store_ref, Arc::clone(&logger_ref)),
+			read_all_objects(
+				&*kv_store_ref,
+				PENDING_PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
+				PENDING_PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
+				Arc::clone(&logger_ref),
+			),
+			read_address_pool(&*kv_store_ref, &*logger_ref),
+		)
+	});
 
 	// Initialize the status fields.
 	let node_metrics = match node_metris_res {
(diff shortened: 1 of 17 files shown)


src/io/utils.rs
```diff
@@ -223,21 +226,18 @@ where
 	})
 }
 
-/// Read previously persisted payments information from the store.
-pub(crate) async fn read_payments<L: Deref>(
-	kv_store: &DynStore, logger: L,
-) -> Result<Vec<PaymentDetails>, std::io::Error>
+/// Generic helper to read persisted items from a KV store namespace.
+async fn read_objects_from_store<T, L: Deref>(
```
tnull: This DRYing up/refactor is very welcome, but please split it out in a separate prefactor commit.

src/io/utils.rs
```diff
@@ -223,21 +226,18 @@ where
...
+	kv_store: &DynStore, logger: L, primary_namespace: &str, secondary_namespace: &str,
+	type_name: &str,
```
tnull: Instead of taking `type_name` explicitly, can we use [`std::any::type_name`](https://doc.rust-lang.org/std/any/fn.type_name.html)?

src/payment/store.rs
```diff
@@ -762,3 +765,114 @@ mod tests {
...
+/// Details of a payment that has been forwarded through this node.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ForwardedPaymentDetails {
+	/// A unique identifier for this forwarded payment.
+	pub id: ForwardedPaymentId,
+	/// The channel id of the incoming channel between the previous node and us.
+	pub prev_channel_id: ChannelId,
+	/// The channel id of the outgoing channel between the next node and us.
+	pub next_channel_id: ChannelId,
+	/// The `user_channel_id` of the incoming channel between the previous node and us.
```
tnull: Can you document when this would be `None`? Same for all other `Option`s below.

tnull (commented):
Thanks! I do wonder if we should really enable storing all forwards by default, or rather make this optional. 

Also more generally I wonder if users really expect us to store all forwarded payments forever, or if we should only keep the last X entries in the store? Also, with general-purpose HTLC interception coming up, maybe storing forwards might even be something we entirely want to leave entirely to the user after all?

What do you think?

src/lib.rs
```diff
@@ -1692,6 +1694,34 @@ impl Node {
...
+	/// });
+	/// ```
+	pub fn list_forwarded_payments_with_filter<F: FnMut(&&ForwardedPaymentDetails) -> bool>(
+		&self, f: F,
+	) -> Vec<ForwardedPaymentDetails> {
+		self.forwarded_payment_store.list_filter(f)
+	}
+
+	/// Retrieves all forwarded payments.
+	pub fn list_forwarded_payments(&self) -> Vec<ForwardedPaymentDetails> {
```
tnull: This makes we wonder if we should wait for https://github.com/lightningdevkit/rust-lightning/pull/4347 (and some follow-up PRs for implementations here ) to land so this can directly use pagination from the start?

benthecarman: I figured we weren't doing a release until the next ldk release so that'd be a given but can hold off on it if that's not the case

TheBlueMatt (commented):
IMO storing granular forwarding information long-term is dangerous. Instead, can we store information that allows for easier compaction? eg total forwarding on a per-X basis between channel pairs?

benthecarman:
I started working on a version where you set an interval, say 12 hours, and we store individual events and after the time has lapsed we combine them into a single entry. Does that sound good?

src/lib.rs
tnull: Yeah, given that the `PaginatedKVStore` is already pretty far along, let's wait on in here. Thankfully we can then directly bump the LDK dependency.

tnull:
It seems this should be unblocked by now, but needs a considerable rebase?

benthecarman:
Rebased and updated for pagination.

However probably worth holding off on until https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/issues/4766 is fixed

tnull:
> Rebased and updated for pagination.
> 
> However probably worth holding off on until https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/issues/4766 is fixed

Ugh, okay, that means that the fix needs to happen for 0.3 still then.

benthecarman:
Rebased and now uses the updated `HTLCLocator`. Main open questions for this is that we currently persist the aggregate bucket length and don't let the user change it as there's no real way to reconcile previous ones. This is kinda ugly, we could alternatively not allow this to be configured and just have a hard coded number.

src/payment/store.rs
```diff
@@ -149,7 +149,7 @@ impl Readable for PaymentDetails {
 }
 
 impl StorableObjectId for PaymentId {
-	fn encode_to_hex_str(&self) -> String {
```
tnull, line 185: I'm a bit dubious why we need this refactor? Can you elaborate?

src/ffi/types.rs
```diff
@@ -910,6 +911,22 @@ uniffi::custom_type!(PaymentId, String, {
 	},
 });
 
+uniffi::custom_type!(ForwardedPaymentId, String, {
```
tnull, line 1064: Would be good to avoid the custom types if we can. Note that they are error prone as if conversion fails in a path where `Error` can't be emitted (i.e., we can't bubble up a `Result<X, Error>`), it will have the node panic. While most callsites should be safe, we therefore still aim to reduce the number of custom type implementations over time.

src/config.rs
```diff
@@ -129,6 +129,84 @@ pub(crate) const HRN_RESOLUTION_TIMEOUT_SECS: u64 = 5;
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+pub enum ForwardedPaymentTrackingMode {
+	/// Store individual forwarded payments for the current and previous aggregation buckets.
+	///
+	/// Payments from older buckets are aggregated into channel-pair statistics and removed.
+	Detailed {
+		/// The length of each aggregation bucket, in minutes. Must be non-zero.
+		///
+		/// The first detailed configuration persists this value. Later detailed configurations for
```
tnull: This seems like an error prone API. If users can't change this config after the fact we maybe shouldn't allow them to configure it in the first place *or* provide an API design that communicates that outside of docs (who users tend to read after the fact).

src/event.rs
```diff
@@ -84,18 +91,187 @@ impl_writeable_tlv_based!(HTLCLocator, {
...
+			next_htlc_index: 0,
+			inbound_amount_forwarded_msat: prev_amounts[0],
+			total_fee_earned_msat,
+			skimmed_fee_msat,
+			claim_from_onchain_tx,
+			outbound_amount_forwarded_msat: next_amounts[0],
+		}]);
+	}
+
+	// LDK reports the incoming and outgoing HTLCs, but not which incoming HTLC funded each
```
tnull: It seems like here we're trying to re-establish an API contract that LDK doesn't provide, and it seems error prone. Can we avoid this? If we think this matching is important we might need to change upstream first?

src/lib.rs
```diff
@@ -636,6 +679,23 @@ impl Node {
 			chain_source.continuously_process_broadcast_queue(stop_tx_bcast).await
 		});
 
+		if let Some(forwarded_payment_bucket_size_secs) = self.forwarded_payment_bucket_size_secs {
+			let stop_aggregation = self.stop_sender.subscribe();
+			let forwarded_payment_store = Arc::clone(&self.forwarded_payment_store);
+			let channel_pair_stats_store = Arc::clone(&self.channel_pair_forwarding_stats_store);
+			let logger = Arc::clone(&self.logger);
+			self.runtime.spawn_cancellable_background_task(async move {
+				run_forwarded_payment_aggregation(
```
tnull: I'm not sure aggregating stats needs its own background task? I mean they are relatively cheap, but there's still some overhead. Do we think doing it on-demand would be very costly?

Or maybe, if we want to do it in the background, doing some prefactoring to have some of these tasks merged into a single 'maintenance task' could make sense (as we have a few by now).

tnull (commented):
Needs a rebase, some comments.

src/event.rs
```diff
@@ -1538,13 +1722,185 @@ where
...
+								inbound_stats.total_fee_earned_msat.as_mut(),
+								total_fee_earned_msat,
+							) {
+								*total += fee;
+							}
+							inbound_stats.total_skimmed_fee_msat += skimmed_fee_msat.unwrap_or(0);
+						}
+					}
+
+					for inbound_stats in inbound_stats_by_channel.into_values() {
```

>   1. [P1] Forwarding accounting is not idempotent across event replay.
>      The handler performs additive channel-stat updates and creates detailed records with fresh random IDs before all later persistence—including the user event queue write—has succeeded. If a later write fails, ReplayEvent causes the same forward to be processed again, doubling counts,
>      amounts, and fees. LDK also explicitly permits duplicate PaymentForwarded events when the fee is unknown. See /home/tnull/worktrees/ldk-node/pr-772-latest-20260721/src/event.rs:1769, /home/tnull/worktrees/ldk-node/pr-772-latest-20260721/src/payment/store.rs:1419, and LDK’s event contract
>      (https://github.com/lightningdevkit/rust-lightning/blob/f680fd37d8bea69383e1b18730423f347e6a8341/lightning/src/events/mod.rs#L1489-L1514). A stable idempotency marker or atomic persistence strategy is needed.

True, though unclear how to tackle without an idempotency marker / event ID.

src/event.rs

>   2. [P2] Many-to-many forwards fabricate channel-pair attribution.
>      LDK does not identify which incoming HTLC funded which outgoing HTLC, but the implementation assumes both arrays are FIFO-aligned and assigns the complete fee to the final inferred pair. Reordering either set changes the resulting channel-pair profitability despite representing the same
>      forward. This also contradicts the public documentation, which says such detailed records are omitted. See /home/tnull/worktrees/ldk-node/pr-772-latest-20260721/src/event.rs:144 and /home/tnull/worktrees/ldk-node/pr-772-latest-20260721/src/payment/store.rs:1123.

src/payment/store.rs
benthecarman: `ChannelPairStatsId` is 80 bytes, so its hex would be 160 characters, which is larger than the KVStore key limit of 120. For `ChannelPairStatsId` I used base64 instead, which is 107 characters so it can fit. Couldn't really trim `ChannelPairStatsId` either

src/ffi/types.rs
benthecarman: Removed

src/config.rs
benthecarman: Switched to a fixed retention time

src/event.rs
benthecarman: yeah i dont really know a way around this unless we add the payment hash to the forwarded event

src/event.rs
benthecarman: There isn't a mapping of the sats for when we receive them to when we send them, they are all in a single batch. I did FIFO because it's a simple way to split the amounts while keeping the totals correct. I don't know if this is something we really can upstream in ldk because there really is no "pairs" its just one batch of htlcs coming in and going out.

src/lib.rs
benthecarman: I'd rather not do it on-demand, that always feels flaky and confusing to me, especially when handling potential storage errors.

I think a maintenance task could make sense as a follow up

src/event.rs
benthecarman: fixed docs

src/payment/store.rs
```diff
@@ -1086,3 +1096,1117 @@ mod tests {
...
+
+/// Details of a payment that has been forwarded through this node.
+///
+/// For multi-HTLC forwards, records use FIFO bookkeeping to divide the reported amounts between
+/// incoming and outgoing channel pairs. The incoming and outgoing HTLCs are separate batches, so
+/// FIFO defines an accounting allocation between them. It preserves the event totals, but the
+/// allocation can depend on the order of the HTLCs.
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct ForwardedPaymentDetails {
```
tnull: These types should live either in `forwarding.rs` or a new `forwarding_store.rs` file, not just in the payment store file.

src/payment/store.rs
tnull: Why do `bucket_start_timestamp: u64, bucket_size_secs: u64` need to be part of that ID? Can we just simply use something like `channel_id_1 XOR channel_id_2` for the key?

src/data_store.rs
```diff
@@ -179,13 +196,83 @@ where
 		self.objects.lock().expect("lock").values().filter(f).cloned().collect::<Vec<SO>>()
 	}
 
+	pub(crate) fn is_empty(&self) -> bool {
+		self.objects.lock().expect("lock").is_empty()
+	}
+
+	pub(crate) async fn list_page(
```
tnull, line 479: Seems all the discussion over at #959 around mixing in-memory and on-disk stores applies here as well. So we probably need to wait with this PR until the comments there were addressed and the discussion reached a conclusion.

src/event.rs
tnull: To be honest together with https://github.com/lightningdevkit/ldk-node/pull/772#discussion_r3620695548 this seems like a blocker for this PR. If the upstream API doesn't provide sufficient data for us to do the level of detailed aggregation we want to do, and also doesn't allow us to aggregate safely without overcounting on restarts, we probably either a) start with a version that exposes fewer details or b) first change the upstream API altogether so it provides all the data we need to do this correctly and safely.

tnull (commented):
Needs a rebase. As mentioned below, I'm not sure if we can move forward with this before we changed the upstream APIs. If we need it right away in LDK server, we probably need to start with a simplified version that doesn't lean on the very brittle re-derivation of channel pairs, for which we simply don't have sufficient information available.

Also here are some comments from Claude:

> Findings
> 
>   Bugs / must-fix
> 
>   1. Leftover doc comment, src/io/mod.rs:50 — /// The immutable forwarded-payment bucket size will be persisted under this key. sits stacked on top of the NODE_METRICS doc comment. It's a remnant of the removed bucket-size persistence and should be deleted.
>   2. Event-replay double counting (known, still unresolved) — the handler at src/event.rs:1736–1830 persists inbound stats, then outbound stats, then detail records, then enqueues the user event; any later failure returns ReplayEvent, and the replay re-applies the already-persisted increments
>   (and inserts duplicate detail records under fresh random IDs). The PR discussion acknowledges this. Counters can't be made idempotent without an LDK-side event ID, but the detail records could be: derive the ID deterministically from event contents (e.g., hash of HTLC set + amounts + fees)
>   so replays overwrite rather than duplicate. Worth at least a doc caveat that stats are at-least-once.
> 
>   Design concerns
> 
>   3. Channel-pair buckets grow without bound — detail records are pruned, but the per-(pair, hour) ChannelPairForwardingStats buckets are never compacted or expired, and all of them are loaded into memory at every startup (read_all_objects in builder.rs) and held in the DataStore HashMap
>   forever. An LSP forwarding across many pairs accrues up to ~8,760 buckets per active pair per year. This partially defeats the bounded-storage goal that motivated the aggregation design. Suggest a coarser second-tier compaction (daily/weekly) or a documented retention policy.
>   4. Pagination reads disk while everything is already in memory — DataStore::list_page (src/data_store.rs) does a paginated KVStore list plus one KVStore read per key, yet the same objects all sit in the in-memory map. For VSS that's a network round-trip per record for data already in RAM.
>   Either serve pages from memory, or (better long-term, given point 3) stop retaining these datasets fully in memory — but the current halfway state has the costs of both.
>   5. Startup blocks on aggregation, src/lib.rs:362 — start() runs aggregate_expired_forwarded_payments synchronously via block_on, and each detail removal is an individually awaited KVStore op. A Detailed-mode node that was offline for a while, on a remote store, pays that as startup latency.
>   Since the periodic task exists anyway, consider making it run one pass immediately when spawned instead of only at the next bucket boundary, and dropping the blocking pass from start().
> 
>   Semantics / documentation
> 
>   6. Fee attribution is asymmetric and undocumented — fees are credited to the inbound channel's ChannelForwardingStats only (src/event.rs:1758–1766), while outbound entries hardcode total_fee_earned_msat: Some(0) even when the event's fee is unknown (inbound uses total_fee_earned_msat.map(|_|
>   0) and correctly poisons to None). Defensible as an attribution convention, but the field doc ("Total forwarding fees earned through this channel") doesn't say it, and users summing outbound channels will conclude those channels never earn fees. Similarly, onchain_claims_count goes once to
>   the first outbound channel while the detail/pair records set claim_from_onchain_tx on every pair of the forward — the two stores won't reconcile. Both conventions should be stated on the field docs.
>   7. Clock-skew edge case — timestamps come from SystemTime::now(); a backwards NTP step can insert a detail record into a bucket already committed via the contains_key marker (src/payment/store.rs:1646), and the next pass silently deletes it without counting it. Low severity, but worth a
>   comment. Also inconsistent: event.rs uses .expect(...) on duration_since(UNIX_EPOCH) while store.rs uses unwrap_or(0).
>   8. list_channel_pair_stats_in_range can scan the whole store — keys sort by channel pair, not time, so a sparse time range walks every page (each a blocking KVStore round-trip) before returning. The per-call behavior is documented, but callers on the FFI thread may be surprised; worth a doc
>   note about cost.

src/payment/store.rs
benthecarman: Originally did it like this because so we could reverse look up but seems like don't need that anymore. Need `bucket_start_timestamp` so each bucket doesn't overwrite the last but `bucket_size_secs` is fixed now so removed that. Did a hash instead of XOR so the ordering is preserved ie a -> b vs b -> a

src/event.rs
benthecarman: Changed back to only handle prev/next_hltc len = 1 events again. LDK doesn't forward trampoline payments so we can handle these once we do

src/payment/store.rs
benthecarman: moved

src/data_store.rs
benthecarman: What do you think about making these on-disk only? These shouldn't be read out very often so we don't really need the in-memory store. Did this for now

tnull: > What do you think about making these on-disk only?

What do you mean by 'these'?

benthecarman: the saved forwarded payment data

tnull:
Please rebase now that #1024 landed.

src/io/mod.rs
```diff
@@ -33,6 +33,20 @@ pub(crate) const PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE: &str = "";
 pub(crate) const PENDING_PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE: &str = "pending_payments";
 pub(crate) const PENDING_PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE: &str = "";
 
+/// The forwarded payment information will be persisted under this prefix.
```
tnull: Shouldn't all forwarded-payment-store related objects be stored under the same primary namespace, and be discerned by secondary?

src/payment/forwarding_store.rs
```diff
@@ -0,0 +1,1190 @@
...
+use crate::data_store::{StorableObject, StorableObjectId, StorableObjectUpdate, UpdatableObject};
+use crate::hex_utils;
+use crate::logger::{log_debug, log_error, Logger};
+use crate::types::{DynStore, DynStoreRef};
+use crate::Error;
+
+pub(crate) const FORWARDED_PAYMENT_AGGREGATION_BUCKET_SIZE_SECS: u64 = 60 * 60;
+
+/// A disk-backed store for forwarding data that is too large to keep in memory.
+pub(crate) struct DiskStore<SO: StorableObject> {
```
tnull: I don't understand why we need this? The idea of `DataStore` being generic is exactly to avoid adding specialized stores for every instance?

src/data_store.rs
```diff
@@ -264,7 +273,7 @@ pub(crate) struct DataStorePage<SO> {
 	pub next_page_token: Option<PageToken>,
 }
 
-pub(crate) struct DataStore<SO: StorableObject, L: Deref, P: CachePolicy = KeepAllEntries>
+pub(crate) struct DataStore<SO: UpdatableObject, L: Deref, P: CachePolicy = KeepAllEntries>
```
tnull: I think if we leave the general bound `SO: StorableObject` and move only `update`, `insert_or_update`, and related merge methods in a separate `impl` block bound by `UO: UpdatableObject`, we should be good to just reuse `DataStore` for the forwarded payment store, even without adding dummy implementations for `update`.

src/event.rs
tnull: Hmm, @TheBlueMatt recently indicated the HTLC ID was added to `PaymentForwarded`, but it still doesn't seem to be the case. So AFAIU we have no idempotency token available here. Begs the question if we're fine shipping this with a known bug or how we'd want to deal with this.

src/payment/forwarding.rs
```diff
@@ -0,0 +1,403 @@
...
+	pub next_page_token: Option<PageToken>,
+}
+
+/// A handler allowing to query forwarded payments and forwarding statistics.
+///
+/// Should be retrieved by calling [`Node::forwarding`].
+///
+/// [`Node::forwarding`]: crate::Node::forwarding
+#[cfg_attr(feature = "uniffi", derive(uniffi::Object))]
+pub struct Forwarding {
```
tnull: Ngl, `Forwarding` is a bit of an odd name - "forwarding what?"

Maybe `ForwardingAnalytics` or `ForwardedPayments`?

tnull (commented):
Thanks! 

Some comments, I think the two main things remaining is that we still don't have a way to handle replayed events and that we should probably reuse `DataStore` for this (fwiw, there are a few fixed bugs in `DiskStore` that we can just avoid when we drop the fork/DRY up the code).

src/payment/forwarding.rs
benthecarman: fixed

src/payment/forwarding_store.rs
benthecarman: Removed, goal was to not cache all these and just write to store

benthecarman:
Addressed most review comments, however, can hold off until https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/pulls/4912 is merged

tnull (commented):
> Addressed most review comments, however, can hold off until https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/pulls/4912 is merged

Cool. That should land shortly, though we'll also need a backport PR to land. That said, it seems all fixups now happened in the second commit. Can you split all changes to `DataStore` out to 1-2 prefactor commits, also to avoid switching to `UpdatableObject` and back?

src/builder.rs
```diff
@@ -1509,6 +1525,39 @@ fn build_with_store_internal(
 		},
 	};
 
+	let forwarded_payment_store = Arc::new(ForwardedPaymentStore::new(
```
tnull: Rather than adding three stores to `Node` directly, could we add a wrapper object that keeps the internal API a bit cleaner and is more clearly documenting what the different stores are for? 

Somehting like:

```rust
  pub(crate) struct ForwardingStore {
      details: ForwardedPaymentStore,
      channel_stats: ChannelForwardingStatsStore,
      channel_pair_stats: ChannelPairForwardingStatsStore,
  }
```
Possibly with the API like:

```rust
  impl ForwardingStore {
      async fn record_forward(&self, forward: ForwardRecord) -> Result<(), Error>;
      async fn payment(&self, id: &str) -> Result<Option<ForwardedPaymentDetails>, Error>;
      async fn channel_stats(&self, id: &ChannelId) -> Result<Option<ChannelForwardingStats>, Error>;
      async fn aggregate_expired(&self, retention_secs: u64) -> Result<(u64, u64), Error>;
  }
```

src/event.rs
```diff
@@ -1751,6 +1761,105 @@ where
 						.await;
 				}
 
+				if let ([prev_htlc], [next_htlc], outbound_amount_msat, Some(fee_earned_msat)) = (
```
tnull: See above, IMO it would be cleaner to keep this forwarding-store specific logic in a separate wrapper object, not just add it here in the event handler.

src/builder.rs
benthecarman: done, much cleaner, thanks

src/event.rs
benthecarman: done, much cleaner, thanks

src/data_store.rs
```diff
@@ -428,6 +451,37 @@ where
 		self.contains(id).await
 	}
 
+	/// Returns whether this store contains no objects.
+	pub(crate) async fn is_empty(&self) -> Result<bool, Error> {
+		let mut page_token = None;
+		loop {
```
tnull: Why do we need this loop here? That seems just like a risk that we eventually could add a bug that loops forever? Whether the store is empty should be a simple O(1) decision based on the first response, no?

src/data_store.rs
```diff
@@ -428,6 +451,37 @@ where
...
+			}
+			let Some(next_page_token) = response.next_page_token else {
+				return Ok(true);
+			};
+			page_token = Some(next_page_token);
+		}
+	}
+
+	/// Returns all stored objects matching `f` by reading each page from the backing store.
+	pub(crate) async fn list_filter_from_store<F: FnMut(&&SO) -> bool>(
```
tnull: No, one of the main design goals (i.e., why we introduced the typestate pattern and dropped `list_payments_with_filter`) of the recent refactor was to disallow any internal callers from making prohibitively expensive listing calls. For a long running node this will be insanely expensive as it always interatively walks the entire uncached store. We can't do this, IMO hence have to drop this and all dependent callers.

tnull:
Needs a rebase now.

tnull (approved):

PR #774, opened 2026-01-29
Make sync timeouts configurable

tnull:
Previously, the timeouts applied to chain syncing weren't configurable
to users. While historically there were good reasons for this (mostly to
avoid leaving the Node in a blocked state for extended periods during
chain syncing), by now we should be good to let the users configure
timeouts ~freely, if they deem it fit. Here we allow for exactly that.

We also bump the previous defaults considerably, in the hopes that users now normally
shouldn't hit them ever.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -30,16 +30,16 @@ const DEFAULT_PROBING_LIQUIDITY_LIMIT_MULTIPLIER: u64 = 3;
 const DEFAULT_ANCHOR_PER_CHANNEL_RESERVE_SATS: u64 = 25_000;
 
 // The default timeout after which we abort a wallet syncing operation.
-const DEFAULT_BDK_WALLET_SYNC_TIMEOUT_SECS: u64 = 20;
+const DEFAULT_BDK_WALLET_SYNC_TIMEOUT_SECS: u64 = 60;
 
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -30,16 +30,16 @@ const DEFAULT_PROBING_LIQUIDITY_LIMIT_MULTIPLIER: u64 = 3;
 const DEFAULT_ANCHOR_PER_CHANNEL_RESERVE_SATS: u64 = 25_000;
 
 // The default timeout after which we abort a wallet syncing operation.
-const DEFAULT_BDK_WALLET_SYNC_TIMEOUT_SECS: u64 = 20;
+const DEFAULT_BDK_WALLET_SYNC_TIMEOUT_SECS: u64 = 60;
```
jkczyz, line 33: If there is an outstanding sync, do we skip a newly requested one?

tnull: Actually no, additional sync calls would subscribe and wait for the the pending sync to finish. The result will then be propagated via `propagate_result_to_subscribers`:

https://github.com/lightningdevkit/ldk-node/blob/5d830001c732dd1dc417fee6063eebee6e91493b/src/chain/mod.rs#L56-L83 

In case of a timeout that would be an `Error::TxSyncTimeout`.

Issue #178, opened 2023-10-30
Connection to tor nodes


FlorianHubl:
Is there a way to connect to lightning nodes which run on tor? I run tor in a iOS app. How can I set the proxy in ldk-node?

tnull:
> Would love to see this, and the ability to query the Esplora nodes over Tor too

Noted. I now opened a separate issue as first we'd need to add a separate networking crate to `rust-lightning`: https://github.com/lightningdevkit/rust-lightning/issues/4299

tnull:
> I think with [lightningdevkit/rust-lightning#4305](https://github.com/lightningdevkit/rust-lightning/pull/4305) this could be unblocked?

Yes, indeed, it should be.

> I have a node based on 0.7.0 that successfully connected over Tor, with this patch:
> 
> [jharveyb@ab23a78](https://github.com/jharveyb/ldk-node/commit/ab23a78597a3da87d84cee8a073511e81791a516)

Awesome, already looks pretty good! Do you want to open a PR for this?

> One awkward bit is that the peer is stored internally with the tor proxy socketaddress, not the original onionv3. Will keep testing.

Ah, alright. Happy to brainstorm/discuss options on the PR.

PR #778, opened 2026-01-30
Add Tor support for outbound connections via SOCKS

jharveyb:
Addresses https://github.com/lightningdevkit/ldk-node/issues/178 . Builds on https://github.com/lightningdevkit/rust-lightning/pull/4305 .

As mentioned there, I see two drawbacks with this patch as is:

- We record the peer remote address as the proxy address, not the actual onionv3 of that peer:

https://github.com/lightningdevkit/rust-lightning/blob/f9ad3450b7d8b722b440f0a5e3d9be8bd7a696ae/lightning-net-tokio/src/lib.rs#L332

Which affects `list_peers()` output:

```json
{
			"pubkey": {
				"pubkey": "03864ef025fde8fb587d989186ce6a4a186895ee44a926bfc370e2c366597a3f8f"
			},
			"socket_addr": {
				"address": "3.33.236.230:9735"
			},
			"is_connected": true
		},
		{
			"pubkey": {
				"pubkey": "03fe47fdfea0f25fad0013498e8d6cec348ae3d673841ec25ee94f87c21af16ed8"
			},
			"socket_addr": {
				"address": "127.0.0.1:9050"
			},
			"is_connected": true
		}
```

But more significantly, AFAICT affects our setup messages with our peer:

https://github.com/lightningdevkit/rust-lightning/blob/f9ad3450b7d8b722b440f0a5e3d9be8bd7a696ae/lightning/src/ln/peer_handler.rs#L1905

The `filter_addresses()` call here means we won't include the proxy address, but we also won't include that peer's onion address, which we may want to do?

I'm testing this actively now and haven't hit any issues yet, though I haven't had much traffic with that peer either.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -45,7 +45,7 @@ use vss_client::headers::VssHeaderProvider;
 use crate::chain::ChainSource;
 use crate::config::{
 	default_user_config, may_announce_channel, AnnounceError, AsyncPaymentsRole,
-	BitcoindRestClientConfig, Config, ElectrumSyncConfig, EsploraSyncConfig,
+	BitcoindRestClientConfig, Config, ElectrumSyncConfig, EsploraSyncConfig, TorConfig,
 	DEFAULT_ESPLORA_SERVER_URL, DEFAULT_LOG_FILENAME, DEFAULT_LOG_LEVEL,
 };
 use crate::connection::ConnectionManager;
@@ -165,6 +165,8 @@ pub enum BuildError {
 	InvalidListeningAddresses,
 	/// The given announcement addresses are invalid, e.g. too many were passed.
 	InvalidAnnouncementAddresses,
+	/// The given tor proxy address is invalid, e.g. an onion address was passed.
+	InvalidTorProxyAddress,
 	/// The provided alias is invalid.
 	InvalidNodeAlias,
 	/// An attempt to setup a runtime has failed.
@@ -206,6 +208,7 @@ impl fmt::Display for BuildError {
 			Self::InvalidAnnouncementAddresses => {
 				write!(f, "Given announcement addresses are invalid.")
 			},
+			Self::InvalidTorProxyAddress => write!(f, "Given Tor proxy address is invalid."),
 			Self::RuntimeSetupFailed => write!(f, "Failed to setup a runtime."),
 			Self::ReadFailed => write!(f, "Failed to read from store."),
 			Self::WriteFailed => write!(f, "Failed to write to store."),
@@ -520,6 +523,23 @@ impl NodeBuilder {
 		}
 
 		self.config.announcement_addresses = Some(announcement_addresses);
+		Ok(self)
+	}
+
+	/// Configures the [`Node`] instance to use a Tor SOCKS proxy for outbound connections to peers with OnionV3 addresses.
+	/// Connections to clearnet addresses are not affected, and are not made over Tor.
+	/// The proxy address must not itself be an onion address.
+	///
+	/// **Note**: If unset, connecting to peer OnionV3 addresses will fail.
+	pub fn set_tor_config(&mut self, tor_config: TorConfig) -> Result<&mut Self, BuildError> {
+		match tor_config.proxy_address {
+			SocketAddress::OnionV2 { .. } | SocketAddress::OnionV3 { .. } => {
+				return Err(BuildError::InvalidTorProxyAddress);
+			},
+			_ => {},
+		}
+
+		self.config.tor_config = Some(tor_config);
 		Ok(self)
 	}
 
@@ -957,6 +977,15 @@ impl ArcedNodeBuilder {
 		self.inner.write().unwrap().set_announcement_addresses(announcement_addresses).map(|_| ())
 	}
 
+	/// Configures the [`Node`] instance to use a Tor SOCKS proxy for outbound connections to peers with OnionV3 addresses.
+	/// Connections to clearnet addresses are not affected, and are not made over Tor.
+	/// The proxy address must not itself be an onion address.
+	///
+	/// **Note**: If unset, connecting to peer OnionV3 addresses will fail.
+	pub fn set_tor_config(&self, tor_config: TorConfig) -> Result<(), BuildError> {
+		self.inner.write().unwrap().set_tor_config(tor_config).map(|_| ())
+	}
+
 	/// Sets the node alias that will be used when broadcasting announcements to the gossip
 	/// network.
 	///
@@ -1143,6 +1172,15 @@ fn build_with_store_internal(
 		if config.node_alias.is_some() {
 			log_error!(logger, "Node alias was set but some required configuration options for node announcement are missing: {}", err);
 			return Err(BuildError::InvalidListeningAddresses);
+		}
+	}
+
+	if let Some(tor_config) = &config.tor_config {
+		match tor_config.proxy_address {
+			SocketAddress::OnionV2 { .. } | SocketAddress::OnionV3 { .. } => {
+				return Err(BuildError::InvalidTorProxyAddress);
+			},
+			_ => {},
 		}
 	}
 
@@ -1779,8 +1817,12 @@ fn build_with_store_internal(
 
 	liquidity_source.as_ref().map(|l| l.set_peer_manager(Arc::downgrade(&peer_manager)));
 
-	let connection_manager =
-		Arc::new(ConnectionManager::new(Arc::clone(&peer_manager), Arc::clone(&logger)));
+	let connection_manager = Arc::new(ConnectionManager::new(
+		Arc::clone(&peer_manager),
+		config.tor_config.clone(),
+		Arc::clone(&keys_manager),
+		Arc::clone(&logger),
+	));
 
 	let output_sweeper = match sweeper_bytes_res {
 		Ok(output_sweeper) => Arc::new(output_sweeper),
--- a/src/config.rs
+++ b/src/config.rs
@@ -131,7 +131,8 @@ pub(crate) const LNURL_AUTH_TIMEOUT_SECS: u64 = 15;
 /// | `probing_liquidity_limit_multiplier`   | 3                  |
 /// | `log_level`                            | Debug              |
 /// | `anchor_channels_config`               | Some(..)           |
-/// | `route_parameters`                   | None               |
+/// | `route_parameters`                     | None               |
+/// | `tor_config`                           | None               |
 ///
 /// See [`AnchorChannelsConfig`] and [`RouteParametersConfig`] for more information regarding their
 /// respective default values.
@@ -196,6 +197,13 @@ pub struct Config {
 	/// **Note:** If unset, default parameters will be used, and you will be able to override the
 	/// parameters on a per-payment basis in the corresponding method calls.
 	pub route_parameters: Option<RouteParametersConfig>,
+	/// Configuration options for enabling peer connections via the Tor network.
+	///
+	/// Setting [`TorConfig`] enables connecting to peers with OnionV3 addresses. No other connections
+	/// are routed via Tor. Please refer to [`TorConfig`] for further information.
+	///
+	/// **Note**: If unset, connecting to peer OnionV3 addresses will fail.
+	pub tor_config: Option<TorConfig>,
 }
 
 impl Default for Config {
@@ -208,6 +216,7 @@ impl Default for Config {
 			trusted_peers_0conf: Vec::new(),
 			probing_liquidity_limit_multiplier: DEFAULT_PROBING_LIQUIDITY_LIMIT_MULTIPLIER,
 			anchor_channels_config: Some(AnchorChannelsConfig::default()),
+			tor_config: None,
 			route_parameters: None,
 			node_alias: None,
 		}
@@ -485,6 +494,16 @@ pub struct BitcoindRestClientConfig {
 	pub rest_host: String,
 	/// Host port.
 	pub rest_port: u16,
+}
+
+/// Configuration for connecting to peers via the Tor Network.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct TorConfig {
+	/// Tor daemon SOCKS proxy address. Only connections to OnionV3 peers will be made
+	/// via this proxy; other connections (IPv4 peers, Electrum server) will not be
+	/// routed over Tor.
+	pub proxy_address: SocketAddress,
 }
 
 /// Options which apply on a per-channel basis and may change at runtime or based on negotiation
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -14,8 +14,9 @@ use std::time::Duration;
 use bitcoin::secp256k1::PublicKey;
 use lightning::ln::msgs::SocketAddress;
 
+use crate::config::TorConfig;
 use crate::logger::{log_error, log_info, LdkLogger};
-use crate::types::PeerManager;
+use crate::types::{KeysManager, PeerManager};
 use crate::Error;
 
 pub(crate) struct ConnectionManager<L: Deref + Clone + Sync + Send>
@@ -25,6 +26,8 @@ where
 	pending_connections:
 		Mutex<HashMap<PublicKey, Vec<tokio::sync::oneshot::Sender<Result<(), Error>>>>>,
 	peer_manager: Arc<PeerManager>,
+	tor_proxy_config: Option<TorConfig>,
+	keys_manager: Arc<KeysManager>,
 	logger: L,
 }
 
@@ -32,9 +35,13 @@ impl<L: Deref + Clone + Sync + Send> ConnectionManager<L>
 where
 	L::Target: LdkLogger,
 {
-	pub(crate) fn new(peer_manager: Arc<PeerManager>, logger: L) -> Self {
+	pub(crate) fn new(
+		peer_manager: Arc<PeerManager>, tor_proxy_config: Option<TorConfig>,
+		keys_manager: Arc<KeysManager>, logger: L,
+	) -> Self {
 		let pending_connections = Mutex::new(HashMap::new());
-		Self { pending_connections, peer_manager, logger }
+
+		Self { pending_connections, peer_manager, tor_proxy_config, keys_manager, logger }
 	}
 
 	pub(crate) async fn connect_peer_if_necessary(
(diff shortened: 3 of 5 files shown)


src/connection.rs
```diff
@@ -64,27 +72,73 @@ where
...
-			.to_socket_addrs()
-			.map_err(|e| {
-				log_error!(self.logger, "Failed to resolve network address {}: {}", addr, e);
-				self.propagate_result_to_subscribers(&node_id, Err(Error::InvalidSocketAddress));
-				Error::InvalidSocketAddress
-			})?
-			.next()
-			.ok_or_else(|| {
-				log_error!(self.logger, "Failed to resolve network address {}", addr);
+		let res = if let SocketAddress::OnionV2(old_onion_addr) = addr {
```
tankyleo: nit: did you consider using a `match` statement here?

src/connection.rs
```diff
@@ -64,27 +72,73 @@ where
...
+			let proxy_addr = self.tor_proxy_addr.ok_or_else(|| {
+				log_error!(
+					self.logger,
+					"Failed to resolve network address {:?}: Tor proxy address is unset.",
+					addr
+				);
 				self.propagate_result_to_subscribers(&node_id, Err(Error::InvalidSocketAddress));
 				Error::InvalidSocketAddress
 			})?;
+			let connection_future = lightning_net_tokio::tor_connect_outbound(
```
tankyleo: Do we want to allow people to proxy the other types over tor as well ? Have to be careful with adding more settings, but I'm thinking of an analog to the `always-use-proxy=true` CLN setting.

Could be done in a follow-up if preferred.

src/builder.rs
```diff
@@ -1683,8 +1702,12 @@ fn build_with_store_internal(
 
 	liquidity_source.as_ref().map(|l| l.set_peer_manager(Arc::downgrade(&peer_manager)));
 
-	let connection_manager =
-		Arc::new(ConnectionManager::new(Arc::clone(&peer_manager), Arc::clone(&logger)));
+	let connection_manager = Arc::new(ConnectionManager::new(
+		Arc::clone(&peer_manager),
+		config.tor_proxy_address.clone(),
+		ephemeral_bytes,
```
tankyleo: This will be used to derive passwords sent in plaintext to the SOCKS5 proxy, so out of caution I would have passed a seed here that is different from the seed passed to the `PeerManager` to derive BOLT 8 per-connection ephemeral keys.

The `KeysManager` entropy source is also a chacha20 so seems to me we can cheaply do another `keys_manager.get_secure_random_bytes();`

src/connection.rs
```diff
@@ -64,27 +72,73 @@ where
...
+				Arc::clone(&self.peer_manager),
+				node_id,
+				addr.clone(),
+				proxy_addr,
+				self.tor_proxy_rng.clone(),
```
tankyleo: nit: Would have used `Arc::clone(&self.tor_proxy_rng)` here, same style as the peer manager above, just to make it very clear we are only creating a new pointer to the same state, and not cloning the state itself.

src/connection.rs
jharveyb: Good point, that would be a useful setting...

I'm open to add that on to this PR, would need another cfg setting I suppose.

tankyleo:
@jharveyb Thanks for pointing out these quirks. From my reading, we are still onside, so I would expect everything to work on on the current main branch.

For the outbound connection case, we could add an `Option<SocketAddr>` parameter to `fn setup_outbound` in `lightning-net-tokio`, and we would use that parameter in `fn tor_connect_outbound` to override the default `remote_addr` with the onion V3 address we are connecting to.

One clarification from the spec:
```
SHOULD set remote_addr to reflect the remote IP address (and port) of an incoming connection, if the node is the receiver and the connection was done via IP.
```
Does this mean that if the connection was done "via Tor" we should *not* set the `remote_addr` field in the `Init` message ?

While we don't have native support for inbound tor connections yet,  `list_peers` will always display some `127.0.0.1:xxxxx` for those connections.

tankyleo:
this is the patch I have in mind @tnull curious your thoughts.

```diff
diff --git a/lightning-net-tokio/src/lib.rs b/lightning-net-tokio/src/lib.rs
index eec0e424e..104fd8595 100644
--- a/lightning-net-tokio/src/lib.rs
+++ b/lightning-net-tokio/src/lib.rs
@@ -378,12 +378,12 @@ where
 /// futures are freed, though, because all processing futures are spawned with tokio::spawn, you do
 /// not need to poll the provided future in order to make progress.
 pub fn setup_outbound<PM: Deref + 'static + Send + Sync + Clone>(
-       peer_manager: PM, their_node_id: PublicKey, stream: StdTcpStream,
+       peer_manager: PM, their_node_id: PublicKey, stream: StdTcpStream, remote_addr_override: Option<SocketAddress>,
 ) -> impl std::future::Future<Output = ()>
 where
        PM::Target: APeerManager<Descriptor = SocketDescriptor>,
 {
-       let remote_addr = get_addr_from_stream(&stream);
+       let remote_addr = remote_addr_override.or(get_addr_from_stream(&stream));
        let (reader, mut write_receiver, read_receiver, us) = Connection::new(stream);
        #[cfg(test)]
        let last_us = Arc::clone(&us);
@@ -469,7 +469,7 @@ where
        if let Ok(Ok(stream)) =
                time::timeout(Duration::from_secs(CONNECT_OUTBOUND_TIMEOUT), connect_fut).await
        {
-               Some(setup_outbound(peer_manager, their_node_id, stream))
+               Some(setup_outbound(peer_manager, their_node_id, stream, None))
        } else {
                None
        }
@@ -488,12 +488,12 @@ where
        PM::Target: APeerManager<Descriptor = SocketDescriptor>,
 {
        let connect_fut = async {
-               tor_connect(addr, tor_proxy_addr, entropy_source).await.map(|s| s.into_std().unwrap())
+               tor_connect(addr.clone(), tor_proxy_addr, entropy_source).await.map(|s| s.into_std().unwrap())
        };
        if let Ok(Ok(stream)) =
                time::timeout(Duration::from_secs(TOR_CONNECT_OUTBOUND_TIMEOUT), connect_fut).await
        {
-               Some(setup_outbound(peer_manager, their_node_id, stream))
+               Some(setup_outbound(peer_manager, their_node_id, stream, Some(addr)))
        } else {
                None
        }
@@ -1015,7 +1015,7 @@ mod tests {
                // 127.0.0.1.
                let (conn_a, conn_b) = make_tcp_connection();

-               let fut_a = super::setup_outbound(Arc::clone(&a_manager), b_pub, conn_a);
+               let fut_a = super::setup_outbound(Arc::clone(&a_manager), b_pub, conn_a, None);
                let fut_b = super::setup_inbound(b_manager, conn_b);

                tokio::time::timeout(Duration::from_secs(10), a_connected.recv()).await.unwrap();
@@ -1085,7 +1085,7 @@ mod tests {
                // Call connection setup inside new tokio tasks.
                let manager_reference = Arc::clone(&a_manager);
                tokio::spawn(async move { super::setup_inbound(manager_reference, conn_a).await });
-               tokio::spawn(async move { super::setup_outbound(a_manager, b_pub, conn_b).await });
+               tokio::spawn(async move { super::setup_outbound(a_manager, b_pub, conn_b, None).await });
        }

        #[tokio::test(flavor = "multi_thread")]
```

src/builder.rs
```diff
@@ -904,6 +914,15 @@ impl ArcedNodeBuilder {
 		self.inner.write().unwrap().set_announcement_addresses(announcement_addresses).map(|_| ())
 	}
 
+	/// Set the address which [`Node`] will use as a Tor proxy to connect to peer OnionV3 addresses.
+	///
+	/// **Note**: If unset, connecting to peer OnionV3 addresses will fail.
+	///
+	/// [`tor_proxy_address`]: Config::tor_proxy_address
+	pub fn set_tor_proxy_address(&mut self, tor_proxy_address: core::net::SocketAddr) {
+		self.config.tor_proxy_address = Some(tor_proxy_address);
```
tnull: Rather than replicating the functionality, please just call through to `inner` as we do elsewhere in this builder. This avoids their behavior getting out-of-sync over time.

src/config.rs
```diff
@@ -149,6 +149,12 @@ pub struct Config {
 	///
 	/// [`listening_addresses`]: Config::listening_addresses
 	pub announcement_addresses: Option<Vec<SocketAddress>>,
+	/// The address which the node will use as a Tor proxy to connect to peer OnionV3 addresses.
+	///
+	/// **Note**: If unset, connecting to peer OnionV3 addresses will fail.
+	///
+	/// [`tor_proxy_address`]: Config::tor_proxy_address
+	pub tor_proxy_address: Option<core::net::SocketAddr>,
```
tnull: Hmm, I think it would be preferable to drop this field and just add a separate `Option<TorConfig>` or similar. The `struct TorConfig` would then also allow to add tor-specific settings, such as the ones mentioned below.

Also note that we'll want to expose all of the new APIs to bindings, so you'll need to make the corresponding changes in `bindings/ldk_node.udl` and use `SocketAddress`, not `SocketAddr`.

src/connection.rs
tnull: > Good point, that would be a useful setting...
> 
> I'm open to add that on to this PR, would need another cfg setting I suppose.

Let's just add this as a flag in above mentioned `TorConfig` struct?

src/lib.rs
```diff
@@ -123,7 +123,7 @@ pub use builder::BuildError;
 pub use builder::NodeBuilder as Builder;
 use chain::ChainSource;
 use config::{
-	default_user_config, may_announce_channel, AsyncPaymentsRole, ChannelConfig, Config,
+	default_user_config, may_announce_channel, AsyncPaymentsRole, ChannelConfig, TorConfig, Config,
```
jharveyb: Building the Kotlin bindings locally failed without this. Can move this 'use' to a new line with a warning suppression if there's no other way to address that behavior.

jharveyb:
> @jharveyb upstream PR merged here [lightningdevkit/rust-lightning#4372](https://github.com/lightningdevkit/rust-lightning/pull/4372) should fix the issues you raised above

Thanks for that! Updated here, added the `TorConfig` struct, and updated bindings. I haven't tested the `proxy_all_outbound` behavior live yet.

src/config.rs
```diff
@@ -478,6 +486,16 @@ pub struct BitcoindRestClientConfig {
 	pub rest_port: u16,
 }
 
+/// Configuration for connecting to peers via the Tor Network.
+#[derive(Debug, Clone)]
+pub struct TorConfig {
+	/// Tor daemon SOCKS proxy address.
+	pub proxy_address: SocketAddress,
+
+	/// If set, all outbound peer connections will be made via the Tor SOCKS proxy.
```
tnull: Hmm, I do wonder if we want to expose this right now, as it could be misleading given that *all other* connections (like Esplora/Eleectrum syncing, RGS updates, etc) would still go over clearnet? Maybe we should make this clear in the `TorConfig` docs more generally, and only expose this flag once we're positive we can switch to tunnel everything over Tor?

src/connection.rs
```diff
@@ -25,16 +27,23 @@ where
...
 }
 
 impl<L: Deref + Clone + Sync + Send> ConnectionManager<L>
 where
 	L::Target: LdkLogger,
 {
-	pub(crate) fn new(peer_manager: Arc<PeerManager>, logger: L) -> Self {
+	pub(crate) fn new(
+		peer_manager: Arc<PeerManager>, tor_proxy_config: Option<TorConfig>,
+		ephemeral_random_data: [u8; 32], logger: L,
```
tnull: Let's just have this take a `KeysManager` reference and rederive `RandomBytes` when needed?

src/lib.rs
```diff
@@ -122,6 +122,8 @@ pub use builder::BuildError;
 #[cfg(not(feature = "uniffi"))]
 pub use builder::NodeBuilder as Builder;
 use chain::ChainSource;
+#[cfg(feature = "uniffi")]
+use config::TorConfig;
```
tnull: Please just move this to the exports in the beginning of `src/ffi/types.rs` IIUC that should suffice to make `uniffi` happy.

src/config.rs
jharveyb: Mm, good point - and incoming traffic from other nodes would also still be over clearnet.

I'll remove the flag + make the comments for TorConfig more explicit + remove the flag. It can also be a follow-up if we really want it.

src/connection.rs
jharveyb: Sure - is there a reason that isn't done for `PeerManager`? out of curiousity

tnull: It does take a `NodeSigner`, but not quite sure why it doesn't also take  an `EntropySource`.

src/builder.rs
```diff
@@ -521,6 +524,23 @@ impl NodeBuilder {
 		Ok(self)
 	}
 
+	/// Configures the [`Node`] instance to use a Tor SOCKS proxy for some (or all) outbound connections.
+	/// The proxy address must not itself be an onion address.
+	///
+	///
+	/// **Note**: If unset, connecting to peer OnionV3 addresses will fail.
+	pub fn set_tor_config(&mut self, tor_config: TorConfig) -> Result<&mut Self, BuildError> {
+		match tor_config.proxy_address {
```
tnull, line 535: Maybe we should move this check to the `build` step so it happens also when set via `Builder::from_config`? Not sure if we even need this setter then?

src/builder.rs
```diff
@@ -521,6 +524,23 @@ impl NodeBuilder {
 		Ok(self)
 	}
 
+	/// Configures the [`Node`] instance to use a Tor SOCKS proxy for some (or all) outbound connections.
+	/// The proxy address must not itself be an onion address.
```
tankyleo: I think it would be worth underscoring again here that setting this proxy **only** enables connecting to Onion V3 addresses; it is not used to connect to clearnet addresses.

src/builder.rs
```diff
@@ -521,6 +524,23 @@ impl NodeBuilder {
...
+	///
+	///
```
tankyleo: nit: extra line

src/builder.rs
```diff
@@ -918,6 +938,14 @@ impl ArcedNodeBuilder {
 		self.inner.write().unwrap().set_announcement_addresses(announcement_addresses).map(|_| ())
 	}
 
+	/// Configures the [`Node`] instance to use a Tor SOCKS proxy for some (or all) outbound connections.
+	/// The proxy address must not itself be an onion address.
```
tankyleo: Likewise here, maybe something along "...to use a Tor SOCKS proxy to connect to Onion V3 addresses"?

src/config.rs
```diff
@@ -192,6 +192,13 @@ pub struct Config {
 	/// **Note:** If unset, default parameters will be used, and you will be able to override the
 	/// parameters on a per-payment basis in the corresponding method calls.
 	pub route_parameters: Option<RouteParametersConfig>,
+	/// Configuration options for enabling peer connections via the Tor network.
+	///
+	/// Setting [`TorConfig`] enables connecting to Tor-only peers. Please refer to [`TorConfig`]
+	/// for further information.
```
tankyleo: "Setting [`TorConfig`] only enables connecting to OnionV3 peers, it does not route IP connections via the tor network" ?

src/builder.rs
jharveyb: I'm fine with also having the check there, though I'll note that there are checks in other setters, like `set_announcement_addresses()` and `set_listening_addresses()`, that aren't mirrored in `build_with_store_internal()` right now.

I'd prefer to keep the setter (since I'm using the others as well), but can remove it if needed.

src/builder.rs
jharveyb: Thanks, forgot to update comments after dropping the logic for clearnet addresses.

src/builder.rs
jharveyb: Using the same comment for the setter in both places; lmk if you think I should revise further!

jharveyb:
Rebased + updated docs.

src/builder.rs
```diff
@@ -1076,6 +1105,15 @@ fn build_with_store_internal(
...
 	}
 
+	if let Some(tor_config) = &config.tor_config {
+		match tor_config.proxy_address {
+			SocketAddress::OnionV2 { .. } | SocketAddress::OnionV3 { .. } => {
+				return Err(BuildError::InvalidTorProxyAddress);
+			},
+			_ => {},
+		}
+	}
```
tankyleo, line 1185: Do we put this here in addition to the validation in `set_tor_config` because of the `NodeBuilder::from_config` path ?

src/config.rs
```diff
@@ -208,6 +215,7 @@ impl Default for Config {
 			trusted_peers_0conf: Vec::new(),
 			probing_liquidity_limit_multiplier: DEFAULT_PROBING_LIQUIDITY_LIMIT_MULTIPLIER,
 			anchor_channels_config: Some(AnchorChannelsConfig::default()),
+			tor_config: None,
```
tankyleo, line 219: In the doc for `pub struct Config` we'll want to add the `tor_config` default

src/builder.rs
jharveyb: Yes exactly.

src/config.rs
jharveyb: ah didn't notice that bit; updated + rebased

src/builder.rs
```diff
@@ -1779,8 +1817,12 @@ fn build_with_store_internal(
 
 	liquidity_source.as_ref().map(|l| l.set_peer_manager(Arc::downgrade(&peer_manager)));
 
-	let connection_manager =
-		Arc::new(ConnectionManager::new(Arc::clone(&peer_manager), Arc::clone(&logger)));
+	let connection_manager = Arc::new(ConnectionManager::new(
+		Arc::clone(&peer_manager),
+		config.tor_config.clone(),
+		Arc::clone(&keys_manager),
```
tankyleo, line 1823: I want to confirm we are ok with using the same `EntropySource` for both passwords sent in plaintext to the SOCKS proxy, and the BOLT 8 per-connection ephemeral keys ? I had raised this concern earlier in this PR, but I double checked the security properties of `EntropySource`, and it now sounds reasonable to me.

tankyleo (approved):
Thank you

src/builder.rs
jharveyb: Looking at this again, we could also create a new set of random bytes from `keys_manager` here, use that to seed an instance of `lightning::sign::RandomBytes`, and then pass that in inside an Arc. IIUC a ChaCha20 stream is random enough for these passwords.

Then this entropy source should have no interaction with the one used for BOLT8 ephemeral keys.

Lmk if you prefer that approach, happy to make the change if so.

tankyleo: Right @tnull you suggested we pass the `KeysManager` directly here, let me know what you think.

tnull: Yeah, I think either is fine.

PR #780, opened 2026-02-02
Update to upstream LDK after 4263 and 4370

TheBlueMatt:
https://github.com/lightningdevkit/rust-lightning/pull/4263 and https://github.com/lightningdevkit/rust-lightning/pull/4370 changed the `lightning` API, which we update to here.

We also switch to using a `Cargo.toml` `[patch...]` to avoid having to update the `bitcoin-payment-instructions` dep after each bump.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -154,17 +154,20 @@ harness = false
 #lightning-liquidity = { git = "https://github.com/lightningdevkit/rust-lightning", branch = "main" }
 #lightning-macros = { git = "https://github.com/lightningdevkit/rust-lightning", branch = "main" }
 
-#lightning = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "21e9a9c0ef80021d0669f2a366f55d08ba8d9b03" }
-#lightning-types = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "21e9a9c0ef80021d0669f2a366f55d08ba8d9b03" }
-#lightning-invoice = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "21e9a9c0ef80021d0669f2a366f55d08ba8d9b03" }
-#lightning-net-tokio = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "21e9a9c0ef80021d0669f2a366f55d08ba8d9b03" }
-#lightning-persister = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "21e9a9c0ef80021d0669f2a366f55d08ba8d9b03" }
-#lightning-background-processor = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "21e9a9c0ef80021d0669f2a366f55d08ba8d9b03" }
-#lightning-rapid-gossip-sync = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "21e9a9c0ef80021d0669f2a366f55d08ba8d9b03" }
-#lightning-block-sync = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "21e9a9c0ef80021d0669f2a366f55d08ba8d9b03" }
-#lightning-transaction-sync = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "21e9a9c0ef80021d0669f2a366f55d08ba8d9b03" }
-#lightning-liquidity = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "21e9a9c0ef80021d0669f2a366f55d08ba8d9b03" }
-#lightning-macros = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "21e9a9c0ef80021d0669f2a366f55d08ba8d9b03" }
+[patch.'https://github.com/lightningdevkit/rust-lightning']
+# Cargo won't let us patch a git dependency with another rev from the same git dependency, so we
+# have to pick another copy of `rust-lightning` and pin the rev.
+lightning = { git = "https://git.bitcoin.ninja/rust-lightning", rev = "0bc5c95484345c6289dffd9a379982a9b66d688f" }
+lightning-types = { git = "https://git.bitcoin.ninja/rust-lightning", rev = "0bc5c95484345c6289dffd9a379982a9b66d688f" }
+lightning-invoice = { git = "https://git.bitcoin.ninja/rust-lightning", rev = "0bc5c95484345c6289dffd9a379982a9b66d688f" }
+lightning-net-tokio = { git = "https://git.bitcoin.ninja/rust-lightning", rev = "0bc5c95484345c6289dffd9a379982a9b66d688f" }
+lightning-persister = { git = "https://git.bitcoin.ninja/rust-lightning", rev = "0bc5c95484345c6289dffd9a379982a9b66d688f" }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -154,17 +154,20 @@ harness = false
...
-#lightning-background-processor = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "21e9a9c0ef80021d0669f2a366f55d08ba8d9b03" }
-#lightning-rapid-gossip-sync = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "21e9a9c0ef80021d0669f2a366f55d08ba8d9b03" }
-#lightning-block-sync = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "21e9a9c0ef80021d0669f2a366f55d08ba8d9b03" }
-#lightning-transaction-sync = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "21e9a9c0ef80021d0669f2a366f55d08ba8d9b03" }
-#lightning-liquidity = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "21e9a9c0ef80021d0669f2a366f55d08ba8d9b03" }
-#lightning-macros = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "21e9a9c0ef80021d0669f2a366f55d08ba8d9b03" }
+[patch.'https://github.com/lightningdevkit/rust-lightning']
+# Cargo won't let us patch a git dependency with another rev from the same git dependency, so we
+# have to pick another copy of `rust-lightning` and pin the rev.
+lightning = { git = "https://git.bitcoin.ninja/rust-lightning", rev = "0bc5c95484345c6289dffd9a379982a9b66d688f" }
```
tnull, line 160: I don't see how this is preferable. We'll still need to keep that third repository in-sync.

joostjager: It doesn't work for me out of the box also

```
    Updating git repository `https://git.bitcoin.ninja/rust-lightning`
warning: spurious network error (3 tries remaining): unexpected http status code: 413; class=Http (34)
warning: spurious network error (2 tries remaining): unexpected http status code: 413; class=Http (34)
warning: spurious network error (1 try remaining): unexpected http status code: 413; class=Http (34)
error: failed to load source for dependency `lightning`

Caused by:
  Unable to update https://git.bitcoin.ninja/rust-lightning?rev=0bc5c95484345c6289dffd9a379982a9b66d688f

Caused by:
  failed to fetch into: /Users/joost/.cargo/git/db/rust-lightning-94d3943701373680

Caused by:
  network failure seems to have happened
  if a proxy or similar is necessary `net.git-fetch-with-cli` may help here
  https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli

Caused by:
  unexpected http status code: 413; class=Http (34)
  ```

TheBlueMatt: > I don't see how this is preferable. We'll still need to keep that third repository in-sync.

FWIW this repository is ~always in sync. Annoying, yes, but this shouldn't be an issue. We could also have a repo that's more automated.

TheBlueMatt: > It doesn't work for me out of the box also

Wat? I've never seen this....I bumped the max but...weird, honestly.

tnull: > > It doesn't work for me out of the box also
> 
> Wat? I've never seen this....I bumped the max but...weird, honestly.

I think at the time Joost tested it https://github.com/lightningdevkit/rust-lightning/pull/4370 wasn't merged. I then went reviewed and landed that PR, bumped `bitcoin-payment-instructions`, opened the other PR, and asked him to review so we can be unblocked, as he in particular was held up today due to LDK's API breaks. IMO it shows once more that we need to go back to the drawing board on depedency mgmt and merge policy. Maybe we need a hard CI check in LDK that requires fixes are up before we merge *any* breaking change.

TheBlueMatt: Hmm, should still have worked. the Commit here was the top commit on that PR, rather than the future merge commit. I think it was just a git oddity hitting an nginx max request body size issue.

As for the general issue, I'm a bit confused why someone was *blocked* on an update to LDK Node syncing to two trivial upstream PRs? Testing against upstream without those PRs should have been pretty easy, and waiting a day for the update doesn't seem like an issue to me.

But, either way, the limitation in this case wasn't a desire for rust-lightning devs to update LDK Node, but rather that it currently requires you to do something on your `bitcoin-payment-instructions` fork, which means you're the only one who actually can update LDK Node.

I wonder if we can't automate that part - just have a git repo that maintains a `bitcoin-payment-instructions` that gets its `Cargo.toml` auto-updated (as long as no API changes have impacted it) and maintains branches for the last few week's worth of commits to rust-lightning so that there's always a commit to point to.

tnull: > As for the general issue, I'm a bit confused why someone was _blocked_ on an update to LDK Node syncing to two trivial upstream PRs? Testing against upstream without those PRs should have been pretty easy, and waiting a day for the update doesn't seem like an issue to me.

I'll let @joostjager speak to that, but: Yes, it's all doable of course, but if you have an LDK PR that's based on current `main` that you want to test, it costs you at least 10-15 minutes every time to figure out what's going on on the LDK Node side, which exact PRs broke something, check which API changes have already been accounted for, then create a new branch with the changes you want to test, rebase that, account for potential rebase conflicts. Then every time you make changes to that forked branch you'll need to re-apply them to your actual PR branch, potentially running into conflicts again, which is all pretty messy an error prone. Again, all doable, but I can see that it's just an annoying procedure to deal with that forces you to pay attention to things that you actually don't want spend time/brain resources on right now and you end up cleaning up after the API breaks that others introduced.

> But, either way, the limitation in this case wasn't a desire for rust-lightning devs to update LDK Node, but rather that it currently requires you to do something on your `bitcoin-payment-instructions` fork, which means you're the only one who actually can update LDK Node.

Yeah, all ears for alternatives.

> I wonder if we can't automate that part - just have a git repo that maintains a `bitcoin-payment-instructions` that gets its `Cargo.toml` auto-updated (as long as no API changes have impacted it) and maintains branches for the last few week's worth of commits to rust-lightning so that there's always a commit to point to.

Right, that's basically automating what I've been doing over at https://github.com/tnull/bitcoin-payment-instructions/tree/2025-12-ldk-node-base. I generally like the idea of just automating that branch an ensuring that there is a bump commit for every LDK commit, so anybody can just pick the right commit. *However*, that will break as soon as we break any of the APIs actually used by `bitcoin-payment-instructions`, and then it might become even more confusing?

joostjager: >I'll let @joostjager speak to that, but: Yes, it's all doable of course [...]

Nothing to add.

I still think many of these pain points could be alleviated by moving to a monorepo. We'd trade the current problems (git rev juggling, forked branches, mirror sync, after the fact breakage cleanup) for a simpler one: whoever changes an API updates ldk-node in the same PR, enforced by CI. We might lose some release cadence flexibility, but I think we can come up with a reasonable plan for that.

TheBlueMatt: > Right, that's basically automating what I've been doing over at https://github.com/tnull/bitcoin-payment-instructions/tree/2025-12-ldk-node-base. I generally like the idea of just automating that branch an ensuring that there is a bump commit for every LDK commit, so anybody can just pick the right commit. However, that will break as soon as we break any of the APIs actually used by bitcoin-payment-instructions, and then it might become even more confusing?

I was thinking something like automating a new branch for each merge. ie if we merge commit 1234567890 it would create a branch called based-on-ldk-1234567890. Then, when we update LDK Node you don't have to do anything the branch can just be updated.

PR #781, opened 2026-02-03
Update to upstream LDK after 4263 and 4370

tnull:
lightningdevkit/rust-lightning#4263 and
lightningdevkit/rust-lightning#4370 changed the `lightning` API, which we update to here.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,17 +39,17 @@ default = []
 #lightning-liquidity = { version = "0.2.0", features = ["std"] }
 #lightning-macros = { version = "0.2.0" }
 
-lightning = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "7fe3268475551b0664d315bfbc860416ca8fc774", features = ["std"] }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -78,13 +78,13 @@ log = { version = "0.4.22", default-features = false, features = ["std"]}
 vss-client = { package = "vss-client-ng", version = "0.4" }
 prost = { version = "0.11.6", default-features = false}
 #bitcoin-payment-instructions = { version = "0.6" }
-bitcoin-payment-instructions = { git = "https://github.com/tnull/bitcoin-payment-instructions", rev = "b9f9991b42e9d71b3ca966818a93b158cf8f6c40" }
+bitcoin-payment-instructions = { git = "https://github.com/tnull/bitcoin-payment-instructions", rev = "7ee3958834aba575e04b91a862f790d798a7d578" }
```
joostjager, line 81: No fan of this. Wish there was a better way.

joostjager (approved):

Cargo.toml
tnull: Yeah, i agree. All ears for a better solution, but IMO `patch` is not it.

PR #782, opened 2026-02-03
Enable deferred writing

joostjager:
Integrate deferred chain monitor writes ([rust-lightning#4345](https://github.com/lightningdevkit/rust-lightning/pull/4345)) into ldk-node.

- Enable deferred writes mode in `ChainMonitor`.
- Switch from the sync `MonitorUpdatingPersister` to `MonitorUpdatingPersisterAsync`.
- Re-claim inbound payments when the preimage is already known.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -23,9 +23,7 @@ use lightning::routing::gossip;
 use lightning::routing::router::DefaultRouter;
 use lightning::routing::scoring::{CombinedScorer, ProbabilisticScoringFeeParameters};
 use lightning::sign::InMemorySigner;
-use lightning::util::persist::{
-	KVStore, KVStoreSync, MonitorUpdatingPersister, MonitorUpdatingPersisterAsync,
-};
+use lightning::util::persist::{KVStore, KVStoreSync, MonitorUpdatingPersisterAsync};
 use lightning::util::ser::{Readable, Writeable, Writer};
 use lightning::util::sweep::OutputSweeper;
 use lightning_block_sync::gossip::GossipVerifier;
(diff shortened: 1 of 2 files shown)


joostjager:
After offline discussion, we've landed on using DeferredChainMonitor for all backends (including VSS). Pushed the simplified version.

Next step is further parallelization of the background processor loop to reduce the impact of deferred writing. If more performance improvements are needed after that, we can again look into multi-key writes to bundle everything in a single round trip.

joostjager:
Opened https://github.com/lightningdevkit/rust-lightning/pull/4424 to take one non-essential channel manager persist out of the critical path.

src/types.rs
```diff
@@ -135,6 +133,39 @@ impl<'a> KVStoreSync for dyn DynStoreTrait + 'a {
 
 pub(crate) type DynStore = dyn DynStoreTrait;
 
+// Newtype wrapper that implements `KVStore` for `Arc<DynStore>`. This is needed because `KVStore`
+// methods return `impl Future`, which is not object-safe. `DynStoreTrait` works around this by
+// returning `Pin<Box<dyn Future>>` instead, and this wrapper bridges the two by delegating
+// `KVStore` methods to the corresponding `DynStoreTrait::*_async` methods.
+#[derive(Clone)]
+pub(crate) struct DynStoreRef(pub(crate) Arc<DynStore>);
```
tnull, line 141: Ugh, yet another boilerplate wrapper around `KVStore`. Note that this might conflict with #692 and/or needs to be DRYed up with the work there or any uniffi-related wrappers we introduce there or in a follow up. Ofc. depends on what ends up landing first. Also cc @enigbe.

tests/common/mod.rs
```diff
@@ -1211,8 +1211,10 @@ pub(crate) async fn do_channel_full_cycle<E: ElectrumApi>(
 	);
 
 	println!("\nB close_channel (force: {})", force_close);
+	// Allow the background processor to flush deferred monitor writes so that
```
tnull: Hmm, do we think we could get a way to poll for this rather than introducing this potentially flaky guess?

src/event.rs
```diff
@@ -673,6 +673,26 @@ where
 					if info.status == PaymentStatus::Succeeded
 						|| matches!(info.kind, PaymentKind::Spontaneous { .. })
 					{
+						let stored_preimage = match info.kind {
```
tnull, line 694: This might or might not conflict with #811, we'll also need to think through this depending on the order these land.

tests/common/mod.rs
joostjager: Had to add a pub method to get the number of pending ops. Done

src/lib.rs
```diff
@@ -775,6 +775,12 @@ impl Node {
 		}
 	}
 
+	/// Returns the number of pending deferred chain monitor operations that have not
+	/// been flushed yet by the background processor.
+	pub fn pending_monitor_operation_count(&self) -> usize {
```
tnull: We def. shouldn't expose this in the API, let alone with that little documentation. Please make it private, or add some `cfg` flag to enable this for tests. Alternatively, it could be added as another field to `NodeStatus`?

src/event.rs
tnull: Yeah, concern still stands, I'll see to continue work on #811 in the next days so we can decide better.

src/types.rs
tnull: Likewise still an unresolved question to me.

src/lib.rs
joostjager: If we don't want to expose this in the API, I think we also shouldn't expose it in `NodeStatus`?

I looked at somehow making it test only, but I don't think it is possible for the integration tests? Open to other suggestions.

tnull: Well, we could hide it behind some specific `cfg`-flag. FWIW, `NodeStatus` could make some sense as it's maybe a decent proxy metric for whether the node is IO bottlenecked?

joostjager: Discussed offline to revert to fixed delay for now.

PR #784, opened 2026-02-04
Finish migration to tracking all payments by `PaymentId`

### LINKED ISSUE #298: Track inbound payments by also unique payment IDs

tnull:
Closes #298.

Since by now LDK added the capabilities for inbound payment ids, we can now lean on them and remove any legacy instances where we assumed `payment_id` to be derived from `payment_hash`.

This is currently still blocked on us adding a payment metadata store as for the JIT case we need a place to store the `LSPFeeLimits` upon invoice creation (and with this PR, we stop creating payment store entries in `Bolt11Payment::receive*`, but rather create new entries only when handling `PaymentClaimable`, as we do for BOLT12).

Draft until we can move forward here.


tnull:
Closing in favor of doing it all in #811.

PR #788, opened 2026-02-10
Add ability to open channel / splice with all on-chain funds

### LINKED ISSUE #777: splice-in-all-onchain-funds API
Orange-sdk currently does something rather hacky when trying to splice-in after receiving on chain funds where it subtracts 1000 sats to reserve for fees. Really there needs to be an upstream API here (and I guess on LDK itself? or is that already sufficient) to just say "hey, please splice in all of my on-chain funds, minus fees".

benthecarman:
Closes #777

Uses the same fee estimation logic that `send_all_to_address` uses which can be hack but it works today.

Tests generated with claude.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1194,17 +1194,8 @@ impl Node {
 			.peer_by_node_id(peer_node_id)
 			.ok_or(Error::ConnectionFailed)?
 			.init_features;
-		let sats = self.config.anchor_channels_config.as_ref().map_or(0, |c| {
-			if init_features.requires_anchors_zero_fee_htlc_tx()
-				&& !c.trusted_peers_no_reserve.contains(peer_node_id)
-			{
-				c.per_channel_reserve_sats
-			} else {
-				0
-			}
-		});
-
-		Ok(sats)
+		let anchor_channel = init_features.requires_anchors_zero_fee_htlc_tx();
+		Ok(new_channel_anchor_reserve_sats(&self.config, peer_node_id, anchor_channel))
 	}
 
 	fn check_sufficient_funds_for_channel(
@@ -2008,5 +1999,21 @@ pub(crate) fn total_anchor_channels_reserve_sats(
 			})
 			.count() as u64
 			* anchor_channels_config.per_channel_reserve_sats
+	})
+}
+
+pub(crate) fn new_channel_anchor_reserve_sats(
+	config: &Config, peer_node_id: &PublicKey, anchor_channel: bool,
+) -> u64 {
+	if !anchor_channel {
+		return 0;
+	}
+
+	config.anchor_channels_config.as_ref().map_or(0, |c| {
+		if c.trusted_peers_no_reserve.contains(peer_node_id) {
+			0
+		} else {
+			c.per_channel_reserve_sats
+		}
 	})
 }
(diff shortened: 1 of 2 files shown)


src/wallet/mod.rs
```diff
@@ -459,6 +461,87 @@ impl Wallet {
...
+				self.logger,
+				"Unable to determine max funding amount: no spendable funds available."
+			);
+			return Err(Error::InsufficientFunds);
+		}
+
+		// Use a dummy P2WSH script (34 bytes) to match the size of a real funding output.
+		let dummy_p2wsh_script = ScriptBuf::new().to_p2wsh();
+
+		let tmp_tx = if cur_anchor_reserve_sats > DUST_LIMIT_SATS {
```
tnull: Can we DRY this up by pulling the 'spend all but reserves' part into a helper method in a prefactor commit? Basically this `get_max_funding_amount` should already be ~that helper method, no? We'd just need to reuse it in `send_all`?

tests/integration_tests_rust.rs
```diff
@@ -2506,3 +2506,107 @@ async fn persistence_backwards_compatibility() {
 
 	node_new.stop().unwrap();
 }
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn open_channel_with_all_with_anchors() {
+	let (bitcoind, electrsd) = setup_bitcoind_and_electrsd();
+	let chain_source = TestChainSource::Esplora(&electrsd);
```
tnull, line 2476: Given that this will likely land after #769, we should probably use `random_chain_source` here and below.

src/lib.rs
```diff
@@ -1091,7 +1091,7 @@ impl Node {
 	}
 
 	fn open_channel_inner(
-		&self, node_id: PublicKey, address: SocketAddress, channel_amount_sats: u64,
+		&self, node_id: PublicKey, address: SocketAddress, channel_amount_sats: Option<u64>,
```
tnull: In the onchain send case we introduced the `OnchainSendAmount` enum to avoid possibly ambiguous `Option<u64>` field in the internal API. Can we use something similar here?

src/lib.rs
```diff
@@ -1171,21 +1220,8 @@ impl Node {
...
-				if init_features.requires_anchors_zero_fee_htlc_tx()
-					&& !c.trusted_peers_no_reserve.contains(peer_node_id)
-				{
-					c.per_channel_reserve_sats
-				} else {
-					0
-				}
-			});
+		let required_funds_sats =
+			amount_sats + self.new_channel_anchor_reserve_sats(peer_node_id)?;
```
tnull, line 1220: Can we DRY this up and reuse (a variant of) `new_channel_anchor_reserve_sats` for the `OpenChannelRequest` handling in event.rs?

src/lib.rs
```diff
@@ -1380,17 +1380,15 @@ impl Node {
 	///
 	/// This API is experimental. Currently, a splice-in will be marked as an outbound payment, but
 	/// this classification may change in the future.
-	pub fn splice_in(
+	fn splice_in_inner(
```
tnull, line 1369: nit: Let's drop the redundant docs here if we make this non-pub now.

PR #791, opened 2026-02-11
Classify onchain transactions

### LINKED ISSUE #447: Create new `PaymentKind` variants for channel funding/closing on-chain transactions
In #432 we exposed on-chain transactions in the payment store.

Once we're able to discern channel-related transactions, we should create dedicated `PaymentKind`s for them and use `Onchain` only for payments/as a catchall

Blocked on  https://github.com/lightningdevkit/rust-lightning/issues/3548

tnull:
Fixes #447.

In #888 we started exposing `TransactionType` for funding-related transactions. Here, we continue that work and also classify the remaining transaction types so they are shown in the payment store.

### DIFF
--- a/src/payment/store.rs
+++ b/src/payment/store.rs
@@ -288,7 +288,9 @@ impl StorableObject for PaymentDetails {
 		if let Some(tx_type_update) = update.tx_type {
 			match self.kind {
 				PaymentKind::Onchain { ref mut tx_type, .. } => {
-					update_if_necessary!(*tx_type, tx_type_update);
+					if tx_type.is_none() || tx_type_update.is_some() {
+						update_if_necessary!(*tx_type, tx_type_update);
+					}
 				},
 				_ => {},
 			}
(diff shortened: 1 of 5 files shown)


src/payment/store.rs
```diff
@@ -351,6 +499,8 @@ pub enum PaymentKind {
 		txid: Txid,
 		/// The confirmation status of this payment.
 		status: ConfirmationStatus,
+		/// The type of the on-chain transaction, if known.
+		tx_type: Option<TransactionType>,
```
tnull, line 516: Hmm, we should probably just default to `Onchain` if we don't know what it is.

src/tx_broadcaster.rs
```diff
@@ -21,6 +25,7 @@ where
 {
 	queue_sender: mpsc::Sender<Vec<Transaction>>,
 	queue_receiver: Mutex<mpsc::Receiver<Vec<Transaction>>>,
+	tx_types: std::sync::Mutex<HashMap<Txid, TransactionType>>,
```
jkczyz, line 28: If we broadcast the transaction but restart before the wallet syncs, won't we lose this data?

tnull: Yeah, we need to double-check whether we need to persist this. I think most channel-related transactions will be regenerated/rebroadcast after restart, but not sure if that would be the case for splice-related data?

src/wallet/mod.rs
jkczyz: Should we persist before broadcasting like we do for splices?

tnull: Done in a fixup.

jkczyz (commented):
LGTM. Please squash

jkczyz (dismissed):

jkczyz:
Oops, looks like this needs a rebase.

tnull:
> Oops, looks like this needs a rebase.

Rebased

tnull:
Failing test is unrelated flake (passes locally), need to fix in another PR

jkczyz (approved):

PR #792, opened 2026-02-12
Refactor liquidity source to support multiple LSP nodes

Camillarhi:
The current setup ties you to a single LSP per protocol via `set_liquidity_source_lsps1` / `set_liquidity_source_lsps2`. This refactor replaces that with a unified `Vec<LspNode>` model where LSP nodes are added via `add_lsp()` and protocol support is discovered at runtime through LSPS0 `list_protocols`. Multi-LSP support has been requested previously in #529.

* Deprecated `set_liquidity_source_lsps1` / `set_liquidity_source_lsps2` in favor of `add_lsp()`
* Replaced the per-protocol `LSPS1Client` / `LSPS2Client` with global pending request maps keyed by `LSPSRequestId`
* Added LSPS0 protocol discovery with event handling for `ListProtocolsResponse`
* Background discovery task spawns on `Node::start()`
* LSPS2 JIT channels now query all LSPS2-capable LSPs and automatically select the cheapest fee offer
* Added `request_channel_from_lsp()` for explicit LSPS1 LSP selection
* Updated event handling to use `is_lsps_node()` for multi-LSP counterparty checks

This sets the foundation for LSPS5 support currently being worked on in #729


src/builder.rs
```diff
@@ -422,16 +413,32 @@ impl NodeBuilder {
...
+	/// Configures the [`Node`] instance to source inbound liquidity from the given LSP, without specifying
+	/// the exact protocol used (e.g., LSPS1 or LSPS2).
+	///
+	/// Will mark the LSP as trusted for 0-confirmation channels, see [`Config::trusted_peers_0conf`].
+	///
+	/// The given `token` will be used by the LSP to authenticate the user.
+	/// This method is useful when the user wants to connect to an LSP but does not want to be concerned with
+	///  the specific protocol used for liquidity provision. The node will automatically detect and use the
+	/// appropriate protocol supported by the LSP.
+	pub fn add_lsp(
```
tnull: Hmm, I think there previously were users that required to be able to configure the LSPs at runtime. Not sure if they solved it differently by now, but we could at least explore whether this is possible, e.g., by allowing to set this on a sub-API struct retrievable from `Node` (similar to what we do for payments), e.g. via `node.liquidity().add_lsp()`.

src/liquidity.rs
```diff
@@ -267,8 +221,23 @@ where
...
+				.lsp_nodes
+				.into_iter()
+				.map(|cfg| LspNode {
+					node_id: cfg.node_id,
+					address: cfg.address,
+					token: cfg.token,
+					supported_protocols: Mutex::new(None),
+				})
+				.collect(),
+			pending_lsps1_opening_params_requests: Mutex::new(HashMap::new()),
```
tnull: Uh, no, I don't think we should just merge all into one. Especially given that we intend to add more logic on a per-spec basis, this will be will become even more confusing going forward. If anything, we should maybe start the refactoring by *first* moving the LSPS1/LSPS2 specific parts to `src/liquidity/{lsps1,lsps2}.rs`, or maybe even to client/service specific sub-modules like `src/liquidity/{client,service}/{lsps1,lsps2}.rs`.

src/liquidity.rs
```diff
@@ -947,15 +896,18 @@ where
 
 	pub(crate) async fn lsps1_request_channel(
 		&self, lsp_balance_sat: u64, client_balance_sat: u64, channel_expiry_blocks: u32,
-		announce_channel: bool, refund_address: bitcoin::Address,
+		announce_channel: bool, refund_address: bitcoin::Address, node_id: &PublicKey,
```
tnull: Hmm, I do wonder if it would make sense to create something like an `struct ServiceProvider` and move the API methods to there. Then, each registered LSP would have a corresponding `ServiceProvider` that exposes a bunch of public and internal APIs, which would make the modularization cleaner and would avoid having to give `node_id` everywhere?

tnull (commented):
Thanks! Some initial questions..

src/builder.rs
Camillarhi: I will explore exposing this at runtime

src/liquidity.rs
Camillarhi: Alright, I will move the LSPS1/LSPS2 specific parts

src/liquidity.rs
Camillarhi: Makes sense, I will have a look and see

Camillarhi:
Hello @tnull, apologies for the delay. This is ready for another round of review

src/liquidity.rs
```diff
@@ -1,1542 +0,0 @@
-// This file is Copyright its original authors, visible in version control history.
```
tnull, line 1: Would be good to structure this PR in a way that (as far as possible) makes any code moves dedicated commits that can be picked up by `git diff --color-moved --patience`, as otherwise reviewing this in detail will be very hard.

src/lib.rs
```diff
@@ -674,6 +675,29 @@ impl Node {
...
+							discovery_logger,
+							"Failed to connect to LSP {} for protocol discovery: {}",
+							node_id,
+							e
+						);
+						continue;
+					}
+				}
+
+				discovery_ls.discover_all_lsp_protocols().await;
```
tnull: I think it would be good to make this part of the background task above? Also, can we spawn the discovery tasks in parallel rather than doing them sequentially?

src/lib.rs
```diff
@@ -1949,6 +1973,39 @@ impl Node {
 			Error::PersistenceFailed
 		})
 	}
+
+	/// Configures the [`Node`] instance to source inbound liquidity from the given LSP at runtime,
+	/// without specifying the exact protocol used (e.g., LSPS1 or LSPS2).
```
tnull: I think we can drop the remark regarding 'without specifying the exact protocol' here and elsewhere, as the API already communicates that due to being generic. I do however wonder if we'd want to move this method to an API-extension object similar to what we do for the payment types? I.e., retrieve the API object via `Node::liquidity()`?

src/liquidity/client/lsps1.rs
```diff
@@ -0,0 +1,548 @@
...
+/// To open [bLIP-52 / LSPS2] JIT channels, please refer to
+/// [`Bolt11Payment::receive_via_jit_channel`].
+///
+/// [bLIP-51 / LSPS1]: https://github.com/lightning/blips/blob/master/blip-0051.md
+/// [bLIP-52 / LSPS2]: https://github.com/lightning/blips/blob/master/blip-0052.md
+/// [`Node::lsps1_liquidity`]: crate::Node::lsps1_liquidity
+/// [`Bolt11Payment::receive_via_jit_channel`]: crate::payment::Bolt11Payment::receive_via_jit_channel
+#[derive(Clone)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Object))]
+pub struct LSPS1Liquidity {
```
tnull: It seems this is currently not exposed in the API anymore?

src/builder.rs
```diff
@@ -440,17 +436,12 @@ impl NodeBuilder {
 	/// The given `token` will be used by the LSP to authenticate the user.
 	///
 	/// [bLIP-51 / LSPS1]: https://github.com/lightning/blips/blob/master/blip-0051.md
+	#[deprecated(note = "Use `add_lsp` instead")]
```
tnull: Hmm, I think so far we've been fine with just breaking the APIs without deprecating them first. If we find a better API I'd be fine with just dropping the old ones to clean up.

tnull (commented):
Cool, already looks pretty good! Did a first higher-level pass, have yet to look into the details.

src/lib.rs
Camillarhi: Thanks. I will remove the remark and move this to a `Node::liquidity()` API-extension

tnull (commented):
Cool. Docs CI is currently failing. 

Would be great if you try once more to breakup the second commit into smaller chunks, but let me know if it's too cumbersome/impossible.

Feel free to mark ready for review then.

tnull (commented):
Cool to see this works, from what I see this already looks pretty good! 

However, it's still very hard to check that most of the changes are actually just code moves and do not change the behavior. Actually reviewing this would mean to re-review ~all liquidity related logic. So, please split the second commit into multiple ones, preferably checking that `git diff --color-moved --ignore-all-space [--patience]` does indeed pick up the code moves as such.

Could e.g., do these commits, rougly:

- Just move LSPS1-related logic into new module, simply redirecting calls
- Then move all LSPS2-related logic into new module, simply redirecting calls
- Then start making behavioral changes, e.g., switching to the service discovery.

This might also need another rebase as we're now disallowing any `unwrap`s since #868.

Camillarhi:
> Cool to see this works, from what I see this already looks pretty good!
> 
> However, it's still very hard to check that most of the changes are actually just code moves and do not change the behavior. Actually reviewing this would mean to re-review ~all liquidity related logic. So, please split the second commit into multiple ones, preferably checking that `git diff --color-moved --ignore-all-space [--patience]` does indeed pick up the code moves as such.
> 
> Could e.g., do these commits, rougly:
> 
> * Just move LSPS1-related logic into new module, simply redirecting calls
> * Then move all LSPS2-related logic into new module, simply redirecting calls
> * Then start making behavioral changes, e.g., switching to the service discovery.
> 

Thanks for the detailed breakdown! I'll split the second commit along the lines you suggested. 


> This might also need another rebase as we're now disallowing any `unwrap`s since #868.

The branch has also been rebased and all `unwrap` usages have been updated to align with the changes introduced in #868.

Camillarhi:
> Cool to see this works, from what I see this already looks pretty good!
> 
> However, it's still very hard to check that most of the changes are actually just code moves and do not change the behavior. Actually reviewing this would mean to re-review ~all liquidity related logic. So, please split the second commit into multiple ones, preferably checking that `git diff --color-moved --ignore-all-space [--patience]` does indeed pick up the code moves as such.
> 
> Could e.g., do these commits, rougly:
> 
> * Just move LSPS1-related logic into new module, simply redirecting calls
> * Then move all LSPS2-related logic into new module, simply redirecting calls
> * Then start making behavioral changes, e.g., switching to the service discovery.
> 

Hi @tnull, I've split the second commit along the lines you suggested:

src/lib.rs
```diff
@@ -679,7 +680,51 @@ impl Node {
 			let mut stop_liquidity_handler = self.stop_sender.subscribe();
 			let liquidity_handler = Arc::clone(&liquidity_source);
 			let liquidity_logger = Arc::clone(&self.logger);
+			let discovery_ls = Arc::clone(&liquidity_source);
+			let discovery_cm = Arc::clone(&self.connection_manager);
 			self.runtime.spawn_background_task(async move {
+				let discovery_logger = Arc::clone(&liquidity_logger);
+				tokio::spawn(async move {
```
tnull, line 687: We need to make sure that we cancel all spawned tasks on shutdown based on the stop signal.

src/liquidity/mod.rs
```diff
@@ -0,0 +1,478 @@
...
+	/// LSP nodes are automatically trusted for 0-confirmation channels.
+	///
+	/// The given `token` will be used by the LSP to authenticate the user.
+	pub fn add_lsp(
+		&self, node_id: PublicKey, address: SocketAddress, token: Option<String>,
+	) -> Result<(), Error> {
+		let liquidity_source =
+			self.liquidity_source.as_ref().ok_or(Error::LiquiditySourceUnavailable)?;
+		let lsp_config = LspConfig { node_id, address, token };
+		liquidity_source.add_lsp_node(lsp_config.clone())?;
```
tnull, line 143: Can we just inline this helper and other very short helpers like `is_lsps_node`? They seem to just increas the clutter.

We should also abort here if we already have an entry for the `node_id`, not just add another entry at the end.

src/liquidity/mod.rs
```diff
@@ -0,0 +1,478 @@
...
+	pub(crate) async fn discover_lsp_protocols(
+		&self, node_id: &PublicKey,
+	) -> Result<Vec<u16>, Error> {
+		let lsps0_handler = self.liquidity_manager.lsps0_client_handler();
+
+		let (sender, receiver) = oneshot::channel();
+		{
+			let mut pending_discovery = self.pending_lsps0_discovery.lock().expect("lock");
+			lsps0_handler.list_protocols(node_id);
+			pending_discovery.insert(*node_id, sender);
```
tnull, line 445: Hmm, should we also abort here if we find a discovery is already in-flight before we just override it? Also, when would we expire stale requests, i.e., `sender`s we add here but which never finish, e.g., by hitting an error below?

src/liquidity/client/lsps2.rs
```diff
@@ -0,0 +1,507 @@
...
+{
+	pub(crate) async fn lsps2_receive_to_jit_channel(
+		&self, amount_msat: u64, description: &Bolt11InvoiceDescription, expiry_secs: u32,
+		max_total_lsp_fee_limit_msat: Option<u64>, payment_hash: Option<PaymentHash>,
+		connection_manager: &ConnectionManager<L>,
+	) -> Result<(Bolt11Invoice, u64, LspConfig), Error> {
+		let lsps2_nodes = select_all_lsps_for_protocol(&self.lsp_nodes, 2);
+		if lsps2_nodes.is_empty() {
+			log_error!(self.logger, "No LSPs available for LSPS2 protocol.",);
+			return Err(Error::LiquiditySourceUnavailable);
```
tnull, line 62: Hmm, given the discovery will now take a bit after startup, do we see any way to make this not immediately fail, e.g., when a node comes up and calls `receive_via_jit_channel`? Maybe we could block on the discovery task, of course applying a timeout?

src/builder.rs
```diff
@@ -435,45 +431,24 @@ impl NodeBuilder {
...
-	pub fn set_liquidity_source_lsps2(
+	/// This method is useful when the user wants to connect to an LSP but does not want to be concerned with
+	///  the specific protocol used for liquidity provision. The node will automatically detect and use the
+	/// appropriate protocol supported by the LSP.
+	pub fn add_lsp(
 		&mut self, node_id: PublicKey, address: SocketAddress, token: Option<String>,
 	) -> &mut Self {
 		// Mark the LSP as trusted for 0conf
-		self.config.trusted_peers_0conf.push(node_id.clone());
+		self.config.trusted_peers_0conf.push(node_id);
```
tnull, line 447: It seems we only set this here, but not for the `Liquidity::add_lsp` variant? Maybe they both should now take `trust_peer_0conf: bool`?

src/builder.rs
```diff
@@ -435,45 +431,24 @@ impl NodeBuilder {
...
 	/// Will mark the LSP as trusted for 0-confirmation channels, see [`Config::trusted_peers_0conf`].
 	///
 	/// The given `token` will be used by the LSP to authenticate the user.
-	///
-	/// [bLIP-52 / LSPS2]: https://github.com/lightning/blips/blob/master/blip-0052.md
-	pub fn set_liquidity_source_lsps2(
+	/// This method is useful when the user wants to connect to an LSP but does not want to be concerned with
+	///  the specific protocol used for liquidity provision. The node will automatically detect and use the
+	/// appropriate protocol supported by the LSP.
+	pub fn add_lsp(
```
tnull, line 443: Hmm, the previous API (`set_liquidity_source_lsps2`) tried to somewhat disambiguate the case where we are the client vs. where we are the provider. We also intentionally had most builder methods start with `set_` to reach a more homogeneous API. Breaking the latter is fine, but if we rename this it would be good to find a name that also works in context of `set_liquidity_provider_lsps2`, as `add_lsp` and `set_liquidity_provider_lsps2` read now like they do very similar things, but they actually don't.

src/event.rs
```diff
@@ -1263,14 +1267,15 @@ where
 						.try_into()
 						.expect("slice is exactly 16 bytes"),
 				);
-				let allow_0conf = self.config.trusted_peers_0conf.contains(&counterparty_node_id);
-				let mut channel_override_config = None;
-				if let Some((lsp_node_id, _)) = self
+				let is_lsp_node = self
```
tnull, line 1270: Super confused here: why are we even updating the list in `add_lsp` if we now decide we trust *every* LSP for 0conf? Shouldn't we still maintain the list and only trust individual LSPs if the users configures them?

tnull (commented):
Thanks, def. more approachable!

src/liquidity/mod.rs
Camillarhi: > Can we just inline this helper and other very short helpers like `is_lsps_node`? They seem to just increas the clutter.
> 

I have Inlined `add_lsp_node` directly into `add_lsp`, and kept `LiquiditySource::is_lsps_node` as a method since `event.rs` only holds an `Arc<LiquiditySource>` and can't reach the private `lsp_nodes` field without widening visibility.

> We should also abort here if we already have an entry for the `node_id`, not just add another entry at the end.

A duplicate check has been added that returns `Error::DuplicateLspNode`, so re-adding the same `node_id` aborts early instead of silently appending another entry.

src/liquidity/mod.rs
Camillarhi: Thanks for pointing that out. I hadn't considered either case. I'll look into it and push a fix.

src/builder.rs
Camillarhi: LSPs added via the builder were always marked as trusted for 0-conf. Instead of mutating the config at runtime, I took that as the intended default and added the `|| is_lsp_node` check in `event.rs` so runtime-added LSPs get the same treatment, which is why `Liquidity::add_lsp` doesn't touch `trusted_peers_0conf`.

If we're moving to an explicit `trust_peer_0conf: bool` parameter, my plan would be to store it on `LspConfig` and have the event handler check that alongside `trusted_peers_0conf`, rather than trying to mutate Config at runtime. The `|| is_lsp_node` auto-trust would go away. What do you think?

src/builder.rs
Camillarhi: We now support multiple LSPs, so `set_` felt misleading, implying a single value, and discovery determines the protocol. How does renaming it to `add_liquidity_source` sound? That would restore the `source/provider` contrast with `set_liquidity_provider_lsps2`.

src/event.rs
Camillarhi: Yeah, this is the same inconsistency you flagged on the other comment https://github.com/lightningdevkit/ldk-node/pull/792#discussion_r3110890719. I trusted every LSP via `|| is_lsp_node` because at compile time, LSPs added through the builder were already always trusted for `0-conf`

src/builder.rs
tnull: That sounds good, and maybe since we're already breaking the pattern it would then be clearer to rename the other one to `enable_liquidity_service` or `enable_liquidity_provider` or similar and have it take an `lsps2_service_config: Option<LSPS2ServiceConfig>` argument, given that we'll soon also want to add an LSPS1 provider?

Camillarhi: I went with `enable_liquidity_provider` and also renamed `Liquidity::add_lsp` to `add_liquidity_source` to match the builder.

src/builder.rs
Camillarhi: I went with `trust_peer_0conf: bool` on both `add_liquidity_source` variants (builder and runtime) and dropped the `|| is_lsp_node` auto-trust in event.rs, now `allow_0conf` is `trusted_peers_0conf.contains(..) || lsp2_node.trust_peer_0conf`, so the config list stays authoritative and the LSP flag only adds trust on top.

src/liquidity/mod.rs
```diff
@@ -10,49 +10,203 @@
...
+	///
+	/// LSP nodes are automatically trusted for 0-confirmation channels.
+	///
+	/// The given `token` will be used by the LSP to authenticate the user.
+	pub fn add_liquidity_source(
+		&self, node_id: PublicKey, address: SocketAddress, token: Option<String>,
+		trust_peer_0conf: bool,
+	) -> Result<(), Error> {
+		let liquidity_source =
+			self.liquidity_source.as_ref().ok_or(Error::LiquiditySourceUnavailable)?;
```
tnull: Hmm, wouldn't erroring out here mean that we can't add LSPs only at runtime, if we don't also add some during build time?

src/liquidity/mod.rs
```diff
@@ -10,49 +10,203 @@
...
+
+		{
+			let mut lsp_nodes = liquidity_source.lsp_nodes.write().expect("lock");
+			if lsp_nodes.iter().any(|n| n.node_id == node_id) {
+				log_error!(self.logger, "LSP node {} is already added.", node_id);
+				return Err(Error::DuplicateLspNode);
```
tnull: Maybe we should just silently continue rather than erroring here?

src/event.rs
```diff
@@ -1263,31 +1267,33 @@ where
 						.try_into()
 						.expect("slice is exactly 16 bytes"),
 				);
-				let allow_0conf = self.config.trusted_peers_0conf.contains(&counterparty_node_id);
+				let mut allow_0conf =
+					self.config.trusted_peers_0conf.contains(&counterparty_node_id);
 				let mut channel_override_config = None;
-				if let Some((lsp_node_id, _)) = self
+				if let Some(lsp2_node) = self
```
tnull: nit: Should be `lsps2_node` or similar?

src/liquidity/client/lsps2.rs
```diff
@@ -316,4 +400,139 @@ where
...
+			// LSP protocol discovery may still be in flight, we wait briefly for it to finish, then re-check.
+			let mut rx = self.discovery_done_rx.clone();
+			if !*rx.borrow() {
+				log_debug!(
+					self.logger,
+					"No LSPS2 node available yet, waiting for protocol discovery to complete."
+				);
+				let _ = tokio::time::timeout(
+					Duration::from_secs(LSPS_DISCOVERY_WAIT_TIMEOUT_SECS),
+					rx.wait_for(|done| *done),
```
tnull, line 529: Hmm, if there's nothing in-flight, we might just be waiting unnecessarily, no?

src/liquidity/client/lsps1.rs
```diff
@@ -242,34 +228,242 @@ where
 
 		Ok(response)
 	}
+
+	pub(crate) async fn handle_next_event(&self, event: LSPS1ClientEvent) {
```
tnull: Can make moving the event handling to the sub-modules 3 more prefactor commits (at least LSPS1 client, LSPS2 client, LSPS2 service, maybe also for the LSPS0 variant)?

src/liquidity/client/lsps1.rs
```diff
@@ -7,82 +7,54 @@
...
-	pub(crate) pending_opening_params_requests:
+pub(crate) struct LSPS1LiquiditySource<L: Deref>
+where
+	L::Target: LdkLogger,
+{
+	pub(crate) lsp_nodes: Arc<RwLock<Vec<LspNode>>>,
+	pub(crate) pending_lsps1_opening_params_requests:
 		Mutex<HashMap<LSPSRequestId, oneshot::Sender<LSPS1OpeningParamsResponse>>>,
-	pub(crate) pending_create_order_requests:
+	pub(crate) pending_lsps1_create_order_requests:
```
tnull: Why are these field renames necessary? They seem to only increase the noise in the diff?

tnull (commented):
Thanks for splitting things out into prefactors! There's still a lot of things going on at the same time in the last commit, please break it up further if you see the opportunity. A few more comments, also some suggestions on how to reduce noisiness in the last commit.

src/liquidity/client/lsps2.rs
Camillarhi: That is true. The wait should only kick in when there's actually something in flight worth waiting for. `LspNode.supported_protocols` is `None` precisely while a node's protocol discovery hasn't completed or failed. So we can use that as the in-flight signal, if every node already has its protocols populated, there's nothing to wait for.

The previous shape selected `all_lsps_for_protocol` first and only waited if the result was empty. But that means if 1 of N LSPS2 capable nodes happened to respond quickly while the rest were still mid-discovery, we'd return a partial set and skip waiting which isn't ideal since we want to pick the best LSP from the full set. So I've inverted the structure to wait first if anything's pending, then select once.

src/liquidity/client/lsps1.rs
Camillarhi: I have broken this down into prefactor commits for LSPS1 client, LSPS2 client and LSPS2 service

src/liquidity/client/lsps1.rs
Camillarhi: The field renames have been reverted

src/liquidity/client/lsps1.rs
```diff
@@ -0,0 +1,534 @@
...
+		})?
+		.map_err(|e| {
+			log_error!(self.logger, "Failed to handle response from liquidity service: {}", e);
+			Error::LiquidityRequestFailed
+		})?;
+
+		Ok(response)
+	}
+
+	pub(crate) async fn handle_next_event(&self, event: LSPS1ClientEvent) {
```
tnull: nit: This should probably simply be called `handle_event` or `handle_lsps1_event`, same for the LSPS2 counterparts ofc.

src/lib.rs
```diff
@@ -666,25 +666,73 @@ impl Node {
...
+		let discovery_cm = Arc::clone(&self.connection_manager);
+		self.runtime.spawn_background_task(async move {
+			// Spawn discovery for configured LSPs in parallel.
+			let discovery_logger = Arc::clone(&liquidity_logger);
+			let mut discovery_set = tokio::task::JoinSet::new();
+			for (node_id, address) in liquidity_handler.get_all_lsp_details() {
+				let cm = Arc::clone(&discovery_cm);
+				let logger = Arc::clone(&discovery_logger);
+				let ls = Arc::clone(&liquidity_handler);
+				discovery_set.spawn(async move {
```
tnull, line 690: We should probably make sure to cancel/abort all tasks we spawn here in case of a shutdown.

src/builder.rs
```diff
@@ -443,45 +439,35 @@ impl NodeBuilder {
...
-	/// Configures the [`Node`] instance to source inbound liquidity from the given
-	/// [bLIP-51 / LSPS1] service.
+	/// Configures the [`Node`] instance to source inbound liquidity from the given LSP.
 	///
-	/// Will mark the LSP as trusted for 0-confirmation channels, see [`Config::trusted_peers_0conf`].
+	/// The node will discover the LSP's supported protocols (LSPS1/LSPS2) on startup via [bLIP-50 / LSPS0]
+	/// and select the appropriate protocol per request automatically.
 	///
 	/// The given `token` will be used by the LSP to authenticate the user.
+	/// `trust_peer_0conf` controls whether the node will accept 0-confirmation channels opened by this
```
tnull: Codex:

>   - Medium: trust_peer_0conf = false does not actually supersede Config::trusted_peers_0conf as documented. The event handler initializes allow_0conf from the global config and ORs in the LSP flag, so an LSP present in the global trust list is still accepted as zero-conf even when added with
    trust_peer_0conf = false. See docs at src/builder.rs:448 and behavior at src/event.rs:1294.

src/event.rs
```diff
@@ -680,9 +679,9 @@ where
 				}
 			},
 			LdkEvent::FundingTxBroadcastSafe { user_channel_id, counterparty_node_id, .. } => {
-				self.liquidity_source.as_ref().map(|ls| {
-					ls.lsps2_funding_tx_broadcast_safe(user_channel_id, counterparty_node_id);
-				});
+				self.liquidity_source
```
tnull, line 712: Codex:

>   - Low: Non-LSPS2-service nodes now call the LSPS2 service broadcast-safe hook for every FundingTxBroadcastSafe event. Because lsps2_funding_tx_broadcast_safe treats lsps2_service: None as “do not early-return,” it reaches the missing service handler path and logs an error for ordinary nodes/
    channels. See src/event.rs:681 and src/liquidity/service/lsps2.rs:164.

src/lib.rs
```diff
@@ -666,25 +666,73 @@ impl Node {
...
+							  liquidity_logger,
+							  "Stopping processing liquidity events.",
+						  );
+						  return;
+					  }
+					  _ = liquidity_handler.handle_next_event() => {}
+					  res = discovery_set.join_next(), if !discovery_done => {
+						if res.is_none() {
+							liquidity_handler.mark_discovery_done();
+							discovery_done = true;
```
tnull, line 736: Codex:

>   - High: A builder-configured LSP can become permanently unusable after a transient startup discovery failure. Node::start logs connect/discovery errors and still marks discovery done, while LSPS1/LSPS2 selection later only considers nodes with supported_protocols = Some(...) and never
    retries undiscovered nodes. Runtime add_liquidity_source also skips the same node ID, so users cannot recover in-process. See src/lib.rs:682, src/lib.rs:725, src/liquidity/client/lsps2.rs:506, src/liquidity/client/lsps1.rs:400, src/liquidity/mod.rs:126.

Do we see an easy  way to retry discovery during the runtime? If not/if it gets too complicated I'd rather leave it for a follow-up.

Camillarhi: I will look into it and if it gets too complicated, I will do it in a follow up PR

Camillarhi: I will handle this in a follow-up PR. I will have a background retry for failed startup discovery and also expose an API, something like `retry_discovery(node_id)`, to cover the case where a user knows their configured LSP has rolled out a new protocol.

tnull (commented):
Seems CI is unfortunately failing right now (CLN failure is unrelated though).

Camillarhi:
> Seems CI is unfortunately failing right now (CLN failure is unrelated though).

The failing CI has been resolved

src/liquidity/client/lsps1.rs
```diff
@@ -0,0 +1,534 @@
...
+}
+
+#[cfg_attr(feature = "uniffi", uniffi::export)]
+impl LSPS1Liquidity {
+	/// Connects to the configured LSP and places an order for an inbound channel.
+	///
+	/// The channel will be opened after one of the returned payment options has successfully been
+	/// paid.
+	///
+	/// If `node_id` is `None` and multiple LSPs support LSPS1, the first one registered
```
f3r10, line 469: I have one question: Was best-price comparison across multiple LSPS1 providers considered? LSPS2 queries all providers and picks the cheapest, but LSPS1 falls back to the first registered. Is there a reason for this asymmetry, or is price comparison for LSPS1 planned for the future?

f3r10 (commented):
Great work. Still trying to understand all the changes.

src/event.rs
```diff
@@ -1291,35 +1291,37 @@ where
...
-						// check that they don't take too much before claiming.
-						channel_override_config = Some(ChannelConfigOverrides {
-							update_overrides: Some(ChannelConfigUpdate {
-								accept_underpaying_htlcs: Some(true),
-								..Default::default()
-							}),
+					allow_0conf = allow_0conf || lsp.trust_peer_0conf;
+				}
+
+				if self.liquidity_source.get_lsp_config(&counterparty_node_id, Some(2)).is_some() {
```
f3r10: If an LSPS2-capable LSP opens a channel before its discovery completes (i.e. supported_protocols is still None), `get_lsp_config(Some(2))` returns `None` and accept_underpaying_htlcs is never enabled. Could this silently break the JIT channel flow for fast LSPs?

src/liquidity/client/lsps1.rs
Camillarhi: The behavioral difference is intentional and rests on two things:
- The API exposure, `LSPS1Liquidity::request_channel` takes `node_id: Option<PublicKey>` and is the user's primary entry point, while `lsps2_receive_to_jit_channel` for example is invoked internally from `Bolt11Payment::receive_via_jit_channel`, therefore the choice is automated
- The Protocol shape, LSPS2 has a `request_opening_params` which returns a fee menu, so cost is computable. LSPS1 has a `request_supported_options` which returns only capability ranges , no fee fields. The price will only materialize from `create_order`, which commits to a specific order with a specific LSP.

src/event.rs
Camillarhi: Thanks for pointing this out, `discovery_done` wait was not covered here. It has been updated

src/event.rs
```diff
@@ -1315,35 +1315,36 @@ where
...
+				let mut allow_0conf =
+					self.config.trusted_peers_0conf.contains(&counterparty_node_id);
 				let mut channel_override_config = None;
-				if let Some((lsp_node_id, _)) = self
-					.liquidity_source
-					.as_ref()
-					.and_then(|ls| ls.as_ref().get_lsps2_lsp_details())
+
+				// If the peer is a configured LSP node, additionally honor its trust_peer_0conf flag.
+				if let Some(lsp) =
```
tnull, line 1323: Codex:

>   - [P1] trust_peer_0conf is only honored for LSPS2-discovered LSPs: /home/tnull/worktrees/ldk-node/pr-792/src/event.rs:1323. add_liquidity_source(..., true) documents generic 0-conf trust for the LSP (/home/tnull/worktrees/ldk-node/pr-792/src/builder.rs:448, /home/tnull/worktrees/ldk-node/pr-
>     792/src/liquidity/mod.rs:117), but inbound acceptance only looks up protocol 2. LSPS1-only LSPs, or trusted LSPs before/after failed LSPS0 discovery, now fall through to accept_inbound_channel instead of accept_inbound_channel_from_trusted_peer. The trust lookup should be by node id
>     independent of protocol; keep the LSPS2 protocol check only for accept_underpaying_htlcs.
>

src/message_handler.rs
```diff
@@ -18,24 +18,19 @@ use lightning_types::features::{InitFeatures, NodeFeatures};
 
 use crate::liquidity::LiquiditySource;
 
-pub(crate) enum NodeCustomMessageHandler<L: Deref>
+pub(crate) struct NodeCustomMessageHandler<L: Deref>
 where
 	L::Target: Logger,
 {
-	Ignoring,
```
tnull, line 25: Hmm, in conjunction with #900 we might want to revisit this, i.e., whether we'll want to allow to disable LSP functionalities based on a feature entirely.

src/liquidity/client/mod.rs
```diff
@@ -0,0 +1,11 @@
...
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
+// accordance with one or both of these licenses.
+
+pub(crate) mod lsps1;
+pub(crate) mod lsps2;
+
+pub use lsps1::LSPS1OrderStatus;
```
tnull, line 11: nit:  

>  - [P3] Three new files have CRLF line endings, so git diff --check upstream/main...HEAD reports trailing whitespace on every line: /home/tnull/worktrees/ldk-node/pr-792/src/liquidity/client/mod.rs:1, /home/tnull/worktrees/ldk-node/pr-792/src/liquidity/service/mod.rs:1, /home/tnull/worktrees/
>     ldk-node/pr-792/src/liquidity/service/lsps2.rs:1. cargo fmt tolerates it, but diff whitespace checks will fail. Normalize those files to LF.

tnull (approved):
Thanks! Excuse the considerable delay here! I'm going ahead landing this, we can address / revisit minor things in follow-ups (or as part of #900 for instance).

tnull:
Now documented them over at https://github.com/lightningdevkit/ldk-node/issues/936.

Camillarhi:
Thanks for landing it, and no worries about the delay! I'll pick up the follow-ups in #936

PR #793, opened 2026-02-16
Async persistence (explore only)

joostjager:
**This changeset was generated with the help of AI tooling (Claude Code) to get an impression of the work involved in migrating to fully async persistence. It is not intended to be merged as-is.**

This branch migrates all persistence operations in ldk-node from the synchronous `KVStoreSync` trait to LDK's async `KVStore` trait, eliminating blocking I/O from the runtime entirely.

The migration is broken into 8 incremental commits, each converting one subsystem and propagating async through its callers:

- **Use async ChainMonitor persister via `new_async_beta`**: Switch from `MonitorUpdatingPersister` to `MonitorUpdatingPersisterAsync`, using a single async persister for both startup reads and runtime persistence. Introduces `DynStoreRef` to work around higher-ranked lifetime issues with `Arc<dyn DynStoreTrait>`.

- **Convert `write_node_metrics` to async `KVStore`**: Split in-memory mutations from async persist calls to avoid holding write locks across `.await` points.

- **Convert `DataStore` persistence to async `KVStore`**: The largest commit, propagating async through all payment modules (bolt11, bolt12, spontaneous, unified), event handling, wallet, and chain sync. Uses the pattern of doing in-memory work under a lock, dropping the lock, then awaiting the persist.

- **Convert `PeerStore` persistence to async `KVStore`**: Makes peer management operations async, restructuring event handlers to avoid holding non-Send locks across await points.

- **Convert `StaticInvoiceStore` persistence to async `KVStore`**: Straightforward async conversion of static invoice read/write operations.

- **Convert BDK wallet persistence to `AsyncWalletPersister`**: Converts the BDK wallet layer from sync `WalletPersister` to `AsyncWalletPersister`, switching the persister to a `tokio::sync::Mutex`. Uses `block_in_place`+`block_on` for sync trait impls (`Listen`, `SignerProvider`).

- **Convert `export_pathfinding_scores` to async `KVStore`**.

- **Remove `KVStoreSync` from core traits and store implementations**: Final cleanup removing the `SyncAndAsyncKVStore` supertrait, all `KVStoreSync` impls from `SqliteStore`/`VssStore`/`InMemoryStore`, and the sync test infrastructure. Simplifies `DynStoreTrait` to async-only methods.

Across 31 files, 1301 insertions and 1516 deletions, the key recurring patterns are:
- Do in-memory work under lock, drop lock, then `.await` the persist
- Use `block_in_place`+`block_on` where sync trait boundaries require it (e.g., `Listen`, `SignerProvider`)
- Wrap `SqliteStore` calls in `async { ... .await }` inside `block_on` to ensure the tokio runtime context is available for `spawn_blocking`

### DIFF
--- a/benches/payments.rs
+++ b/benches/payments.rs
@@ -35,12 +35,10 @@ fn spawn_payment(node_a: Arc<Node>, node_b: Arc<Node>, amount_msat: u64) {
 				tokio::time::sleep(std::time::Duration::from_millis(100)).await;
 			}
 
-			let payment_id = node_a.spontaneous_payment().send_with_preimage(
-				amount_msat,
-				node_b.node_id(),
-				preimage,
-				None,
-			);
+			let payment_id = node_a
+				.spontaneous_payment()
+				.send_with_preimage(amount_msat, node_b.node_id(), preimage, None)
+				.await;
 
 			match payment_id {
 				Ok(payment_id) => {
@@ -110,6 +108,7 @@ async fn send_payments(node_a: Arc<Node>, node_b: Arc<Node>) -> std::time::Durat
 			PaymentPreimage(preimage_bytes),
 			None,
 		)
+		.await
 		.ok()
 		.unwrap();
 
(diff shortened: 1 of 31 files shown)


tnull:
Superseded by #919

PR #800, opened 2026-02-20
Replace `rand` with `getrandom` and `KeysManager` entropy in non-test…

TheBlueMatt:
… code

`rand` is basically just a way to wrap `getrandom` with a chacha stream plus a bag of unrelated random utilities. Given we already have a way to wrap `getrandom` with a chacha stream, there's no point in relying on `rand` for it, so we simply drop it here.

Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>


tnull (approved):
Alright. No strong objection here, so I'm landing this before it has a chance to get stale.

Issue #797, opened 2026-02-20
Follow-ups to PR #628


tnull:
See https://github.com/lightningdevkit/ldk-node/pull/628#pullrequestreview-3831183024 and https://github.com/lightningdevkit/ldk-node/pull/628#discussion_r2832444797

(cc @Camillarhi)

tnull:
Addressed by #802.

PR #802, opened 2026-02-23
Minor follow-ups to #628

Camillarhi:
Some minor follow-ups after #628 landed

- Use user-provided fee rate for bumping transaction and return an error if it is too low, otherwise use system estimated fee rate and retry mechanism.

- Switches the RBF test to use `random_chain_source`.

Closes -  #797

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -378,8 +378,8 @@ impl Wallet {
 					);
 					let payment =
 						self.payment_store.get(&payment_id).ok_or(Error::InvalidPaymentId)?;
-					let pending_payment_details = self
-						.create_pending_payment_from_tx(payment.clone(), conflict_txids.clone());
+					let pending_payment_details =
+						self.create_pending_payment_from_tx(payment, conflict_txids.clone());
 
 					self.pending_payment_store.insert_or_update(pending_payment_details)?;
 				},
@@ -1044,7 +1044,10 @@ impl Wallet {
 
 		if let Some(replaced_details) = self
 			.pending_payment_store
-			.list_filter(|p| p.conflicting_txids.contains(&target_txid))
+			.list_filter(|p| {
+				matches!(p.details.kind, PaymentKind::Onchain { txid, .. } if txid == target_txid)
+					|| p.conflicting_txids.contains(&target_txid)
+			})
 			.first()
 		{
 			return Some(replaced_details.details.id);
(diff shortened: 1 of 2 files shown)


tests/integration_tests_rust.rs
```diff
@@ -2522,10 +2519,10 @@ async fn onchain_fee_bump_rbf() {
 	// Successful fee bump
 	let new_txid = node_b.onchain_payment().bump_fee_rbf(payment_id, None).unwrap();
 	wait_for_tx(&electrsd.client, new_txid).await;
-
-	// Sleep to allow for transaction propagation
+	// Give the chain source time to index the unconfirmed transaction before syncing.
```
Camillarhi, line 2522: The sleep is needed here because of test-environment-specific reasons,  `wait_for_tx` uses the Electrum API to confirm the tx is in the mempool, but when Esplora is the chain source, Esplora may not have indexed it yet. If `sync_wallets()` runs in that window, BDK misses the tx and `bump_fee_rbf` can't find it via tx_details.

In practice, the fee of a transaction will not be bumped the instant they send. By the time a user decides to bump, the wallet has gone through multiple sync cycles, and the tx is long indexed. The sleep simulates that gap.

tests/integration_tests_rust.rs
```diff
@@ -2620,10 +2587,20 @@ async fn onchain_fee_bump_rbf() {
...
 	assert_eq!(node_a_received_payment.len(), 1);
+	match &node_a_received_payment[0].kind {
+		PaymentKind::Onchain { txid: inbound_txid, .. } => {
+			assert_eq!(
+				*inbound_txid, second_bump_txid,
+				"node_a inbound payment txid should be updated to the second replacement txid"
+			);
+		},
+		_ => panic!("Unexpected payment kind"),
+	}
```
Camillarhi, line 2603: Due to sync timing, the assertion verifying that node a's inbound payment `txid` has been updated to the `replacement_txid` is now done after the transaction is confirmed rather than immediately after each bump. This still fully validates the behaviour. Once confirmed, we check that the `txid` reflects the final bumped transaction, confirming that node A's sync correctly processed the replacement.

src/wallet/mod.rs
```diff
@@ -1074,6 +1042,16 @@ impl Wallet {
 			return Some(direct_payment_id);
 		}
 
+		if let Some(replaced_details) = self
+			.pending_payment_store
+			.list_filter(
```
tnull: Hmm, we should probably combine that with the iteration below, no?

tests/integration_tests_rust.rs
```diff
@@ -2497,6 +2490,10 @@ async fn onchain_fee_bump_rbf() {
 	let txid =
 		node_b.onchain_payment().send_to_address(&addr_a, amount_to_send_sats, None).unwrap();
 	wait_for_tx(&electrsd.client, txid).await;
+	// Give the chain source time to index the unconfirmed transaction before syncing.
+	// Without this, Esplora may not yet have the tx, causing sync to miss it and
+	// leaving the BDK wallet graph empty.
+	tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
tnull, line 2496: I think a second should be enough for this?

tests/integration_tests_rust.rs
tnull: Alright, I guess that makes sense (interesting we never ran into this edge case so far). Though, a second should be more than enough for Esplora to catch up to Electrum I assume?

Camillarhi: sometimes a second is enough but not always, locally it worked fine, but CI failed, so bumped it to 5 seconds be safe

src/wallet/mod.rs
Camillarhi: Yes, that is correct. I will make the update

tnull (approved):

PR #805, opened 2026-02-24
Refactoring binding python test

heyolaniran:
Hello @tnull x @enigbe 

In this Pull Request, I'd refactored the python binding test by introducing parallelism and by avoiding code duplication specially with event tracking and handling.

Thanks for having time to review the topic and send me a feedback if i need to make some specific action.

### DIFF
--- a/bindings/python/src/ldk_node/test_ldk_node.py
+++ b/bindings/python/src/ldk_node/test_ldk_node.py
@@ -112,6 +112,16 @@ def get_esplora_endpoint():
         return str(os.environ['ESPLORA_ENDPOINT'])
     return DEFAULT_ESPLORA_SERVER_URL
 
+
+def expect_event(node, expected_event_type):
+    event = node.wait_next_event()
+    assert isinstance(event, expected_event_type)
+    print("EVENT:", event)
+    node.event_handled()
+    return event 
+
+
+
 class TestLdkNode(unittest.TestCase):
     def setUp(self):
         bitcoin_cli("createwallet ldk_node_test")
@@ -175,16 +185,9 @@ class TestLdkNode(unittest.TestCase):
 
         node_1.open_channel(node_id_2, listening_addresses_2[0], 50000, None, None)
 
-        channel_pending_event_1 = node_1.wait_next_event()
-        assert isinstance(channel_pending_event_1, Event.CHANNEL_PENDING)
-        print("EVENT:", channel_pending_event_1)
-        node_1.event_handled()
 
-        channel_pending_event_2 = node_2.wait_next_event()
-        assert isinstance(channel_pending_event_2, Event.CHANNEL_PENDING)
-        print("EVENT:", channel_pending_event_2)
-        node_2.event_handled()
-
+        channel_pending_event_1 = expect_event(node_1, Event.CHANNEL_PENDING)
+        channel_pending_event_2 = expect_event(node_2, Event.CHANNEL_PENDING)
         funding_txid = channel_pending_event_1.funding_txo.txid
         wait_for_tx(esplora_endpoint, funding_txid)
         mine_and_wait(esplora_endpoint, 6)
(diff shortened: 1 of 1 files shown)


bindings/python/src/ldk_node/test_ldk_node.py
```diff
@@ -112,6 +115,25 @@ def get_esplora_endpoint():
...
 
+# handling expect events
+
+def expect_event(node, expected_event_type):
+    event = node.wait_next_event()
+    assert isinstance(event, expected_event_type)
+    print("EVENT:", event)
+    node.event_handled()
+    #According to the event type, we may want to return some data from the event for further processing
+    match expected_event_type:
```
tnull: No, rather than doing that match, let's just introduce dedicated variants (like `expect_channel_ready_event`, etc) for the cases where we need to return something? For all other cases we can keep `expect_event`, without the `match`.

tnull (commented):
Okay, but there are still a lot of changes in this PR that aren't connected to the `expect_event` method transition. Please revert them for now.

bindings/python/src/ldk_node/test_ldk_node.py
heyolaniran: yeah i could be a great idea, thanks.  I've made some update in that way.

bindings/python/src/ldk_node/test_ldk_node.py
```diff
@@ -112,6 +112,33 @@ def get_esplora_endpoint():
         return str(os.environ['ESPLORA_ENDPOINT'])
     return DEFAULT_ESPLORA_SERVER_URL
 
+# handling expected event
```
tnull: nit: Please drop these comments, they are very redundant to the method names.

bindings/python/src/ldk_node/test_ldk_node.py
```diff
@@ -112,6 +112,33 @@ def get_esplora_endpoint():
...
+    node.event_handled()
+
+# handling channel pending event
+
+def expect_channel_pending_event(node):
+    event = node.wait_next_event()
+    assert isinstance(event, Event.CHANNEL_PENDING)
+    print("EVENT:", event)
+    node.event_handled()
+    return event
```
tnull: I'm confused: if we now simply return the event, we can just reduce that to one `expect_event` method again that makes the `assert isinstance` check and returns the event, leaving the matching to the callsite? That's also fine by me, but I don't see why we now need three different methods if they don't do the matching and don't end up having different return values?

heyolaniran: Sure. I should just return the event in the `expect_event`, @tnull i made some fix int that way. I just keep the  `expected_event` with event  return then call the needed attributes on that event. Should be good now, thanks for your improvments

bindings/python/src/ldk_node/test_ldk_node.py
```diff
@@ -175,59 +185,36 @@ def test_channel_full_cycle(self):
...
-        channel_pending_event_1 = node_1.wait_next_event()
-        assert isinstance(channel_pending_event_1, Event.CHANNEL_PENDING)
-        print("EVENT:", channel_pending_event_1)
-        node_1.event_handled()
 
-        channel_pending_event_2 = node_2.wait_next_event()
-        assert isinstance(channel_pending_event_2, Event.CHANNEL_PENDING)
-        print("EVENT:", channel_pending_event_2)
-        node_2.event_handled()
+        channel_pending_event_1 = expect_event(node_2, Event.CHANNEL_PENDING)
```
tnull: Why are we now getting the `_1` event from `node_2`, and what happened to handling the even of `node_1`? 

Please don't reorder any of these statements / generally don't make any unrelated changes.

heyolaniran: Noted, sorry for the disconvenance on that. It is fixed right now.  No more general unrelated changes in the code base

PR #807, opened 2026-02-25
Paginated sqlite

### LINKED ISSUE #804: Implement `PaginatedKVStore` for `SqliteStore`
Now that we have it in LDK, we should do this.

benthecarman:
Closes #804

Implement rust-lightning's PaginatedKVStore/PaginatedKVStoreSync traits
on SqliteStore with a v2→v3 schema migration that adds a created_at
column for tracking insertion order.

### DIFF
--- a/src/io/sqlite_store/migrations.rs
+++ b/src/io/sqlite_store/migrations.rs
@@ -9,11 +9,15 @@ use lightning::io;
 use rusqlite::Connection;
 
 pub(super) fn migrate_schema(
-	connection: &mut Connection, kv_table_name: &str, from_version: u16, to_version: u16,
+	connection: &mut Connection, kv_table_name: &str, mut from_version: u16, to_version: u16,
 ) -> io::Result<()> {
 	assert!(from_version < to_version);
-	if from_version == 1 && to_version == 2 {
+	if from_version == 1 && to_version >= 2 {
 		migrate_v1_to_v2(connection, kv_table_name)?;
+		from_version = 2;
+	}
+	if from_version == 2 && to_version >= 3 {
+		migrate_v2_to_v3(connection, kv_table_name)?;
 	}
 
 	Ok(())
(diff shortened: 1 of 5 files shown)


src/io/sqlite_store/migrations.rs
```diff
@@ -65,11 +69,82 @@ fn migrate_v1_to_v2(connection: &mut Connection, kv_table_name: &str) -> io::Res
...
+	})?;
+
+	let old_table = format!("{}_v2_old", kv_table_name);
+	let map_err = |e: rusqlite::Error| -> io::Error {
+		let msg = format!("Failed to migrate table {} from v2 to v3: {}", kv_table_name, e);
+		io::Error::new(io::ErrorKind::Other, msg)
+	};
+
+	// Recreate the table to ensure the correct PRIMARY KEY regardless of migration history.
+	// Tables migrated from v1 have PK (primary_namespace, key) only — missing
```
tnull, line 79: Huh, I'm confused, why is this true? We explictly have

```rust
	// Add new 'secondary_namespace' column
	let sql = format!(
		"ALTER TABLE {}
			ADD secondary_namespace TEXT DEFAULT \"\" NOT NULL;",
		kv_table_name
	);
```
in `migrate_v1_to_v2`? Am I missing something, or is this Claude hallucinating?

src/io/sqlite_store/migrations.rs
```diff
@@ -65,11 +69,82 @@ fn migrate_v1_to_v2(connection: &mut Connection, kv_table_name: &str) -> io::Res
...
+	// secondary_namespace. Recreating normalizes the schema for all databases.
+	tx.execute(&format!("ALTER TABLE {} RENAME TO {}", kv_table_name, old_table), [])
```
tnull: nit: Please pull out any SQL statments in corresponding `sql` variables?

src/io/sqlite_store/migrations.rs
```diff
@@ -65,11 +69,82 @@ fn migrate_v1_to_v2(connection: &mut Connection, kv_table_name: &str) -> io::Res
...
+		.map_err(map_err)?;
+
+	tx.execute(
+		&format!(
+			"CREATE TABLE {} (
```
tnull, line 111: Honestly not sure if we need all that recreation logic, why can't we just do:

```
let sql = format!(
		"ALTER TABLE {}
			ADD created_at INTEGER NOT NULL DEFAULT 0,
		kv_table_name
	);
```

?

tnull (commented):
Thanks! Can we also extend the `persistence_backwards_compatibility` test to setup an `ldk_node_070::Node` and check we can continue to upgrade from it, too? (Could be refactored into `do_persistence_backwards_compatibility` taking the necessary parameters to DRY up the test logic)

src/io/sqlite_store/migrations.rs
benthecarman: In v1 to v2 is added but its not a part of the primary key so we can't use `ON CONFLICT` on writes for just updating the value. Before we could use replace but now we want to keep the `created_at` date. This rewrites the table to make it a part of the primary key, otherwise we need to do a query inside to look up the `created_at` date when replacing. Could add a unique index instead but this felt cleaner

tnull: Still confused

> This rewrites the table to make it a part of the primary key

But it doesn't?
```
			PRIMARY KEY (primary_namespace, secondary_namespace, key)
```

And arguably `created_at` really shouldn't be part of the key?

src/io/sqlite_store/mod.rs
```diff
@@ -58,7 +63,10 @@ impl SqliteStore {
 		data_dir: PathBuf, db_file_name: Option<String>, kv_table_name: Option<String>,
 	) -> io::Result<Self> {
 		let inner = Arc::new(SqliteStoreInner::new(data_dir, db_file_name, kv_table_name)?);
-		let next_write_version = AtomicU64::new(1);
+
+		// Initialize next_write_version from the DB to ensure monotonicity across restarts.
+		let initial_version = inner.get_max_created_at()? + 1;
```
tnull: Ugh, really dislike mixing the concepts here, as the lock versioning is really only a crutch for the async-KVStore so far.

src/io/sqlite_store/mod.rs
```diff
@@ -289,7 +324,9 @@ impl SqliteStoreInner {
 			primary_namespace TEXT NOT NULL,
 			secondary_namespace TEXT DEFAULT \"\" NOT NULL,
 			key TEXT NOT NULL CHECK (key <> ''),
-			value BLOB, PRIMARY KEY ( primary_namespace, secondary_namespace, key )
+			value BLOB,
+			created_at INTEGER NOT NULL DEFAULT 0,
```
Camillarhi: Can we use SQLite's built-in `ROWID` instead of introducing `created_at`? The name is a bit misleading too since it's not actually a timestamp, it's just the write version counter.

src/io/sqlite_store/mod.rs
```diff
@@ -367,7 +425,9 @@ impl SqliteStoreInner {
 			let locked_conn = self.connection.lock().unwrap();
 
 			let sql = format!(
-				"INSERT OR REPLACE INTO {} (primary_namespace, secondary_namespace, key, value) VALUES (:primary_namespace, :secondary_namespace, :key, :value);",
+				"INSERT INTO {} (primary_namespace, secondary_namespace, key, value, created_at) \
+				 VALUES (:primary_namespace, :secondary_namespace, :key, :value, :created_at) \
+				 ON CONFLICT(primary_namespace, secondary_namespace, key) DO UPDATE SET value = excluded.value;",
```
Camillarhi: With `ON CONFLICT DO UPDATE`, updating a record won't move it in the list. Was insertion-order intentional, or do we want `last updated first` like the old `INSERT OR REPLACE` gave us?

src/io/sqlite_store/migrations.rs
benthecarman: When the v1->v2 migration added `secondary_namespace` but didn't update the primary key, so databases that started with v1 had PK (primary_namespace, key) instead of PK (primary_namespace, secondary_namespace, key). 

This issue didn't cause problems because we used `INSERT OR REPLACE` so we'd silently overwrite rows without errors (besides potentially losing data). However now we want to use `ON CONFLICT` so we don't replace the `created_at` and just update the value. This however now breaks for databases that started in v1 because we'll get conflicts even when the secondary namespace is different because the primary namespace and key are the primary key. So instead of silently replacing data, we now actually error because we're trying to do the `ON CONFLICT` updates

src/io/sqlite_store/mod.rs
benthecarman: We can rename it, but just using `ROWID` isn't safe because you can get repeats after deletes.

src/io/sqlite_store/mod.rs
benthecarman: Idk if this really matters? Only thing where i would see it is the payment store but we load all of those into a hashmap when initing. I could see us moving to the paginated version in the future anyways which will have its own sorting already.

benthecarman:
Gave better explanation around the v2 migration bug and why v3 is the way it is. Also changed `created_at` to just use the current timestamp rather than reusing the write version counter.

src/io/sqlite_store/migrations.rs
tnull: Hmm, right. But can we check whether we already have the right primary key, before we initiate a potentially costly/laggy migration? Node that still have the v1 schema would need to have been upgraded since LDK Node v0.2. I doubt there would be (m)any nodes requiring this migration at all, given this is such an early version that almost nobody ran in production.

AFAICT you should be able to use ` PRAGMA table_info(table_name)` and check the `pk` column.

benthecarman: done!

PR #809, opened 2026-02-25
wallet: expose next unused address

jharveyb:
I had an issue funding a node due to the existing `get_new_address` behavior:

- Call `get_new_address()` on a timer / loop when reporting the status of the node.
- After some # of calls, send funds to the returned address.
- On-chain balance isn't detected by the node.

IIUC, this is due to the gap limit here: https://github.com/lightningdevkit/ldk-node/blob/7ad0d630a4cd22c5527eadda843e8b80155daac1/src/chain/electrum.rs#L471

After manually recovering the funds with Sparrow and sending them to the first unused address, funds were picked up successfully.

The `next_unused_address` behavior seems much less surprising so I've added it here + used it successfully with my nodes. Even if it's not recommended in the doc examples for creating a channel, a doc note similar to what's on the `bdk_wallet` page may be useful.

### DIFF
--- a/src/payment/onchain.rs
+++ b/src/payment/onchain.rs
@@ -64,6 +64,13 @@ impl OnchainPayment {
 		Ok(funding_address)
 	}
 
+	/// Retrieve the next unused on-chain/funding address.
+	pub fn next_address(&self) -> Result<Address, Error> {
+		let funding_address = self.wallet.get_next_address()?;
+		log_info!(self.logger, "Generated next funding address: {}", funding_address);
+		Ok(funding_address)
+	}
+
 	/// Send an on-chain payment to the given address.
 	///
 	/// This will respect any on-chain reserve we need to keep, i.e., won't allow to cut into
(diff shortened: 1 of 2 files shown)


src/payment/onchain.rs
```diff
@@ -64,6 +64,13 @@ impl OnchainPayment {
 		Ok(funding_address)
 	}
 
+	/// Retrieve the next unused on-chain/funding address.
+	pub fn next_address(&self) -> Result<Address, Error> {
```
tnull, line 68: Exposing this API will be super confusing, IMO, as it will be entirely unclear which variant to use. And we intentionally only expose APIs that advance the descriptor index to avoid address reuse.

tnull:
> Ok - this was using ldk-node v0.2.0

Oh wow, that's ancient :grin: 

>  I can try and validate this behavior though.

Yeah feel free if you think that it still persists, my guess however is that it's related to other bugs we since fixed.

jharveyb:
> > Ok - this was using ldk-node v0.2.0
> 
> Oh wow, that's ancient 😁

Ah I was mixing up version numbers - it was my fork based on 0.7.0,  with my fork of rust-lightning based on 0.2.0.

But it seems like some sync / ChainSource fixes have landed in rust-lightning since then, not sure if they would affect this situation.

Issue #305, opened 2024-06-07
Expose a way to access channel or reserve type


tnull:
In #141 there [was discussion around](https://github.com/lightningdevkit/ldk-node/pull/141#discussion_r1627959567) whether and how to best expose the type (Anchors or not) of a channel and the reserve we keep (upfront/legacy, adaptive anchors, no reserve anchors) in the API. For now, we dropped the previously included `ChannelType`, but should revisit this question again eventually.

PR #810, opened 2026-02-26
Expose node-invariant features via NodeStatus

enigbe:
## What this PR does
For users interested in features announced in `node_announcement`, we make node_features available via `Node::status()`. 

The plan is to modify `ldk-server`'s `handle_get_node_info_request` and make features available on `GetNodeInfoResponse` so that the node's supported features are available to any requesting client. 

`lnrpc` and `cln-grpc` make these fields available in the response to `get_node_info`.

This is in relation to an exploratory `sim-ln` [PR](https://github.com/enigbe/sim-ln/pull/1) seeking to support/integrate `ldk-node/server`.

### DIFF
--- a/src/ffi/types.rs
+++ b/src/ffi/types.rs
@@ -44,6 +44,7 @@ pub use lightning_liquidity::lsps0::ser::LSPSDateTime;
 pub use lightning_liquidity::lsps1::msgs::{
 	LSPS1ChannelInfo, LSPS1OrderId, LSPS1OrderParams, LSPS1PaymentState,
 };
+use lightning_types::features::NodeFeatures as LdkNodeFeatures;
 pub use lightning_types::payment::{PaymentHash, PaymentPreimage, PaymentSecret};
 pub use lightning_types::string::UntrustedString;
 use vss_client::headers::{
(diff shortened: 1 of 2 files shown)


tnull (commented):
I'm not convinced we should expose this as freestanding APIs. Note that for channels and invoices these might also might be confusing as people really would be interested in the *negotiated* features, not what we generally, maybe support.

I guess we could consider to find a way to expose the invariant node features via an API (or maybe `Node::status`) if that would be useful, but for channels we should do https://github.com/lightningdevkit/ldk-node/issues/305 and maybe expose features in a sane way through `list_channels`. But note all of this should also work in bindings, so reusing LDK types might be challenging.

enigbe:
> I guess we could consider to find a way to expose the invariant node features via an API (or maybe Node::status) if that would be useful

For node features, I've decided to expose via `NodeStatus` and done some cleanup of the original FFI-exposed type. The PR ready for another review round.

src/ffi/types.rs
```diff
@@ -1519,6 +1520,301 @@ pub enum ClosureReason {
...
+	pub fn from_bytes(bytes: &[u8]) -> Self {
+		Self { inner: LdkNodeFeatures::from_be_bytes(bytes.to_vec()) }
+	}
+
+	/// Returns the BOLT 9 big-endian encoded representation of these features.
+	pub fn to_bytes(&self) -> Vec<u8> {
+		self.inner.encode()
+	}
+
+	/// Whether this node's `node_announcement` advertises support for `option_data_loss_protect` (bit 1).
```
tnull: Hmm, maybe we should drop mentioning the bit explicitly. For some features they might change over time in particular (e.g. when new features start out as experimental before finalization), and then we'll risk our bindings docs being off (which will be hard to catch)

src/lib.rs
```diff
@@ -2053,6 +2063,28 @@ impl Node {
...
+			},
+			lightning_background_processor::GossipSync::Rapid(_) => LdkNodeFeatures::empty(),
+			lightning_background_processor::GossipSync::None => {
+				unreachable!("We must always have a gossip sync!")
+			},
+		};
+		self.channel_manager.node_features()
+			| self.chain_monitor.provided_node_features()
+			| self.onion_messenger.provided_node_features()
+			| gossip_features
```
tnull, line 2081: Is this redundant to the features in channel manager/monitor etc?

tnull (commented):
Looks good I think, please squash.

src/lib.rs
enigbe: Yes, it is redundant in the sense that we reconstruct it from the underlying LDK components, but not redundant as an API. The idea with `NodeStatus::node_features` is to expose the aggregate node-level feature set that ldk-node advertises in `node_announcement`. LDK computes that aggregate internally in `PeerManager::broadcast_node_announcement`, but `PeerManager` does not expose the already-computed value, so we reconstruct it here.

tnull (approved):

PR #812, opened 2026-02-27
Upgrade `uniffi` dependency to v0.29.5 switch to proc macros (mostly)

tnull:
We here upgrade to `uniffi` v0.29.5. This upgrade requires a few changes and workarounds, in particular `UniffiCustomTypeConverter` are gone and replaced with corresponding macros. (see [`uniffi-rs` CHANGELOG](https://github.com/mozilla/uniffi-rs/blob/main/CHANGELOG.md#v0290-backend-crates-v0290---2025-02-06)). We also had to find workarounds for remote enum interface definitions, and discovered that `uniffi` lost the ability to use remote trait interfaces (see https://github.com/mozilla/uniffi-rs/issues/2823).

The latter therefore requires us to add some duplicative code as we added a local copy of `VssHeaderProvider`/`VssHeaderProviderError` and used an adaptor to make it work for the bindings side of the builder.

Since we've now already utilized the newer proc macro approach to make this work, we also included a list of commits that switch over some types from the manual definitions in the UDL  file to have the bindings auto-derived via the proc macros, which will reduce maintenance overhead going forward. It also means that we will finally get docstrings working in the target language, which is another big win.

Changeset looks large, but most of it are code moves (especially in the last commit) as we split some of the impl blocks into parts that are exposed and parts that are not exposed via `uniffi`-procmacros.

Best reviewed with:
```
git diff --diff-algorithm=patience --color-moved --color-moved-ws=ignore-all-space
```
(cc @vincenzopalazzo)

### DIFF
--- a/bindings/ldk_node.udl
+++ b/bindings/ldk_node.udl
@@ -5,21 +5,11 @@ namespace ldk_node {
 
 typedef dictionary Config;
 
-typedef dictionary AnchorChannelsConfig;
-
-typedef dictionary BackgroundSyncConfig;
-
-typedef dictionary SyncTimeoutsConfig;
-
 typedef dictionary EsploraSyncConfig;
 
 typedef dictionary ElectrumSyncConfig;
 
-typedef dictionary LSPS2ServiceConfig;
-
 typedef interface NodeEntropy;
-
-typedef enum EntropyError;
 
 typedef enum WordCount;
 
@@ -135,8 +125,6 @@ interface Node {
 	[Throws=NodeError]
 	bytes export_pathfinding_scores();
 };
-
-typedef enum Bolt11InvoiceDescription;
 
 typedef interface Bolt11Payment;
 
@@ -254,18 +242,6 @@ enum PaymentFailureReason {
 	"BlindedPathCreationFailed",
 };
 
-typedef enum PaymentKind;
-
-typedef enum UnifiedPaymentResult;
-
-typedef enum PaymentDirection;
-
-typedef enum PaymentStatus;
-
-typedef dictionary LSPFeeLimits;
-
-typedef enum ConfirmationStatus;
-
 typedef dictionary PaymentDetails;
 
 [Remote]
@@ -275,8 +251,6 @@ dictionary RouteParametersConfig {
 	u8 max_path_count;
 	u8 max_channel_saturation_power_of_half;
 };
-
-typedef dictionary CustomTlvRecord;
 
 [Remote]
 dictionary LSPS1OrderStatus {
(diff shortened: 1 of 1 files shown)


src/ffi/types.rs
```diff
@@ -41,7 +41,101 @@ pub use lightning_liquidity::lsps1::msgs::{
...
 pub use lightning_types::string::UntrustedString;
-pub use vss_client::headers::{VssHeaderProvider, VssHeaderProviderError};
+use std::collections::HashMap;
+
+use vss_client::headers::VssHeaderProvider as VssClientHeaderProvider;
+use vss_client::headers::VssHeaderProviderError as VssClientHeaderProviderError;
+
+/// Errors around providing headers for each VSS request.
+#[derive(Debug)]
+pub enum VssHeaderProviderError {
```
jkczyz, line 51: May want to leave a comment on why these are duplicated or if they need to be kept in sync. Or would we just get a compilation error? Likewise for other enums.

src/ffi/types.rs
```diff
@@ -41,7 +41,101 @@ pub use lightning_liquidity::lsps1::msgs::{
...
+			},
+			VssHeaderProviderError::AuthorizationError { error } => {
+				VssClientHeaderProviderError::AuthorizationError { error }
+			},
+			VssHeaderProviderError::InternalError { error } => {
+				VssClientHeaderProviderError::InternalError { error }
+			},
+		}
+	}
+}
```
jkczyz, line 108: I take it wrapping the underlying enum and providing `From` doesn't really help clients needing to use a variant?

src/payment/bolt11.rs
```diff
@@ -478,64 +625,6 @@ impl Bolt11Payment {
 		Ok(maybe_wrap(invoice))
 	}
 
-	pub(crate) fn receive_inner(
```
jkczyz, line 481: Could just use multiple impl blocks to avoid the code moves, fwiw.

tnull: True, though that might be more confusing w.r.t. what's exposed and what not going forward.

src/ffi/types.rs
tnull: Doc changes won't be directly detected, but actual signature changes would due to the match statements below.

src/ffi/types.rs
tnull: Ah, yes. I initially also attempted to just keep using the upstream/remote enum, but couldn't make it work for now. I hope we can drop all of the copied code as soon as https://github.com/mozilla/uniffi-rs/issues/2823 is resolved.

Issue #765, opened 2026-01-22
Background probing service


tnull:
We should allow LDK Node (and then eventually Server) to run in a background probing service mode, i.e., have a background task running that periodically probes paths with certain characteristics. To that end, users should be able to configure:

- Probing strategy (probably choosing from some pre-defined strategies, possibly also a custom strategy)
- How much of the local liquidity should be used for probing at max
- How often probes should be sent / how often old data should be refreshed

Relevant discussion: https://github.com/lightningdevkit/ldk-node/pull/678 was a first shot at this but didn't get very far.

adamjonas:
My two cents --- This is open source so I don't think we need to reserve anything. Best code wins!

tnull:
Done in #815.

PR #815, opened 2026-03-01
Add probing service

randomlogin:
Added a probing service which is used to send probes to estimate channels' capacities. 

Related issue: https://github.com/lightningdevkit/ldk-node/issues/765.

Probing is intended to be used in two ways: 
 - on a 'normal' node that allocates some liquidity to probe channels; 
 - on a probing node which doesn't make any real payments, just observes the network.

\
For probing a new abstraction `Prober` is defined and is (optionally) created during node building.
Prober periodically sends probes to feed the data to the scorer.
Prober sends probes using a ProbingStrategy.

ProbingStrategy trait has only one method: ` fn next_probe(&self) -> Option<Probe>`; every tick it generates a probe,  where `Probe` represents how to send a probe. 

To accommodate two different ways the probing is used, we either construct a probing route manually (`Probe::PrebuiltRoute`) or rely on the router/scorer (`Probe::Destination`).

Prober tracks how much liquidity is locked in-flight in probes, prevents the new probes from firing if the cap is reached. 

There are two probing strategies implemented: 

- Random probing strategy, it picks a random route from the current node, the route is probed via `send_probe`, thus ignores scoring parameters (what hops to pick), it also ignores `liquidity_limit_multiplier` which prohibits taking a hop if its capacity is too small. It is a true random route.

- High degree probing strategy, it examines the graph and finds the nodes with the biggest number of (public) channels and probes routes to them using `send_spontaneous_preflight_probes` which uses the current router/scorer. 

The former is meant to be used on payment nodes, while the latter on probing nodes. For the HighDegreeStrategy to work it is recommended to set `probing_diversity_penalty_msat` to some nonzero value to prevent routes reuse, however it may fail to find any available routes.

There are three tests added:
- check the probing locked amount increases/decreases
- check that the new probes are not fired if the current locked amount cap is reached
- performance testing which sets up a network of nodes and 4 observing nodes

Example output (runs for ~1 minute, needs `--nocapture` flag):

```
SCID            Direction         Probing Random               Probing HiDeg                Probing HiDeg+P              Probing nostrat              Real outbound msat
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------
114349209354241 C→Random          [0, 90 100 235]              unknown                      unknown                      unknown                      0
114349209354241 Random→C          [9 899 765, 100 000 000]     unknown                      unknown                      unknown                      917 184 187
114349209419777 D→HiDeg+P         unknown                      unknown                      [0, 91 127 725]              unknown                      0
114349209419777 HiDeg+P→D         unknown                      unknown                      [8 872 275, 100 000 000]     unknown                      905 313 697
114349209485312 B→F               [0, 3 418 855]               unknown                      unknown                      unknown                      160 288 291
114349209485312 F→B               [96 581 145, 100 000 000]    unknown                      unknown                      unknown                      819 711 709
114349209550848 B→D               [0, 6 141 420]               [0, 2 988 399]               [0, 91 127 725]              unknown                      30 052 300
114349209550848 D→B               [93 858 580, 100 000 000]    [97 011 601, 100 000 000]    [8 872 275, 100 000 000]     unknown                      874 555 512
114349209616384 E→HiDeg           [0, 2 097 950]               [0, 90 455 967]              unknown                      unknown                      0
114349209616384 HiDeg→E           [97 902 050, 100 000 000]    [9 544 033, 100 000 000]     unknown                      unknown                      913 366 836
114349209681920 B→C               [98 411 698, 100 000 000]    unknown                      unknown                      unknown                      86 511 108
114349209681920 C→B               [0, 1 588 302]               unknown                      unknown                      unknown                      812 521 355
114349209747457 B→E               [0, 6 625 595]               [0, 90 455 967]              [0, 4 795 311]               unknown                      0
114349209747457 E→B               [93 374 405, 100 000 000]    [9 544 033, 100 000 000]     [95 204 689, 100 000 000]    unknown                      907 339 924
114349209812993 C→nostrat         [0, 1 113 591]               unknown                      unknown                      unknown                      0
114349209812993 nostrat→C         [98 886 409, 100 000 000]    unknown                      unknown                      unknown                      990 000 000
114349209878528 C→E               [2 097 950, 5 335 740]       [0, 90 808 929]              unknown                      unknown                      55 569 306
114349209878528 E→C               [94 664 260, 97 902 050]     [9 191 071, 100 000 000]     unknown                      unknown                      878 423 826
114349209944065 B→C               [97 243 524, 100 000 000]    unknown                      [1 698 550, 6 086 572]       unknown                      938 012 264
114349209944065 C→B               [0, 2 756 476]               unknown                      [93 913 428, 98 301 450]     unknown                      34 367 637
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Known directions                   18/20                        8/20                         8/20                         0/20                        

```

For performance testing I had to expose the scoring data (`scorer_channel_liquidity`).
Also exposed `scoring_fee_params: ProbabilisticScoringFeeParameters` to `Config`.

TODOs: 
- adjust default parameters
- improve HighDegree strategy to take into account channel capacities
- improve HighDegree strategy to cache results
- improve performance test to better estimate real channel capacity

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -50,6 +50,7 @@ use crate::config::{
 	default_user_config, may_announce_channel, AnnounceError, AsyncPaymentsRole,
 	BitcoindRestClientConfig, Config, ElectrumSyncConfig, EsploraSyncConfig, HRNResolverConfig,
 	TorConfig, DEFAULT_ESPLORA_SERVER_URL, DEFAULT_LOG_FILENAME, DEFAULT_LOG_LEVEL,
+	DEFAULT_MAX_PROBE_AMOUNT_MSAT, DEFAULT_MIN_PROBE_AMOUNT_MSAT,
 };
 use crate::connection::ConnectionManager;
 use crate::entropy::NodeEntropy;
@@ -74,6 +75,10 @@ use crate::logger::{log_error, LdkLogger, LogLevel, LogWriter, Logger};
 use crate::message_handler::NodeCustomMessageHandler;
 use crate::payment::asynchronous::om_mailbox::OnionMessageMailbox;
 use crate::peer_store::PeerStore;
+use crate::probing::{
+	HighDegreeStrategy, Prober, ProbingConfig, ProbingStrategy, ProbingStrategyKind,
+	RandomWalkStrategy,
+};
 use crate::runtime::{Runtime, RuntimeSpawner};
 use crate::tx_broadcaster::TransactionBroadcaster;
 use crate::types::{
@@ -306,6 +311,7 @@ pub struct NodeBuilder {
 	async_payments_role: Option<AsyncPaymentsRole>,
 	runtime_handle: Option<tokio::runtime::Handle>,
 	pathfinding_scores_sync_config: Option<PathfindingScoresSyncConfig>,
+	probing_config: Option<ProbingConfig>,
 }
 
 impl NodeBuilder {
@@ -323,6 +329,7 @@ impl NodeBuilder {
 		let log_writer_config = None;
 		let runtime_handle = None;
 		let pathfinding_scores_sync_config = None;
+		let probing_config = None;
 		Self {
 			config,
 			chain_data_source_config,
@@ -332,6 +339,7 @@ impl NodeBuilder {
 			runtime_handle,
 			async_payments_role: None,
 			pathfinding_scores_sync_config,
+			probing_config,
 		}
 	}
 
@@ -629,6 +637,31 @@ impl NodeBuilder {
 		Ok(self)
 	}
 
+	/// Sets background probing config.
+	///
+	/// Use [`ProbingConfigBuilder`] to build the configuration:
+	/// ```no_run
+	/// # #[cfg(not(feature = "uniffi"))]
+	/// # {
+	/// use std::time::Duration;
+	/// use ldk_node::Builder;
+	/// use ldk_node::probing::ProbingConfigBuilder;
+	///
+	/// let mut builder = Builder::new();
+	/// builder.set_probing_config(
+	///     ProbingConfigBuilder::high_degree(100)
+	///         .interval(Duration::from_secs(30))
+	///         .build()
+	/// );
+	/// # }
+	/// ```
+	///
+	/// [`ProbingConfigBuilder`]: crate::probing::ProbingConfigBuilder
+	pub fn set_probing_config(&mut self, config: ProbingConfig) -> &mut Self {
+		self.probing_config = Some(config);
+		self
+	}
+
 	/// Builds a [`Node`] instance with a [`SqliteStore`] backend and according to the options
 	/// previously configured.
 	pub fn build(&self, node_entropy: NodeEntropy) -> Result<Node, BuildError> {
@@ -868,6 +901,7 @@ impl NodeBuilder {
 			self.gossip_source_config.as_ref(),
 			self.liquidity_source_config.as_ref(),
 			self.pathfinding_scores_sync_config.as_ref(),
+			self.probing_config.as_ref(),
 			self.async_payments_role,
 			seed_bytes,
 			runtime,
@@ -1166,6 +1200,15 @@ impl ArcedNodeBuilder {
 		self.inner.write().expect("lock").set_async_payments_role(role).map(|_| ())
 	}
 
+	/// Configures background probing.
+	///
+	/// Use [`ProbingConfigBuilder`] to build the configuration.
+	///
+	/// [`ProbingConfigBuilder`]: crate::probing::ProbingConfigBuilder
+	pub fn set_probing_config(&self, config: Arc<ProbingConfig>) {
+		self.inner.write().expect("lock").set_probing_config((*config).clone());
+	}
+
 	/// Builds a [`Node`] instance with a [`SqliteStore`] backend and according to the options
 	/// previously configured.
 	pub fn build(&self, node_entropy: Arc<NodeEntropy>) -> Result<Arc<Node>, BuildError> {
@@ -1361,8 +1404,8 @@ fn build_with_store_internal(
 	gossip_source_config: Option<&GossipSourceConfig>,
 	liquidity_source_config: Option<&LiquiditySourceConfig>,
 	pathfinding_scores_sync_config: Option<&PathfindingScoresSyncConfig>,
-	async_payments_role: Option<AsyncPaymentsRole>, seed_bytes: [u8; 64], runtime: Arc<Runtime>,
-	logger: Arc<Logger>, kv_store: Arc<DynStore>,
+	probing_config: Option<&ProbingConfig>, async_payments_role: Option<AsyncPaymentsRole>,
+	seed_bytes: [u8; 64], runtime: Arc<Runtime>, logger: Arc<Logger>, kv_store: Arc<DynStore>,
 ) -> Result<Node, BuildError> {
 	optionally_install_rustls_cryptoprovider();
 
@@ -2219,6 +2262,51 @@ fn build_with_store_internal(
 		_leak_checker.0.push(Arc::downgrade(&wallet) as Weak<dyn Any + Send + Sync>);
 	}
 
+	let prober = probing_config.map(|probing_cfg| {
+		let strategy: Arc<dyn ProbingStrategy> = match &probing_cfg.kind {
+			ProbingStrategyKind::HighDegree { top_node_count } => {
+				// Dedicated router for probing so the diversity penalty doesn't interfere
+				// with real payments; shares the scorer so probe results still train it.
+				let mut probing_fee_params = ProbabilisticScoringFeeParameters::default();
+				if let Some(penalty) = probing_cfg.diversity_penalty_msat {
+					probing_fee_params.probing_diversity_penalty_msat = penalty;
+				}
+				let probing_router = Arc::new(DefaultRouter::new(
+					Arc::clone(&network_graph),
+					Arc::clone(&logger),
+					Arc::clone(&keys_manager),
+					Arc::clone(&scorer),
+					probing_fee_params,
+				));
+				Arc::new(HighDegreeStrategy::new(
+					Arc::clone(&network_graph),
+					Arc::clone(&channel_manager),
+					probing_router,
+					*top_node_count,
+					DEFAULT_MIN_PROBE_AMOUNT_MSAT,
+					DEFAULT_MAX_PROBE_AMOUNT_MSAT,
+					probing_cfg.cooldown,
+					config.probing_liquidity_limit_multiplier,
+				))
+			},
+			ProbingStrategyKind::RandomWalk { max_hops } => Arc::new(RandomWalkStrategy::new(
+				Arc::clone(&network_graph),
+				Arc::clone(&channel_manager),
+				*max_hops,
+				DEFAULT_MIN_PROBE_AMOUNT_MSAT,
+				DEFAULT_MAX_PROBE_AMOUNT_MSAT,
+			)),
+			ProbingStrategyKind::Custom(s) => Arc::clone(s),
+		};
+		Arc::new(Prober {
+			channel_manager: Arc::clone(&channel_manager),
+			logger: Arc::clone(&logger),
+			strategy,
+			interval: probing_cfg.interval,
+			max_locked_msat: probing_cfg.max_locked_msat,
+		})
+	});
+
 	Ok(Node {
 		runtime,
 		stop_sender,
@@ -2252,6 +2340,7 @@ fn build_with_store_internal(
 		om_mailbox,
 		async_payments_role,
 		hrn_resolver,
+		prober,
 		#[cfg(cycle_tests)]
 		_leak_checker,
 	})
--- a/src/config.rs
+++ b/src/config.rs
@@ -28,6 +28,12 @@ const DEFAULT_BDK_WALLET_SYNC_INTERVAL_SECS: u64 = 80;
 const DEFAULT_LDK_WALLET_SYNC_INTERVAL_SECS: u64 = 30;
 const DEFAULT_FEE_RATE_CACHE_UPDATE_INTERVAL_SECS: u64 = 60 * 10;
 const DEFAULT_PROBING_LIQUIDITY_LIMIT_MULTIPLIER: u64 = 3;
+pub(crate) const DEFAULT_PROBING_INTERVAL_SECS: u64 = 10;
+pub(crate) const MIN_PROBING_INTERVAL: Duration = Duration::from_millis(100);
+pub(crate) const DEFAULT_PROBED_NODE_COOLDOWN_SECS: u64 = 60 * 60; // 1 hour
+pub(crate) const DEFAULT_MAX_PROBE_LOCKED_MSAT: u64 = 100_000_000; // 100k sats
+pub(crate) const DEFAULT_MIN_PROBE_AMOUNT_MSAT: u64 = 1_000_000; // 1k sats
+pub(crate) const DEFAULT_MAX_PROBE_AMOUNT_MSAT: u64 = 10_000_000; // 10k sats
 const DEFAULT_ANCHOR_PER_CHANNEL_RESERVE_SATS: u64 = 25_000;
 
 // The default timeout after which we abort a wallet syncing operation.
--- a/src/event.rs
+++ b/src/event.rs
@@ -52,6 +52,7 @@ use crate::payment::store::{
 	PaymentDetails, PaymentDetailsUpdate, PaymentDirection, PaymentKind, PaymentStatus,
 };
 use crate::payment::PaymentMetadata;
+use crate::probing::Prober;
 use crate::runtime::Runtime;
 use crate::types::{
 	CustomTlvRecord, DynStore, KeysManager, OnionMessenger, PaymentStore, Sweeper, Wallet,
@@ -536,12 +537,13 @@ where
 	payment_store: Arc<PaymentStore>,
 	peer_store: Arc<PeerStore<L>>,
 	keys_manager: Arc<KeysManager>,
-	runtime: Arc<Runtime>,
-	logger: L,
-	config: Arc<Config>,
 	static_invoice_store: Option<StaticInvoiceStore>,
 	onion_messenger: Arc<OnionMessenger>,
 	om_mailbox: Option<Arc<OnionMessageMailbox>>,
+	prober: Option<Arc<Prober>>,
+	runtime: Arc<Runtime>,
+	logger: L,
+	config: Arc<Config>,
 }
 
 impl<L: Deref + Clone + Sync + Send + 'static> EventHandler<L>
@@ -556,8 +558,8 @@ where
 		liquidity_source: Arc<LiquiditySource<Arc<Logger>>>, payment_store: Arc<PaymentStore>,
 		peer_store: Arc<PeerStore<L>>, keys_manager: Arc<KeysManager>,
 		static_invoice_store: Option<StaticInvoiceStore>, onion_messenger: Arc<OnionMessenger>,
-		om_mailbox: Option<Arc<OnionMessageMailbox>>, runtime: Arc<Runtime>, logger: L,
-		config: Arc<Config>,
+		om_mailbox: Option<Arc<OnionMessageMailbox>>, prober: Option<Arc<Prober>>,
+		runtime: Arc<Runtime>, logger: L, config: Arc<Config>,
 	) -> Self {
 		Self {
 			event_queue,
@@ -571,12 +573,13 @@ where
 			payment_store,
 			peer_store,
 			keys_manager,
-			logger,
-			runtime,
-			config,
 			static_invoice_store,
 			onion_messenger,
 			om_mailbox,
+			prober,
+			runtime,
+			logger,
+			config,
 		}
 	}
 
@@ -1208,8 +1211,16 @@ where
 
 			LdkEvent::PaymentPathSuccessful { .. } => {},
 			LdkEvent::PaymentPathFailed { .. } => {},
-			LdkEvent::ProbeSuccessful { .. } => {},
-			LdkEvent::ProbeFailed { .. } => {},
+			LdkEvent::ProbeSuccessful { path, payment_id, .. } => {
+				if let Some(prober) = &self.prober {
+					prober.handle_background_probe_successful(&path, payment_id);
+				}
+			},
+			LdkEvent::ProbeFailed { path, payment_id, .. } => {
+				if let Some(prober) = &self.prober {
+					prober.handle_background_probe_failed(&path, payment_id);
+				}
+			},
 			LdkEvent::HTLCHandlingFailed { failure_type, .. } => {
 				self.liquidity_source
 					.lsps2_service()
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -101,10 +101,12 @@ pub mod logger;
 mod message_handler;
 pub mod payment;
 mod peer_store;
+pub mod probing;
 mod runtime;
 mod scoring;
 mod tx_broadcaster;
 mod types;
+mod util;
 mod wallet;
 
 use std::default::Default;
@@ -172,6 +174,9 @@ use payment::{
 	UnifiedPayment,
 };
 use peer_store::{PeerInfo, PeerStore};
+#[cfg(feature = "uniffi")]
+pub use probing::ArcedProbingConfigBuilder as ProbingConfigBuilder;
+use probing::{run_prober, Prober};
 use runtime::Runtime;
 pub use tokio;
 use types::{
@@ -250,6 +255,7 @@ pub struct Node {
 	om_mailbox: Option<Arc<OnionMessageMailbox>>,
 	async_payments_role: Option<AsyncPaymentsRole>,
 	hrn_resolver: HRNResolver,
+	prober: Option<Arc<Prober>>,
 	#[cfg(cycle_tests)]
 	_leak_checker: LeakChecker,
 }
@@ -610,10 +616,18 @@ impl Node {
 			static_invoice_store,
 			Arc::clone(&self.onion_messenger),
 			self.om_mailbox.clone(),
+			self.prober.clone(),
 			Arc::clone(&self.runtime),
 			Arc::clone(&self.logger),
 			Arc::clone(&self.config),
 		));
+
+		if let Some(prober) = self.prober.clone() {
+			let stop_rx = self.stop_sender.subscribe();
+			self.runtime.spawn_cancellable_background_task(async move {
+				run_prober(prober, stop_rx).await;
+			});
+		}
 
 		// Setup background processing
 		let background_persister = Arc::clone(&self.kv_store);
@@ -1143,6 +1157,11 @@ impl Node {
 			Arc::clone(&self.liquidity_source),
 			Arc::clone(&self.logger),
 		))
+	}
+
+	/// Returns a reference to the [`Prober`], or `None` if no probing strategy is configured.
+	pub fn prober(&self) -> Option<&Prober> {
+		self.prober.as_deref()
 	}
 
 	/// Retrieve a list of known channels.
--- a/src/probing.rs
+++ b/src/probing.rs
@@ -1,0 +1,834 @@
+// This file is Copyright its original authors, visible in version control history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
+// accordance with one or both of these licenses.
+
+//! Background probing for training the payment scorer.
+//!
+//! Lightning Network nodes only know channels' capacities via their initially announced limits;
+//! the real values change unpredictably after payments have been sent, which makes some of
+//! the channels inoperable (capacity has been depleted). The only way to know about channel
+//! depletion is to attempt sending a payment through it. Thus, sending a live payment
+//! might involve a significant time delay for finding an appropriate channel with enough capacity,
+//! up to complete failure when a route with enough capacity cannot be found.
+//!
+//! The background probing service fires probes to learn about the live state of channels and
+//! their capacities, providing accurate data to the scorer and router.
+//!
+//! This module provides the configuration for such a service. There are two pre-built strategies,
+//! [`RandomWalkStrategy`] and [`HighDegreeStrategy`], as well as a [`ProbingStrategy`] trait which
+//! allows defining a custom probing strategy (for example if there is an established payment
+//! pattern).
+//!
+//! # Configuration
+//!
+//! Probing is opt-in: a node only runs the service if a [`ProbingConfig`] has been registered
+//! on the [`Builder`] via [`Builder::set_probing_config`] before [`Builder::build`]. Without a
+//! config, no probes are sent.
+//!
+//! # Example
+//!
+//! ```no_run
+//! # #[cfg(not(feature = "uniffi"))]
+//! # {
+//! use std::time::Duration;
+//! use ldk_node::Builder;
+//! use ldk_node::probing::ProbingConfigBuilder;
+//!
+//! let probing_config = ProbingConfigBuilder::high_degree(100)
+//!     .interval(Duration::from_secs(30))
+//!     .max_locked_msat(500_000)
+//!     .diversity_penalty_msat(250)
+//!     .build();
+//!
+//! let mut builder = Builder::new();
+//! builder.set_probing_config(probing_config);
+//! # }
+//! ```
+//!
+//! # Caution
+//!
+//! Probes send real HTLCs along real paths. If an intermediate hop is offline or
+//! misbehaving, the probe HTLC can remain in-flight — locking outbound liquidity
+//! on the first-hop channel until the HTLC timeout elapses (potentially hours).
+//! `max_locked_msat` caps the total outbound capacity that in-flight probes may
+//! hold at any one time; tune it conservatively for nodes with tight liquidity.
+//!
+//! [`Builder`]: crate::Builder
+//! [`Builder::set_probing_config`]: crate::Builder::set_probing_config
+//! [`Builder::build`]: crate::Builder::build
+
+use std::collections::HashMap;
+use std::fmt;
+#[cfg(feature = "uniffi")]
+use std::sync::RwLock;
+use std::sync::{Arc, Mutex};
+use std::time::{Duration, Instant};
+
+use bitcoin::secp256k1::PublicKey;
+use lightning::ln::channelmanager::{PaymentId, RecentPaymentDetails};
+use lightning::routing::gossip::NodeId;
+use lightning::routing::router::Router as LdkRouter;
+use lightning::routing::router::{
+	Path, PaymentParameters, RouteHop, RouteParameters, MAX_PATH_LENGTH_ESTIMATE,
+};
+use lightning_invoice::DEFAULT_MIN_FINAL_CLTV_EXPIRY_DELTA;
+use lightning_types::features::{ChannelFeatures, NodeFeatures};
+
+use crate::config::{
+	DEFAULT_MAX_PROBE_LOCKED_MSAT, DEFAULT_PROBED_NODE_COOLDOWN_SECS,
+	DEFAULT_PROBING_INTERVAL_SECS, MIN_PROBING_INTERVAL,
+};
+use crate::logger::{log_debug, LdkLogger, Logger};
+use crate::types::{ChannelManager, Graph, Router};
+use crate::util::random_range;
+
+/// Which built-in probing strategy to use, or a custom one.
+#[derive(Clone)]
+pub(crate) enum ProbingStrategyKind {
+	HighDegree { top_node_count: usize },
+	RandomWalk { max_hops: usize },
+	Custom(Arc<dyn ProbingStrategy>),
+}
+
+impl fmt::Debug for ProbingStrategyKind {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		match self {
+			Self::HighDegree { top_node_count } => {
+				f.debug_struct("HighDegree").field("top_node_count", top_node_count).finish()
+			},
+			Self::RandomWalk { max_hops } => {
+				f.debug_struct("RandomWalk").field("max_hops", max_hops).finish()
+			},
+			Self::Custom(_) => f.write_str("Custom(<probing strategy>)"),
+		}
+	}
+}
+
+/// Configuration for the background probing subsystem.
+///
+/// Instances are produced by [`ProbingConfigBuilder`], which exposes three strategy
+/// constructors: [`ProbingConfigBuilder::high_degree`], [`ProbingConfigBuilder::random_walk`],
+/// and [`ProbingConfigBuilder::custom`].
+///
+/// Optional setters on the builder tune timing and liquidity limits, and
+/// [`ProbingConfigBuilder::build`] finalizes the value.
+///
+/// # Examples
+///
+/// Using pre-built strategy:
+/// ```no_run
+/// # #[cfg(not(feature = "uniffi"))]
+/// # {
+/// use std::time::Duration;
+/// use ldk_node::Builder;
+/// use ldk_node::probing::ProbingConfigBuilder;
+///
+/// let config = ProbingConfigBuilder::high_degree(100)
+///     .interval(Duration::from_secs(30))
+///     .max_locked_msat(500_000)
+///     .diversity_penalty_msat(250)
+///     .build();
+///
+/// let mut builder = Builder::new();
+/// builder.set_probing_config(config);
+/// # }
+/// ```
+///
+/// Creating a custom strategy that always probes the same path:
+/// ```
+/// use ldk_node::lightning::routing::router::Path;
+/// use ldk_node::probing::ProbingStrategy;
+///
+/// struct FixedPathStrategy {
+///     path: Path,
+/// }
+/// impl ProbingStrategy for FixedPathStrategy {
+///     fn next_probe(&self) -> Option<Path> {
+///         if self.path.hops.len() > 1 {
+///             Some(self.path.clone())
+///         } else {
+///             None
+///         }
+///     }
+/// }
+/// ```
+#[derive(Clone, Debug)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Object))]
+pub struct ProbingConfig {
+	pub(crate) kind: ProbingStrategyKind,
+	pub(crate) interval: Duration,
+	pub(crate) max_locked_msat: u64,
+	pub(crate) diversity_penalty_msat: Option<u64>,
+	pub(crate) cooldown: Duration,
+}
+
+/// Builder for [`ProbingConfig`].
+///
+/// A new instance starts from one of three strategy constructors — [`high_degree`],
+/// [`random_walk`], or [`custom`] — and is finalized through [`build`]. Optional setters
+/// in between override the timing and liquidity defaults.
+///
+/// [`high_degree`]: Self::high_degree
+/// [`random_walk`]: Self::random_walk
+/// [`custom`]: Self::custom
+/// [`build`]: Self::build
+pub struct ProbingConfigBuilder {
+	kind: ProbingStrategyKind,
+	interval: Duration,
+	max_locked_msat: u64,
+	diversity_penalty_msat: Option<u64>,
+	cooldown: Duration,
+}
+
+impl ProbingConfigBuilder {
+	fn with_kind(kind: ProbingStrategyKind) -> Self {
+		Self {
+			kind,
+			interval: Duration::from_secs(DEFAULT_PROBING_INTERVAL_SECS),
+			max_locked_msat: DEFAULT_MAX_PROBE_LOCKED_MSAT,
+			diversity_penalty_msat: None,
+			cooldown: Duration::from_secs(DEFAULT_PROBED_NODE_COOLDOWN_SECS),
+		}
+	}
+
+	/// Start building a config that probes toward the highest-degree nodes in the graph.
+	///
+	/// `top_node_count` controls how many of the most-connected nodes are cycled through.
+	pub fn high_degree(top_node_count: usize) -> Self {
+		Self::with_kind(ProbingStrategyKind::HighDegree { top_node_count })
+	}
+
+	/// Start building a config that probes via random graph walks.
+	///
+	/// `max_hops` is the upper bound on the number of hops in a randomly constructed path.
+	/// Values below `2` are clamped to `2`.
+	pub fn random_walk(max_hops: usize) -> Self {
+		Self::with_kind(ProbingStrategyKind::RandomWalk { max_hops })
+	}
+
+	/// Start building a config with a custom [`ProbingStrategy`] implementation.
+	pub fn custom(strategy: Arc<dyn ProbingStrategy>) -> Self {
+		Self::with_kind(ProbingStrategyKind::Custom(strategy))
+	}
+
+	/// Overrides the interval between probe attempts.
+	///
+	/// Defaults to 10 seconds.
+	pub fn interval(&mut self, interval: Duration) -> &mut Self {
+		self.interval = interval;
+		self
+	}
+
+	/// Overrides the maximum millisatoshis that may be locked in in-flight probes at any time.
+	///
+	/// Defaults to 100 000 000 msat (100k sats).
+	pub fn max_locked_msat(&mut self, max_msat: u64) -> &mut Self {
+		self.max_locked_msat = max_msat;
+		self
+	}
+
+	/// Sets the probing diversity penalty applied by the probabilistic scorer.
+	///
+	/// When set, the scorer will penalize channels that have been recently probed,
+	/// encouraging path diversity during background probing. The penalty decays
+	/// quadratically over 24 hours.
+	///
+	/// This is only useful for probing strategies that route through the scorer
+	/// (e.g., [`HighDegreeStrategy`]). Strategies that build paths manually
+	/// (e.g., [`RandomWalkStrategy`]) bypass the scorer entirely.
+	///
+	/// If unset, LDK's default of `0` (no penalty) is used.
+	pub fn diversity_penalty_msat(&mut self, penalty_msat: u64) -> &mut Self {
+		self.diversity_penalty_msat = Some(penalty_msat);
+		self
+	}
+
+	/// Sets how long a probed node stays ineligible before being probed again.
+	///
+	/// Only applies to [`HighDegreeStrategy`]. Defaults to 1 hour.
+	pub fn cooldown(&mut self, cooldown: Duration) -> &mut Self {
+		self.cooldown = cooldown;
+		self
+	}
+
+	/// Builds the [`ProbingConfig`].
+	pub fn build(&self) -> ProbingConfig {
+		ProbingConfig {
+			kind: self.kind.clone(),
+			interval: self.interval.max(MIN_PROBING_INTERVAL),
+			max_locked_msat: self.max_locked_msat,
+			diversity_penalty_msat: self.diversity_penalty_msat,
+			cooldown: self.cooldown,
+		}
+	}
+}
+
+/// Builder for [`ProbingConfig`].
+///
+/// A new instance starts from one of two strategy constructors — [`high_degree`] or
+/// [`random_walk`] — and is finalized through [`build`]. Optional setters in between
+/// override the timing and liquidity defaults.
+///
+/// [`high_degree`]: Self::high_degree
+/// [`random_walk`]: Self::random_walk
+/// [`build`]: Self::build
+#[cfg(feature = "uniffi")]
+pub struct ArcedProbingConfigBuilder {
+	inner: RwLock<ProbingConfigBuilder>,
+}
+
+#[cfg(feature = "uniffi")]
+impl ArcedProbingConfigBuilder {
+	/// Start building a config that probes toward the highest-degree nodes in the graph.
+	///
+	/// `top_node_count` controls how many of the most-connected nodes are cycled through.
+	pub fn high_degree(top_node_count: u64) -> Self {
+		Self { inner: RwLock::new(ProbingConfigBuilder::high_degree(top_node_count as usize)) }
+	}
+
+	/// Start building a config that probes via random graph walks.
+	///
+	/// `max_hops` is the upper bound on the number of hops in a randomly constructed path.
+	/// Values below `2` are clamped to `2`.
+	pub fn random_walk(max_hops: u64) -> Self {
+		Self { inner: RwLock::new(ProbingConfigBuilder::random_walk(max_hops as usize)) }
+	}
+
+	/// Overrides the interval between probe attempts.
+	///
+	/// Defaults to 10 seconds.
+	pub fn set_interval(&self, secs: u64) {
+		self.inner.write().expect("lock").interval(Duration::from_secs(secs));
+	}
+
+	/// Overrides the maximum millisatoshis that may be locked in in-flight probes at any time.
+	///
+	/// Defaults to 100 000 000 msat (100k sats).
+	pub fn set_max_locked_msat(&self, max_msat: u64) {
+		self.inner.write().expect("lock").max_locked_msat(max_msat);
+	}
+
+	/// Sets the probing diversity penalty applied by the probabilistic scorer.
+	///
+	/// When set, the scorer will penalize channels that have been recently probed,
+	/// encouraging path diversity during background probing. The penalty decays
+	/// quadratically over 24 hours.
+	///
+	/// This is only useful for probing strategies that route through the scorer
+	/// (e.g., [`HighDegreeStrategy`]). Strategies that build paths manually
+	/// (e.g., [`RandomWalkStrategy`]) bypass the scorer entirely.
+	///
+	/// If unset, LDK's default of `0` (no penalty) is used.
+	pub fn set_diversity_penalty_msat(&self, penalty_msat: u64) {
+		self.inner.write().expect("lock").diversity_penalty_msat(penalty_msat);
+	}
+
+	/// Sets how long a probed node stays ineligible before being probed again.
+	///
+	/// Only applies to [`HighDegreeStrategy`]. Defaults to 1 hour.
+	pub fn set_cooldown(&self, secs: u64) {
+		self.inner.write().expect("lock").cooldown(Duration::from_secs(secs));
+	}
+
+	/// Builds the [`ProbingConfig`].
+	pub fn build(&self) -> Arc<ProbingConfig> {
+		Arc::new(self.inner.read().expect("lock").build())
+	}
+}
+
+/// A strategy that decides which path the probing service should probe next.
+pub trait ProbingStrategy: Send + Sync + 'static {
+	/// Returns the next probe path to run, or `None` to skip this tick.
+	fn next_probe(&self) -> Option<Path>;
+}
+
+/// Probes toward the most-connected nodes in the graph.
+///
+/// On each tick the strategy reads the current gossip graph, sorts nodes by
+/// channel count, and picks the highest-degree node from the top
+/// `top_node_count` that has not been probed within `cooldown`.
+/// Nodes probed more recently are skipped so that the strategy
+/// naturally spreads across the top nodes and picks up graph changes.
+/// If all top nodes are on cooldown, the cooldown map is cleared and a new cycle begins
+/// immediately.
+///
+/// The probe amount is chosen uniformly at random from
+/// `[min_amount_msat, max_amount_msat]`.
+///
+/// `HighDegreeStrategy` can only use publicly announced channels for probing.
+pub struct HighDegreeStrategy {
+	network_graph: Arc<Graph>,
+	channel_manager: Arc<ChannelManager>,
+	router: Arc<Router>,
+	/// How many of the highest-degree nodes to cycle through.
+	pub top_node_count: usize,
+	/// Lower bound for the randomly chosen probe amount.
+	pub min_amount_msat: u64,
+	/// Upper bound for the randomly chosen probe amount.
+	pub max_amount_msat: u64,
+	/// How long a node stays ineligible after being probed.
+	pub cooldown: Duration,
+	/// Skip a path when the first-hop outbound liquidity is less than
+	/// `path_value * liquidity_limit_multiplier`.
+	pub liquidity_limit_multiplier: u64,
+	/// Nodes probed recently, with the time they were last probed.
+	recently_probed: Mutex<HashMap<NodeId, Instant>>,
+}
+
+impl HighDegreeStrategy {
+	/// Creates a new high-degree probing strategy.
+	pub(crate) fn new(
+		network_graph: Arc<Graph>, channel_manager: Arc<ChannelManager>, router: Arc<Router>,
+		top_node_count: usize, min_amount_msat: u64, max_amount_msat: u64, cooldown: Duration,
+		liquidity_limit_multiplier: u64,
+	) -> Self {
+		assert!(
+			min_amount_msat <= max_amount_msat,
+			"min_amount_msat must not exceed max_amount_msat"
+		);
+		Self {
+			network_graph,
+			channel_manager,
+			router,
+			top_node_count,
+			min_amount_msat,
+			max_amount_msat,
+			cooldown,
+			liquidity_limit_multiplier,
+			recently_probed: Mutex::new(HashMap::new()),
+		}
+	}
+}
+
+impl ProbingStrategy for HighDegreeStrategy {
+	fn next_probe(&self) -> Option<Path> {
+		let graph = self.network_graph.read_only();
+
+		let mut nodes_by_degree: Vec<(NodeId, usize)> =
+			graph.nodes().unordered_iter().map(|(id, info)| (*id, info.channels.len())).collect();
+
+		if nodes_by_degree.is_empty() {
+			return None;
+		}
+
+		nodes_by_degree.sort_unstable_by(|a, b| b.1.cmp(&a.1));
+
+		let top_node_count = self.top_node_count.min(nodes_by_degree.len());
+		let now = Instant::now();
+
+		let mut probed = self.recently_probed.lock().unwrap_or_else(|e| e.into_inner());
+
+		// We could check staleness when we use the entry, but that way we'd not clear cache at
+		// all. For hundreds of top nodes it's okay to call retain each tick.
+		probed.retain(|_, probed_at| now.duration_since(*probed_at) < self.cooldown);
+
+		// If all top nodes are on cooldown, reset and start a new cycle.
+		let final_node_id = match nodes_by_degree[..top_node_count]
+			.iter()
+			.find(|(node_id, _)| !probed.contains_key(node_id))
+		{
+			Some((node_id, _)) => *node_id,
+			None => {
+				probed.clear();
+				nodes_by_degree[0].0
+			},
+		};
+
+		probed.insert(final_node_id, now);
+		drop(probed);
+		drop(graph);
+
+		let final_node = PublicKey::try_from(final_node_id).ok()?;
+
+		let amount_msat = random_range(self.min_amount_msat, self.max_amount_msat);
+		let payment_params =
+			PaymentParameters::from_node_id(final_node, DEFAULT_MIN_FINAL_CLTV_EXPIRY_DELTA as u32);
+		let route_params =
+			RouteParameters::from_payment_params_and_value(payment_params, amount_msat);
+
+		let payer = self.channel_manager.get_our_node_id();
+		let usable_channels = self.channel_manager.list_usable_channels();
+		let first_hops: Vec<&_> = usable_channels.iter().collect();
+		let inflight_htlcs = self.channel_manager.compute_inflight_htlcs();
+
+		let route = self
+			.router
+			.find_route(&payer, &route_params, Some(&first_hops), inflight_htlcs)
+			.ok()?;
+
+		let path = route.paths.into_iter().next()?;
+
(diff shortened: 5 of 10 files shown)


src/builder.rs
```diff
@@ -72,6 +74,8 @@ use crate::logger::{log_error, LdkLogger, LogLevel, LogWriter, Logger};
 use crate::message_handler::NodeCustomMessageHandler;
 use crate::payment::asynchronous::om_mailbox::OnionMessageMailbox;
 use crate::peer_store::PeerStore;
+use crate::probing;
+use crate::probing::ProbingStrategy;
```
enigbe: nit: `ProbingStrategy` is imported but never used because it qualified at all usage sites. You could either remove it and use the qualified form, or do the opposite. The rest of builder.rs follows the convention of importing types and using short names and I suggest doing the same here just for consistency.

src/probing.rs
```diff
@@ -0,0 +1,355 @@
...
+use lightning::routing::router::{Path, RouteHop, MAX_PATH_LENGTH_ESTIMATE};
+use lightning_types::features::NodeFeatures;
+
+use bitcoin::secp256k1::PublicKey;
+
+use rand::Rng;
+
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
```
enigbe: nit: Import ordering doesn't match the codebase's convention. The rest of the repo follows: std -> external deps -> crate (see e.g. builder.rs, event.rs) but here it's reversed. Please reorder to match.

src/probing.rs
```diff
@@ -0,0 +1,355 @@
...
+
+/// Probes toward the most-connected nodes in the graph.
+///
+/// Sorts all graph nodes by channel count descending, then cycles through the
+/// top-`top_n` entries using `Destination` so the router finds the actual path.
+/// The probe amount is chosen uniformly at random from `[min_amount_msat, max_amount_msat]`.
+pub struct HighDegreeStrategy {
+    network_graph: Arc<Graph>,
+    /// How many of the highest-degree nodes to cycle through.
+    pub top_n: usize,
```
enigbe: Could `top_n` be renamed to `num_top_nodes`? The latter reads less generic to me but up to you to modify or not.

src/probing.rs
```diff
@@ -0,0 +1,355 @@
...
+        Self { network_graph, top_n, min_amount_msat, max_amount_msat, cursor: Mutex::new(0) }
+    }
+}
+
+impl ProbingStrategy for HighDegreeStrategy {
+    fn next_probe(&self) -> Option<Probe> {
+        let graph = self.network_graph.read_only();
+
+        // Collect (pubkey, channel_count) for all nodes.
+        // wtf it does why we need to iterate here and then sort? maybe we can go just once?
```
enigbe: Here, and in other locations, we have what looks like leftover development notes. Could you fix-up its removal into this commit so it doesn't end up in git history?

src/builder.rs
```diff
@@ -273,6 +309,7 @@ impl NodeBuilder {
 			async_payments_role: None,
 			pathfinding_scores_sync_config,
 			recovery_mode,
+			probing_strategy: None,
```
enigbe: nit: Most optional fields here are initialized as local variables before the struct literal (e.g. let `chain_data_source_config` = `None`;). Could `probing_strategy `follow the same pattern for consistency? While at it, `async_payments_role` has the same inconsistency. Would be nice to align both.

src/builder.rs
```diff
@@ -557,6 +594,64 @@ impl NodeBuilder {
...
+		let kind = ProbingStrategyKind::Random { max_hops };
+		self.probing_strategy = Some(self.make_probing_config(kind));
+		self
+	}
+
+	/// Configures a custom probing strategy for background channel probing.
+	///
+	/// When set, the node will periodically call [`ProbingStrategy::next_probe`] and dispatch the
+	/// returned probe via the channel manager.
+	pub fn set_probing_strategy(
```
enigbe: Maybe `set_custom_probing_strategy`?

src/probing.rs
```diff
@@ -0,0 +1,355 @@
...
+                                        log_debug!(self.logger, "Prebuilt path probe failed: {:?}", e);
+                                    }
+                                }
+                            }
+                        }
+                        Some(Probe::Destination { final_node, amount_msat }) => {
+                            if self.locked_msat.load(Ordering::Relaxed) + amount_msat > self.max_locked_msat {
+                                log_debug!(self.logger, "Skipping probe: locked-msat budget exceeded.");
+                            } else {
+                                match self.channel_manager.send_spontaneous_preflight_probes(final_node, amount_msat, 18, self.liquidity_limit_multiplier) {
```
enigbe: Why 18?

src/probing.rs
```diff
@@ -0,0 +1,355 @@
+use crate::logger::{log_debug, LdkLogger, Logger};
```
enigbe: Can you add a license header to match the files in the repo?

src/lib.rs
```diff
@@ -1050,9 +1050,35 @@ impl Node {
 		self.prober.as_ref().map(|p| p.locked_msat.load(std::sync::atomic::Ordering::Relaxed))
 	}
 
-    /// Gives access to the scorer; needed to valuate the probing tests.
-    #[cfg(test)]
-    pub fn scorer(&self) -> &Arc<Mutex<Scorer>> { &self.scorer }
```
enigbe: `scorer` in the parent commit (7f3ce11) is only there to be deleted here. Please add a fix-up that removes it from the parent, so this commit's diff is a clean addition of `scorer_channel_liquidity` without the unrelated deletion.

src/config.rs
```diff
@@ -195,6 +198,12 @@ pub struct Config {
 	/// **Note:** If unset, default parameters will be used, and you will be able to override the
 	/// parameters on a per-payment basis in the corresponding method calls.
 	pub route_parameters: Option<RouteParametersConfig>,
+	/// Parameters for the probabilistic scorer used when computing payment routes.
+	///
+	/// These correspond to [`ProbabilisticScoringFeeParameters`] in LDK. If unset, LDK defaults
+	/// are used. Notably, [`ProbabilisticScoringFeeParameters::probing_diversity_penalty_msat`]
+	/// should be set to a non-zero value for some of the probing strategies.
+	pub scoring_fee_params: ProbabilisticScoringFeeParameters,
```
enigbe: I don't think I agree with this addition of `ProbabilisticScoringFeeParameters` to the node's config. Node operators who don't use probing have no reason to interact with this field, yet it's non-optional. Probing itself is off by default and requires an explicit `set_*_probing_strategy()` call in the builder, so surfacing scorer parameters unconditionally in `Config` seems inconsistent, even if there's a default to fall back to. More broadly, do we expect nodes to ever probe by default? If not, could this live on a probing-specific config instead, or at minimum be optional?

src/lib.rs
```diff
@@ -1050,9 +1050,35 @@ impl Node {
...
+	/// Returns the scorer's estimated `(min, max)` liquidity range for the given channel in the
+	/// direction toward `target`, or `None` if the scorer has no data for that channel.
+	///
+	/// Works by serializing the `CombinedScorer` (which writes `local_only_scorer`) and
+	/// deserializing it as a plain `ProbabilisticScorer` to call `estimated_channel_liquidity_range`.
+	pub fn scorer_channel_liquidity(
```
enigbe: If this is being exposed solely for testing then maybe it should be feature-gated with `_test_utils`. Except we see a benefit to exposing this publicly. Any thoughts on this @tnull?

tests/probing_tests.rs
```diff
@@ -0,0 +1,730 @@
...
+}
+
+impl ProbingStrategy for FixedDestStrategy {
+	fn next_probe(&self) -> Option<Probe> {
+		Some(Probe::Destination { final_node: self.destination, amount_msat: self.amount_msat })
+	}
+}
+
+/// CapturingLogger — buffers every log line for post-hoc assertion.  Used by budget tests.
+struct CapturingLogger {
```
enigbe: This is basically a duplicate of an already existing test logger (see `MockLogFacadeLogger` in `tests/common/logging.rs` ). Could you reuse that instead?

tests/probing_tests.rs
```diff
@@ -0,0 +1,730 @@
...
+impl LogWriter for CapturingLogger {
+	fn log(&self, record: LogRecord) {
+		if record.level >= LogLevel::Debug {
+			let msg = format!("[{}] {}", record.module_path, record.args);
+			self.messages.lock().unwrap().push(msg);
+		}
+	}
+}
+
+// helpers
```
enigbe: There are a lot of helpers defined here. Could we move the general-purpose ones to `tests/common/` and the probing-specific to `tests/common/probing.rs` to keep this file focused on the tests? For example, `configure_chain_source` duplicates the chain source setup already in `setup_node` `(tests/common/mod.rs)`.

tests/probing_tests.rs
```diff
@@ -0,0 +1,730 @@
...
+	let mut config = config_with_label("HiDeg+P");
+	config.node_config.scoring_fee_params.probing_diversity_penalty_msat = probing_diversity_penalty_msat;
+	setup_builder!(builder, config.node_config);
+	configure_chain_source(&mut builder, chain_source);
+	builder
+		.set_high_degree_probing_strategy(top_n)
+		.set_probing_interval(Duration::from_millis(PROBING_INTERVAL_MILLISECONDS))
+		.set_max_probe_locked_msat(MAX_LOCKED_MSAT);
+	build_and_start(builder, config)
+}
```
enigbe: `build_and_start` and `configure_chain_source`  (and other `build_node_*` functions) duplicate logic that `setup_node` (chain source setup, store selection, custom logger, node start) already handles. Could we augment `setup_node` / `TestConfig` to support probing configuration and reuse it here instead?

tests/probing_tests.rs
```diff
@@ -0,0 +1,730 @@
...
+	print!("{:<15} {:<width$}", "Known dirs", "", width = DIR_W);
+	for count in &known_counts {
+		print!(" {:<width$}", format!("{}/{}", count, total_dirs), width = SCORER_W);
+	}
+	println!();
+}
+
+/// Test change of locked_msat amount
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn test_probe_budget_increments_and_decrements() {
```
enigbe: We can drop the `test_` prefix since `#[tokio::test]` already marks it as one.

tests/probing_tests.rs
```diff
@@ -0,0 +1,730 @@
...
+	println!("Probe resolved; locked_msat = 0");
+
+	node_a.stop().unwrap();
+	node_b.stop().unwrap();
+	node_c.stop().unwrap();
+}
+
+/// Test that probing stops if the upper locked in flight probe limit is reached
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn test_probe_budget_blocks_when_node_offline() {
```
enigbe: How about renaming to `exhausted_probe_budget_blocks_new_probes` and then updating the documentation to highlight how exhaustion occurs, i.e. node going offline?

tests/probing_tests.rs
```diff
@@ -0,0 +1,730 @@
...
+	println!("Node B stopped — HTLC should be stuck in-flight.");
+
+	// The probe HTLC is now consuming outbound capacity on the A→B channel.
+	let stuck_capacity = node_a
+		.list_channels()
+		.into_iter()
+		.find(|ch| ch.counterparty_node_id == node_b.node_id())
+		.map(|ch| ch.outbound_capacity_msat)
+		.expect("A→B channel not found after B stopped");
+	assert!(
```
enigbe: This assertion fails with:
```sh
HTLC not visible in channel state: capacity unchanged (988999000 msat)
```
I think this is because the prober dispatches a probe every 250ms in the background as soon as a viable path exists, which happens before `initial_capacity` is read. By the time the test reads `initial_capacity`, a probe is already in-flight, so the value is already reduced. The `wait_until` resolves quickly because `locked_msat` is already > 0. Thus, by the time we stop `node_b`, the same probe is still in-flight, and `stuck_capacity` reads the same value as `initial_capacity`. Possibly why the assertion fails.

enigbe (changes requested):
Hi @randomlogin, thanks for the work on this! I've reviewed the first two commits:
1. 7f3ce1140c27648de98f673813caffcdcdc695cd: "Create probing service" and 
2. 6574bf9d7145d89bd9ef5444695768379dae69ea: "Add probing tests"

I've left a bunch of inline comments addressing configuration and public API, commit hygiene, testing infrastructure, and test flakiness.

In summary:

1. A couple of items are exposed publicly that seem like they should be scoped to probing or gated for tests only (see `scoring_fee_params` in `Config` and `scorer_channel_liquidity` on `Node`).
3. The probing tests duplicate existing test helpers (`setup_node`, `MockLogFacadeLogger`). Reusing and extending what's already in `tests/common/` would reduce duplication and keep the test file focused on the tests themselves.
5. `test_probe_budget_blocks_when_node_offline` has a race condition where the prober dispatches probes before the baseline capacity is measured, causing the assertion between the baseline and stuck capacities to fail. Details in the inline comment.
7. A few nits about commit hygiene, import structure, and suggestions for renaming stuff.

Also  needs to be rebased.

src/builder.rs
randomlogin: fixed

src/builder.rs
randomlogin: changed

src/config.rs
randomlogin: We only need it for `probing_diversity_penalty_msat`, which itself is only used for HighDegreeStrategy. 

There are two options:
1. To transfer it to the ProbingStrategyConfig, despite it is not used for all probing strategies.
2. To leave it as a NodeBuilder field with a default value. Router will use the default value, unless a new value is provided in during building.

The question is will anything else try to set this value:

Thoughts: 
- if we use 1st we have router-level setting set in probing config, which might conflict in future with something else, also it is used only by one strategy, it is not global probing setting.
- if we use 2nd variant we multiply entities, having 2 separate fields that are somehow probing-related (unless some other scenario needs to set the penalty, that it will be valid)

Still I stick to the second, as it in the end it is a router setting.

src/probing.rs
randomlogin: fixed

src/probing.rs
randomlogin: I'd leave it as is (maybe `top_k`, as somehow it is more common in algorithms to describe the number of samplings). 

What about `top_node_count`? 

<sup>Personally I don't like 'num' as a short for 'number'</sup>

tests/probing_tests.rs
randomlogin: fixed

tests/probing_tests.rs
randomlogin: fixed

tests/probing_tests.rs
randomlogin: fixed

tests/probing_tests.rs
randomlogin: renamed

tests/probing_tests.rs
randomlogin: renamed

tests/probing_tests.rs
randomlogin: fixed

src/probing.rs
randomlogin: squashed intermediate commits

src/builder.rs
```diff
@@ -281,6 +317,8 @@ pub struct NodeBuilder {
 	runtime_handle: Option<tokio::runtime::Handle>,
 	pathfinding_scores_sync_config: Option<PathfindingScoresSyncConfig>,
 	recovery_mode: bool,
+	probing_strategy: Option<ProbingStrategyConfig>,
+	probing_diversity_penalty_msat: Option<u64>,
```
tnull: Can we track this field as part of `ProbingStrategyConfig`?

src/builder.rs
```diff
@@ -614,6 +657,87 @@ impl NodeBuilder {
 		self
 	}
 
+	/// Configures background probing toward the highest-degree nodes in the network graph.
+	///
+	/// `top_n` controls how many of the most-connected nodes are cycled through.
+	#[cfg_attr(feature = "uniffi", allow(dead_code))]
```
tnull: Rather than adding all these on `NodeBuilder`, it might be preferable to create a dedicated builder in `src/probing.rs` and then only have the output of that (`ProbingStrategyConfig`? Or something similar?) settable on `Node`.

src/builder.rs
```diff
@@ -614,6 +657,87 @@ impl NodeBuilder {
...
+	pub fn set_high_degree_probing_strategy(&mut self, top_n: usize) -> &mut Self {
+		let kind = ProbingStrategyKind::HighDegree { top_n };
+		self.probing_strategy = Some(self.make_probing_config(kind));
+		self
+	}
+
+	/// Configures background probing via random graph walks of up to `max_hops` hops.
+	#[cfg_attr(feature = "uniffi", allow(dead_code))]
```
tnull: Please avoid any `dead_code` exemptions, I don't think you'd need them as these methods will be called by the `ArcNodeBuilder` under `uniffi`, no?

src/builder.rs
```diff
@@ -1226,6 +1382,7 @@ fn build_with_store_internal(
 	pathfinding_scores_sync_config: Option<&PathfindingScoresSyncConfig>,
 	async_payments_role: Option<AsyncPaymentsRole>, recovery_mode: bool, seed_bytes: [u8; 64],
 	runtime: Arc<Runtime>, logger: Arc<Logger>, kv_store: Arc<DynStore>,
+	probing_config: Option<&ProbingStrategyConfig>, probing_diversity_penalty_msat: Option<u64>,
```
tnull: nit: Please move the `probing_config` argument up to be next to the other `_config` arguments (here and elsewhere).

src/builder.rs
```diff
@@ -1626,7 +1783,10 @@ fn build_with_store_internal(
 		},
 	}
 
-	let scoring_fee_params = ProbabilisticScoringFeeParameters::default();
+	let mut scoring_fee_params = ProbabilisticScoringFeeParameters::default();
```
tnull: I do wonder if we should allow the user to set the entire `ProbabilisticScoringFeeParameters` and `ProbabilisticScoringDecayParameters` via the `ProbingConfigBuilder` mentioned above? Do you see any reason where that would conflict with other API design decisions?

src/builder.rs
```diff
@@ -1965,6 +2125,36 @@ fn build_with_store_internal(
 		_leak_checker.0.push(Arc::downgrade(&wallet) as Weak<dyn Any + Send + Sync>);
 	}
 
+	let prober = probing_config.map(|probing_cfg| {
+		let strategy: Arc<dyn probing::ProbingStrategy> = match &probing_cfg.kind {
+			ProbingStrategyKind::HighDegree { top_n } => {
+				Arc::new(probing::HighDegreeStrategy::new(
+					network_graph.clone(),
```
tnull: nit: When cloning `Arc`, please always use `Arc::clone` rather than `.clone()` to make it immediately visible that it's 'just' cloning a reference, not the actual data.

src/event.rs
```diff
@@ -515,6 +516,7 @@ where
 	static_invoice_store: Option<StaticInvoiceStore>,
 	onion_messenger: Arc<OnionMessenger>,
 	om_mailbox: Option<Arc<OnionMessageMailbox>>,
+	probe_locked_msat: Option<Arc<AtomicU64>>,
```
tnull: Rather than tracking this field which is rather obscure in the context of the event handler, should the event handler rather take a reference to the prober object and then delegate to some kind of `handle_probe_{successful,failed}` method on there? This would keep the prober internals more isolated to that module, IMO.

src/lib.rs
```diff
@@ -1067,6 +1079,42 @@ impl Node {
 		))
 	}
 
+	/// Returns the total millisatoshis currently locked in in-flight probes, or `None` if no
```
tnull: I think these should live on the `Prober` (we could consider an API extension pattern like we have for the payments, i.e., call `Node::prober` to acquire a reference to `Prober` that exposes more probing-specific APIs) or the latter method here on `NetworkGraph`, maybe? Arguably `probe_locked_msat` could also be a field on `NodeMetrics` rather than accessible via a dedicated method.

src/probing.rs
```diff
@@ -0,0 +1,391 @@
...
+use lightning::routing::gossip::NodeId;
+use lightning::routing::router::{Path, RouteHop, MAX_PATH_LENGTH_ESTIMATE};
+use lightning_invoice::DEFAULT_MIN_FINAL_CLTV_EXPIRY_DELTA;
+use lightning_types::features::NodeFeatures;
+
+use crate::logger::{log_debug, LdkLogger, Logger};
+use crate::types::{ChannelManager, Graph};
+
+/// Returns a random `u64` uniformly distributed in `[min, max]` (inclusive).
+fn random_range(min: u64, max: u64) -> u64 {
```
tnull: Can we move this to a new `src/util/random.rs` or similar module, as other places might need the same utility at some point?

src/probing.rs
```diff
@@ -0,0 +1,391 @@
...
+		if nodes_by_degree.is_empty() {
+			return None;
+		}
+
+		nodes_by_degree.sort_unstable_by(|a, b| b.1.cmp(&a.1));
+
+		let top_node_count = self.top_node_count.min(nodes_by_degree.len());
+
+		let cursor = self.cursor.fetch_add(1, Ordering::Relaxed);
+		let (final_node, _degree) = nodes_by_degree[cursor % top_node_count];
```
tnull: Hmm, this is a bit odd as the graph might move from down under us, no? So as the graph might change, cursor would index a different entry each time potentially? Wouldn't it be preferable to always cache `top_node_count` nodes in a list, and only refresh the list once we probed all of them?

src/probing.rs
```diff
@@ -0,0 +1,391 @@
...
+		Some(Probe::Destination { final_node, amount_msat })
+	}
+}
+
+/// Explores the graph by walking a random number of hops outward from one of our own
+/// channels, constructing the [`Path`] explicitly.
+///
+/// On each tick:
+///   1. Picks one of our confirmed, usable channels to start from.
+///   2. Performs a deterministic walk of a randomly chosen depth (up to
```
tnull: Doing a random walk is a great idea! Do you think it would be a worthwhile extension (or another strategy idea) to use some kind of weighted random walk, e.g. to favor channels of certain criteria (smaller/larger)? 

Maybe a bounded breadth first search strategy could also make sense, which would allow the peer to explore it's closer vincinity?

src/probing.rs
```diff
@@ -0,0 +1,391 @@
...
+		let graph = self.network_graph.read_only();
+		let first_hop =
+			&initial_channels[random_range(0, initial_channels.len() as u64 - 1) as usize];
+		let first_hop_scid = first_hop.short_channel_id.unwrap();
+		let next_peer_pubkey = first_hop.counterparty.node_id;
+		let next_peer_node_id = NodeId::from_pubkey(&next_peer_pubkey);
+
+		// Track the tightest HTLC limit across all hops to cap the probe amount.
+		// The first hop limit comes from our live channel state; subsequent hops use htlc_maximum_msat from the gossip channel update.
+		let mut route_least_htlc_upper_bound = first_hop.next_outbound_htlc_limit_msat;
```
tnull, line 555: Do we also need a lower bound?

src/probing.rs
```diff
@@ -0,0 +1,391 @@
...
+}
+
+/// Runs the probing loop for the given [`Prober`] until `stop_rx` fires.
+pub(crate) async fn run_prober(prober: Arc<Prober>, mut stop_rx: tokio::sync::watch::Receiver<()>) {
+	let mut ticker = tokio::time::interval(prober.interval);
+	ticker.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+
+	loop {
+		tokio::select! {
+			_ = stop_rx.changed() => {
```
tnull, line 798: nit: Probably doesn't matter all too much as the second arm here won't block for too long (I assume) on shutdown, but probably still good to add `biased;` here so we'd always poll for the stop signal first.

tests/common/mod.rs
```diff
@@ -713,6 +749,41 @@ pub async fn open_channel(
 	open_channel_push_amt(node_a, node_b, funding_amount_sat, None, should_announce, electrsd).await
 }
 
+/// Like [`open_channel`] but skips the `wait_for_tx` electrum check so that
+/// multiple channels can be opened back-to-back before any blocks are mined.
+/// The caller is responsible for mining blocks and confirming the funding txs.
+pub async fn open_channel_no_electrum_wait(
```
tnull: Hmm, if we want this, could we call this `open_channel_no_wait` or similar and then have `open_channel` call into `open_channel_no_wait` to DRY up the functionality. Note that this helper is also a bit risky to introduce as currently we do not have UTXO locking, i.e., if this is used to open multiple subsequent channels for the same node it will double-spend itself.

tests/probing_tests.rs
```diff
@@ -0,0 +1,571 @@
...
+}
+
+impl ProbingStrategy for FixedDestStrategy {
+	fn next_probe(&self) -> Option<Probe> {
+		Some(Probe::Destination { final_node: self.destination, amount_msat: self.amount_msat })
+	}
+}
+
+// helpers
+async fn wait_until(timeout: Duration, predicate: impl Fn() -> bool) -> bool {
```
tnull: I think we should look to use `tokio::time::timeout` instead?

tnull (commented):
Thanks for taking this on and excuse the delay here!

Did a first review pass and this already looks great! Here are some relatively minor comments, mostly concerning the API design.

src/builder.rs
randomlogin: fixed

src/probing.rs
randomlogin: fixed

src/probing.rs
randomlogin: Moved it to `src/util.rs`. I doubt that there will be a lot of other random functions there, so it seems there is no need for a distinct 'random' util module, however other utilities might will require their own modules (then `src/util.rs` will move to `src/util/mod.rs`.

I've checked the other usages of randomness throughout the codebase, in all the places it's just filling some bytes with random data and using those bytes as is (or as integers), for instance:
```
let user_channel_id: u128 = u128::from_ne_bytes(
	self.keys_manager.get_secure_random_bytes()[..16].try_into().unwrap(),
);
```

Also made distribution indeed uniform (the current one had bias for big ranges like 2**63+1 because of modulo).

src/builder.rs
randomlogin: fixed

src/builder.rs
randomlogin: removed

src/builder.rs
randomlogin: moved, now it is in `ProbingConfig`

src/builder.rs
randomlogin: Created a dedicated builder.

src/probing.rs
randomlogin: Added a lower bound.

src/probing.rs
randomlogin: Right, the cursor could point at the node we already probed. 

I think it's better to cache nodes we already probed instead of current top nodes at some point in time, then we can just skip them if we try to probe them the second time. This way we still probe current top nodes if the graph changes (versus we still probe original cached nodes, despite the graph has changed).

Implemented it, added a default cooldown for probing the same node of 1 hour.

src/event.rs
randomlogin: removed this field, added a reference to the prober object

src/lib.rs
randomlogin: For now moved this getter to the prober object. 

About moving to NodeMetrics: currently the probed_locked_msat is not persisted and is lost on node restart, if there are in flight probes when our node becomes offline, then on restart the probes would just resolve and locked amount stay at zero:

```
	pub(crate) fn handle_probe_successful(&self, path: &lightning::routing::router::Path) {
		let amount: u64 = path.hops.iter().map(|h| h.fee_msat).sum();
		let _ = self.locked_msat.fetch_update(Ordering::AcqRel, Ordering::Acquire, |v| {
			Some(v.saturating_sub(amount))
		});
	}
```

This way we can briefly exceed the probing budget and falsely fire new probes, which seems to be more or less fine. 

If we persist it, we would need to perform IO on each probe event, they can be quite frequent, say several times per second, is it okay to do so? 
Other fields from NodeMetrics change more rarely, for example the wallet is synced every 30 seconds, others fields once every ~30 minutes.

About NetworkGraph: I don't think we should have probing info there at all, it feels like it's just for the network topology and not related to our node info.

tests/probing_tests.rs
randomlogin: Can you please elaborate? I've changed it to use timeout, but not sure it's what you meant.

tests/common/mod.rs
randomlogin: Renamed and made other function call this one. 

About locking:
If we have enough UTXOs we're fine, as BDK tracks unconfirmed txs and thus knows about previous funding txs.

src/builder.rs
randomlogin: I think we should expose these settings (in NodeBuilder, not probing builder). 
However these are for fine tuning and should be used only by advanced users. Also I would expose `UserConfig`, as for example user cannot decide what features do advertise. 

I can add builder methods for scoring parameters, though maybe it should be in another PR aimed on exposing settings for an advanced user?

src/probing.rs
randomlogin: Yes, I think these are nice ideas. 

I think the best would be to somehow identify 'hub' routing nodes through which we're likely to route a payment and to probe paths both to this node from ours and also from this 'hub' node for several hops around it.

I think we need some live network data to understand which probing strategy gives the best results.

Let's return back to it after the basic probing framework is established?

tnull (commented):
Seems tests are failing right now:

```
thread 'exhausted_probe_budget_blocks_new_probes' (167312) panicked at tests/probing_tests.rs:381:5:
no probe dispatched within 15 s


failures:
    exhausted_probe_budget_blocks_new_probes
    probe_budget_increments_and_decrements
```

tests/probing_tests.rs
tnull: Ah, I just meant that we don't need to reimplement the same functionality or introduce a dedicated helper, as we can just use `tokio::time::timeout`, no?

src/probing.rs
tnull: Sounds good!

src/builder.rs
tnull: > Also I would expose UserConfig, as for example user cannot decide what features do advertise.

Yes, this is very intentional, as we want to provide a sane/safe API. For example letting user freely choose to set certain parameters if we don't implement them properly will just lead to a lot of footguns, in some circumstances even with the potential for money loss.

src/lib.rs
tnull: Ah, yeah, sorry, I indeed meant `NodeStatus`, not `NodeMetrics`. But probably easier to leave it as is as a getter on the `Prober`, now that it's accesible.

src/builder.rs
randomlogin: In that case I think we don't need to expose ProbabilisticScoringFeeParameters and ProbabilisticScoringFeeParameters by the very reason you mentioned.

src/builder.rs
```diff
@@ -73,12 +75,13 @@ use crate::logger::{log_error, LdkLogger, LogLevel, LogWriter, Logger};
 use crate::message_handler::NodeCustomMessageHandler;
 use crate::payment::asynchronous::om_mailbox::OnionMessageMailbox;
 use crate::peer_store::PeerStore;
+use crate::probing;
```
tnull: nit: Please import the objects you're using directly, also avoiding the `probing::` prefix in docs.

src/builder.rs
```diff
@@ -1960,6 +2001,39 @@ fn build_with_store_internal(
...
+				))
+			},
+			probing::ProbingStrategyKind::Custom(s) => Arc::clone(s),
+		};
+		Arc::new(probing::Prober {
+			channel_manager: Arc::clone(&channel_manager),
+			logger: Arc::clone(&logger),
+			strategy,
+			interval: probing_cfg.interval,
+			liquidity_limit_multiplier: Some(config.probing_liquidity_limit_multiplier),
```
tnull: I do wonder if we should move the `probing_liquidity_limit_multiplier` config field to `ProbingConfig` now?

src/event.rs
```diff
@@ -530,7 +532,7 @@ where
 		payment_store: Arc<PaymentStore>, peer_store: Arc<PeerStore<L>>,
 		keys_manager: Arc<KeysManager>, static_invoice_store: Option<StaticInvoiceStore>,
 		onion_messenger: Arc<OnionMessenger>, om_mailbox: Option<Arc<OnionMessageMailbox>>,
-		runtime: Arc<Runtime>, logger: L, config: Arc<Config>,
+		runtime: Arc<Runtime>, logger: L, config: Arc<Config>, prober: Option<Arc<Prober>>,
```
tnull: Please move the `prober` after `om_mailbox` here, as `runtime`, `logger`, and `config` are usually the trailing fields in most constructors, etc.

tests/probing_tests.rs
```diff
@@ -0,0 +1,616 @@
...
+	}
+	println!();
+}
+
+/// Verifies that `locked_msat` increases when a probe is dispatched and returns
+/// to zero once the probe resolves (succeeds or fails).
+#[tokio::test(flavor = "multi_thread")]
+async fn probe_budget_increments_and_decrements() {
+	let (bitcoind, electrsd) = setup_bitcoind_and_electrsd();
+	let chain_source = TestChainSource::Electrum(&electrsd);
```
tnull: Please use `random_chain_source()` for all tests that aren't chainsource specific.

src/lib.rs
```diff
@@ -101,10 +101,12 @@ pub mod logger;
 mod message_handler;
 pub mod payment;
 mod peer_store;
+mod probing;
```
tnull: Let's make this a `pub mod` and expose `Prober` and related objects under `crate::probing`, *not* at the root level. Note that for uniffi to still build you might need to add corresponding re-exports to the list in the beginning of `ffi/types.rs`, as `uniffi` indeed expects all objects to live under the crate root.

tnull (commented):
Also asked Claude to do a review, some of this might be worth addressing:

```
 1. Bug: locked_msat accounting mismatch for Probe::Destination (most significant)

  In run_prober (src/probing.rs:593-609), for a Destination probe, amount_msat is added to locked_msat once. But send_spontaneous_preflight_probes
  can dispatch multiple probe paths. Each resolved path triggers a separate ProbeSuccessful/ProbeFailed event, and
  handle_probe_successful/handle_probe_failed each subtract path.hops.iter().map(|h| h.fee_msat).sum() — which is the amount plus routing fees for
  that path.

  Two problems compound:
  - Multiple paths: added once, subtracted N times (one per path event). With saturating_sub this silently clamps to 0, but the budget will appear
  to clear prematurely.
  - Fee mismatch: even for a single path, the subtracted amount includes routing fees (amount_msat + fees), while only amount_msat was added. So
  the subtraction is always ≥ what was added.

  The PrebuiltRoute variant is fine — it consistently uses path.hops.iter().map(|h| h.fee_msat).sum() on both sides.

  2. Bug: Fee calculation in RandomStrategy::try_build_path uses final amount instead of forwarded amount

  At src/probing.rs:462:
  let fee = update.fees.base_msat as u64
      + (amount_msat * update.fees.proportional_millionths as u64 / 1_000_000);

  This computes each intermediate hop's fee based on the delivery amount_msat, but each hop actually charges fees on the amount it forwards
  (delivery + all downstream fees). For routes > 2 hops, this underestimates intermediate fees. The total_outgoing check at the end could then let
  through routes that are actually too large for the first-hop HTLC limit. This is minor for probing purposes since probes fail deliberately, but
  it's still incorrect.

  3. Doc inconsistency on HighDegreeStrategy

  The struct-level doc comment says "Returns None (skips the tick) if all top nodes are on cooldown." But the implementation at line 276-286 does
  the opposite — when all top nodes are on cooldown, it clears the cooldown map and returns the first node, never returning None on cooldown
  exhaustion. The None return only happens when the graph is empty.

  4. random_range overflow edge case

  In src/util.rs:14, let range = max - min + 1 overflows to 0 when max - min == u64::MAX, causing a division-by-zero panic at u64::MAX % range.
  This can't happen with current callers, but the function has no guard against it.

  5. Race condition in budget check (minor/theoretical)

  In run_prober, the budget check (locked_msat.load() + amount > max_locked_msat) and the subsequent fetch_add are not atomic. Since the prober
  loop is single-threaded and the event handler only decreases locked_msat, this can't cause the budget to be exceeded — the worst case is a
  slightly stale read that rejects a probe that would have fit. So this is benign, but worth noting.

  6. scorer_channel_liquidity is very expensive

  src/lib.rs:1090-1108 serializes the entire scorer to a Vec<u8> and then deserializes it as a ProbabilisticScorer just to call
  estimated_channel_liquidity_range. This is O(scorer size) per call. It's only used in tests currently, but it's a public API — worth a doc
  warning or a #[cfg(test)] gate if it's not intended for production use.

  ---
  Issues #1 and #2 are real bugs. #1 in particular means the budget tracking for Destination probes is systematically wrong.
```

src/builder.rs
randomlogin: fixed

src/event.rs
randomlogin: moved

tests/probing_tests.rs
randomlogin: changed

src/lib.rs
randomlogin: done

src/builder.rs
randomlogin: Made it a HighDegreeStrategy parameter.

src/lib.rs
```diff
@@ -1067,6 +1079,42 @@ impl Node {
...
+		self.prober.as_deref()
+	}
+
+	/// Returns the scorer's estimated `(min, max)` liquidity range for the given channel in the
+	/// direction toward `target`, or `None` if the scorer has no data for that channel.
+	///
+	/// **Warning:** This is expensive — O(scorer size) per call. It works by serializing the
+	/// entire `CombinedScorer` and deserializing it as a plain `ProbabilisticScorer` to access
+	/// `estimated_channel_liquidity_range`. Intended for testing and debugging, not hot paths.
+	pub fn scorer_channel_liquidity(&self, scid: u64, target: PublicKey) -> Option<(u64, u64)> {
```
tnull: I don't think we'll want to expose this here. If we think we need this we'd need to work out an API that is more reasonable than returning plain `(u64, u64)` (which are rather cryptic to begin with) and would also work in bindings. I also don't think the re-serializing of the scorer here is acceptable - if we need access to `estimated_channel_liquidity_range`, we probably need to work out how we can expose that in a reasonable way on `CombinedScorer` upstream

tnull (commented):
Have yet to do another full pass. However, please start looking into how we can expose the adding probing APIs as part of the `uniffi` bindings in the meantime. We try to keep feature parity between Rust and bindings APIs as far as possible.

src/lib.rs
randomlogin: Removed it from this PR, instead created a PR upstream: https://github.com/lightningdevkit/rust-lightning/pull/4568 

After some considerations I think it's indeed needed to be exposed, as it is needed to gather data about probing strategies to polish them. 

Even despite the fact we cannot get the real liquidity data on live network to have 2 figures to compare (real vs estimate), it can be used in for example measuring the number of failed payments against the liquidity estimates we had on the path. Or to have a bunch of our own routing nodes with exact channel liquidity data (on signet) to compare with the estimations.

Also removed the test that naively compared strategy performances (and used this method).

tnull: That PR now landed. You might need to wait for other PRs that fix pre-existing API breakages before you can use the API here. Until then, feel free to change the dependency to a branch of yours that depends on the same LDK base as our current main (but that includes the fix from LDK#4568), but note you'll also have to update `bitcoin-payment-instructions` to depend on the same revision accordingly.

src/probing.rs
```diff
@@ -0,0 +1,716 @@
...
+			.field("cooldown", &self.cooldown)
+			.finish()
+	}
+}
+
+impl ProbingConfig {
+	/// Start building a config that probes toward the highest-degree nodes in the graph.
+	///
+	/// `top_node_count` controls how many of the most-connected nodes are cycled through.
+	pub fn high_degree(top_node_count: usize) -> ProbingConfigBuilder {
```
tnull: In the builder pattern it's a bit uncommon to have these constructors directly on the thing that is being built. Usually, you'd have a `ProbingConfig::builder` method that returns a `ProbingConfigBuilder`. I guess it is fine to have multiple constructors, just noting.

src/probing.rs
```diff
@@ -0,0 +1,716 @@
...
+
+#[cfg(feature = "uniffi")]
+#[uniffi::export]
+impl ProbingConfig {
+	/// Creates a probing config that probes toward the highest-degree nodes in the graph.
+	///
+	/// `top_node_count` controls how many of the most-connected nodes are cycled through.
+	/// All other parameters are optional and fall back to sensible defaults when `None`.
+	#[uniffi::constructor]
+	pub fn new_high_degree(
```
tnull: However, we should def. not duplicate `new_high_degree` here. It seems it would make more sense to expose the `ProbingConfigBuilder` itself in bindings, and then allow users to use that rather than replicating the builder wiring here in `ProbingConfig`?

tnull (commented):
Thanks for the update. It seems this is blocked now until we can bump the LDK dependency to include the upstream PR?

src/probing.rs
randomlogin: Changed in favor of exposing `ProbingConfigBuilder` in bindings.

src/probing.rs
randomlogin: Changed.

src/lib.rs
randomlogin: The test which I removed was more an diagnostic, not a correctness, whether the scorer gets its estimations, it didn't assert anything, so I'm not sure if I need to return it to the PR, should not be a blocker for now. 
(still of course the upstream change is needed)

src/probing.rs
```diff
@@ -0,0 +1,734 @@
...
+				fee_msat: amount_msat,
+				cltv_expiry_delta: 0,
+				maybe_announced_channel: true,
+			});
+		}
+
+		// Non-final hops, from second-to-last back to first.
+		for i in (0..last).rev() {
+			let (node_id, via_scid, pubkey) = route[i];
+			let channel_info = graph.channel(via_scid)?;
```
kaloudis, line 576: The `random_walk` strategy seems to be usable for any probing node that is using unannounced channels.

`try_build_path` always fails when the first hop is a private/unannounced channel. The first entry in route uses the local channel's SCID, which isn't in the gossip graph, so `graph.channel(via_scid)?` returns None and the entire function bails.

Could we fall back to default features instead?

kaloudis (changes requested):

src/probing.rs
randomlogin: A nice find, thanks!

Fixed, now we use local information for the first hop which makes strategy usable without announced channels.

Made a refactor for getting the features via a closure (method would require carrying &graph and first channel id).

.github/workflows/audit.yml
tnull, line 23: Please drop all these unrelated changes.

tnull (commented):
It seems your git commit history got mangled and this PR now touches almost every file. Please clean it up and rebase on current main. Please refer to https://cbea.ms/git-commit/ for some guidance on how to structure commits and how to write good descriptions.

.github/workflows/audit.yml
randomlogin: Dropped.

tnull (commented):
This still need a rebase.

randomlogin:
Now it should be fine. Previously I've accidentally absorbed changes from the main without a proper merge.
Currently CI fails because of the changes from the main branch.

Should I squash my commits (especially with the abundance of empty CI re-trigger commits I made)?

Also regarding the commit messages, I'll to be more accurate (and follow case styling). Please, don't hesitate to tell me once again if it occurs. Also ideally the commits should be more granular, right?

tnull:
> Now it should be fine. Previously I've accidentally absorbed changes from the main without a proper merge. Currently CI fails because of the changes from the main branch.

Yes, fixed in #891.

> Should I squash my commits (especially with the abundance of empty CI re-trigger commits I made)?

Yes, please generally restructure your commit history so it has a few logical 'feature' commit and preferably only add fixups  right after them, so they can be squashed in once reviewed. Usually you'd prefix their commit description with `fixup!` or `f`.

> Also regarding the commit messages, I'll to be more accurate (and follow case styling). Please, don't hesitate to tell me once again if it occurs. Also ideally the commits should be more granular, right?

Yes, that would be great. Preferably code introduced in earlier commit doesn't get changed again in later commits, and all revisions should build test, and format independently.

tnull (commented):
The `exhausted_probe_budget_blocks_new_probes` test and the linting CI job are failing unfortunately. Regarding the latter, note that we recently banned `unwrap` project-wide in non-test code, so you'll need to use `except` everywhere, eg., `except("lock")` when acquiring locks.

randomlogin:
Removed usage of unwrap(), reworked the exhaust test not to rely on a node stopping race condition.

src/builder.rs
```diff
@@ -625,6 +635,25 @@ impl NodeBuilder {
 		self
 	}
 
+	/// Configures background probing.
+	///
+	/// Use [`ProbingConfigBuilder`] to build the configuration:
```
tnull, line 642: We should probably add another paragraph before this example that gives some context on what background probing is and why users would want to enable it.

src/lib.rs
```diff
@@ -113,16 +115,17 @@ use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 #[cfg(cycle_tests)]
 use std::{any::Any, sync::Weak};
 
+#[cfg(feature = "uniffi")]
+use crate::probing::ProbingConfig;
```
tnull: nit: We do these `uniffi`-specific re-exports via use statements in `ffi/types.rs`.

src/lib.rs
```diff
@@ -171,6 +174,9 @@ use payment::{
 	UnifiedPayment,
 };
 use peer_store::{PeerInfo, PeerStore};
+#[cfg(feature = "uniffi")]
+pub use probing::ArcedProbingConfigBuilder as ProbingConfigBuilder;
```
tnull, line 178: Same here, please expose this in `ffi/types.rs`

src/probing.rs
```diff
@@ -0,0 +1,749 @@
...
+			},
+			Self::Random { max_hops } => {
+				f.debug_struct("Random").field("max_hops", max_hops).finish()
+			},
+			Self::Custom(_) => f.write_str("Custom(<probing strategy>)"),
+		}
+	}
+}
+
+/// Configuration for the background probing subsystem.
```
tnull, line 110: See above: given this is the main probing-related object it might be worth spending a paragraph here (or on the module docs) to explain what probing is and why users would want to enable it in the first place.

src/event.rs
```diff
@@ -1186,8 +1161,16 @@ where
 
 			LdkEvent::PaymentPathSuccessful { .. } => {},
 			LdkEvent::PaymentPathFailed { .. } => {},
-			LdkEvent::ProbeSuccessful { .. } => {},
-			LdkEvent::ProbeFailed { .. } => {},
+			LdkEvent::ProbeSuccessful { path, .. } => {
+				if let Some(prober) = &self.prober {
+					prober.handle_probe_successful(&path);
```
tnull: If we handle all probes the same here to reduce the locked amount, we probably also need to mark any amounts that we send manually via preflight probes as locked, as otherwise the accounting is off? Or, maybe it would be preferable to keep track of which paths were for background probing and only reduced the locked amount for them?

src/probing.rs
```diff
@@ -0,0 +1,749 @@
...
+			let (node_id, via_scid, pubkey) = route[last];
+			let (node_features, channel_features, maybe_announced_channel) =
+				hop_features(&node_id, via_scid)?;
+			hops.push(RouteHop {
+				pubkey,
+				node_features,
+				short_channel_id: via_scid,
+				channel_features,
+				fee_msat: amount_msat,
+				cltv_expiry_delta: 0,
```
tnull: Codex:

>   - High: RandomStrategy builds invalid final-hop CLTVs. src/probing.rs:584 sets the last hop’s cltv_expiry_delta to 0, but LDK expects the last hop delta to be the destination final CLTV. For multi-hop random probes, the penultimate node can reject the forward as outgoing CLTV too soon, so
    the strategy mostly trains failures instead of probing liquidity. Use a real final CLTV delta, e.g. DEFAULT_MIN_FINAL_CLTV_EXPIRY_DELTA.

Seems valid?

src/probing.rs
```diff
@@ -0,0 +1,749 @@
...
+			return None;
+		}
+
+		Some(Path { hops, blinded_tail: None })
+	}
+}
+
+impl ProbingStrategy for RandomStrategy {
+	fn next_probe(&self) -> Option<Path> {
+		let target_hops = random_range(1, self.max_hops as u64) as usize;
```
tnull: Codex:

>   - Medium: RandomStrategy can produce one-hop paths that send_probe rejects. src/probing.rs:636 samples target_hops starting at 1, and random_walk(1) will only ever emit single-hop paths, which LDK rejects with “No need probing a path with less than two hops.” Clamp/sample to at least two
    hops or return None for that config.

Seems also valid, probably makes sense to start at 2?

tnull (commented):
Excuse the delay here. Looks pretty good, but there are some minor things we might want to address before this can land.

src/lib.rs
randomlogin: Moved to `ffi/types.rs`

src/lib.rs
randomlogin: Moved to `ffi/types.rs`

src/builder.rs
randomlogin: Added documentation on the module level as well as expanded/corrected docs for particular objects (builder).

src/probing.rs
randomlogin: Added documentation on the module level as well as expanded/corrected docs for particular objects (builder).

src/probing.rs
randomlogin: Fixed.

src/event.rs
randomlogin: Addressed this by adding `inflight_probes` hashmap to track our probes. 
Changed event handling to fire background probing events only on these (background) probes.
Code may be DRYed (as event handlers share parts of code), but I find the current version more clear.

randomlogin:
> Excuse the delay here. Looks pretty good, but there are some minor things we might want to address before this can land.

Addressed all issues mentioned, they have their respective commits. If they are fine, I'm going to squash them.

src/probing.rs
randomlogin: Fixed

src/builder.rs
```diff
@@ -1109,6 +1132,13 @@ impl ArcedNodeBuilder {
 		self.inner.write().expect("lock").set_wallet_recovery_mode();
 	}
 
+	/// Configures background probing.
+	///
+	/// See [`ProbingConfig`] for details.
```
tnull: Please make sure the docs on the `ArcedNodeBuilder` mirror the ones on the regular `NodeBuilder`.

src/event.rs
```diff
@@ -1186,8 +1161,24 @@ where
 
 			LdkEvent::PaymentPathSuccessful { .. } => {},
 			LdkEvent::PaymentPathFailed { .. } => {},
-			LdkEvent::ProbeSuccessful { .. } => {},
-			LdkEvent::ProbeFailed { .. } => {},
+			LdkEvent::ProbeSuccessful { path, payment_id, .. } => {
+				if let Some(prober) = &self.prober {
+					if let Some(amount) =
```
tnull: Maybe just move this check into `handle_background_probe_successful`/`handle_background_probe_failed`? Then you could keep `inflight_probes` visibility at the module level, too.

src/probing.rs
```diff
@@ -0,0 +1,843 @@
...
+pub struct Prober {
+	pub(crate) channel_manager: Arc<ChannelManager>,
+	pub(crate) logger: Arc<Logger>,
+	/// The strategy that decides what to probe.
+	pub strategy: Arc<dyn ProbingStrategy>,
+	/// How often to fire a probe attempt.
+	pub interval: Duration,
+	/// Maximum total millisatoshis that may be locked in in-flight probes at any time.
+	pub max_locked_msat: u64,
+	pub(crate) locked_msat: Arc<AtomicU64>,
```
tnull: Isn't this redudant to the sum over all `inflight_probes` now? Probably it's better to keep one state for accounting purposes, to avoid the possibility for them to get out-of-sync.

src/builder.rs
```diff
@@ -2036,6 +2065,40 @@ fn build_with_store_internal(
...
+			ProbingStrategyKind::Custom(s) => Arc::clone(s),
+		};
+		Arc::new(Prober {
+			channel_manager: Arc::clone(&channel_manager),
+			logger: Arc::clone(&logger),
+			strategy,
+			interval: probing_cfg.interval,
+			max_locked_msat: probing_cfg.max_locked_msat,
+			locked_msat: Arc::new(AtomicU64::new(0)),
+			inflight_probes: Mutex::new(HashMap::new()),
```
tnull: Hmm, it seems accounting will still be off if we send out probes and then restart, as we'll re-init with an empty map that forgot about the previously-sent probes. Do we think that's acceptable or do we need to somehow persist this map? Seems like persistence would add some considerable complication on top? Thoughts?

randomlogin: Initially I wanted probing to be without persistence for the sake of simplicity, but now it seems it is really important, because if the node really goes out of the budget then likely this will occur after a restart and eventually it will drain the whole balance.

The corner case is the situation when a probing node:
1) has a lot of unresolved probes (or small probing budget)
2) restarts frequently

In that case the budget would be falsely shown as near-zero, whereas in fact there some probes sent and unresolved.

If we restart node more rarely than the time needed for locked inflight probes to release, we're more or less are fine. So if a node restarts more rarely than once in two hours it should be okay (it would not drain the balance completely, but there could be time periods where the probing budget is in reality around twice as bigger as it should be). 



There is an option to add to the public API this: https://github.com/lightningdevkit/rust-lightning/blob/03ab73da95af34f3ae3d28d7e83383cfa0b04fc1/lightning/src/ln/channelmanager.rs#L6143
Then we could iterate through recent payments on restart and populate the `inflight_probes` with `max_locked_msat` without persisting anything new.


Also some calculations when when the budget limit hits: assume probing is done every 500ms, or equivalently it has frequency of $\nu = 2 s^{-1}$ probes per second.
Let's assume the probability of a probe to get locked for $T=2$ hours is $p$.
Let's assume node has $PB$ probing budget, for ordinary user-node it is $PB=10A$ (max 10 probes can be fired simultaneously), and for example for a routing node it can be something like $100A-200A$.

That would mean during time $T$ we got locked $A\cdot p \cdot \nu \cdot T$, if it is greater than $PB=10A$ then we have hit the ceiling of the budget. So $A\cdot p \cdot \nu \cdot T \geq 10A$
With figures from the above if the rate of failure of probes is bigger than 0.0007 (0.07%), then we would hit the budget bound. If the node restarts. And actually it's not that big (1 out of 1440 probes is locked for 2 hours). 

All in all, I think it's worth adding some way of persistence, because for 'ordinary' nodes the cap seems to be easily reached if the settings (frequency, amount) is not tuned properly.

src/probing.rs
randomlogin: yeah, if we have all the probes details now, we can just calculate it when needed (with a tiny worsening of performance, but I guess it's fine).

src/event.rs
randomlogin: Moved

src/builder.rs
randomlogin: I think it does so already. The builder methods do have (short) docs on them for both normal and arced builders.

src/builder.rs
tnull: Is there something unaddressed after https://github.com/lightningdevkit/rust-lightning/pull/4603 ? If so, mind opening an issue for it, so we can see if we want to follow-up?

tnull:
As discussed above this is now blocked on https://github.com/lightningdevkit/rust-lightning/pull/4603 becoming available. Will revisit once earlier LDK bumps have landed.

src/builder.rs
randomlogin: I think it is the only missing piece, the rest is fine. If anything will come to my mind, I'll post and open an issue.

randomlogin:
After the dependency update we have everything needed for this PR. 

Also in the last commit I renamed `RandomStrategy` into `RandomWalkStrategy` to make it more clear.

src/probing.rs
```diff
@@ -0,0 +1,749 @@
...
+			kind: self.kind.clone(),
+			interval: self.interval.max(MIN_PROBING_INTERVAL),
+			max_locked_msat: self.max_locked_msat,
+			diversity_penalty_msat: self.diversity_penalty_msat,
+			cooldown: self.cooldown,
+		}
+	}
+}
+
+/// A UniFFI-compatible wrapper around [`ProbingConfigBuilder`] that uses interior mutability
```
tnull: nit: Lets drop the mention of UniFFI from these docs, they should just be identical to the ones on `ProbingConfigBuilder` as the bindings users surely don't care about UniFFI or whatever bindings generator we use. They only care about how the API works, so the UniFFI objects should just be treated as normal objects and explain the API.

While we're at it, we should probably also do the typedef trick we do on `Builder` to expose this as `ProbingConfigBuilder` rather than `ArcedProbingConfigBuilder`

src/builder.rs
```diff
@@ -1661,7 +1698,10 @@ fn build_with_store_internal(
 		},
 	}
 
-	let scoring_fee_params = ProbabilisticScoringFeeParameters::default();
+	let mut scoring_fee_params = ProbabilisticScoringFeeParameters::default();
+	if let Some(penalty) = probing_config.and_then(|c| c.diversity_penalty_msat) {
+		scoring_fee_params.probing_diversity_penalty_msat = penalty;
```
tnull: Hmm, this will install the `probing_diversity_penalty_msat` globally, which is probably not what we want when a node enables the background probing but still wants to send regular payments. 

This probably means that we can't use the default router for probing, but rather need to call https://docs.rs/lightning/latest/lightning/routing/router/fn.find_route.html manually (as it allows us to pass the score params directly per call) to create a route and then hand the pre-built route to https://docs.rs/lightning/latest/lightning/ln/channelmanager/struct.ChannelManager.html#method.send_payment_with_route

src/probing.rs
```diff
@@ -0,0 +1,833 @@
...
+
+impl Prober {
+	/// Returns the total millisatoshis currently locked in in-flight probes.
+	pub fn locked_msat(&self) -> u64 {
+		return self
+			.channel_manager
+			.list_recent_payments()
+			.into_iter()
+			.filter_map(|p| match p {
+				RecentPaymentDetails::Pending { is_probe: true, total_msat, .. } => {
```
tnull: Codex:

>   - [P2] Include route fees in pending probe budget — /home/tnull/workspace/ldk-node/src/probing.rs:762-763
>     For pending probes this uses RecentPaymentDetails::Pending.total_msat, but LDK defines that value as excluding routing fees while the first-hop HTLC locks
>     the final amount plus all route fees. On routes or custom strategies with non-trivial fees, locked_msat() under-reports in-flight liquidity, so subsequent
>     budget checks can allow max_locked_msat to be exceeded and the public locked amount is misleading.
>

tnull (commented):
Excuse the delay here! This mostly looks good, I think the most important thing that still needs to be addressed is not setting the diversity multiplier globally, as it could unexpected consequences for users. This unfortunately also needs a rebase, and while you rebase please squash the fixup commits so that you don't touch the same codepaths multiple times (e.g. squash the rename commit)

src/builder.rs
randomlogin: In that case we will not be able to identify probes in recent payments if we use `send_payment_with_route`. We could store sent probes' payment ids and persist them manually, but that's what I tried to avoid and use existing ways.

Though we still can find a path via `find_route` storing the needed parameters (e.g. scorer) in Prober at build time, but as this is only relevant for one strategy, I'd propose another solution.

Another solution is to have a dedicated router just for probing of `HighDegreeStrategy`, which shares the reference to the same scorer (so the results from probing are visible for "ordinary payment scorer"). It feels as a smaller change which is relevant only to this strategy.

src/probing.rs
randomlogin: Fixed.

src/probing.rs
randomlogin: For that I think we need to have a change upstream, to be able deduce the fee of payment. It feels that other accounting scenarios can benefit from this change, and overall it seems weird to have just amount, without total amount paid.

Opened a PR upstream https://github.com/lightningdevkit/rust-lightning/pull/4674

randomlogin:
Fixed mentioned above discrepancy between uniffi and ordinary docs and squashed commits.
Added a dedicated router to `HighDegreeStrategy`.

src/probing.rs
tnull: That PR landed now, probably worth waiting for it to be backported to 0.3.

src/builder.rs
tnull: > In that case we will not be able to identify probes in recent payments if we use `send_payment_with_route`. We could store sent probes' payment ids and persist them manually, but that's what I tried to avoid and use existing ways.

Ah, duh, yes, sorry, main point was using a separate router.

> Another solution is to have a dedicated router just for probing of HighDegreeStrategy, which shares the reference to the same scorer (so the results from probing are visible for "ordinary payment scorer"). It feels as a smaller change which is relevant only to this strategy.

Yes, that makes sense, arguably you could use the router for all probing-related operations.

src/probing.rs
```diff
@@ -308,14 +308,14 @@ impl ArcedProbingConfigBuilder {
 	///
 	/// Defaults to 10 seconds.
 	pub fn set_interval(&self, secs: u64) {
-		self.inner.write().unwrap().interval(Duration::from_secs(secs));
+		self.inner.write().expect("lock").interval(Duration::from_secs(secs));
```
tnull, line 304: nit: These fixups should have been squashed into the first commit where the original change happened. In general this commit should only have test changes.

src/builder.rs
```diff
@@ -2202,10 +2200,23 @@ fn build_with_store_internal(
 	let prober = probing_config.map(|probing_cfg| {
 		let strategy: Arc<dyn ProbingStrategy> = match &probing_cfg.kind {
 			ProbingStrategyKind::HighDegree { top_node_count } => {
+				// Dedicated router for probing so the diversity penalty doesn't interfere
+				// with real payments; shares the scorer so probe results still train it.
+				let mut probing_fee_params = ProbabilisticScoringFeeParameters::default();
```
tnull, line 2270: These changes should be part of the initial commit.

tnull (commented):
Looks good I think, but this needs a small rebase by now. We'll probably also want to wait until the backport of the upstream PR for accurate accounting landed.

In the meantime please fix the commit history, many fixups seem to have been squashed into the wrong commit. Maybe having two commits (add probing, add probing tests) would be easiest.

src/probing.rs
```diff
@@ -0,0 +1,829 @@
...
+				let path = match prober.strategy.next_probe() {
+					Some(p) => p,
+					None => continue,
+				};
+				let amount: u64 = path.hops.iter().map(|h| h.fee_msat).sum();
+				if prober.locked_msat() + amount > prober.max_locked_msat {
+					log_debug!(prober.logger, "Skipping probe: locked-msat budget exceeded.");
+					continue;
+				}
+				match prober.channel_manager.send_probe(path.clone()) {
```
tnull, line 812: Codex:

>  - [P2] Skip one-hop paths before probing — /home/tnull/workspace/ldk-node/src/probing.rs:807-807
>     When a strategy returns a non-blinded path with fewer than two hops—for example, high-degree routing to a directly connected LSP or a random walk that dead-ends at the first hop—ChannelManager::send_probe always rejects it as unnecessary. In those configurations the background loop just
>     wakes up and logs send failures instead of dispatching probes, so filter these paths or have the strategies choose another target before calling send_probe.

src/probing.rs
```diff
@@ -0,0 +1,829 @@
...
+
+impl ProbingStrategy for HighDegreeStrategy {
+	fn next_probe(&self) -> Option<Path> {
+		let graph = self.network_graph.read_only();
+
+		let mut nodes_by_degree: Vec<(PublicKey, usize)> = graph
+			.nodes()
+			.unordered_iter()
+			.filter_map(|(id, info)| {
+				PublicKey::try_from(*id).ok().map(|pubkey| (pubkey, info.channels.len()))
```
TheBlueMatt: Might be better to convert things to `NodeId` vs converting the entire graph's `NodeId`s to pubkeys here. The pubkey conversion stuff isn't free as it validates the points are on the curve.

randomlogin: Changed.

src/probing.rs
randomlogin: I'd leave current setup: as there can be genuine situations when strategy returns 1-hop path (for example we're connected to a dead end). On the current tick we have no choice but to do nothing, it is a legitimate failure. 

On the level of strategy we can try to make this situations rare (added checks to both strategies, now they return None). 

About logging/handling: we could move it to `run_prober` but it is essentially the same (perhaps without a logging):
```
		tokio::select! {
			biased;
			_ = stop_rx.changed() => {
				log_debug!(prober.logger, "Stopping background probing.");
				return;
			}
			_ = ticker.tick() => {
				let path = match prober.strategy.next_probe() {
					Some(p) => p,
					None => continue,
				};
				...
				//something like this
				if path.hops.len() < 2 {
				   continue;
				}
				...
```

src/builder.rs
randomlogin: fixed

src/probing.rs
randomlogin: fixed

tnull:
This needs a rebase, preferably after #971 landed which will include the pending LDK changes.

tnull:
#971 landed, so should be good to rebase.

randomlogin:
Rebased. Also recreated `set_probing_config` method that I accidentally deleted during one of rebases.
Let's wait for CI to complete.

src/builder.rs
```diff
@@ -2219,6 +2262,51 @@ fn build_with_store_internal(
 		_leak_checker.0.push(Arc::downgrade(&wallet) as Weak<dyn Any + Send + Sync>);
 	}
 
+	let prober = probing_config.map(|probing_cfg| {
```
tnull, line 2265: nit: Would be better to move this up before the leak checker setup.

src/probing.rs
```diff
@@ -0,0 +1,834 @@
...
+			_ = stop_rx.changed() => {
+				log_debug!(prober.logger, "Stopping background probing.");
+				return;
+			}
+			_ = ticker.tick() => {
+				let path = match prober.strategy.next_probe() {
+					Some(p) => p,
+					None => continue,
+				};
+				let amount: u64 = path.hops.iter().map(|h| h.fee_msat).sum();
```
tnull, line 807: Codex:

> 
> Medium: /home/tnull/worktrees/ldk-node/pr-815/src/probing.rs:807 undercounts custom blinded probe paths. sum(hop.fee_msat) only equals total locked value for unblinded paths. A custom ProbingStrategy can return a blinded Path, where the blinded final value is excluded, so max_locked_msat
>     can be exceeded before locked_msat() catches up.

src/probing.rs
```diff
@@ -0,0 +1,834 @@
...
+			} {
+				Some(u) => u,
+				None => break,
+			};
+
+			if !update.enabled {
+				break;
+			}
+
+			route_least_htlc_upper_bound =
```
tnull, line 609: Codex:

>   - Medium: /home/tnull/worktrees/ldk-node/pr-815/src/probing.rs:609 caps the random-walk delivered amount against each channel’s htlc_maximum_msat, but later adds downstream fees at /home/tnull/worktrees/ldk-node/pr-815/src/probing.rs:691 and only rechecks the first hop at /home/tnull/
>     worktrees/ldk-node/pr-815/src/probing.rs:714. Intermediate HTLCs can exceed advertised maxima; the fee math also uses unchecked u64 arithmetic on gossip-sourced fee fields.

tests/probing_tests.rs
```diff
@@ -0,0 +1,480 @@
...
+				maybe_announced_channel: true,
+			},
+		],
+		blinded_tail: None,
+	}
+}
+
+/// Verifies that `locked_msat` increases when a probe is dispatched and returns
+/// to zero once the probe resolves (succeeds or fails).
+#[tokio::test(flavor = "multi_thread")]
```
tnull, line 124: Codex:    
>   - Test reliability: /home/tnull/worktrees/ldk-node/pr-815/tests/probing_tests.rs:124 is bind-sensitive. cargo test --test probing_tests failed twice with InvalidSocketAddress from /home/tnull/worktrees/ldk-node/pr-815/tests/common/mod.rs:739. The failed test passed when run alone, so this
>     looks like fixed-port test setup contention, but the added test target is not reliably green as-is.

tnull (commented):
Looks good, mod some minor nits/comments that can happen in a follow-up. 

Will land this once CI passes (a bit more). Kicked it once more for now, unfortunately it's a bit flaky right now, so the failures so far seem unrelated.

tnull (approved):

src/probing.rs
randomlogin: fixed

src/probing.rs
randomlogin: fixed

src/builder.rs
randomlogin: moved

Issue #767, opened 2026-01-22
Add chain syncing via `bip157` crate


tnull:
We should add support for syncing via BIP157/158 via the [`kyoto`/`bip157` ](https://docs.rs/bip157/0.3.4/bip157/) crate.

(cc @rustaceanrob)

PR #821, opened 2026-03-09
Add syncing via compact block filters (bip157)

randomlogin:
Related issue: https://github.com/lightningdevkit/ldk-node/issues/767

Introduce new backend for syncing via compact filters (bip157).

Create new chain source which start kyoto under the hood.

Fees retrieval has two options: either esplora/electrum or a rough estimate via the latest block from kyoto.

Test bitcoind node is run with filter flags.


rustaceanrob:
Epic. Version `0.4.1` is released which is configurable to fetch witnesses when requesting blocks.

PR #822, opened 2026-03-10
Add wallet birthday height for seed recovery on pruned nodes

### LINKED ISSUE #818: Wallet birthday height for seed recovery on pruned nodes
When restoring an ldk-node wallet from seed, the node starts syncing from the current chain tip. Any on-chain deposits made before the restore are never seen by `synchronize_listeners`, so the wallet balance shows zero even though funds exist on-chain.

This affects all seed restores, but pruned nodes are hit hardest: historical blocks are gone, so there is no workaround via rescan.

There is an existing TODO acknowledging this:

```rust
// TODO: Use a proper wallet birthday once BDK supports it.
```
(`src/builder.rs:1327`)

### Impact

- On-chain balance shows 0 after seed restore
- Cannot open Lightning channels (no visible funds)
- Funds are safe on-chain but inaccessible through ldk-node
- Mobile wallets (where reinstalls and data loss are common) are the primary use case

### What we did

We hit this building [Bitcoin Pocket Node](https://github.com/FreeOnlineUser/bitcoin-pocket-node), an Android app running bitcoind + ldk-node on GrapheneOS with pruned storage.

Our workaround overrides `chain_tip_opt` in the Builder when a birthday height is provided. Caller flow:

1. User restores seed words
2. App uses bitcoind `scantxoutset` to find UTXOs and their block heights (works on pruned nodes)
3. Sets birthday to `min(heights) - 10` as safety margin
4. ldk-node syncs from there, balance appears with zero transaction fees

Working fork with the changes (~100 lines): [FreeOnlineUser/ldk-node@upstream/wallet-birthday](https://github.com/FreeOnlineUser/ldk-node/tree/upstream/wallet-birthday)

Verified end-to-end on real hardware: seed restore, UTXO scan, birthday set, 110,628 sats recovered in seconds.

### Upstream context

BDK is working on related pieces:
- [bdk#2126](https://github.com/bitcoindevkit/bdk/issues/2126): Adding `start_height` to `FilterIter::new`
- [bdk#2050](https://github.com/bitcoindevkit/bdk/issues/2050): Pruned node support via `scantxoutset`

Once BDK ships native birthday support, ldk-node can use that directly (which is what the TODO anticipates). Our fork is a pragmatic solution in the meantime.

We are also the folks working on the watchtower improvements discussed in #813.

FreeOnlineUser:
Adds `set_wallet_birthday_height(height)` to the builder, allowing the on-chain wallet to sync from a specific block height instead of the current tip or genesis.

This complements the existing `set_wallet_recovery_mode()` from #769 by supporting pruned nodes where scanning from genesis fails due to missing blocks.

### Motivation

When restoring a wallet from seed on a pruned node:
- `recovery_mode` (scan from genesis) fails: blocks are pruned
- Default behavior (checkpoint at tip) misses historical funds

With a birthday height, the wallet checkpoints at a known block before the first transaction, recovering funds without needing the full chain.

### How it works

Three-way logic in wallet creation:
1. **Birthday set**: fetch block hash at birthday height, checkpoint there
2. **No birthday, no recovery mode**: checkpoint at current tip (existing behavior)
3. **Recovery mode without birthday**: sync from genesis (existing behavior)

Falls back to current tip if the birthday block hash cannot be fetched.

### Real-world use

Built and verified in [Bitcoin Pocket Node](https://github.com/FreeOnlineUser/bitcoin-pocket-node), an Android app running bitcoind + ldk-node on GrapheneOS with pruned storage. Recovery flow:

1. User restores seed
2. App runs `scantxoutset` to find UTXOs and their block heights
3. Sets birthday to `min(heights) - 10`
4. ldk-node syncs from that height, balance appears

Tested end-to-end on real hardware: 10,844 sats recovered in seconds.

Closes #818

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -639,6 +639,11 @@ impl NodeBuilder {
 	/// first transaction. If unknown, use a conservative estimate.
 	///
 	/// This only takes effect when creating a new wallet (not when loading existing state).
+	///
+	/// If [`set_wallet_recovery_mode`] is also set, the birthday height takes precedence and
+	/// the wallet is checkpointed at the birthday block.
+	///
+	/// [`set_wallet_recovery_mode`]: Self::set_wallet_recovery_mode
 	pub fn set_wallet_birthday_height(&mut self, height: u32) -> &mut Self {
 		self.wallet_birthday_height = Some(height);
 		self
@@ -1386,8 +1391,8 @@ fn build_with_store_internal(
 	liquidity_source_config: Option<&LiquiditySourceConfig>,
 	pathfinding_scores_sync_config: Option<&PathfindingScoresSyncConfig>,
 	async_payments_role: Option<AsyncPaymentsRole>, recovery_mode: bool,
-	wallet_birthday_height: Option<u32>, seed_bytes: [u8; 64],
-	runtime: Arc<Runtime>, logger: Arc<Logger>, kv_store: Arc<DynStore>,
+	wallet_birthday_height: Option<u32>, seed_bytes: [u8; 64], runtime: Arc<Runtime>,
+	logger: Arc<Logger>, kv_store: Arc<DynStore>,
 ) -> Result<Node, BuildError> {
 	optionally_install_rustls_cryptoprovider();
 
@@ -1619,10 +1624,8 @@ fn build_with_store_internal(
 							birthday_hash
 						);
 						let mut latest_checkpoint = wallet.latest_checkpoint();
-						let block_id = bdk_chain::BlockId {
-							height: birthday_height,
-							hash: birthday_hash,
-						};
+						let block_id =
+							bdk_chain::BlockId { height: birthday_height, hash: birthday_hash };
 						latest_checkpoint = latest_checkpoint.insert(block_id);
 						let update = bdk_wallet::Update {
 							chain: Some(latest_checkpoint),
@@ -1634,30 +1637,17 @@ fn build_with_store_internal(
 						})?;
 					},
 					Err(e) => {
+						// A birthday was explicitly set but we couldn't fetch the block hash
+						// for it. Silently checkpointing at the chain tip would defeat the
+						// feature: the wallet would scan from tip and recover nothing, leaving
+						// the user to assume their seed is wrong. Fail loudly instead.
 						log_error!(
 							logger,
-							"Failed to fetch block hash at birthday height {}: {:?}. \
-							 Falling back to current tip.",
+							"Failed to fetch block hash at birthday height {}: {}",
 							birthday_height,
 							e
 						);
-						// Fall back to current tip
-						if let Some(best_block) = chain_tip_opt {
-							let mut latest_checkpoint = wallet.latest_checkpoint();
-							let block_id = bdk_chain::BlockId {
-								height: best_block.height,
-								hash: best_block.block_hash,
-							};
-							latest_checkpoint = latest_checkpoint.insert(block_id);
-							let update = bdk_wallet::Update {
-								chain: Some(latest_checkpoint),
-								..Default::default()
-							};
-							wallet.apply_update(update).map_err(|e| {
-								log_error!(logger, "Failed to apply fallback checkpoint: {}", e);
-								BuildError::WalletSetupFailed
-							})?;
-						}
+						return Err(BuildError::WalletSetupFailed);
 					},
 				}
 			} else if !recovery_mode {
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -72,7 +72,7 @@ use crate::liquidity::{
 	LSPS1ClientConfig, LSPS2ClientConfig, LSPS2ServiceConfig, LiquiditySourceBuilder,
 };
 use crate::lnurl_auth::LnurlAuth;
-use crate::logger::{log_error, LdkLogger, LogLevel, LogWriter, Logger};
+use crate::logger::{log_error, log_info, LdkLogger, LogLevel, LogWriter, Logger};
 use crate::message_handler::NodeCustomMessageHandler;
 use crate::payment::asynchronous::om_mailbox::OnionMessageMailbox;
 use crate::peer_store::PeerStore;
@@ -292,6 +292,7 @@ pub struct NodeBuilder {
 	runtime_handle: Option<tokio::runtime::Handle>,
 	pathfinding_scores_sync_config: Option<PathfindingScoresSyncConfig>,
 	recovery_mode: bool,
+	wallet_birthday_height: Option<u32>,
 }
 
 impl NodeBuilder {
@@ -310,6 +311,7 @@ impl NodeBuilder {
 		let runtime_handle = None;
 		let pathfinding_scores_sync_config = None;
 		let recovery_mode = false;
+		let wallet_birthday_height = None;
 		Self {
 			config,
 			chain_data_source_config,
@@ -320,6 +322,7 @@ impl NodeBuilder {
 			async_payments_role: None,
 			pathfinding_scores_sync_config,
 			recovery_mode,
+			wallet_birthday_height,
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


src/builder.rs
```diff
@@ -1359,10 +1387,65 @@ fn build_with_store_internal(
...
+							let update = bdk_wallet::Update {
+								chain: Some(latest_checkpoint),
+								..Default::default()
+							};
+							wallet.apply_update(update).map_err(|e| {
+								log_error!(logger, "Failed to apply fallback checkpoint: {}", e);
+								BuildError::WalletSetupFailed
+							})?;
+						}
+					},
```
Jolah1, line 1651: If the user set a wallet birthday they almost certainly don't want to be silently checkpointed at chain tip on a transient hash-fetch failure, that defeats the whole point of the feature. A mobile recovery user would see 0 sats restored, assume their seed is wrong, and likely never look at the logs.
Two safer alternatives:
1. Return BuildError::WalletSetupFailed so the caller can retry (or fall back at the application layer).
2. Fall back to no checkpoint, i.e. the recovery_mode genesis-sync behavior, so the user still gets historical coverage at the cost of a full scan.
I'd lean toward option 1 (surface the error). Falling back to tip is the worst outcome for the user. WDYT?

src/chain/mod.rs
```diff
@@ -214,6 +215,24 @@ impl ChainSource {
 		}
 	}
 
+	/// Fetches the block hash at the given height from the chain source.
+	pub(crate) async fn get_block_hash_by_height(
```
Jolah1, line 219: Every other method on ChainSource in this file returns Result<_, Error> (e.g. start at L188, update_fee_rate_estimates at L446). Result<_, ()> here drops useful context from all three backends: UtxoSource::get_block_hash_by_height returns BlockSourceResult<BlockHash>, esplora_client.get_block_hash returns Result<_, EsploraError>, and block_header_raw returns Result<_, electrum_client::Error>. Surfacing those via crate::Error would be consistent and would let the call site reason about the failure mode (relevant to the comment on builder.rs:1419-1444).

src/builder.rs
```diff
@@ -579,6 +582,22 @@ impl NodeBuilder {
 		self
 	}
 
+	/// Sets the wallet birthday height for seed recovery on pruned nodes.
```
Jolah1, line 631: The current code makes birthday take precedence over set_wallet_recovery_mode (the if let Some(birthday_height) = … else if !recovery_mode { … } chain in build_with_store_internal). Worth a sentence here noting the precedence so callers who set both know what to expect.
"Use a conservative estimate" is direction-ambiguous. A height after the first tx would miss funds. 
Maybe: "use a height earlier than your first transaction,  lower is safer than higher."

src/chain/mod.rs
```diff
@@ -17,6 +17,7 @@ use bitcoin::{Script, Txid};
 use lightning::chain::{BestBlock, Filter};
```
Jolah1: Heads up: cargo fmt --all -- --check fails on this branch in 4 spots (builder.rs:1181, builder.rs:1402, chain/mod.rs:17, chain/mod.rs:29), which is what's red on build (ubuntu-latest, stable) in CI. A cargo fmt --all pass + rebase should clear it.

Jolah1 (commented):
Nice Work taking this up!

Took a look at the changes, the feature is well-motivated and the real-hardware validation is great, but a few things I'd like to see addressed before this lands.

- The silent fallback to chain tip when the birthday hash fetch fails (src/builder.rs:1419-1444) silently nullifies the feature for the exact users it's meant to help. Detail inline.
- Result<_, ()> on ChainSource::get_block_hash_by_height is inconsistent with the rest of the file and drops error context that would help the call site above.
- Precedence between set_wallet_birthday_height and set_wallet_recovery_mode isn't documented.
- set_wallet_birthday_height isn't exposed in bindings/ldk_node.udl — and the motivating use case is mobile (Android via Kotlin FFI).
- CI is red on cargo fmt --check in four places.

Missing: there are no tests for the new setter, the build-time branching, or the new chain-source method. The branching has four distinct paths (birthday set / birthday + fetch failure / no birthday + tip / no birthday + 
recovery) and at minimum a unit test that the setter updates the field would help. Integration coverage for the birthday-checkpoint path would be ideal given the existing regtest harness, but I won't block on that.

Also: branch will need a rebase onto current main once the above is sorted rebaseable: false on the API right now.

FreeOnlineUser:
@Jolah1 Thank you, this was a genuinely useful review and I appreciate you picking up a PR that had gone quiet. The silent-fallback catch especially: you're right that degrading to chain tip turns a recovery feature into a 0-sats surprise for exactly the user it's meant to help.

Up front: I use AI assistance (Claude) in this work and verify its output against the code before it goes up. Flagging it so you know what you're reviewing.

All points addressed in the latest push:

- **Silent fallback (builder.rs):** now returns `BuildError::WalletSetupFailed` when a birthday is set but the block hash can't be fetched, instead of silently checkpointing at the tip.
- **`Result<_, ()>`:** `ChainSource::get_block_hash_by_height` now returns a typed `Error::ChainAccessFailed`, consistent with the other `ChainSource` methods, and mirrored into the UDL error enum.
- **Precedence:** documented that the birthday height takes precedence over recovery mode.
- **UDL exposure:** `set_wallet_birthday_height` is now in `bindings/ldk_node.udl`.
- **fmt:** `cargo fmt --all` pass; the four flagged spots are clean.
- **Tests:** added `onchain_wallet_recovery_with_birthday`, which recovers funds via a wallet birthday over `random_chain_source` (exercised against bitcoind, esplora, and electrum).
- **Rebase:** rebased onto current main.

Thanks again.

src/builder.rs
FreeOnlineUser: Done. Now returns `BuildError::WalletSetupFailed` when a birthday is set but the block hash can't be fetched, rather than silently checkpointing at the tip.

src/chain/mod.rs
FreeOnlineUser: Done. Returns `Error::ChainAccessFailed` now, consistent with the other `ChainSource` methods, and mirrored into the UDL error enum.

src/builder.rs
FreeOnlineUser: Documented: the birthday height takes precedence over recovery mode.

src/chain/mod.rs
FreeOnlineUser: Fixed with a `cargo fmt --all` pass, and regrouped the `lightning_block_sync` import while here.

Jolah1:
> @Jolah1 Thank you, this was a genuinely useful review and I appreciate you picking up a PR that had gone quiet. The silent-fallback catch especially: you're right that degrading to chain tip turns a recovery feature into a 0-sats surprise for exactly the user it's meant to help.
> 
> Up front: I use AI assistance (Claude) in this work and verify its output against the code before it goes up. Flagging it so you know what you're reviewing.
> 
> All points addressed in the latest push:
> 
> * **Silent fallback (builder.rs):** now returns `BuildError::WalletSetupFailed` when a birthday is set but the block hash can't be fetched, instead of silently checkpointing at the tip.
> * **`Result<_, ()>`:** `ChainSource::get_block_hash_by_height` now returns a typed `Error::ChainAccessFailed`, consistent with the other `ChainSource` methods, and mirrored into the UDL error enum.
> * **Precedence:** documented that the birthday height takes precedence over recovery mode.
> * **UDL exposure:** `set_wallet_birthday_height` is now in `bindings/ldk_node.udl`.
> * **fmt:** `cargo fmt --all` pass; the four flagged spots are clean.
> * **Tests:** added `onchain_wallet_recovery_with_birthday`, which recovers funds via a wallet birthday over `random_chain_source` (exercised against bitcoind, esplora, and electrum).
> * **Rebase:** rebased onto current main.
> 
> Thanks again.

Thanks, the updates look good and address the points raised. Two things before this is ready:
- CI Linting is red on the "Ban unwrap in library code" check two new unwraps need to be .expect("…") to match the codebase convention: src/builder.rs (self.inner.write().unwrap()…) and src/chain/electrum.rs (self.electrum_runtime_status.read().unwrap()).
- The branch now conflicts with main (mergeable: CONFLICTING) — needs a rebase onto current main, which will also re-run the stale VSS jobs.

tnull:
This is/will be superseded by https://github.com/lightningdevkit/ldk-node/pull/884

FreeOnlineUser:
Closing in favor of #884.

@tnull noted above that #884 supersedes this, and having read it over I agree it's the more complete approach: it treats a failed chain-tip fetch as a hard error instead of silently falling back to tip, validates the rescan height against the tip, keeps the config in `config.rs`, and has broader test coverage. No sense keeping two parallel versions open.

@Jolah1 you were right on both your last points (the library-unwrap lint and the rebase), and the silent-fallback catch in the first round was a good one. Thank you for both reviews. Rather than have you spend more time here, #884 is where this is landing, so follow it there if you'd like.

Jolah1:
> Closing in favor of #884.
> 
> @tnull noted above that #884 supersedes this, and having read it over I agree it's the more complete approach: it treats a failed chain-tip fetch as a hard error instead of silently falling back to tip, validates the rescan height against the tip, keeps the config in `config.rs`, and has broader test coverage. No sense keeping two parallel versions open.
> 
> @Jolah1 you were right on both your last points (the library-unwrap lint and the rebase), and the silent-fallback catch in the first round was a good one. Thank you for both reviews. Rather than have you spend more time here, #884 is where this is landing, so follow it there if you'd like.
> 

Agreed, #884 is the more complete approach. Glad the reviews were useful.
Thanks.

PR #825, opened 2026-03-13
Bump rust-lightning to include trampoline changes

carlaKC:
See #838

~Also needs the following patch on top of [2026-02-ldk-node-base](https://github.com/jkczyz/bitcoin-payment-instructions/tree/2026-02-ldk-node-base) in `bitcoin-payment-instructions`.~ 
Pointing to my own `bitcoin-payment-instructions` based on previous branch as discussed offline.


src/event.rs
```diff
@@ -61,6 +64,71 @@ use crate::{
 	UserChannelId,
 };
 
+/// A set of multiple htlcs all associated with same forward.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct HTLCSet(Vec<HTLCLocator>);
```
carlaKC: Interested in some feedback here 🙏 

To use `default_value` to fill in legacy fields when the new field isn't present, we need to implement `Readable`/`Writeable` for `Vec<HTLCLocator>`. We can't do this because we don't own the trait or the type (orphan rule).

The solution I've gone with here is to use a wrapper struct and pull in the code we have in `impl_for_vec` in LDK. 

Alternatives would be:
* Don't use `default_value`, and write manual `Readable`/`Writeable` impls for `Event` so that we can manually fill the legacy values (I can't find a way to do it within macros).
* Just use `required_vec` and don't fill legacy values (seems bad).

src/event.rs
```diff
@@ -323,16 +376,58 @@ impl_writeable_tlv_based_enum!(Event,
...
 	},
 	(7, PaymentForwarded) => {
-		(0, prev_channel_id, required),
-		(1, prev_node_id, option),
-		(2, next_channel_id, required),
-		(3, next_node_id, option),
-		(4, prev_user_channel_id, option),
-		(6, next_user_channel_id, option),
+		// For backwards compatibility, write the first prev/next_htlc to our legacy fields. This
+		// allows use to downgrade with some information loss about the remaining htlcs.
```
carlaKC: Not sure what the project's philosophy is for backwards compatibility, so just gone with what I would have done in LDK - happy to do something else if it's preferred!

Cargo.toml
```diff
@@ -79,13 +79,14 @@ async-trait = { version = "0.1", default-features = false }
 vss-client = { package = "vss-client-ng", version = "0.5" }
 prost = { version = "0.11.6", default-features = false}
 #bitcoin-payment-instructions = { version = "0.6" }
-bitcoin-payment-instructions = { git = "https://github.com/jkczyz/bitcoin-payment-instructions", rev = "0138feb7acefb1e49102a6fb46d7b776bf43265e" }
+# TODO: need to push branch to Jeff's fork?
+bitcoin-payment-instructions = { git = "https://github.com/carlaKC/bitcoin-payment-instructions", rev = "c22c9b836b70d4c915dd28701e11083a8b558d56" }
```
jkczyz: @tnull had been doing most of the updates until recently when my PRs (and others) started breaking LDK Node. I ended up setting up a remote to his branch and based mine off of it. Personally, I'd be fine if you want to do something similar, as that at least gives us a common history for `bitcoin-payment-instructions` bumps.

Cargo.toml
```diff
@@ -182,4 +183,4 @@ harness = false
 #lightning-block-sync = { path = "../rust-lightning/lightning-block-sync" }
 #lightning-transaction-sync = { path = "../rust-lightning/lightning-transaction-sync" }
 #lightning-liquidity = { path = "../rust-lightning/lightning-liquidity" }
-#lightning-macros = { path = "../rust-lightning/lightning-macros" }
+lightning-macros = { path = "../rust-lightning/lightning-macros" }
```
tnull: nit: This seems unrelated.

src/event.rs
```diff
@@ -61,6 +64,70 @@ use crate::{
...
+	/// payment was settled via an on-chain transaction.
+	pub user_channel_id: Option<UserChannelId>,
+	/// The node id of the counterparty for this HTLC.
+	///
+	/// This is only `None` for HTLCs received prior to LDK Node v0.5 or for events serialized by
+	/// versions prior to v0.5.
+	pub node_id: Option<PublicKey>,
+}
+
+impl_writeable_tlv_based!(HTLCLocator, {
```
tnull: Hmm, I'd really like to avoid duplicating upstream's serialization logic, which always risks to get out-of-sync. Can we rather also `impl From<Vec<HTLCLocator>> for Vec<LdkHTLCLocator>` and use upstream's `write`? I admit that is also not great, but maybe preferable to duplicating even more logic?

src/event.rs
```diff
@@ -61,6 +64,70 @@ use crate::{
 	UserChannelId,
 };
 
+/// A set of multiple htlcs all associated with same forward.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct HTLCSet(pub Vec<HTLCLocator>);
```
tnull: Why do we need this additional newtype? I think even for bindings we should be able to leave it as `Vec<HTLCLocator>`?

src/event.rs
```diff
@@ -61,6 +64,70 @@ use crate::{
...
+	pub node_id: Option<PublicKey>,
+}
+
+impl_writeable_tlv_based!(HTLCLocator, {
+	(1, channel_id, required),
+	(3, user_channel_id, option),
+	(5, node_id, option),
+});
+
+impl From<lightning::events::HTLCLocator> for HTLCLocator {
```
tnull: nit: Please import the upstream type as `LdkHTLCLocator`, which is the usual pattern.

src/event.rs
tnull: Yeah, while we def. want to get to support downgrades at some point, we currently don't. So we can just drop this extra logic as long as we're positive backwards compatibility during upgrades is ensured.

tnull (commented):
Thanks for looking into this! Needs a rebase by now as #824 was merged.

src/event.rs
carlaKC: This was to get around the orphan rule for serialization (we can't implement a trait we don't own - `Readable`/`Writeable` - for a type we don't own - `Vec<T>`).

If we don't need backwards compat, then we don't need `default_value` and can clean up a whole bunch of things including this 👌

src/event.rs
carlaKC: > I admit that is also not great, but maybe preferable to duplicating even more logic?

Given this a shot, it's a little unwieldy but I agree probably preferable to duplication.

tnull: Indeed unwieldy and not loving the reallocation either, but oh well...

But we can now remove the `impl_writeable_tlv_based` here, right?

src/event.rs
```diff
@@ -1378,18 +1451,16 @@ where
 
 				if let Some(liquidity_source) = self.liquidity_source.as_ref() {
 					let skimmed_fee_msat = skimmed_fee_msat.unwrap_or(0);
-					liquidity_source
-						.handle_payment_forwarded(next_channel_id, skimmed_fee_msat)
-						.await;
+					for next_htlc in next_htlcs.iter() {
+						liquidity_source
+							.handle_payment_forwarded(Some(next_htlc.channel_id), skimmed_fee_msat)
```
tnull: Hmm, wouldn't this have us / the accounting logic believe that we withheld the skimmed amount multiple times?

src/event.rs
tnull: Does this also need adjustments?

src/event.rs
```diff
@@ -323,16 +352,63 @@ impl_writeable_tlv_based_enum!(Event,
...
+		(4, legacy_prev_user_channel_id, (legacy, u128, |_| Ok(()), |_: &Event| None::<Option<u128>>)),
+		(6, legacy_next_user_channel_id, (legacy, u128, |_| Ok(()), |_: &Event| None::<Option<u128>>)),
 		(8, total_fee_earned_msat, option),
 		(10, skimmed_fee_msat, option),
 		(12, claim_from_onchain_tx, required),
 		(14, outbound_amount_forwarded_msat, option),
+		// We cannot implement Readable/Writeable for Vec<HTLCLocator> because we do not own the
+		// trait or the type (Vec). To work around this, and prevent duplicating serialization code,
+		// we map to the underlying LdkHTLCLocator type for serialization.
+		(15, prev_htlcs, (custom, Vec<LdkHTLCLocator>,
```
tnull: Can we try to streamline this a bit? Maybe introduce a helper method to dedup the code? At the very least it seems we can drop a few lines:

```diff
diff --git a/src/event.rs b/src/event.rs
index 417f5941..0505a870 100644
--- a/src/event.rs
+++ b/src/event.rs
@@ -369,16 +369,14 @@ impl_writeable_tlv_based_enum!(Event,
                // we map to the underlying LdkHTLCLocator type for serialization.
                (15, prev_htlcs, (custom, Vec<LdkHTLCLocator>,
                        |v: Option<Vec<LdkHTLCLocator>>| {
-                               let res: Result<Vec<HTLCLocator>, lightning::ln::msgs::DecodeError> =
-                                       Ok(v.map(|ldk_vec| ldk_vec.into_iter().map(HTLCLocator::from).collect())
+                               Ok(v.map(|ldk_vec| ldk_vec.into_iter().map(HTLCLocator::from).collect())
                                                .unwrap_or_else(|| {
                                                        legacy_prev_channel_id.map(|ch| vec![HTLCLocator {
                                                                channel_id: ch,
                                                                user_channel_id: legacy_prev_user_channel_id.map(UserChannelId),
                                                                node_id: legacy_prev_node_id,
                                                        }]).unwrap_or_default()
-                                               }));
-                               res
+                                               }))
                        },
                        |us: &Event| {
                                if let Event::PaymentForwarded { ref prev_htlcs, .. } = us {
@@ -390,7 +388,6 @@ impl_writeable_tlv_based_enum!(Event,
                )),
                (17, next_htlcs, (custom, Vec<LdkHTLCLocator>,
                        |v: Option<Vec<LdkHTLCLocator>>| {
-                               let res: Result<Vec<HTLCLocator>, lightning::ln::msgs::DecodeError> =
                                        Ok(v.map(|ldk_vec| ldk_vec.into_iter().map(HTLCLocator::from).collect())
                                                .unwrap_or_else(|| {
                                                        legacy_next_channel_id.map(|ch| vec![HTLCLocator {
@@ -398,8 +395,7 @@ impl_writeable_tlv_based_enum!(Event,
                                                                user_channel_id: legacy_next_user_channel_id.map(UserChannelId),
                                                                node_id: legacy_next_node_id,
                                                        }]).unwrap_or_default()
-                                               }));
-                               res
+                                               }))
                        },
                        |us: &Event| {
                                if let Event::PaymentForwarded { ref next_htlcs, .. } = us {
```

src/event.rs
```diff
@@ -61,6 +61,50 @@ use crate::{
...
+	pub user_channel_id: Option<UserChannelId>,
+	/// The node id of the counterparty for this HTLC.
+	///
+	/// This is only `None` for HTLCs received prior to LDK Node v0.5 or for events serialized by
+	/// versions prior to v0.5.
+	pub node_id: Option<PublicKey>,
+}
+
+impl_writeable_tlv_based!(HTLCLocator, {
+	(1, channel_id, required),
```
tnull: Should we make these even, given its a new struct?

carlaKC: In LDK I'd only go with an even TLV if not understanding the field means that we'll lose money, because odd gives us much more flexibility for making changes in the future if needed. Happy to change to even if that's not the philosophy here!

src/event.rs
carlaKC: Hm yes, but I don't think we can hit it unless lsps2 supports opening multiple JIT channels and also being a trampoline at the same time.

Would you be okay with passing multiple `next_htlc.channel_id` to `handle_payment_forwarded` and then just logging an error if there's more than one?

src/event.rs
tnull: > because odd gives us much more flexibility for making changes in the future if needed

Hmm, you probably can't ever really repurpose an odd field in LDK either so, it's mostly about upgrades/backwards compat of pre-existing objects, no?

TheBlueMatt: Yea the reason we've been doing it in the lightning crate is because it means were could remove them in a future version, making it a good default.

PR #826, opened 2026-03-14
Fix inbound on-chain payment txid not updating after RBF replacement

febyeji:
Found via CI failure in the `onchain_fee_bump_rbf` test: 

- When a sender performs multiple RBF fee bumps, the receiver may not update its inbound payment record to reflect the final confirmed replacement transaction. This causes `onchain_fee_bump_rbf` to fail intermittently:
```
assertion `left == right` failed: node_a inbound payment txid should be updated to the second replacement txid
```

**Fix**
- `find_payment_by_txid` now also checks the payment_store (not just `pending_payment_store`) for onchain payments matching the target txid.
- New `find_payment_by_conflicting_tx` method: if a transaction shares inputs with an existing payment's transaction, it must be an RBF replacement. This works regardless of whether intermediate replacements were observed.

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1213,11 +1213,19 @@ impl Wallet {
 		}
 
 		// Also check the payment store for onchain payments with this txid.
+		// First try direct lookup by payment_id, then fall back to scanning by txid.
+		if let Some(payment) = self.payment_store.get(&direct_payment_id) {
+			if payment.status != PaymentStatus::Succeeded {
+				return Some(payment.id);
+			}
+		}
+
 		if let Some(payment) = self
 			.payment_store
-			.list_filter(
-				|p| matches!(&p.kind, PaymentKind::Onchain { txid, .. } if *txid == target_txid),
-			)
+			.list_filter(|p| {
+				matches!(&p.kind, PaymentKind::Onchain { txid, .. } if *txid == target_txid)
+					&& p.status != PaymentStatus::Succeeded
+			})
 			.first()
 		{
 			return Some(payment.id);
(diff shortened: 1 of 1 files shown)


src/wallet/mod.rs
```diff
@@ -1209,6 +1212,48 @@ impl Wallet {
 			return Some(replaced_details.details.id);
 		}
 
+		// Also check the payment store for onchain payments with this txid.
```
Camillarhi, line 1215: We can also first check with the `payment_id` then the `txid`

src/wallet/mod.rs
```diff
@@ -1209,6 +1212,48 @@ impl Wallet {
...
+		if let Some(payment) = self
+			.payment_store
+			.list_filter(
+				|p| matches!(&p.kind, PaymentKind::Onchain { txid, .. } if *txid == target_txid),
```
Camillarhi: Also, ensure the `PaymentStatus` is not `Succeeded`

src/wallet/mod.rs
```diff
@@ -1209,6 +1212,48 @@ impl Wallet {
...
+	fn find_payment_by_conflicting_tx(
+		&self, locked_wallet: &PersistedWallet<KVStoreWalletPersister>, tx: &Transaction,
+	) -> Option<PaymentId> {
+		let target_txid = tx.compute_txid();
+		let spent_outpoints: std::collections::HashSet<OutPoint> =
+			tx.input.iter().map(|input| input.previous_output).collect();
+
+		let onchain_payments = self.payment_store.list_filter(|p| {
+			matches!(p.kind, PaymentKind::Onchain { .. }) && p.status != PaymentStatus::Failed
+		});
```
Camillarhi: This should scan `pending_payment_store` instead. RBF only works on unconfirmed transactions, and `pending_payment_store` already tracks those. On-chain payments get removed from it once they hit `ANTI_REORG_DELAY` confirmations. So anything still replaceable should be there.

In respect to performance, `pending_payment_store` is small and bounded (only in-flight payments), while `payment_store` grows with every payment the node has ever made.

PR #827, opened 2026-03-14
Expose `ChannelDetails::channel_shutdown_state`

### LINKED ISSUE #803: Expose `ChannelDetails::channel_shutdown_state`
We currently don't expose this, so there is good way to tell whether a channel is in the process of shutting down

f3r10:
## Summary

  Closes #803.

  We currently don't expose `ChannelDetails::channel_shutdown_state`, so there is no good way for callers to tell whether a channel is actively going through a cooperative shutdown or merely has its peer temporarily disconnected — both situations cause `is_usable` to become `false`.

  This PR adds:
  - A new `ChannelShutdownState` enum (mirroring LDK's own type) with variants `NotShuttingDown`, `ShutdownInitiated`, `ResolvingHTLCs`, `NegotiatingClosingFee`, and `ShutdownComplete`.
  - An `Option<ChannelShutdownState>` field on `ChannelDetails`, populated from `LdkChannelDetails::channel_shutdown_state` via the existing `From` conversion.
  - A `typedef enum ChannelShutdownState` entry in the UniFFI UDL so the field is available to Swift, Kotlin, and Python bindings.
  - Public re-export of `ChannelShutdownState` from the crate root.


src/types.rs
```diff
@@ -529,6 +565,10 @@ pub struct ChannelDetails {
 	pub inbound_htlc_maximum_msat: Option<u64>,
 	/// Set of configurable parameters that affect channel operation.
 	pub config: ChannelConfig,
+	/// The current shutdown state of the channel, if any.
+	///
+	/// Returns `None` for channels that have not yet started the shutdown process.
```
Camillarhi: Could we replace this to accurately reflect what the source from the library says  `Returns None for ChannelDetails serialized on LDK versions prior to 0.0.116.`

Camillarhi (changes requested):

src/types.rs
```diff
@@ -347,6 +349,40 @@ impl fmt::Display for UserChannelId {
 	}
 }
 
+/// The shutdown state of a channel as returned in [`ChannelDetails::channel_shutdown_state`].
+///
+/// [`ChannelDetails::channel_shutdown_state`]: crate::ChannelDetails::channel_shutdown_state
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+pub enum ChannelShutdownState {
```
tnull: Do we really need to duplicate this enum here, or could we get away with using the upstream one and only using `#[uniffi::remote(Enum)]` for `uniffi`?

f3r10: yeah totally @tnull I am going to do that

src/ffi/types.rs
```diff
@@ -21,10 +21,11 @@ pub use bip39::Mnemonic;
 use bitcoin::hashes::sha256::Hash as Sha256;
 use bitcoin::hashes::Hash;
 use bitcoin::secp256k1::PublicKey;
-pub use bitcoin::{Address, BlockHash, Network, OutPoint, ScriptBuf, Txid};
+pub use bitcoin::{Address, BlockHash, FeeRate, Network, OutPoint, ScriptBuf, Txid};
 pub use lightning::chain::channelmonitor::BalanceSource;
 use lightning::events::PaidBolt12Invoice as LdkPaidBolt12Invoice;
 pub use lightning::events::{ClosureReason, PaymentFailureReason};
+pub use lightning::ln::channel_state::ChannelShutdownState;
```
tnull: nit: I believe we don't need this to be `pub`?

src/types.rs
```diff
@@ -16,6 +16,7 @@ use bitcoin_payment_instructions::onion_message_resolver::LDKOnionMessageDNSSECH
 use lightning::chain::chainmonitor;
 use lightning::impl_writeable_tlv_based;
 use lightning::ln::channel_state::ChannelDetails as LdkChannelDetails;
+pub use lightning::ln::channel_state::ChannelShutdownState;
```
tnull: nit: Let's drop the `pub` here.

src/lib.rs
```diff
@@ -177,7 +173,10 @@ use types::{
 	HRNResolver, KeysManager, OnionMessenger, PaymentStore, PeerManager, Router, Scorer, Sweeper,
 	Wallet,
 };
-pub use types::{ChannelDetails, CustomTlvRecord, PeerDetails, SyncAndAsyncKVStore, UserChannelId};
+pub use types::{
```
tnull: nit: Rather than re-exporting via `crate::types`, lets just add a `pub use lightning::ln::channel_state::ChannelShutdownState;` line.

src/types.rs
```diff
@@ -529,6 +530,10 @@ pub struct ChannelDetails {
 	pub inbound_htlc_maximum_msat: Option<u64>,
 	/// Set of configurable parameters that affect channel operation.
 	pub config: ChannelConfig,
+	/// The current shutdown state of the channel, if any.
+	///
+	/// Returns `None` for `ChannelDetails` serialized on LDK versions prior to 0.0.116.
```
tnull: As this is in LDK Node context, we should provide compatibility notes based on LDK Node's versions. In this case that would mean `LDK Node versions prior to 0.2` (aruably that's so old we could also omit the note altogether).

tests/integration_tests_rust.rs
```diff
@@ -2108,24 +2108,26 @@ async fn lsps2_client_trusts_lsp() {
...
-	assert_eq!(
-		service_node
-			.list_channels()
-			.iter()
-			.find(|c| c.counterparty_node_id == client_node_id)
-			.unwrap()
-			.confirmations,
-		Some(0)
-	);
+	let client_channel = client_node
```
tnull: Could you expand on why the LSPS2 tests in particular do need additional test coverage for channel shutdown?

tests/common/mod.rs
```diff
@@ -1255,6 +1291,20 @@ pub(crate) async fn do_channel_full_cycle<E: ElectrumApi>(
...
+			None | Some(ChannelShutdownState::NotShuttingDown)
+		));
+	}
+	for channel in node_b.list_channels() {
+		assert!(matches!(
+			channel.channel_shutdown_state,
+			None | Some(ChannelShutdownState::NotShuttingDown)
+		));
+	}
+
```
tnull: Hmm, if we add all these assertions for `NotShuttingDown`, should we at least also add one later for a case when we *are* shutting down (i.e., close/force close)?

tests/integration_tests_rust.rs
```diff
@@ -2672,6 +2678,10 @@ async fn open_channel_with_all_with_anchors() {
 	assert!(channel.channel_value_sats > premine_amount_sat - anchor_reserve_sat - 500);
 	assert_eq!(channel.counterparty_node_id, node_b.node_id());
 	assert_eq!(channel.funding_txo.unwrap(), funding_txo);
+	assert!(matches!(
```
tnull: I think we don't need to add all of these assertions, as the actual behavior should be tested in LDK.

src/types.rs
f3r10: I agree, I am going to remove that line

tests/integration_tests_rust.rs
f3r10: you are right, those tests are not necessary. I am going to revert those.

tests/common/mod.rs
f3r10: I have been trying to test this part, but the problem is that for cooperative close, the shutdown completes very quickly in tests, so by the time we call `list_channels()` after `close_channel()`, the channel may already be gone.

 In the force close phase, LDK doesn't use the cooperative shutdown state machine for force closes, so the field would still be `None/NotShuttingDown` right up until the channel disappears.

src/types.rs
```diff
@@ -558,6 +529,8 @@ pub struct ChannelDetails {
 	pub inbound_htlc_maximum_msat: Option<u64>,
 	/// Set of configurable parameters that affect channel operation.
 	pub config: ChannelConfig,
+	/// The current shutdown state of the channel, if any.
```
tnull, line 532: Please add a note that this will be `None` for objects first serialized with LDK Node v0.1.

f3r10:
> One comment, otherwise LGTM
> 
> Though, IMO, you could further reduce the test changes. Maybe just add one at an opportune place? Asserting on `NotShuttingDown` over and over again isn't super helpful and just increases clutter.

You are right, I am going to leave the "shutting down" check after `close_channel()`

PR #828, opened 2026-03-15
Add chain syncing via compact block filters (BIP 157)

### LINKED ISSUE #767: Add chain syncing via `bip157` crate

febyeji:
This PR adds a new compact block filter chain source (via kyoto) to ldk-node. Let's go through its features.

## Design notes

Contrary to other chain sources, we use a push-based model: we don't proactively request data, but rather listen
for events and process them.

### Kyoto

Kyoto is a standalone node which comes in the form of a library. It manages peers and provides users with compact
block filters as well as the ability to request blocks on demand. 

Getting data from kyoto is done via a channel which emits events (e.g. `IndexedFilter` which means we received a new
compact block filter).

## Operation

In broad terms we operate in the following way:

1. On start, build and run the kyoto node
2. Listen for kyoto events 
3. Process kyoto events and respectively apply needed operations (e.g. download full block to apply it).

There is no concept of "active syncing": when the chain source runs, it always processes new events and applies them. 
Method `sync_wallet` waits until the application of all blocks up to the known tip.

## Node aliveness

As we are responsible for always having an operational node, we have its runtime status, and if it goes down by any
reason, we try to restart the kyoto node. If something went wrong and after `MAX_RESTART_RETRIES=5` the node still died,
we give up.

### Scripts 

CBFChainSource tracks watched scripts provided from the onchain wallet as well as the lightning wallet. 

We need to track two types of scripts: ones provided from the onchain wallet (e.g. to fund the lightning node) and
"LDK-scripts", exposed and needed to be tracked during channel operation (e.g. channel force-close transactions). The
former scripts are revealed via bdk wallet. We don't maintain a CBFChainSource-related copy of them, rather we just ask
bdk wallet each time we need them. This is done to have a single source of truth.

Thus, `registered_scripts` are LDK-revealed scripts.

### Block applicator 

This entity applies blocks to the state of the node. (Perhaps its title can be changed to something more concise.)

It runs as its own background task, decoupled from kyoto event processing via an unbounded channel of `ChainOp`s:

- `ConnectFull` — filter matched, apply the downloaded full block (`block_connected`)
- `ConnectFiltered` — filter didn't match, apply header only (`filtered_block_connected` with no transactions)
- `Disconnect` — reorg, disconnect back to the fork point
- `Synced` — kyoto reported `FiltersSynced`, i.e. we caught up to the network tip
- `Failed` — an unrecoverable error occurred while producing ops (surfaces to `sync_wallets` waiters)

For applying blocks we reuse bitcoind's `ChainListener`, which fans each connect/disconnect out to the onchain
wallet, channel manager, chain monitor and output sweeper.

`process_kyoto_events` is a single sequential loop: for a matched filter it awaits the full block fetch (with
retries/timeout) inline, and for an unmatched one it awaits the header fetch inline, before moving on to the next
kyoto event — this is what ensures ops are produced in order. The block applicator then just consumes those ops
in order and applies them. As a safety net it also tracks the next expected height itself and skips any
out-of-sequence op (e.g. duplicates, or stale ops produced before a reorg); a `Disconnect` rewinds the expected
height to the fork point.

Besides applying blocks, the applicator is the single writer of `CbfSyncState`: it advances `applied_tip` /
`synced_to_tip` (which `sync_wallets` waits on) and persists the node's wallet-sync timestamps once the tip is
reached. When the native kyoto fee source is used, it also opportunistically caches the coinbase-derived fee rate
of every full block it applies, so the fee estimator doesn't have to download those blocks again.

There is room for improvement: we could download full blocks in parallel (e.g. using
https://docs.rs/futures/latest/futures/stream/struct.FuturesOrdered.html), which would speed up initial sync.

We could do without this additional entity, by just processing ops in place, but we stick to this design for
clarity.

### Fee source

As we don't have mempool with CBF chain source, we cannot properly estimate the transaction fees. 
There are three options for a fee source provider (`FeeSource`):
- native kyoto estimation via extrapolation from recent blocks' fees
- esplora
- electrum

Using esplora or electrum just for the fees is privacy-preserving and acceptable.
Using native extrapolation is a fallback method, because it has two flaws:
- we need to download all blocks, which defeats kyoto's bandwidth advantage
- the estimation is still rough, as we have no mempool

Also, recently fetched blocks (when we need to download a full block anyway) are cached so they won't be downloaded
twice during a fee update.

### Reorganizations

On reorgs kyoto emits events about disconnected blocks and we disconnect them accordingly.

### CbfSyncState

We need that struct to distinguish between last block that kyoto has and the last block we have applied locally, these
may be different blocks. Also used to understand whether we processed everything up to kyoto's tip.

### Testing 

We tried to adhere to already existing tests, as they cover the vital node operation scenarios. Though some of the tests
are not meant for kyoto (for example tests related to mempool, 0-conf channels, RBF).

As we want to test exclusively the new chain source in this PR, we've added an environment variable overriding the
random chain source choice, it can be used on the following way:

```
LDK_TEST_CHAIN_SOURCE=cbf cargo test --test integration_tests_rust
```

Other available options: `esplora`, `electrum`, `bitcoind-rpc`, `bitcoind-rest`.

Also please note sometimes in test scenarios kyoto bans bitcoind because of this bug:
https://github.com/bitcoin/bitcoin/pull/35354.

Previously tests assumed that the chain source and the mining node are the same entity, i.e. there is no gossip delay.
This is not the case with the kyoto node, that's why one might experience a race condition due to gossip propagation
(we do something, like mine a block with a needed transaction, then it is correctly shown in esplora/electrum/bitcoind
and we try to invoke some action, but the new block has not yet propagated to the kyoto node); the situation is
worsened by the fact that GitHub's CI runners are slower than local machines, and a simple time-based wait does not
work well to circumvent the propagation delay. That's why many functions in the tests have additional waiting for
propagation.

Tests not comptible with the compact block filter node:
- channel_full_cycle_0conf,
- channel_full_cycle_0conf_0reserve
- onchain_send_receive
- onchain_fee_bump_rbf,
- onchain_fee_bump_rbf_respects_anchor_reserve

### History

We switched to bitcoind's sync model (`poll_and_update_listener` vs `sync_onchain_wallet` and `sync_lightning_wallet`).

We tried to follow the patterns for already existing chain sources, more precisely we had distinct `sync_onchain_wallet`
and `sync_lightning_wallet` (dictated by transaction-based syncing sources) functions however it is suboptimal, because
we might fetch the same block twice.

### DIFF
--- a/src/chain/cbf.rs
+++ b/src/chain/cbf.rs
@@ -293,8 +293,18 @@ impl CbfChainSource {
 							event_rx: new_event_rx,
 						} = new_client;
 
-						*restart_status.lock().expect("lock") =
-							CbfRuntimeStatus::Started { requester: new_requester };
+						{
+							let mut status = restart_status.lock().expect("lock");
+							if matches!(*status, CbfRuntimeStatus::Stopped) {
+								let _ = new_requester.shutdown();
+								log_info!(
+									restart_logger,
+									"CBF restart aborted: stop() called during backoff."
+								);
+								break;
+							}
+							*status = CbfRuntimeStatus::Started { requester: new_requester };
+						}
 
 						current_node = new_node;
 						current_info_rx = new_info_rx;
--- a/src/chain/cbf.rs
+++ b/src/chain/cbf.rs
@@ -6,9 +6,10 @@ use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use bip157::chain::ChainState;
 use bip157::{
 	chain::BlockHeaderChanges, error::FetchBlockError, Builder as KyotoBuilder, Client, Event,
-	HashCheckpoint, Header, IndexedBlock, Info, Node as KyotoNode, Requester, TrustedPeer, Warning,
+	HashCheckpoint, Header, IndexedBlock, Info, Node as KyotoNode, Package, Requester, TrustedPeer,
+	Warning,
 };
-use bitcoin::{BlockHash, FeeRate, Network, Script, ScriptBuf, Txid};
+use bitcoin::{BlockHash, FeeRate, Network, Script, ScriptBuf, Transaction, Txid};
 use electrum_client::{Client as ElectrumClient, ConfigBuilder as ElectrumConfigBuilder};
 use lightning::chain::{BlockLocator, Listen, WatchedOutput};
 
@@ -618,7 +619,10 @@ impl CbfChainSource {
 				.await?
 			},
 			FeeSource::Cbf { block_fee_cache } => {
-				let requester = self.requester()?;
+				let requester = match &*self.cbf_runtime_status.lock().expect("lock") {
+					CbfRuntimeStatus::Started { requester } => requester.clone(),
+					CbfRuntimeStatus::Stopped => return Err(Error::FeerateEstimationUpdateFailed),
+				};
 				let mut samples_sat_per_kwu: Vec<u64> = self
 					.refresh_block_fee_window(&requester, block_fee_cache)
 					.await
@@ -662,16 +666,33 @@ impl CbfChainSource {
 		Ok(())
 	}
 
-	/// Returns a clone of the live kyoto requester, or an error if the node isn't running.
-	fn requester(&self) -> Result<Requester, Error> {
-		match &*self.cbf_runtime_status.lock().expect("lock") {
-			CbfRuntimeStatus::Started { requester } => Ok(requester.clone()),
+	pub(crate) async fn process_broadcast_package(&self, package: Vec<Transaction>) {
+		let requester = match &*self.cbf_runtime_status.lock().expect("lock") {
+			CbfRuntimeStatus::Started { requester } => requester.clone(),
 			CbfRuntimeStatus::Stopped => {
-				debug_assert!(
-					false,
-					"We should have started the chain source before updating fees"
-				);
-				Err(Error::FeerateEstimationUpdateFailed)
+				debug_assert!(false, "We should have started the chain source before broadcasting");
+				return;
+			},
+		};
+
+		match Package::from_vec(package.clone()) {
+			Ok(package) => {
+				if let Err(e) = requester.submit_package(package).await {
+					log_error!(self.logger, "Failed to broadcast transaction package: {:?}", e);
+				}
+			},
+			Err(_) => {
+				for tx in package {
+					let txid = tx.compute_txid();
+					if let Err(e) = requester.submit_package(tx).await {
+						log_error!(
+							self.logger,
+							"Failed to broadcast transaction {}: {:?}",
+							txid,
+							e
+						);
+					}
+				}
 			},
 		}
 	}
--- a/src/chain/cbf.rs
+++ b/src/chain/cbf.rs
@@ -101,6 +101,7 @@ enum ChainOp {
 struct BlockApplicator {
 	chain_listener: ChainListener,
 	ops_rx: mpsc::UnboundedReceiver<ChainOp>,
+	next_height: u32,
 	/// Present only for the native CBF fee source: lets us cache the fee rate of blocks we download
 	/// here, so the fee estimator doesn't have to re-download them.
 	block_fee_cache: Option<BlockFeeCache>,
@@ -113,7 +114,17 @@ impl BlockApplicator {
 			match op {
 				ChainOp::ConnectFull { block_rx } => match block_rx.await {
 					Ok(Ok(ib)) => {
+						if ib.height != self.next_height {
+							log_debug!(
+								self.logger,
+								"CBF skipping out-of-sequence block at height {} (expected {})",
+								ib.height,
+								self.next_height
+							);
+							continue;
+						}
 						self.chain_listener.block_connected(&ib.block, ib.height);
+						self.next_height += 1;
 						if let Some(cache) = &self.block_fee_cache {
 							let fee_rate = coinbase_fee_rate(&ib.block, ib.height);
 							cache
@@ -126,10 +137,21 @@ impl BlockApplicator {
 					Err(_) => log_error!(self.logger, "block oneshot dropped"),
 				},
 				ChainOp::ConnectFiltered { header, height } => {
+					if height != self.next_height {
+						log_debug!(
+							self.logger,
+							"CBF skipping out-of-sequence block at height {} (expected {})",
+							height,
+							self.next_height
+						);
+						continue;
+					}
 					self.chain_listener.filtered_block_connected(&header, &[], height);
+					self.next_height += 1;
 				},
 				ChainOp::Disconnect { fork_point } => {
 					self.chain_listener.blocks_disconnected(fork_point);
+					self.next_height = fork_point.height + 1;
 				},
 				ChainOp::Synced { tip_height } => {
 					log_info!(self.logger, "CBF caught up to tip {}", tip_height);
@@ -264,6 +286,7 @@ impl CbfChainSource {
 			_ => None,
 		};
 		let block_applicator = BlockApplicator {
+			next_height: chain_listener.get_best_block().height + 1,
 			chain_listener: chain_listener.clone(),
 			ops_rx,
 			block_fee_cache,
@@ -336,14 +359,13 @@ impl CbfChainSource {
 							backoff_ms,
 						);
 
-						tokio::time::sleep(Duration::from_millis(backoff_ms)).await;
-						backoff_ms = backoff_ms.saturating_mul(2);
-
 						// Abort the old log consumers before rebuilding.
 						info_handle.abort();
 						warn_handle.abort();
 						event_handle.abort();
 
+						tokio::time::sleep(Duration::from_millis(backoff_ms)).await;
+						backoff_ms = backoff_ms.saturating_mul(2);
 						let (new_node, new_client) = Self::build_kyoto(
 							&restart_peers,
 							&restart_config,
@@ -442,9 +464,11 @@ impl CbfChainSource {
 					let matched = indexed_filter.contains_any(all_scripts.iter());
 
 					let chop: ChainOp = if matched {
-						let block_rx =
-							requester.request_block(block_hash).expect("cannot request block");
-						ChainOp::ConnectFull { block_rx }
+						if let Ok(handle) = requester.request_block(block_hash) {
+							ChainOp::ConnectFull { block_rx: handle }
+						} else {
+							break;
+						}
 					} else {
 						let height = indexed_filter.height();
 						//TODO we need to recheck that a particular height has not been
@@ -468,6 +492,8 @@ impl CbfChainSource {
 								}
 							},
 							Ok(None) => {
+								//TODO what do we do?
+								todo!();
 								log_error!(logger, "No header at height {}", height,);
 								continue;
 							},
@@ -478,7 +504,8 @@ impl CbfChainSource {
 									height,
 									e,
 								);
-								continue;
+								break;
+								// continue;
 							},
 						}
 					};
--- a/src/chain/cbf.rs
+++ b/src/chain/cbf.rs
@@ -9,7 +9,7 @@ use bip157::{
 	HashCheckpoint, Header, IndexedBlock, Info, Node as KyotoNode, Requester, TrustedPeer, Warning,
 };
 use bitcoin::{BlockHash, FeeRate, Script, ScriptBuf, Txid};
-use lightning::chain::{Listen, WatchedOutput};
+use lightning::chain::{BlockLocator, Listen, WatchedOutput};
 
 use tokio::sync::{mpsc, oneshot};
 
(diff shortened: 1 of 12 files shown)


src/chain/cbf.rs
```diff
@@ -0,0 +1,1271 @@
...
+					*state.latest_tip.lock().unwrap() = Some(tip.hash);
+					log_info!(
+						logger,
+						"CBF filters synced to tip: height={}, hash={}",
+						tip.height,
+						tip.hash,
+					);
+					let _ = state.sync_update_tx.try_send(sync_update);
+				},
+				Event::Block(indexed_block) => {
```
rustaceanrob: I believe this event can be removed on my end. This was a product of an older API. Apologies for making the client code more confusing!

src/chain/cbf.rs
```diff
@@ -0,0 +1,1271 @@
...
+	fn build_cbf_node(
+		network: Network, data_dir: &std::path::Path, peers: &[TrustedPeer],
+	) -> (bip157::Node, Client) {
+		let mut builder = Builder::new(network);
+		builder = builder.data_dir(data_dir.to_path_buf());
+		if !peers.is_empty() {
+			builder = builder.add_peers(peers.to_vec());
+		}
+		builder = builder.fetch_witness_data();
+		builder = builder.response_timeout(Duration::from_secs(30));
```
rustaceanrob: We will want to make this timeout user configurable

src/chain/cbf.rs
```diff
@@ -0,0 +1,1031 @@
...
+			|m, t| {
+				m.latest_fee_rate_cache_update_timestamp = t;
+			},
+		)?;
+
+		Ok(())
+	}
+
+	/// Broadcast a package of transactions via the P2P network.
+	pub(crate) async fn process_broadcast_package(&self, package: Vec<Transaction>) {
```
rustaceanrob: [`bip157` version `0.5.0`](https://docs.rs/bip157/latest/bip157/) includes a [`Package`](https://docs.rs/bip157/latest/bip157/struct.Package.html) type that may be used here.

@tnull do I understand correctly that `Vec<Transaction>` is used so the traits are forward compatible for arbitrary package lengths? My crate only supports packages of size 1 and 2, as per the current [Bitcoin Core policy](https://github.com/bitcoin/bitcoin/blob/master/doc/policy/packages.md).

tnull: Hmm, well, kinda. LDK's trait uses a `&[&Transaction]` I believe. Generally we prefer to leave any transaction/package validation to Core via. the chain source. Given bip157 is the chain source, it should be fine to use `from_vec` and log an error if that fails for whatever reason.

randomlogin:
@tnull we are ready to go through the review process. The original post has been updated with the relevant information.

tnull:
> @tnull we are ready to go through the review process. The original post has been updated with the relevant information.

Awesome! Note that because it's late in the release cycle I'm tagging this for v0.9, and PRs that need to land for v0.8 will be prioritized when it comes to review capacity. Still will see to get back to this soon to give first feedback on the approach.

Meanwhile CI seems to be failing, so a first step could be to make it pass.

randomlogin:
Yes, we understand that it's too late for including to 0.8, also we expect the review process will be long, so no problem with that.

Some of CI tests are not compatible with CBF node without access to the mempool (e.g. 0-conf channels), but apart from that there are genuinely failing ones, we're going to fix them, but decided to open the PR for the reviews as it already took a long time and constant rebasing has become painful.

src/chain/electrum.rs
```diff
@@ -708,91 +715,84 @@ impl ElectrumRuntimeClient {
...
-
-		let mut batch = Batch::default();
-		let confirmation_targets = get_all_conf_targets();
-		for target in confirmation_targets {
-			let num_blocks = get_num_block_defaults_for_target(target);
-			batch.estimate_fee(num_blocks, None);
-		}
-
-		let spawn_fut = self.runtime.spawn_blocking(move || electrum_client.batch_call(&batch));
+pub(crate) async fn get_electrum_fee_rate_cache_update(
```
rustaceanrob, line 720: To clean up the diff a bit, perhaps this refactor can be opened as a separate pull. I think this could have additional value outside of compact block filters. Bitcoin Core still uses block-level fee estimation, which can be occasionally inaccurate.

src/chain/cbf.rs
```diff
@@ -0,0 +1,1032 @@
...
+use crate::PersistedNodeMetrics;
+
+/// Walk back this many blocks from the wallet's persisted tip when deriving
+/// the kyoto resume checkpoint, so a recent reorg cannot strand the node
+/// above the new best chain.
+const REORG_SAFETY_BLOCKS: u32 = 7;
+const BLOCK_FEE_CACHE_CAPACITY: usize = REORG_SAFETY_BLOCKS as usize * 2;
+
+/// Peer response timeout passed to kyoto's `Builder::response_timeout`.
+const DEFAULT_RESPONSE_TIMEOUT_SECS: u64 = 30;
```
rustaceanrob, line 42: This seems incredibly long. The base crate uses a default of 5 seconds, and even that is quite long. I would recommend something more like 2-5 seconds, which should be sufficient for users even on relatively poor internet. I think this should also be configurable.

src/chain/cbf.rs
```diff
@@ -0,0 +1,1032 @@
...
+					}
+				},
+				Event::FiltersSynced(sync_update) => {
+					//Because application of blocks is async, the fact that kyoto synced up to the
+					//tip does NOT mean that we caught everything up, that's why we send a ChainOp,
+					//only processing of which means we processed all blocks up to the tip.
+					log_info!(logger, "Kyoto synced up to the tip {}", sync_update.tip().height);
+					let _ = ops_tx.send(ChainOp::Synced { tip_height: sync_update.tip().height });
+				},
+				Event::ChainUpdate(BlockHeaderChanges::Connected(indexed_header)) => {
```
rustaceanrob: I think we can make use of these events to save time on syncing. It would be possible to keep a small (least recently used) cache of block headers indexed by the hash, and each time there is no filter match we check that cache first before asking the CBF client for the header corresponding to the block hash.

On my side, it could make sense to just send the block header as part of `IndexedFilter`, since it is only ~48 bytes of extra overhead but is very useful on the client side.

src/chain/cbf.rs
```diff
@@ -0,0 +1,1032 @@
...
+									let _ =
+										ops_tx.send(ChainOp::Failed { error: Error::NotRunning });
+									return;
+								},
+							};
+
+							// Bound the download so an unresponsive peer can't park the fetch forever,
+							// then flatten the three error layers (timeout / receiver dropped / fetch
+							// error) into a single reason so the retry-or-fail decision is written once.
+							let fetched = tokio::time::timeout(
```
rustaceanrob, line 634: This is already handled by the client whereby unresponsive peers are disconnected for stalling (configured by the timeout). I think this can be greatly simplified by removing the retry logic entirely and using an appropriate timeout.

rustaceanrob:
Looking pretty good, nice job! The commit history is hard to follow. Some of the later commits appear to be fixing something which is introduced earlier in the pull request. Perhaps you can have claude apply the fixes to the commit that introduces the structure?

src/chain/cbf.rs
randomlogin: Noted.

src/chain/cbf.rs
randomlogin: Yeah, we can introduce a cache.

Yes, having the header data in the event itself would simplify things. I'll open a PR (next week) to kyoto if you don't mind.

src/chain/cbf.rs
randomlogin: Noted.

randomlogin:
Thank you!

Yes, the history is messy (we had situations with several simultaneous PRs which sometimes depended on one another, current situation is because of that) and at some point we prioritized functionality over commit history cleanness. 

I wonder it would be okay to just squash everything in a few commits after the approve.

src/chain/cbf.rs
rustaceanrob: I'm already working on a patch now. It's only a few lines. You are welcome to review

src/chain/cbf.rs
```diff
@@ -0,0 +1,1002 @@
...
```
rustaceanrob, line 634: This will stop further filter checks even if there are filters available to check, which appears rather unfortunate. We already have a structure here `BlockApplicator` that is asynchronous, so I am wondering why we don't await the network bound task over there and let this task continue checking filters.

I haven't fully implemented proper error propagation, but it would look something like this:

<details>
<summary>diff</summary>

```diff
diff --git a/src/chain/cbf.rs b/src/chain/cbf.rs
index 7baa4a0..6dd7ddd 100644
--- a/src/chain/cbf.rs
+++ b/src/chain/cbf.rs
@@ -4,6 +4,7 @@ use std::sync::{Arc, Mutex};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};

 use bip157::chain::ChainState;
+use bip157::error::FetchBlockError;
 use bip157::{
        chain::BlockHeaderChanges, Builder as KyotoBuilder, Client, Event as KyotoEvent,
        HashCheckpoint, Header, IndexedBlock, Info, Node as KyotoNode, Package, Requester, TrustedPeer,
@@ -122,7 +123,7 @@ pub struct CbfChainSource {

 #[derive(Debug)]
 enum ChainOp {
-       ConnectFull { block: IndexedBlock },
+       ConnectFull { block: tokio::sync::oneshot::Receiver<Result<IndexedBlock, FetchBlockError>> },
        ConnectFiltered { header: Header, height: u32 },
        Disconnect { fork_point: BlockLocator },
        Synced { tip_height: u32 },
@@ -146,7 +147,8 @@ impl BlockApplicator {
        async fn run(mut self) {
                while let Some(op) = self.ops_rx.recv().await {
                        match op {
-                               ChainOp::ConnectFull { block: ib } => {
+                               ChainOp::ConnectFull { block: recv } => {
+                                       let ib = recv.await.unwrap().unwrap();
                                        if ib.height != self.next_height {
                                                log_debug!(
                                                        self.logger,
@@ -611,63 +613,19 @@ impl CbfChainSource {
                                        let matched = indexed_filter.contains_any(all_scripts.iter());

                                        let chop: ChainOp = if matched {
-                                               let mut attempt = 0;
-                                               let block = loop {
-                                                       attempt += 1;
-                                                       let handle = match requester.request_block(block_hash) {
-                                                               Ok(handle) => handle,
-                                                               Err(_) => {
-                                                                       log_error!(
+                                               let handle = match requester.request_block(block_hash) {
+                                                       Ok(handle) => handle,
+                                                       Err(_) => {
+                                                               log_error!(
                                                                                logger,
                                                                                "Failed to obtain receiver for matched CBF block {}; node is stopped",
                                                                                block_hash
                                                                        );
-                                                                       let _ =
-                                                                               ops_tx.send(ChainOp::Failed { error: Error::NotRunning });
-                                                                       return;
-                                                               },
-                                                       };
-
-                                                       // Bound the download so an unresponsive peer can't park the fetch forever,
-                                                       // then flatten the three error layers (timeout / receiver dropped / fetch
-                                                       // error) into a single reason so the retry-or-fail decision is written once.
-                                                       let fetched = tokio::time::timeout(
-                                                               Duration::from_secs(CBF_BLOCK_FETCH_TIMEOUT_SECS),
-                                                               handle,
-                                                       )
-                                                       .await
-                                                       .map_err(|_| {
-                                                               format!("timed out after {}s", CBF_BLOCK_FETCH_TIMEOUT_SECS)
-                                                       })
-                                                       .and_then(|recv| recv.map_err(|_| "receiver was dropped".to_string()))
-                                                       .and_then(|fetch| fetch.map_err(|e| format!("failed: {:?}", e)));
-
-                                                       match fetched {
-                                                               Ok(block) => break block,
-                                                               Err(reason) if attempt < CBF_BLOCK_FETCH_RETRIES => {
-                                                                       log_debug!(
-                                                                               logger,
-                                                                               "CBF block fetch for {} {} on attempt {}; retrying",
-                                                                               block_hash,
-                                                                               reason,
-                                                                               attempt
-                                                                       );
-                                                               },
-                                                               Err(reason) => {
-                                                                       log_error!(
-                                                                               logger,
-                                                                               "CBF block fetch for {} {} after {} attempts; giving up",
-                                                                               block_hash,
-                                                                               reason,
-                                                                               CBF_BLOCK_FETCH_RETRIES
-                                                                       );
-                                                                       let _ =
-                                                                               ops_tx.send(ChainOp::Failed { error: Error::TxSyncFailed });
-                                                                       return;
-                                                               },
-                                                       }
+                                                               let _ = ops_tx.send(ChainOp::Failed { error: Error::NotRunning });
+                                                               return;
+                                                       },
                                                };
-                                               ChainOp::ConnectFull { block }
+                                               ChainOp::ConnectFull { block: handle }
                                        } else {
                                                ChainOp::ConnectFiltered {
                                                        header: indexed_filter.header(),
```
</details>

randomlogin: From readability perspective it makes sense.

My concern here is in that case we could process many new filters before we really got awaited block (e.g. during initial sync). Awaited block might reveal new scripts and therefore when we match filters we might get a false negative result. 

Right now we wait for the block to be downloaded (though not applied) before the next filter is matched. Which is not perfect, but okay. It seems we'll need to synchronize kyoto's producer and block applicator's consumer then.

rustaceanrob: Ah, yes that's right. Perhaps not worth the refactor at this point then

PR #829, opened 2026-03-16
Expose full payment retry strategy via Config

benthecarman:
Replace the `payment_retry_timeout_secs` field with a `PaymentRetryStrategy`enum that mirrors LDK's `Retry` type, allowing users to choose between timeout-based and attempt-count-based retry strategies. Defaults to a 10-second timeout to preserve existing behavior.

This was something I needed when getting ldk-server working with loop.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -14,6 +14,7 @@ use std::time::Duration;
 use bitcoin::secp256k1::PublicKey;
 use bitcoin::Network;
 use lightning::ln::msgs::SocketAddress;
+use lightning::ln::outbound_payment::Retry;
 use lightning::routing::gossip::NodeAlias;
 use lightning::routing::router::RouteParametersConfig;
 use lightning::util::config::{
@@ -29,6 +30,7 @@ const DEFAULT_LDK_WALLET_SYNC_INTERVAL_SECS: u64 = 30;
 const DEFAULT_FEE_RATE_CACHE_UPDATE_INTERVAL_SECS: u64 = 60 * 10;
 const DEFAULT_PROBING_LIQUIDITY_LIMIT_MULTIPLIER: u64 = 3;
 const DEFAULT_ANCHOR_PER_CHANNEL_RESERVE_SATS: u64 = 25_000;
+const DEFAULT_PAYMENT_RETRY_TIMEOUT_SECS: u64 = 10;
 
 // The default timeout after which we abort a wallet syncing operation.
 const DEFAULT_BDK_WALLET_SYNC_TIMEOUT_SECS: u64 = 60;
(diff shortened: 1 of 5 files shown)


tnull (commented):
> An alternative would be to expose the `Retry` completely to allow for attempts based version as well. However, the timeout is generally simpler and more in line with what people expect.

If we do it,  I'd  lean towards exposing `retry_strategy: Retry` in the config. Should be even pretty easy to handle in bindings, as `uniffi` should now expose tuple enum variants.

src/config.rs
```diff
@@ -619,6 +626,45 @@ pub enum AsyncPaymentsRole {
 	Server,
 }
 
+/// Strategies available to retry payment path failures.
+///
+/// See [`Retry`] for details.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+pub enum PaymentRetryStrategy {
```
tnull: Rather than duplicating the object here, can't we reuse LDK's `Retry`?

TheBlueMatt (commented):
Would be nice if this were configurable per-payment rather than only globally, IMO.

src/config.rs
benthecarman: No we can't re-export it to the bindings because of the Duration inside.

tnull: Hmm, then it would be likely preferable to expose a remote type for `Duration`, too, rather than duplicating a lot of code?

tnull (commented):
> Made it so we expose all the bolt 11/12 send params. It's about the same api complexity (a single optional struct) and will be useful for things like the mpp override that we need for orange

Hmm, I have to say I'm not loving this. In the past we spent quite a bit of time to clean up and align BOLT11 and BOLT12 APIs (also opened upstream PRs etc) around `sending_parameters`/`RouteParametersConfig`, so I'd prefer not to simply 'give up', expose everything under the sun and have the APIs diverge a lot just because AI tooling now makes it very easy to generate a bunch of additional structs. If we think we need additional parameters, can we please first at least try to iterate upstream to keep BOLT11 and BOLT12 interfaces aligned and maintain feature parity?

src/config.rs
benthecarman: Looks like we also cannot because `Retry` is a tuple enum `Retry::Attempts(u64)` which isn't supported by uniffi. Would need to be like `Retry::Attempts { attmepts: u64 }` for us to be able to.

I still changed to re-export duration so it is less duplication

tnull: > Looks like we also cannot because `Retry` is a tuple enum `Retry::Attempts(u64)` which isn't supported by uniffi.

They are supported by now, but only in proc macros (see https://mozilla.github.io/uniffi-rs/0.29/udl/enumerations.html). Given it's a remote type you'll want to do something like we did with `ClosureReason` here: https://github.com/lightningdevkit/ldk-node/blob/804f00f6ad763f42074ce2efeef67e48ee772925/src/ffi/types.rs#L1419

I.e., use `Retry` in the normal Rust code, and then in `src/ffi/types.rs` only duplicate the upstream enum with the `#[uniffi::remote(Enum)]` tag. I think that could work (note that the duplicated code doesn't actually be imported anywhere separately, and will just be used by the bindings generator).

Camillarhi:
This now needs a rebase, there's a merge conflict.

m1sterc001guy:
> Would be nice if this were configurable per-payment rather than only globally, IMO.

For what its worth, on the Fedimint side we're looking into allowing clients to customize the timeout that is passed to LND (within a reasonable range). Would be nice if this was possible with ldk-node as well.

PR #832, opened 2026-03-18
Make node's address optional when opening channel

benthecarman:
Previously we always required providing the node's address when opening a channel, this made the api annoying in some cases as you may already be connected to the peer but still need to provide the socket address.

Here we make it optional and will return a `ConnectionFailed` error when the socket address is not provided and we are not connected to the peer.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -25,6 +25,8 @@ pub enum Error {
 	OnchainTxCreationFailed,
 	/// A network connection has been closed.
 	ConnectionFailed,
+	/// The peer is not connected.
+	NotConnected,
 	/// Invoice creation failed.
 	InvoiceCreationFailed,
 	/// Invoice request creation failed.
@@ -148,6 +150,7 @@ impl fmt::Display for Error {
 				write!(f, "On-chain transaction could not be created.")
 			},
 			Self::ConnectionFailed => write!(f, "Network connection closed."),
+			Self::NotConnected => write!(f, "The peer is not connected."),
 			Self::InvoiceCreationFailed => write!(f, "Failed to create invoice."),
 			Self::InvoiceRequestCreationFailed => write!(f, "Failed to create invoice request."),
 			Self::OfferCreationFailed => write!(f, "Failed to create offer."),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -54,7 +54,7 @@
 //!
 //! 	let node_id = PublicKey::from_str("NODE_ID").unwrap();
 //! 	let node_addr = SocketAddress::from_str("IP_ADDR:PORT").unwrap();
-//! 	node.open_channel(node_id, node_addr, 10000, None, None).unwrap();
+//! 	node.open_channel(node_id, Some(node_addr), 10000, None, None).unwrap();
 //!
 //! 	let event = node.wait_next_event();
 //! 	println!("EVENT: {:?}", event);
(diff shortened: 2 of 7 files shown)


tests/common/mod.rs
```diff
@@ -792,7 +792,7 @@ pub async fn splice_in_with_all(
 
 pub(crate) async fn do_channel_full_cycle<E: ElectrumApi>(
 	node_a: TestNode, node_b: TestNode, bitcoind: &BitcoindClient, electrsd: &E, allow_0conf: bool,
-	expect_anchor_channel: bool, force_close: bool,
+	expect_anchor_channel: bool, force_close: bool, connect_first: bool,
```
tnull: Hmm, rather than adding yet another bool here, please create a dedicated test for it. Alternatively, it might better fit to add it to `connection_restart_behavior` or a similar test.

benthecarman: gave it its own test

src/lib.rs
```diff
@@ -1126,39 +1126,47 @@ impl Node {
...
+				self.runtime.block_on(async move {
+					con_cm.connect_peer_if_necessary(node_id, con_addr).await
+				})?;
+				Some(address)
+			},
+			None => {
+				// If we are connected, grab the socket address as we need to make sure we have it persisted
+				// in our peer storage for future reconnections.
+				let peer =
+					self.peer_manager.peer_by_node_id(&node_id).ok_or(Error::ConnectionFailed)?;
```
tnull: Hmm, it's a bit weird that this looks directly at peer manager and doesn't incorporate any connections that are inflight in `ConnectionManager` (i.e., two channel opens in quick succession will still have the second one fail, potentially?). 

Apart from that we probably want to introduce a dedicated error type that communicates *why* we failed to open a channel, and `ConnectionFailed` seems wrong here (as we didn't even attempt the connection in the first place, for example)

benthecarman: > Hmm, it's a bit weird that this looks directly at peer manager and doesn't incorporate any connections that are inflight in ConnectionManager (i.e., two channel opens in quick succession will still have the second one fail, potentially?).

this seems like a good thing to me? We'd want the connection to be fully established before opening the channel.

We also will end up checking the peer manager here anyways: https://github.com/benthecarman/ldk-node/blob/6cfdd4f38607873b7a324b9803fe92c3248db589/src/lib.rs#L1234

benthecarman: added a new error type

src/lib.rs
```diff
@@ -1280,7 +1289,7 @@ impl Node {
 	///
 	/// [`AnchorChannelsConfig::per_channel_reserve_sats`]: crate::config::AnchorChannelsConfig::per_channel_reserve_sats
 	pub fn open_channel(
-		&self, node_id: PublicKey, address: SocketAddress, channel_amount_sats: u64,
+		&self, node_id: PublicKey, address: Option<SocketAddress>, channel_amount_sats: u64,
```
tankyleo, line 1296: I think it's worth adding a quick doc about the `Option` type here.

src/lib.rs
```diff
@@ -1126,39 +1126,47 @@ impl Node {
...
-		})?;
+		// if we don't have the socket address, check if we are already connected
+		let address = match address {
+			Some(address) => {
+				// We need to use our main runtime here as a local runtime might not be around to poll
+				// connection futures going forward.
+				let con_cm = Arc::clone(&self.connection_manager);
+				let con_addr = address.clone();
+				self.runtime.block_on(async move {
+					con_cm.connect_peer_if_necessary(node_id, con_addr).await
```
tankyleo, line 1145: This was already present, but I would find it surprising if I specified an IPV4 address, but I was already connected to the peer via OnionV3, and the IPV4 setting at channel open was silently ignored and replaced with the OnionV3 address. Works also the other way around. WDYT ?

tankyleo:
> Makes sense, though arguably, if you're really after the principle of least surprise, you might want to leave the argument as required. :)

Haha you do have a point here :)

As I said above, I note that we currently silently ignore the address setting of `open_channel` in case we are already connected, which also makes it surprising in case the user specifies an address that is different from the one they are already connected with (ie OnionV3 vs IPV4).

Overall I think for this one I'd go with what's already present out there, ie from Ben's point in the ldk-server PR:

> they all (lnd, cln, eclair) don't require you to include the socket address, I feel it'd be better to follow suit as I imagine this is what users expect and its a pretty easy error message to give back before if they are not connected.

src/lib.rs
benthecarman: fixed

src/lib.rs
benthecarman: While I agree, not really sure what the alternative would be? We could throw an error if the socket addr doesn't match the one we are currently connected with but that seems kinda flaky and not what people would expect.

tankyleo: one potential non-flaky route in case of a mismatch: disconnect from the peer, and reconnect using the address specified in the open parameter ?

tankyleo: though this is likely out-of-scope for this PR as we'd want things to be consistent also with the `Node::connect` method, feel free to dismiss this here.

Curious what @tnull thinks nonetheless.

src/error.rs
```diff
@@ -25,6 +25,8 @@ pub enum Error {
 	OnchainTxCreationFailed,
 	/// A network connection has been closed.
 	ConnectionFailed,
+	/// The peer is not connected.
+	NotConnected,
```
tnull, line 29: Coming back to this after some time, I have to say I still find the new API changes here not great. That we suddenly have this new error case instead of just .. connecting to the peer IMO shows that it just makes the API less explicit, less clear, and more complicated, as users now suddenly will have to handle one more edge case. I'm still not entirely convinced it's worth it tbh. 

However, if you insist, this error variant should like be named more along the lines of `PeerAddressUnknown` or similar, as the issue here really is *not* that we're not connected but that we now  have this API that allows users to call it without all necessary data to connect to the involved peers in certain cases.

tankyleo: Upon further thought, I'd be happy to drop the proposed API changes here too.

>  instead of just .. connecting to the peer

I have a question: what should we do if the user sets an onion v3 address in `open_channel` for the peer, but we are already connected to that peer on IPv4 ?

My initial hunch is to disconnect the peer, and reconnect over onionV3, with perhaps a warning in the logs "hey note we were already connected to that peer on IPv4".

On master, we currently silently drop the onion V3 address, and use the existing IPv4 connection, which I don't think is what we want.

benthecarman: > My initial hunch is to disconnect the peer, and reconnect over onionV3, with perhaps a warning in the logs "hey note we were already connected to that peer on IPv4".

Problem with this is that if we fail to connect over tor, then we've disconnected the peer and interrupted our existing channel for nothing. We can't necessarily just reconnect to them as well as it may be an inbound connection.

tankyleo: thanks that's a good point. i think i'm good to not touch anything then.

benthecarman: The way to have the least surprising api would probably be to remove the socket address from `open_channel` completely and require the user to call `connect_peer` before. However, that is pretty annoying to the user and not what people have come to expect.

This PR is becoming more effort than it is worth. Fine with closing we can't come to agreement

tankyleo: good to close if @tnull is too

tnull: Yeah. Sorry, happy to revisit some time in the future, but right now it seems to me it would just introduce more footguns rather than actually simplify the API.

tnull:
Closing for now.

PR #839, opened 2026-03-24
Add interop integration test harness for LND, CLN, and Eclair

### LINKED ISSUE #766: Rewrite integration test harness ensuring interop with CLN, Eclair, and LND
While we have integration tests with CLN and LND, they so far cover only very basic flows.

We should create a more sophisticated test harness (probably using `proptest`?) that also adds Eclair interop coverage covers more and more error prone scenarios, such as:

a) peer disconnection and reconnection
b) splicing
c) channel closing
d) fee rate disagreements

etc.

febyeji:
Closes lightningdevkit/ldk-node#766

Add a interop test harness for testing LDK-Node against external Lightning implementations (LND, CLN, Eclair).

Each runs against all three implementations:

- `test_basic_channel_cycle`: open, BOLT11 each direction, coop close
- `test_keysend`: keysend each direction, coop close
- `test_force_close_after_payment`: open, payment, force close (LDK-initiated)
- `test_disconnect_during_payment`: disconnect+reconnect mid-payment, coop close
- `test_splice_in`: splice-in, post-splice payment, coop close


tests/docker/docker-compose-cln.yml
tnull, line 1: I do wonder if we should (as a follow-up ?) move all docker-related files to something like `tests/docker`?

.github/workflows/cln-integration.yml
```diff
@@ -13,18 +13,64 @@ jobs:
...
+            exit 1
+          }
 
-      - name: Forward lightningd RPC socket
+      - name: Set permissions for CLN data directory
         run: |
-          docker exec ldk-node-cln-1 sh -c "socat -d -d TCP-LISTEN:9937,fork,reuseaddr UNIX-CONNECT:/root/.lightning/regtest/lightning-rpc&"
-          socat -d -d UNIX-LISTEN:/tmp/lightning-rpc,reuseaddr,fork TCP:127.0.0.1:9937&
+          sudo chown -R $(id -u):$(id -g) $CLN_DATA_DIR
+          sudo chmod -R 755 $CLN_DATA_DIR
```
tnull: nit: Hmm, should we really set all files in there `+x`? I assume it should just be set on the folder?

.github/workflows/lnd-integration.yml
```diff
@@ -33,24 +30,54 @@ jobs:
...
+            echo "ERROR: LND macaroon not found after 60 seconds"
+            docker compose -f docker-compose-lnd.yml logs lnd
+            exit 1
+          }
+
       - name: Set permissions for LND data directory
-        # In PR 4622 (https://github.com/lightningnetwork/lnd/pull/4622),
-        # LND sets file permissions to 0700, preventing test code from accessing them.
-        # This step ensures the test suite has the necessary permissions.
         run: sudo chmod -R 755 $LND_DATA_DIR
```
tnull: Pre-existing, but same nit here?

Cargo.toml
```diff
@@ -79,34 +79,38 @@ async-trait = { version = "0.1", default-features = false }
...
 [target.'cfg(cln_test)'.dev-dependencies]
 clightningrpc = { version = "0.3.0-beta.8", default-features = false }
 
 [target.'cfg(lnd_test)'.dev-dependencies]
-lnd_grpc_rust = { version = "2.10.0", default-features = false }
+lnd_grpc_rust = { version = "2.14.0", default-features = false }
 tokio = { version = "1.37", features = ["fs"] }
 
+[target.'cfg(eclair_test)'.dev-dependencies]
+reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls"] }
```
tnull: Any reason why we need to re-introduce this rather than using `bitreq`?

.github/workflows/eclair-integration.yml
```diff
@@ -0,0 +1,56 @@
...
+          docker compose -f docker-compose-eclair.yml exec bitcoin bitcoin-cli -regtest -rpcuser=user -rpcpassword=pass -rpcwallet=eclair getnewaddress
+          docker compose -f docker-compose-eclair.yml exec bitcoin bitcoin-cli -regtest -rpcuser=user -rpcpassword=pass createwallet ldk_node_test
+
+      - name: Start Eclair
+        run: docker compose -f docker-compose-eclair.yml up -d eclair
+
+      - name: Wait for Eclair to be ready
+        run: |
+          for i in $(seq 1 60); do
+            if curl -s -u :eclairpassword -X POST http://127.0.0.1:8080/getinfo > /dev/null 2>&1; then
```
tnull: Codex reports:

>  - Medium: the Eclair readiness checks can report success on HTTP errors. In tests/integration_tests_eclair.rs:120, readiness is curl ... /getinfo plus status.success() && !stdout.is_empty(), and in .github/workflows/eclair-integration.yml:44 the workflow does the same pattern. curl
     exits 0 for many HTTP 4xx/5xx responses unless --fail is used, so a broken or half-initialized API can be treated as "ready" and the real failure gets deferred into the tests.

tests/common/mod.rs
```diff
@@ -380,6 +462,9 @@ macro_rules! setup_builder {
 
 pub(crate) use setup_builder;
 
+#[cfg(any(cln_test, lnd_test, eclair_test))]
+pub(crate) mod scenarios;
```
tnull, line 470: Please move this up to the other module definitions.

tests/integration_tests_cln.rs
```diff
@@ -9,119 +9,249 @@
...
+	node.stop().unwrap();
+}
+
+/// CLN v24.08+ includes a `payment_secret` in outbound keysend HTLCs.
+/// LDK treats any inbound HTLC with `payment_secret` as a BOLT11 payment and
+/// verifies it against a stored invoice — which fails for spontaneous payments.
+/// Upstream LDK fix needed: skip payment_secret verification when a valid
+/// `keysend_preimage` TLV is present.
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+#[ignore = "CLN v24.08+ sends payment_secret in keysend — LDK rejects (upstream fix needed)"]
```
tnull: Have yet to explore this further.

tests/integration_tests_eclair.rs
```diff
@@ -0,0 +1,366 @@
...
+};
+use common::scenarios::splice::{splice_from_external, splice_then_payment};
+use common::scenarios::{setup_interop_test, CloseType, PayType, Phase, Side};
+use electrsd::corepc_client::client_sync::Auth;
+use electrsd::corepc_node::Client as BitcoindClient;
+use electrum_client::Client as ElectrumClient;
+use ldk_node::{Builder, Event};
+
+/// Run a shell command via `spawn_blocking` to avoid blocking the tokio runtime.
+async fn run_cmd(program: &str, args: &[&str]) -> std::io::Result<std::process::Output> {
```
tnull: Hmm, I think it would be great to avoid this. Any `curl` calls here should be replaced by regular HTTP calls using `bitreq`/`serde`. 

Docker-related things should really be kept to the CI scripts. Do you think there is a way around re-creating the eclair container each time? I'm also a bit dubious why we need this if we're checking in the hotfix for the eclair dockerfile?

tests/common/scenarios/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+
+#[derive(Debug, Clone)]
+pub(crate) enum CloseType {
+	Cooperative,
+	Force,
+}
+
+#[derive(Debug, Clone)]
+pub(crate) enum PayType {
+	Bolt11,
```
tnull: Would be great to add coverage for BOLT12 scenarios (likely, in a follow up?).

tnull (commented):
Thanks, this already looks great on the first pass! I think we should get some minor things out of the way and then move forward with this so we can take advantage of the additional test coverage for our upcoming release already.

Some comments def. can be addressed in a follow-up (e.g. adding BOLT12 test coverage would be great).

tests/common/scenarios/mod.rs
febyeji: Great idea, will add BOLT12 scenario coverage in a follow-up!

tests/docker/docker-compose-cln.yml
febyeji: Good, I'll move them to `tests/docker/` in a follow-up.

Cargo.toml
febyeji: Right, I replaced `reqwest` with `bitreq` which is already a dependency

tests/common/mod.rs
febyeji: Unfortunately this can't be moved up. The scenarios module uses macro_rules! macros defined later in this file (`expect_event!`, `expect_channel_pending_event!`, etc.), and Rust requires macro_rules! definitions to appear before any module that uses them in textual order.

tests/integration_tests_eclair.rs
febyeji: Replaced all `curl` calls with `bitreq` async HTTP requests. Also removed per-test docker container recreation entirely. Tests now reuse a single Eclair instance started by CI, with just a `lockunspent` call between tests to release any UTXOs locked by prior force-close tests. (This can mitigate re-creating the eclair container.)

tnull (commented):
CI is failing right now:

```
thread 'test_eclair_cooperative_close_after_fee_change' (16127) panicked at tests/common/scenarios/channel.rs:275:55:
called `Result::unwrap()` on an `Err` value: ChannelClosingFailed


failures:
    test_eclair_cooperative_close_after_fee_change
```

.github/workflows/cln-integration.yml
```diff
@@ -13,18 +13,64 @@ jobs:
...
+          echo "ERROR: bitcoind not ready"
+          exit 1
+
+      - name: Mine initial block for CLN
         run: |
-          sudo apt-get update -y
-          sudo apt-get install -y socat
+          docker compose -f docker-compose-cln.yml exec bitcoin bitcoin-cli -regtest -rpcuser=user -rpcpassword=pass createwallet miner
+          ADDR=$(docker compose -f docker-compose-cln.yml exec bitcoin bitcoin-cli -regtest -rpcuser=user -rpcpassword=pass -rpcwallet=miner getnewaddress)
+          docker compose -f docker-compose-cln.yml exec bitcoin bitcoin-cli -regtest -rpcuser=user -rpcpassword=pass generatetoaddress 1 "$ADDR"
```
tnull: What are we using the `miner` wallet for? Note that coinbase funds are only maturing after 100 blocks, so they can only be spent then (which is why you usually see `generatetoaddress 101` on setups like this).

.github/workflows/rust.yml
```diff
@@ -49,7 +49,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: bin/bitcoind-${{ runner.os }}-${{ runner.arch }}
-          key: bitcoind-${{ runner.os }}-${{ runner.arch }}
+          key: bitcoind-29.0-${{ runner.os }}-${{ runner.arch }}
```
tnull, line 56: It's a bit odd to only adjust the cache key but not the path here. What's the rationale for this?

tests/common/scenarios/channel.rs
```diff
@@ -0,0 +1,305 @@
...
+use super::{setup_interop_test, CloseType, Side};
+
+/// Wait for a ChannelClosed event and assert the closure reason matches expectations.
+async fn expect_channel_closed(node: &Node, expected_local_initiated: bool) {
+	let event = tokio::time::timeout(
+		Duration::from_secs(super::super::INTEROP_TIMEOUT_SECS),
+		node.next_event_async(),
+	)
+	.await
+	.unwrap_or_else(|_| panic!("{} timed out waiting for ChannelClosed event", node.node_id()));
```
tnull: nit: Could use a newline

tests/common/lnd.rs
```diff
@@ -0,0 +1,424 @@
...
+
+	async fn open_channel(
+		&self, peer_id: PublicKey, _addr: SocketAddress, capacity_sat: u64, push_msat: Option<u64>,
+	) -> Result<String, TestFailure> {
+		let mut client = self.client.lock().await;
+		let local_funding_amount: i64 = capacity_sat
+			.try_into()
+			.map_err(|_| self.make_error(format!("capacity_sat overflow: {}", capacity_sat)))?;
+		let push_sat: i64 = push_msat
+			.map(|m| (m / 1000).try_into())
```
tnull, line 118: Hmm, seems this could lead to surprises in the future, if we simply forget about any sub-sat amounts and then want to assert them? Maybe this means the entire trait needs to take sat denomination? But I guess we have similar issues with other LND APIs, such as list_channels?

tests/common/cln.rs
```diff
@@ -0,0 +1,331 @@
...
+	async fn rpc<F, T>(&self, f: F) -> T
+	where
+		F: FnOnce(&LightningRPC) -> T + Send + 'static,
+		T: Send + 'static,
+	{
+		let client = Arc::clone(&self.client);
+		tokio::task::spawn_blocking(move || f(&*client)).await.expect("CLN RPC task panicked")
+	}
+
+	fn make_error(&self, detail: String) -> TestFailure {
```
tnull: nit: It seems this can just be inlined everywhere, rather than duplicating it a bunch of times for each `ExternalNode`?

tests/common/scenarios/disconnect.rs
```diff
@@ -0,0 +1,109 @@
...
+		Side::External => {
+			peer.disconnect_peer(node.node_id()).await.unwrap();
+		},
+	}
+
+	tokio::time::sleep(Duration::from_secs(1)).await;
+
+	// Reconnect and wait for channel reestablishment
+	node.connect(ext_node_id, ext_addr, true).unwrap();
+	for i in 0..30 {
```
tnull: Can we DRY up these (re-) connection loops (i.e. move to a method and reuse everywhere)?

tests/integration_tests_cln.rs
```diff
@@ -9,119 +9,249 @@
...
-	builder.set_chain_source_esplora("http://127.0.0.1:3002".to_string(), None);
-
+	builder.set_chain_source_electrum("tcp://127.0.0.1:50001".to_string(), None);
 	let node = builder.build(config.node_entropy).unwrap();
 	node.start().unwrap();
+	node
+}
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn test_cln_basic_channel_cycle() {
```
tnull: It seems all these basic tests are identical for all of LND,CLN, Eclair. Maybe we could just use a macro to implement them once and then just call the macro in the individual test files, to declutter/DRY up the code?

.github/workflows/rust.yml
```diff
@@ -60,7 +60,7 @@ jobs:
         if: "matrix.platform != 'windows-latest' && (steps.cache-bitcoind.outputs.cache-hit != 'true' || steps.cache-electrs.outputs.cache-hit != 'true')"
         run: |
           source ./scripts/download_bitcoind_electrs.sh
-          mkdir bin
+          mkdir -p bin
```
tnull, line 67: Why is this suddenly necessary?

tests/integration_tests_cln.rs
```diff
@@ -9,119 +9,249 @@
...
+	let (_user_ch, ext_ch) =
+		open_channel_to_external(&node, &cln, &bitcoind, &electrs, 1_000_000, Some(500_000_000))
+			.await;
+
+	force_close_by_external(&node, &cln, &bitcoind, &electrs, &ext_ch).await;
+	node.stop().unwrap();
+}
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+#[ignore = "CLN splicing requires --experimental-splicing flag and CLN v25+"]
```
tnull: Yes, why can't we enable this? AFAICT CLN v25+ should be availble? E.g., https://hub.docker.com/layers/elementsproject/lightningd/v25.12.1/images/sha256-fb5c956ff969cf7bb3a9ee46240daa9fdc03a9d7bc35931a2f8a888343fd16d2

tests/integration_tests_cln.rs
```diff
@@ -9,119 +9,249 @@
...
+		open_channel_to_external(&node, &cln, &bitcoind, &electrs, 1_000_000, Some(500_000_000))
+			.await;
+
+	receive_bolt11_payment(&node, &cln, 5_000_000).await;
+
+	node.stop().unwrap();
+}
+
+/// CLN v24.08+ includes a `payment_secret` in outbound keysend HTLCs.
+/// LDK treats any inbound HTLC with `payment_secret` as a BOLT11 payment and
```
tnull: Hmm, I'm not positive this is true? Could you un-ignore this so we can see what the error actually is? Is it related to the checks here?: https://github.com/lightningdevkit/rust-lightning/blob/db42ad6ba053d54f98f360f05afad5fee896ed69/lightning/src/ln/channelmanager.rs#L8479

tests/integration_tests_eclair.rs
```diff
@@ -0,0 +1,280 @@
...
+		open_channel_to_external(&node, &eclair, &bitcoind, &electrs, 1_000_000, Some(500_000_000))
+			.await;
+
+	force_close_by_external(&node, &eclair, &bitcoind, &electrs, &ext_ch).await;
+	node.stop().unwrap();
+}
+
+/// Eclair 0.8.0 rejects LDK keysend with `InvalidOnionPayload(8,0)` — LDK includes
+/// `payment_data` (TLV type 8) in keysend onions, which Eclair considers invalid for
+/// spontaneous payments. Eclair 0.14.0+ may handle this differently.
```
tnull: How do you arrive at this conclusion? Are they aware of this incompatibility? Can you link the releveant issues when including such remarks? Given we have Eclair v0.14-SNAPSHOT, can't we un-ignore this then?

tests/integration_tests_eclair.rs
```diff
@@ -0,0 +1,280 @@
...
+
+	let ext_node_id = eclair.get_node_id().await.unwrap();
+	node.spontaneous_payment().send(5_000_000, ext_node_id, None).unwrap();
+	common::expect_event!(node, PaymentSuccessful);
+
+	node.stop().unwrap();
+}
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+#[ignore = "Eclair 0.8.0 does not support splicing (introduced in v0.10.0+)"]
```
tnull: Why is this ignored, we should have Eclair v0.14-SNAPSHOT available, i.e., can test splicing, no?

tnull (commented):
Thanks for addressing the feedback! Here are some more comments.

What stands out to me is that:

1) This PR/the code is very verbose. It would be good to simplify and DRY up where possible to improve maintainability going forward.
2) We currently ignore a lot of test cases, often for reasons that aren't conclusive to me (or seem outdated). Can we double-check whether we really can't make them work?

.github/workflows/rust.yml
febyeji: Changed mkdir bin → mkdir -p bin. Needed because the download step now runs when either cache misses. If bitcoind is cached but electrs isn't, bin/ already exists from the cache restore, and mkdir bin would fail.

.github/workflows/rust.yml
febyeji: The change to add 29.0 only to the cache key is to invalidate caches on bitcoind upgrades. The path can stay as it is.

.github/workflows/eclair-integration.yml
febyeji: Fixed, changed to `curl -sf` so HTTP 4xx/5xx responses correctly fail the readiness check.

.github/workflows/cln-integration.yml
febyeji: Removed the miner wallet entirely. Funding is now handled inside the test code via `premine_and_distribute_funds`, which generates 101+ blocks to ensure coinbase maturity.

tests/common/scenarios/channel.rs
febyeji: Fixed.

tests/common/cln.rs
febyeji: Addressed, `make_error` is now a default method on the `ExternalNode` trait so there's no duplication across implementations.

tests/integration_tests_cln.rs
febyeji: This is the keysend test, not force close. CLN's keysend plugin hardcodes `min_final_cltv_expiry=22` (`plugins/keysend.c:230`), but LDK requires at least 42 (`HTLC_FAIL_BACK_BUFFER(39) + 3`). CLN's keysend RPC has no parameter to override this. Keeping ignored until CLN adds a `final_cltv` parameter.

.github/workflows/cln-integration.yml
febyeji: Fixed, changed from `chmod -R 755` to setting 755 only on directories and 644 on files.

.github/workflows/lnd-integration.yml
febyeji: Fixed alongside the CLN workflow, same approach.

tests/common/scenarios/disconnect.rs
febyeji: Extracted `disconnect_by_side()` helper in `connectivity.rs` and reuse it in all disconnect scenarios.

tests/integration_tests_cln.rs
febyeji: Done, created `interop_tests!` macro in `scenarios/mod.rs` that generates `test_scenarios`, `test_combos`, and `test_send_keysend` for each implementation. Each test file now just calls `interop_tests!(setup_clients)`.

tests/common/lnd.rs
febyeji: Added a comment documenting the truncation behavior. The trait takes msat denomination for consistency across implementations, and LND's API only accepts sats so we truncate here. Agreed that this could cause surprises if we assert sub-sat precision in the future.

tests/integration_tests_cln.rs
febyeji: CLN v25 with `--experimental-splicing` does advertise the standard BOLT bit 62 (alongside experimental bit 162), so feature negotiation with LDK works. However, the actual splice flow panics in rust-lightning during `tx_signatures` exchange (`channel.rs` debug_assert, fixed in rust-lightning#4472 but not yet in the pinned rev). Added `--experimental-splicing` to docker-compose and a `test_cln_splice_in` test, but keeping it `#[ignore]` until the dependency is bumped. Tracking: ldk-node#857.

tests/integration_tests_eclair.rs
febyeji: Eclair uses custom feature bit 154 (`SplicePrototype`) instead of the BOLT-standard bit 62/63. This is documented in their source: [Features.scala](https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala), described as "custom splices implementation for phoenix, to be replaced once splices is spec-ed." Splice interop with LDK isn't possible until Eclair migrates to standard bits. Removed the ignored splice tests and set `supports_splice() = false` with this reference.

.github/workflows/rust.yml
tnull: Makes sense

tests/integration_tests_cln.rs
tnull: Right, but we have a bunch of tests that are failing because of #857 anyways, so might as well have the new tests fail rather than ignoring it as we will bump the LDK version shortly.

febyeji: You are right. I will update the test.

tests/integration_tests_cln.rs
tnull: > This is the keysend test, not force close.

I'm confused what you mean by that?

> CLN's keysend plugin hardcodes `min_final_cltv_expiry=22` (`plugins/keysend.c:230`), but LDK requires at least 42 (`HTLC_FAIL_BACK_BUFFER(39) + 3`). CLN's keysend RPC has no parameter to override this. Keeping ignored until CLN adds a `final_cltv` parameter.

Okay, interesting, but that seems an entirely different issue than originally described in the comment / PR description? Or am I confused? The comment now says "needs CLN upstream fix", but are they even aware of the issue and agreed it should be fixed?

tests/integration_tests_eclair.rs
tnull: Okay, but this again is entirely different from what the comment originally said?

tests/integration_tests_cln.rs
febyeji: > This is the keysend test, not force close.

 Sorry for the confusing wording, I think I misread your original comment as asking about a different test. 
You're right that there's no upstream CLN issue filed for this.

The min_final_cltv_expiry=22 hardcoding in plugins/keysend.c is something I found by reading CLN source, but I haven't confirmed whether the CLN team considers it a bug. will un-ignored the test so we can see the actual error in CI.

tests/integration_tests_eclair.rs
febyeji: > Okay, but this again is entirely different from what the comment originally said?

Ah, my reply was about splice feature on the below code, sorry for the confusion. The original comment was about Eclair 0.8.0, I use 0.14.0, so this is now outdated.

tests/integration_tests_cln.rs
tnull: Now opened https://github.com/ElementsProject/lightning/issues/9034

tnull (commented):
So, this needs another rebase unfortunately. This should then fix the failing VSS and Rust/macOS CI jobs. For the LND job, we should probably disable splicing for the time being, right? 

For the CLN and Eclair issues, I think I'd be fine with re-adding the `ignore` as we have a path to get them fixed, but please open trakcking issues for both so we don't forget to re-enable them once the next CLN and Eclair releases ship.

febyeji:
I pushed an update addressing the latest feedback.

- Rebased onto main (fixes VSS and Rust/macOS).
- Re-added `#[ignore]` on incompatible combos, each
with a tracking issue.
- Simplified test code: use `#[tokio::test]` and `interop_tests!` macro

tnull (commented):
Eclair CI seems failing now unfortunately, other CI failure (Rust stable) is unrelated.

febyeji:
Thanks, I fixed the flaky problems on the Eclair code. Sorry for not catching them before requesting review.

I implemented most of the scenarios in the #766. However, I couldn't figure out what a fee rate disagreement scenario should look like in the current CI. Do you have any thoughts on how to approach this?

tnull:
> I implemented most of the scenarios in the #766. However, I couldn't figure out what a fee rate disagreement scenario should look like in the current CI. Do you have any thoughts on how to approach this?

The most notorious case is usally if two nodes disagree on the acceptable fee ranges during cooperative channel close. If they can't find common ground it leads to them force-closing, which is unnecessarily expensive for the users and also locks up funds for extended periods of time. In the tests it would be good to check that we don't hit such cases on channel close, although most-all implementations might have special fee rate overrides for `regtest`, so the tests might actually only be of limited use.

tnull:
Feel free to rebase now that #891 landed to unbreak CI.

febyeji:
Thanks for the reminder! For the discussion yesterday, I gave it a try on making a scenario. However, regtest fallback fees on both sides land to the same floor, so the test ends up indistinguishable from the existing ones. I am trying to see if adjusting the container configs can make changes as a follow-up.

tnull:
> Thanks for the reminder! For the discussion yesterday, I gave it a try on making a scenario. However, regtest fallback fees on both sides land to the same floor, so the test ends up indistinguishable from the existing ones. I am trying to see if adjusting the container configs can make changes as a follow-up.

Alright, then maybe let's just aim to get this PR landed soon and address further test cases in a follow-up?

febyeji:
Okay, I've pushed the latest changes. The current structure looks like this:

- `tests/common/external_node.rs`: `ExternalNode` trait abstracting CLN/LND/Eclair behind a common interface
- `tests/common/{cln,lnd,eclair}.rs`: implementation of `ExternalNode`
- `tests/common/scenarios/`: shared scenario runners (`connectivity`, `channel`, `payment`)
- `tests/integration_tests_{cln,lnd,eclair}.rs`: thin per-backend entry points that run shared scenarios
- `tests/docker/` + GitHub Actions workflows: container definitions and split CI pipelines

I think this is ready for another look. Happy to address any feedback!

.github/workflows/eclair-integration.yml
```diff
@@ -0,0 +1,56 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  check-eclair:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
tnull, line 14: We've recently bumped our Github actions, would be good to do the same here.

tests/common/scenarios/channel.rs
```diff
@@ -0,0 +1,103 @@
...
+
+	(user_channel_id, ext_channel_id)
+}
+
+/// Cooperative close from the chosen side. Mines 1 block and asserts ChannelClosed.
+pub(crate) async fn cooperative_close<E: ElectrumApi>(
+	node: &Node, peer: &(impl ExternalNode + ?Sized), bitcoind: &BitcoindClient, electrs: &E,
+	user_channel_id: &ldk_node::UserChannelId, ext_channel_id: &str, initiator: Side,
+) {
+	tokio::time::sleep(Duration::from_secs(2)).await;
```
tnull, line 52: Rather than hard-coding 2 seconds once, should we retry a few times?

tests/common/scenarios/mod.rs
```diff
@@ -0,0 +1,251 @@
...
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
+// accordance with one or both of these licenses.
+
+//! Shared interop test scenarios, generic over `ExternalNode`.
+//!
+//! - `channel` / `payment` / `connectivity` -- composable building blocks
+//! - `interop_tests!` macro -- emits one `#[tokio::test]` per scenario
```
tnull, line 11: This comments seems stale? Maybe just drop the comments?

tests/common/scenarios/mod.rs
```diff
@@ -0,0 +1,251 @@
...
+use super::{generate_blocks_and_wait, premine_and_distribute_funds};
+
+#[derive(Debug, Clone, Copy)]
+pub(crate) enum Side {
+	Ldk,
+	External,
+}
+
+/// Retry an async operation with 1s delay; used for ops that may fail due to gossip delay.
+pub(crate) async fn retry_until_ok<F, Fut, T, E>(max_attempts: u32, operation: &str, mut f: F) -> T
```
tnull, line 35: This could likely be DRYed up with the `exponential_backoff_poll` helper we already have.

tnull (approved):
Thanks! 

It would have been great to split this up into 4+ different commits rather than squashing all into one. I think there is also some cleanup/DRYing up potential, but this is test-only code and a clear improvement over before.

So I'll go ahead and land this, but added some comments that would be good to address in a follow-up. And more generally, would be great to further iterate on the code a bit more to improve readability and reduce verbosity if possible.

febyeji:
Thank you for the review! I will apply your feedback in the follow-up commits.

PR #841, opened 2026-03-24
Expose ChannelCounterparty and ReserveType in ChannelDetails

### LINKED ISSUE #305: Expose a way to access channel or reserve type

enigbe:
## What this PR does

In #305, we needed a way to expose channel type and the channel reserve without leaking implementation details not useful to users. The related discussion in #141 proposed a `ReserveType` abstraction that bakes in both in a user-relevant way. This PR introduces the said enumeration to `ChannelDetails` with the following variants: 
- `Legacy`: signifying pre-anchor channel types where on-chain fees paying for broadcast transactions following channel closure were pre-determined 
- `TrustedPeersNoReserve`: for anchor type channels with trusted peers 
- `Adaptive`: indicating anchor channel with adaptive reserve, reflecting dynamic best-effort attempt at fee-bumping.
(see  @jkczyz's [suggestion](https://github.com/lightningdevkit/ldk-node/pull/141#discussion_r1629979502)) 

We modify the `ChannelDetails` constructor to accept an optional anchor channels config to address the challenge of users cross-referencing the channel details and config to determine if counterparties are trusted.

Additionally, following [recommendation](https://github.com/lightningdevkit/ldk-node/pull/810#pullrequestreview-3865820819) in #810 about negotiated features, this PR exposes per-peer `InitFeatures`, and as a consequence, modifies `ChannelDetails` by replacing the flattened `counterparty_*` with `ChannelCounterparty` that encapsulates them and mirror LDK's. 

## Issue Addressed 
Closes #305. 
Builds on discussions and recommendations from #141 and #810.

### DIFF
--- a/src/ffi/types.rs
+++ b/src/ffi/types.rs
@@ -1526,6 +1526,7 @@ pub struct NodeFeatures {
 	pub(crate) inner: LdkNodeFeatures,
 }
 
+#[uniffi::export]
 impl NodeFeatures {
 	/// Constructs node features from big-endian BOLT 9 encoded bytes.
 	#[uniffi::constructor]
@@ -1816,10 +1817,12 @@ impl From<LdkNodeFeatures> for NodeFeatures {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, uniffi::Object)]
+#[uniffi::export(Debug, Eq)]
 pub struct InitFeatures {
 	pub(crate) inner: LdkInitFeatures,
 }
 
+#[uniffi::export]
 impl InitFeatures {
 	/// Constructs init features from big-endian BOLT 9 encoded bytes.
 	#[uniffi::constructor]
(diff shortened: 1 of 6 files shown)


src/types.rs
```diff
@@ -347,6 +349,97 @@ impl fmt::Display for UserChannelId {
 	}
 }
 
+/// Information needed for constructing an invoice route hint for this channel.
```
tnull: Everything related to `uniffi` should live in `ffi/types.rs`, which also avoids the feature-gate on individual items.

src/types.rs
```diff
@@ -347,6 +349,97 @@ impl fmt::Display for UserChannelId {
...
+	/// The minimum difference in cltv_expiry between an ingoing HTLC and its outgoing counterpart,
+	/// such that the outgoing HTLC is forwardable to this counterparty.
+	pub cltv_expiry_delta: u16,
+}
+
+#[cfg(feature = "uniffi")]
+uniffi::custom_type!(InitFeatures, Vec<u8>, {
+	remote,
+	try_lift: |val| Ok(InitFeatures::from_le_bytes(val)),
+	lower: |obj| obj.le_flags().to_vec(),
```
tnull: Generally not the biggest fan of exposing this as `Vec<u8>`, but if we do, why pick little endian over big endian?

src/types.rs
```diff
@@ -347,6 +349,97 @@ impl fmt::Display for UserChannelId {
...
+	/// This value is not included in [`inbound_capacity_msat`] as it can never be spent.
+	///
+	/// [`inbound_capacity_msat`]: ChannelDetails::inbound_capacity_msat
+	pub unspendable_punishment_reserve: u64,
+	/// Information on the fees and requirements that the counterparty requires when forwarding
+	/// payments to us through this channel.
+	pub forwarding_info: Option<CounterpartyForwardingInfo>,
+	/// The smallest value HTLC (in msat) the remote peer will accept, for this channel. This field
+	/// is only `None` before we have received either the `OpenChannel` or `AcceptChannel` message
+	/// from the remote peer, or for `ChannelCounterparty` objects serialized prior to LDK 0.0.107.
```
tnull: All compatibility notes should be formulated referencing LDK Node releases, not LDK releases.

src/types.rs
```diff
@@ -529,13 +600,42 @@ pub struct ChannelDetails {
...
-	fn from(value: LdkChannelDetails) -> Self {
+impl ChannelDetails {
+	pub(crate) fn from_ldk(
+		value: LdkChannelDetails, anchor_channels_config: Option<&AnchorChannelsConfig>,
+	) -> Self {
+		let is_anchor_channel =
+			value.channel_type.as_ref().map_or(false, |ct| ct.supports_anchors_zero_fee_htlc_tx());
+
+		let reserve_type = if is_anchor_channel {
+			let is_trusted = anchor_channels_config.map_or(false, |c| {
```
tnull: nit: If we use `if let Some(anchors_channels_config) = ` once, we avoid `map_or`ing every time and can drop the bool, etc.

tnull (commented):
Thanks! Looks pretty good, have a few initial comments.

src/types.rs
enigbe: I've done this.

src/types.rs
enigbe: I was mistaken about the LE/BE ease of bit indexing vs over the wire serialization. I've replaced the Vec<u8> with a dedicated wrapper type that stores `LdkInitFeatures` directly, so the endianness question no longer applies.

src/types.rs
tnull: This seems unresolved, I think we can move this `CounterpartyForwardingInfo` to `ffi/types.rs`?

tnull (commented):
One comment, otherwise looks good. Feel free to squash (also the new fixup)

src/types.rs
```diff
@@ -558,13 +612,44 @@ pub struct ChannelDetails {
...
+			if value.channel_type.as_ref().is_some_and(|ct| ct.supports_anchors_zero_fee_htlc_tx())
+			{
+				if let Some(config) = anchor_channels_config {
+					if config.trusted_peers_no_reserve.contains(&value.counterparty.node_id) {
+						ReserveType::TrustedPeersNoReserve
+					} else {
+						ReserveType::Adaptive
+					}
+				} else {
+					ReserveType::Adaptive
```
tnull, line 617: Hmm, isn't this wrong? I believe we'd fallback to legacy if we don't have an `anchors_channel_config` set?

I guess we could only land here if we had already accepted an anchor channel and then the user changed `anchors_channels_config` afterwards? 

Might be good to at least leave some comments here to give some rationale.

enigbe: Good catch here. You are right that we only reach this branch if an anchor channel was previously opened with `anchor_channels_config` set, and then later unset. In that case, `Legacy` would be incorrect and `Adaptive` could also be wrong . The channel will still be an anchor type, but if it was originally `TrustedPeersNoReserve`, we'd silently reclassify it as `Adaptive`.

I think the main problem is that we derive the `reserve_type` at query time from the current config rather than recording it when the channel is opened. So any config change could cause a loss of distinction where we can't distinguish between anchor channels that were `Adaptive` vs `TrustedPeersNoReserve`.

We could explore persisting the reserve type (keyed by `ChannelId`) when we open a channel, and have `list_channels` look up the stored value instead of re-deriving it. What do you think?

tnull: I'd rather not persist it. Can we just determine based on the feature, and then decide between `Adaptive` and `TrustedPeersNoReserve` based on whether or not the peer is present in the no reserve list? I guess that's kind of what we're doing already?

enigbe: Yeah, we already do this. There's just that one edge case where a user unsets the anchor channels config after previously setting it. We'd be able to tell if it's an anchor channel but not whether it's adaptive or not.

tnull: I guess that's fine. Could you maybe add a comment to reflect the current rationale/thinking and documenting this edge case? And then we can probably move ahead. Not sure if it's worth over-engineering this.

enigbe: Sure. I'll add the comment.

src/ffi/types.rs
```diff
@@ -1496,6 +1498,114 @@ pub enum ClosureReason {
...
+	/// Returns the BOLT 9 big-endian encoded representation of these features.
+	pub fn to_bytes(&self) -> Vec<u8> {
+		self.inner.encode()
+	}
+
+	/// Whether the peer supports `option_static_remotekey` (bit 13).
+	///
+	/// This ensures the non-broadcaster's output pays directly to their specified key,
+	/// simplifying recovery if a channel is force-closed.
+	pub fn supports_static_remote_key(&self) -> bool {
```
tnull, line 1839: It seems we're missing a bunch of features here? E.g. `supports_keysend` and others seem to be missing?

enigbe: I've added the missing features but `support_keysend` applies to `NodeFeatures` only.

src/ffi/types.rs
```diff
@@ -1496,6 +1498,248 @@ pub enum ClosureReason {
...
+	/// preventing the real channel UTXO from being revealed during routing.
+	pub fn supports_scid_privacy(&self) -> bool {
+		self.inner.supports_scid_privacy()
+	}
+
+	/// Whether the peer requires `option_zeroconf` (bit 51).
+	///
+	/// Zero-conf channels can be used immediately without waiting for
+	/// on-chain funding confirmations.
+	pub fn requires_zero_conf(&self) -> bool {
```
tnull, line 1914: Hmm, this seems to be the only `required_` variant we're exposing. Is this intentional?

enigbe: Yes, intentional. In `rust-lightning`, `supports_zero_conf` actually clears the feature, while `requires_zero_conf` checks whether zero-conf is supported. I kept the upstream naming for parity, even though it’s admittedly confusing.

tnull: > Yes, intentional. In `rust-lightning`, `supports_zero_conf` actually clears the feature

Not sure I understand what you mean by that? Are you referring to this recently-fixed bug? https://github.com/lightningdevkit/rust-lightning/pull/4517

Generally, requiring and supporting a feature are two different things (cf https://github.com/lightning/bolts/blob/master/09-features.md) and we should expose both APIs whenever we can. IIRC, zero conf might be the only (?) exception as it's always deemed required if it's set.

enigbe: > Not sure I understand what you mean by that? Are you referring to this recently-fixed bug? https://github.com/lightningdevkit/rust-lightning/pull/4517

I wasn't aware this was a bug. Due to the macro misalignment, requires_zero_conf currently maps to $supported_getter, so it actually checks support rather than requirement. I can either open a separate PR to bump rust-lightning to https://github.com/lightningdevkit/rust-lightning/commit/47122e8f48b55f1c43b820712603d42d4214c4,  or include a drop-me commit in this PR that updates the revision to one with the fix, which gets cleaned up when rebasing on a planned rust-lightning bump. Which would you prefer?

> Generally, requiring and supporting a feature are two different things (cf https://github.com/lightning/bolts/blob/master/09-features.md) and we should expose both APIs whenever we can. IIRC, zero conf might be the only (?) exception as it's always deemed required if it's set.

Thanks for clarifying. The APIs exposed so far only cover `supports_*` variants. I'll add those for the required features?

tnull: > or include a drop-me commit in this PR that updates the revision to one with the fix, which gets cleaned up when rebasing on a planned rust-lightning bump. Which would you prefer?

Yeah, let's do that!

> Thanks for clarifying. The APIs exposed so far only cover `supports_*` variants. I'll add those for the required features?

Yes please, would be good to offer users the capability to discern what features are optional, and which are required.

src/ffi/types.rs
```diff
@@ -1519,6 +1520,446 @@ pub enum ClosureReason {
...
+
+	/// Whether the peer supports `option_static_remotekey` (bit 13).
+	///
+	/// This ensures the non-broadcaster's output pays directly to their specified key,
+	/// simplifying recovery if a channel is force-closed.
+	pub fn supports_static_remote_key(&self) -> bool {
+		self.inner.supports_static_remote_key()
+	}
+
+	/// Whether the peer requires `option_static_remotekey` (bit 12).
```
tnull: See comment over at #810, would be probably better not to mention the bit numbers explicitly.

tnull (commented):
Looks good I think. One nit.

tnull (commented):
This needs a minor rebase by now.

src/ffi/types.rs
enigbe: I've removed the bit numbers from the docs.

enigbe:
Thanks for the review @tnull.

I've rebased and also simplified the documentation on `InitFeatures` to match the lean style used with `NodeFeatures`.

src/types.rs
```diff
@@ -601,13 +656,54 @@ pub struct ChannelDetails {
...
+				node_id: value.counterparty.node_id,
+				features: maybe_wrap(value.counterparty.features),
+				unspendable_punishment_reserve: value.counterparty.unspendable_punishment_reserve,
+				forwarding_info: value.counterparty.forwarding_info,
+				// unwrap safety: This value will be `None` for objects serialized with LDK versions
+				// prior to 0.0.115.
+				outbound_htlc_minimum_msat: value
+					.counterparty
+					.outbound_htlc_minimum_msat
+					.expect("value is set for objects serialized with LDK v0.0.107+"),
```
tnull: Oh, no, this is no accurate? The original docs are:

```rust
	/// This field is only `None` before we have received either the `OpenChannel` or
	/// `AcceptChannel` message from the remote peer.
```

So this will have us panic if somebody lists channels before a channel is opened.

src/types.rs
```diff
@@ -415,6 +423,75 @@ impl fmt::Display for UserChannelId {
 	}
 }
 
+/// Channel parameters which apply to our counterparty. These are split out from [`ChannelDetails`]
+/// to better separate parameters.
+#[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct ChannelCounterparty {
```
tnull, line 356: This needs to be re-exported via a `pub use` in `src/ffi/types.rs`.

src/types.rs
enigbe: I've fixed this.

src/types.rs
enigbe: Why does this need to be re-exported in `src/ffi/types.rs`?  We don't do this for other `uniffi::Record`s. Do you mean `src/lib.rs`? If it's the latter, I've re-exported in the fix-up commit.

src/ffi/types.rs
```diff
@@ -1815,6 +1815,296 @@ impl From<LdkNodeFeatures> for NodeFeatures {
 	}
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, uniffi::Object)]
+pub struct InitFeatures {
+	pub(crate) inner: LdkInitFeatures,
+}
+
+impl InitFeatures {
```
tnull, line 1826: This seems to be missing `#[uniffi::export]`?

src/types.rs
```diff
@@ -527,13 +585,49 @@ pub struct ChannelDetails {
...
+	pub reserve_type: ReserveType,
 }
 
-impl From<LdkChannelDetails> for ChannelDetails {
-	fn from(value: LdkChannelDetails) -> Self {
+impl ChannelDetails {
+	pub(crate) fn from_ldk(
+		value: LdkChannelDetails, anchor_channels_config: Option<&AnchorChannelsConfig>,
+	) -> Self {
+		let reserve_type =
```
tnull: Hmm, the LDK docs on `channel_type` say: "None until negotiation completes and the channel type is finalized."

This means we'll erroneously report all channels as `Legacy` util the negotiation is fully done, no? Probably will need to make this an optional field, too, or have a state for 'not ready' or similar?

src/ffi/types.rs
enigbe: This has been added.

src/types.rs
enigbe: Leaning more towards making it optional. Also, I updated the documentation to highlight this.

src/ffi/types.rs
tnull: No, the one you added is also fine to keep, but we'll also need to add `uniffi::export` to the impl block, as otherwise the methods won't be available in bindings.

enigbe: That's right. I've added this and extended the coverage in the python test to check the now-exposed methods. I noticed `NodeFeatures` had the same issue, so I fixed and covered that as well.

tnull (approved):
Thanks!

PR #846, opened 2026-03-25
Add cancel_invoice to Bolt11Payment

benthecarman:
Allow cancelling a previously created BOLT11 invoice by payment hash. Enterprise integrators need this when they short-circuit an invoice with an internal database transfer or when an alternative payment method is used (e.g., on-chain payment via unified URI).

The `PaymentClaimable` event handler rejects HTLCs for cancelled invoices only when the preimage is known (auto-claim payments), preserving retry behavior for manual-claim (_for_hash) payments.

`fail_for_hash` already did the logic we needed so `cancel_invoice` just does a little bit extra pre-validation and then calls it. Could make the argument to just use `fail_for_hash`,  update its docs, and the if statement to reject payments mark as failed in the `PaymentClaimable`. The only functionality change would be that if someone retries to do a payment after a `fail_for_hash`, we don't let the user evaluate again.

### DIFF
--- a/src/payment/bolt11.rs
+++ b/src/payment/bolt11.rs
@@ -555,6 +555,40 @@ impl Bolt11Payment {
 		Ok(())
 	}
 
+	/// Allows to cancel a previously created invoice identified by the given payment hash.
+	///
+	/// This will mark the corresponding payment as failed and cause any incoming HTLCs for this
+	/// invoice to be automatically failed back.
+	///
+	/// Will check that the payment is known and has not already been claimed, and will return an
+	/// error otherwise.
+	pub fn cancel_invoice(&self, payment_hash: PaymentHash) -> Result<(), Error> {
+		let payment_id = PaymentId(payment_hash.0);
+
+		if let Some(info) = self.payment_store.get(&payment_id) {
+			if info.direction != PaymentDirection::Inbound {
+				log_error!(
+					self.logger,
+					"Failed to cancel invoice for non-inbound payment with hash {payment_hash}"
+				);
+				return Err(Error::InvalidPaymentHash);
+			}
+
+			if info.status == PaymentStatus::Succeeded {
+				log_error!(
+					self.logger,
+					"Failed to cancel invoice with hash {payment_hash}: payment has already been claimed",
+				);
+				return Err(Error::InvalidPaymentHash);
+			}
+		} else {
+			log_error!(self.logger, "Failed to cancel unknown invoice with hash {payment_hash}");
+			return Err(Error::InvalidPaymentHash);
+		}
+
+		self.fail_for_hash(payment_hash)
+	}
+
 	/// Returns a payable invoice that can be used to request and receive a payment of the amount
 	/// given.
 	///
--- a/src/event.rs
+++ b/src/event.rs
@@ -688,6 +688,19 @@ where
 						};
 					}
 
+					// If the invoice has been canceled, reject the HTLC. We only do this
+					// when the preimage is known to preserve retry behavior for `_for_hash`
+					// manual-claim payments, where `fail_for_hash` may have been a
+					// temporary rejection (e.g., preimage not yet available).
+					if info.status == PaymentStatus::Failed && purpose.preimage().is_some() {
+						log_info!(
+							self.logger,
+							"Refused inbound payment with ID {payment_id}: invoice has been canceled."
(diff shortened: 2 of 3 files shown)


src/payment/bolt11.rs
```diff
@@ -555,6 +555,40 @@ impl Bolt11Payment {
...
 	}
 
+	/// Allows to cancel a previously created invoice identified by the given payment hash.
+	///
+	/// This will mark the corresponding payment as failed and cause any incoming HTLCs for this
+	/// invoice to be automatically failed back.
+	///
+	/// Will check that the payment is known and has not already been claimed, and will return an
+	/// error otherwise.
+	pub fn cancel_invoice(&self, payment_hash: PaymentHash) -> Result<(), Error> {
```
tnull, line 565: Hmm, I always found LND's terminology of 'cancelling an invoice' very confusing. In LDK Node it's even more so confusing as we don't store invoices (yet, cf. #811), so I really don't know what 'cancelling an invoice' is supposed to mean in this context. 

As part of the move towards the payment metadata store we will also finally stop tracking expected inbound BOLT11 payments in the payment store (but then rather store the actual invoices/offers for the user), which means the logic in the event handler won't work as it does now. Let me see to finish #811 and #784 finally, and then we can decide how invoice 'cancellation' could make the most sense. 

Somewhat orthogonally I do wonder if it would be best to actually do this properly in LDK, i.e., add a remove_payment method on `ChannelManager` that  has it forget the preimage, simply?

benthecarman: > Somewhat orthogonally I do wonder if it would be best to actually do this properly in LDK, i.e., add a remove_payment method on ChannelManager that has it forget the preimage, simply?

I don't think that is possible, invoice creation is stateless, its not storing the preimage.

TheBlueMatt: IMO this whole concept is an anti-pattern. If someone wants to pay you extra money in LN you cannot stop them, there are a number of ways in which they can do it, and we're not gonna try to close them all. Further, storing invoices/offers before they're paid is an anti-pattern - creation is stateless for a reason, and tons of people are going to expose APIs where you can request an invoice be created, suddenly subjecting yourself to arbitrary DoS attacks filling your payment store. Unless you actually have to pay for it, free DoS is a really shitty API.

tnull: > I don't think that is possible, invoice creation is stateless, its not storing the preimage.

Ah, lol, true.

> IMO this whole concept is an anti-pattern.

+1 on this, though users seem to expect to have the functionality by now (see elsewhere). But probably we should simply push back on that.

> Further, storing invoices/offers before they're paid is an anti-patter

Well, that in turn has been a big feature request  by users for a long time now. And *someone* will need to store the invoice/offers so just handing the responsibility to the user doesn't actually solve the issue necessarily. In our case we also need to store some metadata based on the payment type (pre-negotiated LSP fees, but preimages could also count towards that). So we will add storage for it, but will attempt to make it optional (cf. #51).

TheBlueMatt: > Well, that in turn has been a big feature request by users for a long time now. And someone will need to store the invoice/offers so just handing the responsibility to the user doesn't actually solve the issue necessarily.

I'd like to better understand the use-cases here. For something like a merchant running an online store, their current architecture probably includes storing invoices/payment hashes in their payments/orders db already, so that when they receive a payment they can match it to the order. LDK Node (or any lightning node) could of course allow the downstream logic to store additional metadata in a payment store around an invoice, allowing the matching to be done by putting the order id in the lightning node, but you still have that order information elsewhere.

If we end up offering storage to downstream code in an invoice store, though, we should actually not store it at all and put it in the invoice payment-metadata. I think its now been more than long enough that that is likely reliably usable on the network, and it avoids storing entirely, at the cost of being rather data limited because it ends up in the QR code and onion. This only works for the "store only an order id" usecase, not full order info.

Are there other reasons why people want to store the full generated-invoice history?

> In our case we also need to store some metadata based on the payment type (pre-negotiated LSP fees, but preimages could also count towards that). So we will add storage for it, but will attempt to make it optional (cf. https://github.com/lightningdevkit/ldk-node/issues/51).

If its small enough, this seems like a great use for payment metadata. We might want to have the payment secret commit to the (encrypted, but not-MAC'd) payment metadata, though...

PR #847, opened 2026-03-26
Fix test flakes from port allocation races

joostjager:


tests/common/mod.rs
```diff
@@ -268,17 +269,14 @@ pub(crate) fn random_storage_path() -> PathBuf {
...
-
-	while listening_addresses.len() < num_addresses {
-		let socket = std::net::TcpListener::bind("127.0.0.1:0").unwrap();
-		let address: SocketAddress = socket.local_addr().unwrap().into();
-		listening_addresses.insert(address);
-	}
+static NEXT_PORT: AtomicU16 = AtomicU16::new(20000);
 
-	listening_addresses.into_iter().collect()
+pub(crate) fn random_listening_addresses() -> Vec<SocketAddress> {
```
tnull: nit: Given it's not random anymore, let's name it `generate_listening_addresses` or similar.

tnull (commented):
Okay, I think we considered that before but you were of the opinion it wouldn't work? Any case, open to test whether it actually makes a difference (I'm not positive the flakes are actually stemming from this at all).

Just one nit.

tests/common/mod.rs
joostjager: renamed

tests/common/mod.rs
```diff
@@ -304,7 +302,7 @@ pub(crate) fn random_config(anchor_channels: bool) -> TestConfig {
 	println!("Setting random LDK storage dir: {}", rand_dir.display());
 	node_config.storage_dir_path = rand_dir.to_str().unwrap().to_owned();
 
-	let rand_listening_addresses = random_listening_addresses();
+	let rand_listening_addresses = generate_listening_addresses();
```
tnull: nit: Likewise please `s/rand_//g` here and below.

joostjager: Fixed

tnull (approved):

PR #853, opened 2026-03-30
Enable 0conf and 0reserve on channels with trusted peers

tankyleo:


Cargo.toml
```diff
@@ -79,13 +79,13 @@ async-trait = { version = "0.1", default-features = false }
 vss-client = { package = "vss-client-ng", version = "0.5" }
 prost = { version = "0.11.6", default-features = false}
 #bitcoin-payment-instructions = { version = "0.6" }
-bitcoin-payment-instructions = { git = "https://github.com/joostjager/bitcoin-payment-instructions", branch = "ldk-dcf0c203e166da2348bef12b2e5eff4a250cdec7" }
+bitcoin-payment-instructions = { git = "https://github.com/tankyleo/bitcoin-payment-instructions", branch = "ldk-688544da72cb348e4405d39a75e4d81102c1278a" }
```
jkczyz: Should we use `rev` here? Seems otherwise the revision could move if the branch is updated.

src/config.rs
```diff
@@ -156,12 +156,19 @@ pub struct Config {
...
+	/// A list of peers that we trust. If a peer on this list opens a channel to us, we will
+	/// forward their HTLCs before any confirmations of the funding transaction (zero-conf), and
+	/// allow them to spend their entire balance (zero-reserve). If we open a channel to a peer
+	/// on this list, we will allow them to spend their entire channel balance (note that for
+	/// channels *we* open, the decision of whether to accept HTLC forwards with no
+	/// confirmations of the funding transaction is *the peer's* decision).
+	///
+	/// **Note:** Allowing payments via zero-confirmation channels is potentially insecure if
+	/// the funding transaction never gets confirmed on-chain. Zero-reserve channels
+	/// allow the counterparty to make cheating attempts with no financial penalty.
```
tnull:
```suggestion
	/// allow the counterparty to attempt cheating with no financial penalty.
```

src/config.rs
```diff
@@ -156,12 +156,19 @@ pub struct Config {
 	/// **Note**: We will only allow opening and accepting public channels if the `node_alias` and the
 	/// `listening_addresses` are set.
 	pub node_alias: Option<NodeAlias>,
-	/// A list of peers that we allow to establish zero confirmation channels to us.
-	///
-	/// **Note:** Allowing payments via zero-confirmation channels is potentially insecure if the
-	/// funding transaction ends up never being confirmed on-chain. Zero-confirmation channels
-	/// should therefore only be accepted from trusted peers.
-	pub trusted_peers_0conf: Vec<PublicKey>,
+	/// A list of peers that we trust. If a peer on this list opens a channel to us, we will
```
tnull: nit: Please expand the first sentence slightly and give it its own paragraph.

src/lib.rs
```diff
@@ -1196,27 +1196,57 @@ impl Node {
...
-				self.peer_store.add_peer(peer_info)?;
-				Ok(UserChannelId(user_channel_id))
-			},
-			Err(e) => {
-				log_error!(self.logger, "Failed to initiate channel creation: {:?}", e);
-				Err(Error::ChannelCreationFailed)
-			},
+		let is_trusted_peer = self.config.trusted_peers_0conf_0reserve.contains(&node_id);
+		if is_trusted_peer {
+			match self.channel_manager.create_channel_to_trusted_peer_0reserve(
```
tnull: Hmm, I'm a bit dubious if we indeed should always opt for 0reserve if the peer is marked trusted?

src/config.rs
```diff
@@ -156,12 +156,19 @@ pub struct Config {
...
+	/// forward their HTLCs before any confirmations of the funding transaction (zero-conf), and
+	/// allow them to spend their entire balance (zero-reserve). If we open a channel to a peer
+	/// on this list, we will allow them to spend their entire channel balance (note that for
+	/// channels *we* open, the decision of whether to accept HTLC forwards with no
+	/// confirmations of the funding transaction is *the peer's* decision).
```
tnull: Hmm, to be honest this is what makes it a bit weird to mix the concepts: yes, both require trust in the counterparty, but the AFAIU the use case is entirely opposite:

- We usually opt into 0conf if we are the client and want to allow the LSP to open 0conf channels
- If we are the client, there is little incentive to allow the LSP to not keep any reserve? So setting 0 reserve only makes sense if we are the LSP and want to allow the *client* to spend their full channel funds, no?

So, rather than also setting 0reserve if we are a LSPS client (see above), shouldn't we only set that if we are the LSP? Or maybe I'm missing something?

tankyleo: > If we are the client, there is little incentive to allow the LSP to not keep any reserve? So setting 0 reserve only makes sense if we are the LSP and want to allow the client to spend their full channel funds, no?

Thanks for the review will mull on this more, a quick reaction to this point: the LSP might want 0-reserve on *their* side of the channel for greater capital efficiency, ie dedicate as much of their capital as possible towards allowing clients to receive payments. If they get caught making even a single cheating attempt, their entire reputation is burned, and their business goes to 0 (hopefully !) so the incentives are aligned. 1000sats (the min reserve) * x users quickly becomes a lot of dead capital.

> So, rather than also setting 0reserve if we are a LSPS client (see above), shouldn't we only set that if we are the LSP? Or maybe I'm missing something?

I see your point yes. Do you have in mind to not offer any public API for setting 0-reserve on outbound channels (assuming the client is usually the opener of the channel, and not the LSP) ?

If we were to offer a completely separate API we have these options:
- Standalone function calls; this would at least require two more `open_channel` functions, one with max funding, and one with a parameterized amount.
- A separate list, similar to what we currently have for inbound channels.
- Some kind of boolean / enum flag on the existing `open_channel` functions.

src/config.rs
tankyleo: went with "to try to steal your funds with no financial penalty", see below

src/config.rs
tankyleo: See below,
```
/// A list of peers that we trust; these are peers that you've had some interaction with
/// out-of-band before clearing them to be on this list.
```

src/lib.rs
tankyleo: For the outbound case, I now add `Node::{open_0reserve_channel, open_0reserve_channel_with_all}`. For now, we do not allow 0-reserve channels to be announced.

src/config.rs
tankyleo: Discussed offline see the revised API below:
- `trusted_peers_0conf_0reserve` now only applies to inbound channels.
- Add `allow_client_0reserve` to `LSPS2ServiceConfig`
- For outbound 0-reserve channels, offer `Node::{open_0reserve_channel, open_0reserve_channel_with_all}`

Question:
    1. in case LSP opens the channel, the client might want 0-conf, but *not 0-reserve* (ie LSP keeps some funds at stake).
    2. In case the client opens the channel, the LSP wants both 0-conf, and 0-reserve (ie client keeps no funds at stake). 

At the moment the list `trusted_peers_0conf_0reserve` accommodates point number 2. For point number 1, it seems to me if you trust the LSP with a 0-conf channel, you would also trust them enough to set 0-reserve.

src/config.rs
```diff
@@ -156,12 +156,18 @@ pub struct Config {
...
+	///
+	/// If a trusted peer opens a channel to us, we will forward their HTLCs before any
+	/// confirmations of the funding transaction (zero-conf), and allow them to spend their
+	/// entire balance (zero-reserve).
+	///
+	/// **Note:** Allowing payments via zero-confirmation channels is insecure if the funding
+	/// transaction never gets confirmed on-chain. Zero-reserve channels allow the peer to try
+	/// to steal your funds with no financial penalty. Zero-confirmation, and zero-reserve
+	/// channels should therefore only be accepted from trusted peers.
+	pub trusted_peers_0conf_0reserve: Vec<PublicKey>,
```
tnull: Hmm, given the potentially very different usecases, I'm still not sure whether it makes sense to mix the two concepts like this. Should this maybe be a `HashMap<PublicKey, TrustedChannelFeatures>` to allow finer-grained control?

src/lib.rs
tnull: So if we mix it in `trusted_peers_0conf_0reserve`, this is somewhat strange, as I'd expect 0reserve to be set by default also for outbound channels if I add the peer to the list. But maybe making it more fine-grained would help, see below?

src/config.rs
tnull: > * `trusted_peers_0conf_0reserve` now only applies to inbound channels.

This is a bit odd to me, we should def. clarify that in the docs. 


How about https://github.com/lightningdevkit/ldk-node/pull/853#discussion_r3031944354 ? Then the user has more fine-grained control, and maybe we can even avoid the dedicated new `open_channel` variants? Or, if we want to keep them, it's more clear that they'd override whatever is in the `trusted_peers` map?

src/lib.rs
```diff
@@ -1395,6 +1419,83 @@ impl Node {
...
+	///
+	/// If Anchor channels are enabled, this will ensure the configured
+	/// [`AnchorChannelsConfig::per_channel_reserve_sats`] is available and will be retained before
+	/// opening the channel.
+	///
+	/// Returns a [`UserChannelId`] allowing to locally keep track of the channel.
+	///
+	/// [`Config::trusted_peers_0conf_0reserve`]: crate::config::Config::trusted_peers_0conf_0reserve
+	/// [`AnchorChannelsConfig::per_channel_reserve_sats`]: crate::config::AnchorChannelsConfig::per_channel_reserve_sats
+	pub fn open_0reserve_channel(
```
tnull, line 1445: If we go ahead with these new APIs, they need to be exposed in `uniffi` bindings.

src/config.rs
tankyleo: - As long as this only applies to inbound channels, I think these are similar use cases no ? If you trust your inbound peer with a 0-conf, you very likely also trust them with a 0-reserve channel, and vice versa.
- On the `main` branch this list already applies only to inbound channels.
- For a value set to `TrustedChannelFeatures::ZeroConfZeroReserve` in the hashmap, we would actually discard the `ZeroConf` bit for outbound channels, would this be a source of confusion ? "Hey I set my peer to trusted, and opened a channel to that peer, why is it not 0conf ?"
- I think this would be quite easy to grok: "Hey this list only applies to inbound channels, if you want to open a zero-reserve yourself, see `open_0reserve_channel`."

src/lib.rs
benthecarman, line 1445: we are getting a combinatorial explosion with these

src/config.rs
benthecarman: > I think this would be quite easy to grok: "Hey this list only applies to inbound channels, if you want to open a zero-reserve yourself, see open_0reserve_channel."

If me and you want to open a 0 reserve channel to eachother, but don't want to accept 0 conf from eachother. Would `open_0reserve_channel` work? Or when I do `open_0reserve_channel` will your node just reject it because its not in the list?

tankyleo: when alice calls `open_0reserve_channel`, she allows bob to spend his entire balance. Bob will accept regardless of whether Alice is on his "trusted peers" list.

If Bob has placed alice on his "trusted peers" list, bob in turn allows alice to spend her entire balance, and at the moment, allows alice to start immediately using the channel with 0-conf.

src/lib.rs
tankyleo: Done below

tankyleo: Discussed offline, we retain separate calls for now, and will ship your `ChannelBuilder::new().with_no_reserve().announced().fund_withall()` suggestion in case there is further expansion of these calls.

src/config.rs
tankyleo: > I think this would be quite easy to grok: "Hey this list only applies to inbound channels, if you want to open a zero-reserve yourself, see open_0reserve_channel."

Added this in the commit below

src/config.rs
tankyleo: > This is a bit odd to me, we should def. clarify that in the docs.

Done below

src/config.rs
tnull: > As long as this only applies to inbound channels, I think these are similar use cases no ? If you trust your inbound peer with a 0-conf, you very likely also trust them with a 0-reserve channel, and vice versa.

Hmm, I see what you mean, but there still is difference in timing. If I trust an LSP for 0conf channels, the trusted period ends when the first confirmation happens. And that might be a trade-off with convenience I might very well be willing to take. But, from an (LSP-client) users's perspective, why would I ever want to let the counterparty set 0reserves? I basically get nothing out of this trade maybe except some minimally higher inbound liquidity that I paid for.

> On the `main` branch this list already applies only to inbound channels.

True, due to the nature of 0conf being decided by the counterparty. Can we follow the same pattern, i.e., always be fine to set our reserve to 0 for outbound channels, and leave it up to the counterparty to decide what it accepts? I.e., why do we need dedicated new APIs for outbound 0reserve at all, if it's the counterparty   that decided what they'll accept in the end?

src/liquidity.rs
```diff
@@ -142,6 +142,10 @@ pub struct LSPS2ServiceConfig {
 	///
 	/// [`bLIP-52`]: https://github.com/lightning/blips/blob/master/blip-0052.md#trust-models
 	pub client_trusts_lsp: bool,
+	/// When set, clients that we open channels to will be allowed to spend their entire channel
+	/// balance. This allows clients to try to steal your funds with no financial penalty, so
+	/// this should only be set if you trust your clients.
+	pub allow_client_0reserve: bool,
```
benthecarman, line 148: this wording kinda suggests that all channels opened to these clients will be effected but in reality it is only the channels we open to them.
```suggestion
	/// When set, channels that we open to clients will be allowed to spend their entire
	/// balance. This allows clients to try to potentially steal your channel funds with no financial penalty, so
	/// this should only be set if you trust your clients.
	pub allow_client_0reserve: bool,
```

tnull: Yeah, might be good to slightly improve docs further in a follow-up.

tests/integration_tests_rust.rs
```diff
@@ -1705,6 +1705,7 @@ async fn do_lsps2_client_service_integration(client_trusts_lsp: bool) {
 		min_channel_opening_fee_msat: 0,
 		max_client_to_self_delay: 1024,
 		client_trusts_lsp,
+		allow_client_0reserve: false,
```
tnull, line 1708: We should def. add a test coverage for the new APIs in a follow-up (preferably a new 0-reserve specific test, maybe (also?) a full-cycle variant like we do for 0conf?).

tnull (commented):
I'm going ahead and landing this since other LDK-bumping PRs are waiting in line. However, might be good to at the very least add some test coverage in a follow-up (cc @tankyleo).

src/liquidity.rs
tankyleo: hmm don't quite see the difference between "clients that we open channels to" and "channels that we open to clients" ? In any case will revisit these in a follow-up

PR #854, opened 2026-03-30
Switch to async DNS resolution

tnull:
Previously, we used `std::net::ToSocketAddrs` for DNS resolution. While this generally worked, we also used the blocking `to_socket_addrs` call in async contexts, blocking a runtime task during resolution. This isn't great, so here we switch to use `tokio::net::lookup_host` for async resolution.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -6,7 +6,6 @@
 // accordance with one or both of these licenses.
 
 use std::collections::hash_map::{self, HashMap};
-use std::net::ToSocketAddrs;
 use std::ops::Deref;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
@@ -15,7 +14,7 @@ use bitcoin::secp256k1::PublicKey;
 use lightning::ln::msgs::SocketAddress;
 
 use crate::config::TorConfig;
-use crate::logger::{log_error, log_info, LdkLogger};
+use crate::logger::{log_debug, log_error, log_info, LdkLogger};
 use crate::types::{KeysManager, PeerManager};
 use crate::Error;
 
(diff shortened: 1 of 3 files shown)


src/connection.rs
```diff
@@ -97,39 +96,48 @@ where
...
+				let mut res = Err(Error::ConnectionFailed);
+				for proxy_addr in resolved_addrs {
+					let connection_future = lightning_net_tokio::tor_connect_outbound(
+						Arc::clone(&self.peer_manager),
+						node_id,
+						addr.clone(),
+						proxy_addr,
+						Arc::clone(&self.keys_manager),
+					);
+					res = self.await_connection(connection_future, node_id, addr.clone()).await;
```
tankyleo, line 132: If `res.is_err()` here, we don't know whether we failed to connect to the proxy, or we failed to connect to the peer. Do you think this is worth fixing on the `tor_connect_outbound` API ?

If we failed to connect to the peer, but nonetheless successfully connected to the proxy, we'd try connecting again using a new resolved address for the proxy, even though the previous one worked fine.

I can see how this is intended behavior, as we could make a successful connection to the peer upon trying again with a different proxy address.

src/connection.rs
```diff
@@ -97,39 +96,48 @@ where
...
+					if res.is_ok() {
```
tankyleo, line 133: In the `err` case, do we want to add a trace / debug log here to give a small heads up "hey we failed here" ? If we eventually succeed on a subsequent iteration, the transient error would go unnoticed.

src/connection.rs
```diff
@@ -139,17 +147,26 @@ where
...
+
+				let mut res = Err(Error::ConnectionFailed);
+				for socket_addr in resolved_addrs {
+					let connection_future = lightning_net_tokio::connect_outbound(
+						Arc::clone(&self.peer_manager),
+						node_id,
+						socket_addr,
+					);
+					res = self.await_connection(connection_future, node_id, addr.clone()).await;
+					if res.is_ok() {
```
tankyleo, line 182: Similar to above, do we want to be more verbose in the logs here ? ie "hey this IP address failed, but eventually succeeded with this other IP address".

src/connection.rs
tnull: Done.

src/connection.rs
tnull: Done.

src/connection.rs
tnull: > If res.is_err() here, we don't know whether we failed to connect to the proxy, or we failed to connect to the peer. Do you think this is worth fixing on the tor_connect_outbound API ?

Not sure it's entirely worth discerning? We generally might want to simply retry, and maybe it's even preferable to do the whole handshake from scratch, i.e., including using a new stream for example?

src/connection.rs
```diff
@@ -88,87 +94,109 @@ where
...
+								"Successfully connected to peer {}@{} via resolved proxy address {} after previous attempts failed.",
+								node_id, addr, proxy_addr
+							);
+						}
+						break;
+					}
+					had_failures = true;
+					log_debug!(
+						self.logger,
+						"Failed to connect to peer {}@{} via resolved proxy address {}, trying next address.",
```
tankyleo: nit: we won't be trying a next address if we are at the end of the iteration, here and below

src/connection.rs
tankyleo: agreed sounds good.

src/connection.rs
tnull: Okay, now dropped the "trying next address part". Not sure it's worth to further complicate the logic just to be precise regarding whether we try one more, especially given we'll be constantly retrying all connections anyways.

PR #860, opened 2026-03-31
Allow binding to port 0 for OS-assigned ports

joostjager:
Test port collision fix.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -541,6 +541,11 @@ impl NodeBuilder {
 			return Err(BuildError::InvalidListeningAddresses);
 		}
 
+		#[cfg(not(test))]
+		if listening_addresses.iter().any(crate::config::has_port_zero) {
+			return Err(BuildError::InvalidListeningAddresses);
+		}
+
 		self.config.listening_addresses = Some(listening_addresses);
 		Ok(self)
 	}
@@ -1944,6 +1949,7 @@ fn build_with_store_internal(
 
 	let (stop_sender, _) = tokio::sync::watch::channel(());
 	let (background_processor_stop_sender, _) = tokio::sync::watch::channel(());
+	let last_bound_addresses: Arc<RwLock<Option<Vec<SocketAddress>>>> = Arc::new(RwLock::new(None));
 	let is_running = Arc::new(RwLock::new(false));
 
 	let pathfinding_scores_sync_url = pathfinding_scores_sync_config.map(|c| c.url.clone());
@@ -1988,6 +1994,7 @@ fn build_with_store_internal(
 		peer_store,
 		payment_store,
 		lnurl_auth,
+		last_bound_addresses,
 		is_running,
 		node_metrics,
 		om_mailbox,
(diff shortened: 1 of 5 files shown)


src/builder.rs
```diff
@@ -1988,6 +1988,8 @@ fn build_with_store_internal(
 		peer_store,
 		payment_store,
 		lnurl_auth,
+		active_listening_addresses: Arc::new(RwLock::new(None)),
```
tnull: nit: Let's initialize the variables outside of the `Node {..}` construction please.

src/lib.rs
```diff
@@ -234,6 +234,11 @@ pub struct Node {
 	peer_store: Arc<PeerStore<Arc<Logger>>>,
 	payment_store: Arc<PaymentStore>,
 	lnurl_auth: Arc<LnurlAuth>,
+	/// Addresses the node is actively listening on. Set on `start()`, cleared on `stop()`.
+	active_listening_addresses: Arc<RwLock<Option<Vec<SocketAddress>>>>,
```
tnull: If we add all of that extra logic, etc, could we move these fields and the logic to `ConnectionManager` or at least `connection.rs`? Seems it would be the appropriate place?

src/lib.rs
```diff
@@ -525,6 +568,8 @@ impl Node {
 
 							let addresses = if let Some(announcement_addresses) = bcast_config.announcement_addresses.clone() {
 								announcement_addresses
+							} else if let Some(bound_addrs) = bcast_active_addrs.read().unwrap().clone() {
+								bound_addrs
```
tnull: No, we should def. never announce with any randomly-bound addresses. If the user announces a node, the address and port needs to be ~static, not something that will be lost on restart.

tnull (commented):
While I know about the flakiness issue, I can't say I'm loving all this additional complexity for very little actual (read: production) gain. Do we really need this?

src/builder.rs
joostjager: Moved to ConnectionManager

src/lib.rs
joostjager: Moved to `ConnectionManager`, but it doesn't seem completely logical.

src/lib.rs
joostjager: Fixed

src/lib.rs
joostjager: Moved it back. ConnectionManager didn't do anything with the field except for offering get and set.

tnull (commented):
Excuse the delay here. I'm still not the biggest fan of the added complexity in production logic for test-only fixes. Are we at least really positive that this will fix the flaky behavior we were seeing? 

Also needs a rebase by now.

joostjager:
We've had some time to observe main builds, and it seems the issue is hardly occurring anymore. Maybe the deterministic ports did fix the bulk of it. Broken pipe still happens now and then, but I think that's a different failure mode.

Will close the PR for now. We can revisit when this issue comes back.

tnull:
> Will close the PR for now. We can revisit when this issue comes back.

SGTM, thanks!

PR #862, opened 2026-04-01
Harden LNURL-auth request handling

benthecarman:
Enforce HTTPS for non-localhost URLs per LNURL spec and disable redirect following since the auth flow is a single GET request.

### DIFF
--- a/src/lnurl_auth.rs
+++ b/src/lnurl_auth.rs
@@ -96,6 +96,13 @@ impl LnurlAuth {
 
 		let domain = url.base_url();
 
+		// Enforce HTTPS for non-localhost URLs per LNURL spec.
+		let is_localhost = domain == "localhost" || domain == "127.0.0.1" || domain == "[::1]";
+		if url.scheme() != "https" && !is_localhost {
+			log_error!(self.logger, "LNURL-auth URL must use HTTPS for non-localhost domains");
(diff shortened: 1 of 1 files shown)


src/lnurl_auth.rs
```diff
@@ -96,6 +96,13 @@ impl LnurlAuth {
 
 		let domain = url.base_url();
 
+		// Enforce HTTPS for non-localhost URLs per LNURL spec.
+		let is_localhost = domain == "localhost" || domain == "127.0.0.1" || domain == "[::1]";
```
tnull, line 100: That's fine, but I do wonder if we should limit access to localhost/local network in general? Otherwise someone could prompt us to authenticate but have us make requests to arbitrary hosts, no? Maybe this method should take an expected hostname and abort if the decoded LNURL auth doesn't match that?

benthecarman: I feel local host and private networks can still be useful for testing.

I don't think the expected host name would really work. Most things you are just scanning qr and authenticating in. It's not really the end of the world if you auth to the "wrong" place. Your auth key is tied to the domain so the only real threat would be if someone dns hijacked you or something.

tnull: Not so much authenticating, but it allows to have us make requests against random endpoints. Anyways, moving on for now.

PR #863, opened 2026-04-01
Add Postgres database

benthecarman:
Add a PostgresStore implementation behind the `postgres` feature flag, mirroring the existing `SqliteStore`. Uses `tokio-postgres` (async-native) with an internal tokio runtime for the sync `KVStoreSync` trait, following the `VssStore` pattern.

Also gates sqlite behind a feature flag, since most tests relied on this, we ended up needing to gate a lot of them behind the flag. Curious on thoughts on how to handle this best

### DIFF
--- a/.github/workflows/postgres-integration.yml
+++ b/.github/workflows/postgres-integration.yml
@@ -1,0 +1,66 @@
+name: CI Checks - PostgreSQL Integration Tests
+
+on: [ push, pull_request ]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+
+    services:
+      postgres:
+        image: postgres:latest
+        ports:
+          - 5432:5432
+        env:
+          POSTGRES_DB: postgres
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+      - name: Install Rust stable toolchain
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --profile=minimal --default-toolchain stable
+      - name: Enable caching for bitcoind
+        id: cache-bitcoind
+        uses: actions/cache@v4
+        with:
+          path: bin/bitcoind-${{ runner.os }}-${{ runner.arch }}
+          key: bitcoind-27_2-${{ runner.os }}-${{ runner.arch }}
+      - name: Enable caching for electrs
+        id: cache-electrs
+        uses: actions/cache@v4
+        with:
+          path: bin/electrs-${{ runner.os }}-${{ runner.arch }}
+          key: electrs-esplora_a33e97e1-${{ runner.os }}-${{ runner.arch }}
+      - name: Download bitcoind/electrs
+        if: "steps.cache-bitcoind.outputs.cache-hit != 'true' || steps.cache-electrs.outputs.cache-hit != 'true'"
+        run: |
+          source ./scripts/download_bitcoind_electrs.sh
+          mkdir -p bin
+          mv "$BITCOIND_EXE" bin/bitcoind-${{ runner.os }}-${{ runner.arch }}
+          mv "$ELECTRS_EXE" bin/electrs-${{ runner.os }}-${{ runner.arch }}
+      - name: Set bitcoind/electrs environment variables
+        run: |
+          echo "BITCOIND_EXE=$( pwd )/bin/bitcoind-${{ runner.os }}-${{ runner.arch }}" >> "$GITHUB_ENV"
+          echo "ELECTRS_EXE=$( pwd )/bin/electrs-${{ runner.os }}-${{ runner.arch }}" >> "$GITHUB_ENV"
+      - name: Run PostgreSQL store tests
+        env:
+          TEST_POSTGRES_URL: "host=localhost user=postgres password=postgres"
+        run: cargo test --features postgres io::postgres_store
+      - name: Run PostgreSQL integration tests
+        env:
+          TEST_POSTGRES_URL: "host=localhost user=postgres password=postgres"
+        run: |
+          RUSTFLAGS="--cfg no_download --cfg cycle_tests" cargo test --features postgres --test integration_tests_postgres
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ panic = 'abort'     # Abort on panic
 
 [features]
 default = []
+postgres = ["dep:tokio-postgres", "dep:native-tls", "dep:postgres-native-tls"]
 
 [dependencies]
 #lightning = { version = "0.2.0", features = ["std"] }
@@ -78,6 +79,9 @@ serde_json = { version = "1.0.128", default-features = false, features = ["std"]
 log = { version = "0.4.22", default-features = false, features = ["std"]}
 
 async-trait = { version = "0.1", default-features = false }
+tokio-postgres = { version = "0.7", default-features = false, features = ["runtime"], optional = true }
+native-tls = { version = "0.2", default-features = false, optional = true }
+postgres-native-tls = { version = "0.5", default-features = false, features = ["runtime"], optional = true }
 vss-client = { package = "vss-client-ng", version = "0.5" }
 prost = { version = "0.11.6", default-features = false}
 #bitcoin-payment-instructions = { version = "0.6" }
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -644,6 +644,50 @@ impl NodeBuilder {
 		self.build_with_store_and_logger(node_entropy, kv_store, logger)
 	}
 
+	/// Builds a [`Node`] instance with a [PostgreSQL] backend and according to the options
+	/// previously configured.
+	///
+	/// Connects to the PostgreSQL database at the given `connection_string`, e.g.,
+	/// `"postgres://user:password@localhost/ldk_db"`.
+	///
+	/// The given `db_name` will be used or default to
+	/// [`DEFAULT_DB_NAME`](io::postgres_store::DEFAULT_DB_NAME). The `connection_string` must
+	/// not include a `dbname` when `db_name` is set, providing both is an error. The database
+	/// will be created automatically if it doesn't already exist. The initial connection is
+	/// made to the target database, and if it fails we fall back to the default `postgres`
+	/// database to create it.
+	///
+	/// The given `kv_table_name` will be used or default to
+	/// [`DEFAULT_KV_TABLE_NAME`](io::postgres_store::DEFAULT_KV_TABLE_NAME).
+	///
+	/// If `certificate_pem` is `Some`, TLS will be used for database connections and the
+	/// provided PEM-encoded CA certificate will be added to the system's default root
+	/// certificates (it does not replace them). If `certificate_pem` is `None`, connections
+	/// will be unencrypted.
+	///
+	/// [PostgreSQL]: https://www.postgresql.org
+	#[cfg(feature = "postgres")]
+	pub fn build_with_postgres_store(
+		&self, node_entropy: NodeEntropy, connection_string: String, db_name: Option<String>,
+		kv_table_name: Option<String>, certificate_pem: Option<String>,
+	) -> Result<Node, BuildError> {
+		let logger = setup_logger(&self.log_writer_config, &self.config)?;
+		let runtime = self.setup_runtime(&logger)?;
+		let kv_store = runtime
+			.block_on(io::postgres_store::PostgresStore::new_with_logger(
+				connection_string,
+				db_name,
+				kv_table_name,
+				certificate_pem,
+				Some(Arc::clone(&logger)),
+			))
+			.map_err(|e| {
+				log_error!(logger, "Failed to set up Postgres store: {e}");
+				BuildError::KVStoreSetupFailed
+			})?;
+		self.build_with_store_runtime_and_logger(node_entropy, kv_store, runtime, logger)
+	}
+
 	/// Builds a [`Node`] instance with a [`FilesystemStoreV2`] backend and according to the options
 	/// previously configured.
 	///
@@ -789,18 +833,27 @@ impl NodeBuilder {
 		self.build_with_store_and_logger(node_entropy, kv_store, logger)
 	}
 
+	fn setup_runtime(&self, logger: &Arc<Logger>) -> Result<Arc<Runtime>, BuildError> {
+		if let Some(handle) = self.runtime_handle.as_ref() {
+			Ok(Arc::new(Runtime::with_handle(handle.clone(), Arc::clone(logger))))
+		} else {
+			Ok(Arc::new(Runtime::new(Arc::clone(logger)).map_err(|e| {
+				log_error!(logger, "Failed to setup tokio runtime: {}", e);
+				BuildError::RuntimeSetupFailed
+			})?))
+		}
+	}
+
 	fn build_with_store_and_logger<S: SyncAndAsyncKVStore + Send + Sync + 'static>(
 		&self, node_entropy: NodeEntropy, kv_store: S, logger: Arc<Logger>,
 	) -> Result<Node, BuildError> {
-		let runtime = if let Some(handle) = self.runtime_handle.as_ref() {
-			Arc::new(Runtime::with_handle(handle.clone(), Arc::clone(&logger)))
-		} else {
-			Arc::new(Runtime::new(Arc::clone(&logger)).map_err(|e| {
-				log_error!(logger, "Failed to setup tokio runtime: {}", e);
-				BuildError::RuntimeSetupFailed
-			})?)
-		};
+		let runtime = self.setup_runtime(&logger)?;
+		self.build_with_store_runtime_and_logger(node_entropy, kv_store, runtime, logger)
+	}
 
+	fn build_with_store_runtime_and_logger<S: SyncAndAsyncKVStore + Send + Sync + 'static>(
+		&self, node_entropy: NodeEntropy, kv_store: S, runtime: Arc<Runtime>, logger: Arc<Logger>,
+	) -> Result<Node, BuildError> {
 		let seed_bytes = node_entropy.to_seed_bytes();
 		let config = Arc::new(self.config.clone());
 
@@ -1114,6 +1167,58 @@ impl ArcedNodeBuilder {
 	/// previously configured.
 	pub fn build(&self, node_entropy: Arc<NodeEntropy>) -> Result<Arc<Node>, BuildError> {
 		self.inner.read().expect("lock").build(*node_entropy).map(Arc::new)
+	}
+
+	/// Builds a [`Node`] instance with a [PostgreSQL] backend and according to the options
+	/// previously configured.
+	///
+	/// Connects to the PostgreSQL database at the given `connection_string`, e.g.,
+	/// `"postgres://user:password@localhost/ldk_db"`.
+	///
+	/// The given `db_name` will be used or default to
+	/// [`DEFAULT_DB_NAME`](io::postgres_store::DEFAULT_DB_NAME). The `connection_string` must
+	/// not include a `dbname` when `db_name` is set, providing both is an error. The database
+	/// will be created automatically if it doesn't already exist. The initial connection is
+	/// made to the target database, and if it fails we fall back to the default `postgres`
+	/// database to create it.
+	///
+	/// The given `kv_table_name` will be used or default to
+	/// [`DEFAULT_KV_TABLE_NAME`](io::postgres_store::DEFAULT_KV_TABLE_NAME).
+	///
+	/// If `certificate_pem` is `Some`, TLS will be used for database connections and the
+	/// provided PEM-encoded CA certificate will be added to the system's default root
+	/// certificates (it does not replace them). If `certificate_pem` is `None`, connections
+	/// will be unencrypted.
+	///
+	/// [PostgreSQL]: https://www.postgresql.org
+	#[cfg(feature = "postgres")]
+	pub fn build_with_postgres_store(
+		&self, node_entropy: Arc<NodeEntropy>, connection_string: String, db_name: Option<String>,
+		kv_table_name: Option<String>, certificate_pem: Option<String>,
+	) -> Result<Arc<Node>, BuildError> {
+		self.inner
+			.read()
+			.unwrap()
+			.build_with_postgres_store(
+				*node_entropy,
+				connection_string,
+				db_name,
+				kv_table_name,
+				certificate_pem,
+			)
+			.map(Arc::new)
+	}
+
+	/// Builds a [`Node`] instance with a [PostgreSQL] backend and according to the options
+	/// previously configured.
+	///
+	/// This requires the `postgres` crate feature.
+	#[cfg(not(feature = "postgres"))]
+	pub fn build_with_postgres_store(
+		&self, _node_entropy: Arc<NodeEntropy>, _connection_string: String,
+		_db_name: Option<String>, _kv_table_name: Option<String>, _certificate_pem: Option<String>,
+	) -> Result<Arc<Node>, BuildError> {
+		Err(BuildError::KVStoreSetupFailed)
 	}
 
 	/// Builds a [`Node`] instance with a [`FilesystemStoreV2`] backend and according to the options
--- a/src/io/mod.rs
+++ b/src/io/mod.rs
@@ -7,6 +7,8 @@
 
 //! Objects and traits for data persistence.
 
+#[cfg(feature = "postgres")]
+pub mod postgres_store;
 pub mod sqlite_store;
 #[cfg(test)]
 pub(crate) mod test_utils;
--- a/src/io/postgres_store/mod.rs
+++ b/src/io/postgres_store/mod.rs
@@ -1,0 +1,1227 @@
+// This file is Copyright its original authors, visible in version control history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
+// accordance with one or both of these licenses.
+
+//! Objects related to [`PostgresStore`] live here.
+use std::collections::HashMap;
+use std::future::Future;
+use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
+use std::sync::{Arc, Mutex};
+
+use lightning::io;
+use lightning::util::persist::{
+	KVStore, KVStoreSync, PageToken, PaginatedKVStore, PaginatedKVStoreSync, PaginatedListResponse,
+};
+use lightning_types::string::PrintableString;
+use native_tls::TlsConnector;
+use postgres_native_tls::MakeTlsConnector;
+use tokio_postgres::config::SslMode;
+use tokio_postgres::{Config, Error as PgError};
+
+use self::pool::{make_config_connection, ClientConnection, PgTlsConnector, SmallPool};
+use crate::io::utils::check_namespace_key_validity;
+use crate::logger::{log_debug, log_info, LdkLogger, Logger};
+
+mod migrations;
+mod pool;
+
+/// The default database name used when none is specified.
+pub const DEFAULT_DB_NAME: &str = "ldk_db";
+
+/// The default table in which we store all data.
+pub const DEFAULT_KV_TABLE_NAME: &str = "ldk_data";
+
+// The current schema version for the PostgreSQL store.
+const SCHEMA_VERSION: u16 = 1;
+
+// The number of entries returned per page in paginated list operations.
+const PAGE_SIZE: usize = 50;
+
+// Keep this small while still allowing progress if one runtime worker blocks on sync store access.
+const INTERNAL_RUNTIME_WORKERS: usize = 2;
+
+fn sql_identifier(identifier: &str) -> io::Result<String> {
+	if identifier.is_empty() || identifier.contains('\0') {
+		return Err(io::Error::new(
+			io::ErrorKind::InvalidInput,
+			format!("Invalid PostgreSQL identifier: {identifier}"),
+		));
+	}
+
+	Ok(format!("\"{}\"", identifier.replace('"', "\"\"")))
+}
+
+fn sql_table_identifier(table_name: &str) -> io::Result<String> {
+	let parts: Vec<&str> = table_name.split('.').collect();
+	if parts.is_empty() || parts.len() > 2 || parts.iter().any(|part| part.is_empty()) {
+		return Err(io::Error::new(
+			io::ErrorKind::InvalidInput,
+			format!(
+				"Invalid PostgreSQL table name: {table_name}. Expected a table name or schema.table"
+			),
+		));
+	}
+
+	let quoted_parts: io::Result<Vec<String>> = parts.into_iter().map(sql_identifier).collect();
+	Ok(quoted_parts?.join("."))
+}
+
+/// Runs a tokio-postgres query and, if the connection dropped mid-flight, reconnects and retries
+/// once. `$store` is the [`PostgresStoreInner`], `$locked` the held client slot guard,
+/// `$err_map` an `Fn(PgError) -> io::Error` (called at most once), and `$query` an expression
+/// that yields a fresh `Future<Output = Result<_, PgError>>` each time it's evaluated. `$query`
+/// may be evaluated up to twice (once normally, once on retry), so it must be side-effect-free
+/// outside of issuing the query itself.
+macro_rules! query_with_retry {
+	($store:expr, $locked:ident, $err_map:expr, $query:expr) => {{
+		match $query.await {
+			Ok(v) => Ok(v),
+			Err(e) if $locked.is_closed() || e.is_closed() => {
+				if let Some(logger) = $store.logger.as_ref() {
+					log_debug!(logger, "Reconnecting to PostgreSQL after error: {e}");
+				}
+				*$locked = make_config_connection(&$store.config, &$store.tls).await?;
+				$query.await.map_err($err_map)
+			},
+			Err(e) => Err($err_map(e)),
+		}
+	}};
+}
+
+/// A [`KVStoreSync`] implementation that writes to and reads from a [PostgreSQL] database.
+///
+/// Maintains an internal runtime for the underlying tokio-postgres connection drivers and for
+/// synchronous [`KVStoreSync`] and [`PaginatedKVStoreSync`] calls.
+///
+/// [PostgreSQL]: https://www.postgresql.org
+pub struct PostgresStore {
+	inner: Arc<PostgresStoreInner>,
+
+	// Version counter to ensure that writes are applied in the correct order. It is assumed that read and list
+	// operations aren't sensitive to the order of execution.
+	next_write_version: AtomicU64,
+
+	// A store-internal runtime used for setup, connection driver tasks, and sync store access.
+	internal_runtime: Option<tokio::runtime::Runtime>,
+}
+
+// tokio::sync::Mutex (used for the DB client) contains UnsafeCell which opts out of
+// RefUnwindSafe. std::sync::Mutex (used by SqliteStore) doesn't have this issue because
+// it poisons on panic. This impl is needed for do_read_write_remove_list_persist which
+// requires K: KVStoreSync + RefUnwindSafe.
+#[cfg(test)]
+impl std::panic::RefUnwindSafe for PostgresStore {}
+
+impl PostgresStore {
+	/// Constructs a new [`PostgresStore`].
+	///
+	/// Connects to the PostgreSQL database at the given `connection_string`, e.g.,
+	/// `"postgres://user:password@localhost/ldk_db"`.
+	///
+	/// The given `db_name` will be used or default to [`DEFAULT_DB_NAME`]. The
+	/// `connection_string` must not include a `dbname` when `db_name` is set, providing both
+	/// is an error. The database will be created automatically if it doesn't already exist.
+	/// The initial connection is made to the target database, and if it fails we fall back to
+	/// the default `postgres` database to create it.
+	///
+	/// The given `kv_table_name` will be used or default to [`DEFAULT_KV_TABLE_NAME`].
+	///
+	/// If `certificate_pem` is `Some`, TLS will be used for database connections and the
+	/// provided PEM-encoded CA certificate will be added to the system's default root
+	/// certificates (it does not replace them). If `certificate_pem` is `None`, connections
+	/// will be unencrypted.
+	pub async fn new(
+		connection_string: String, db_name: Option<String>, kv_table_name: Option<String>,
+		certificate_pem: Option<String>,
+	) -> io::Result<Self> {
+		Self::new_with_logger(connection_string, db_name, kv_table_name, certificate_pem, None)
+			.await
+	}
+
+	/// Like [`Self::new`], but lets crate-internal callers route setup logs through the node
+	/// logger. This stays separate because [`Logger`] is crate-private and cannot be part of the
+	/// public [`PostgresStore::new`] signature.
+	pub(crate) async fn new_with_logger(
+		connection_string: String, db_name: Option<String>, kv_table_name: Option<String>,
+		certificate_pem: Option<String>, logger: Option<Arc<Logger>>,
+	) -> io::Result<Self> {
+		let internal_runtime = tokio::runtime::Builder::new_multi_thread()
+			.enable_all()
+			.thread_name_fn(|| {
+				static ATOMIC_ID: AtomicUsize = AtomicUsize::new(0);
+				let id = ATOMIC_ID.fetch_add(1, Ordering::SeqCst);
+				format!("ldk-node-postgres-runtime-{}", id)
+			})
+			.worker_threads(INTERNAL_RUNTIME_WORKERS)
+			.max_blocking_threads(INTERNAL_RUNTIME_WORKERS)
+			.build()
+			.map_err(|e| {
+				io::Error::new(
+					io::ErrorKind::Other,
+					format!("Failed to build PostgreSQL runtime: {e}"),
+				)
+			})?;
+		let tls = Self::build_tls_connector(certificate_pem)?;
+		let runtime_handle = internal_runtime.handle();
+		let inner = tokio::task::block_in_place(|| {
+			runtime_handle.block_on(async {
+				PostgresStoreInner::new(connection_string, db_name, kv_table_name, tls, logger)
+					.await
+			})
+		})?;
+		let inner = Arc::new(inner);
+		let next_write_version = AtomicU64::new(1);
+		Ok(Self { inner, next_write_version, internal_runtime: Some(internal_runtime) })
+	}
+
+	fn build_tls_connector(certificate_pem: Option<String>) -> io::Result<PgTlsConnector> {
+		match certificate_pem {
+			Some(pem) => {
+				let crt = native_tls::Certificate::from_pem(pem.as_bytes()).map_err(|e| {
+					io::Error::new(
+						io::ErrorKind::InvalidInput,
+						format!("Failed to parse PEM certificate: {e}"),
+					)
+				})?;
+				let connector =
+					TlsConnector::builder().add_root_certificate(crt).build().map_err(|e| {
+						io::Error::new(
+							io::ErrorKind::Other,
+							format!("Failed to build TLS connector: {e}"),
+						)
+					})?;
+				Ok(PgTlsConnector::NativeTls(MakeTlsConnector::new(connector)))
+			},
+			None => Ok(PgTlsConnector::Plain),
+		}
+	}
+
+	fn build_locking_key(
+		&self, primary_namespace: &str, secondary_namespace: &str, key: &str,
+	) -> String {
+		format!("{primary_namespace}#{secondary_namespace}#{key}")
+	}
+
+	fn get_new_version_and_lock_ref(
+		&self, locking_key: String,
+	) -> (Arc<tokio::sync::Mutex<u64>>, u64) {
+		let version = self.next_write_version.fetch_add(1, Ordering::Relaxed);
+		if version == u64::MAX {
+			panic!("PostgresStore version counter overflowed");
+		}
+
+		let inner_lock_ref = self.inner.get_inner_lock_ref(locking_key);
+
+		(inner_lock_ref, version)
+	}
+}
+
+impl Drop for PostgresStore {
+	fn drop(&mut self) {
+		if let Some(internal_runtime) = self.internal_runtime.take() {
+			internal_runtime.shutdown_background();
+		}
+	}
+}
+
+impl KVStore for PostgresStore {
+	fn read(
+		&self, primary_namespace: &str, secondary_namespace: &str, key: &str,
+	) -> impl Future<Output = Result<Vec<u8>, io::Error>> + 'static + Send {
+		let primary_namespace = primary_namespace.to_string();
+		let secondary_namespace = secondary_namespace.to_string();
+		let key = key.to_string();
+		let inner = Arc::clone(&self.inner);
+		async move { inner.read_internal(&primary_namespace, &secondary_namespace, &key).await }
+	}
+
+	fn write(
+		&self, primary_namespace: &str, secondary_namespace: &str, key: &str, buf: Vec<u8>,
+	) -> impl Future<Output = Result<(), io::Error>> + 'static + Send {
+		let locking_key = self.build_locking_key(primary_namespace, secondary_namespace, key);
+		let (inner_lock_ref, version) = self.get_new_version_and_lock_ref(locking_key.clone());
+		let primary_namespace = primary_namespace.to_string();
+		let secondary_namespace = secondary_namespace.to_string();
+		let key = key.to_string();
+		let inner = Arc::clone(&self.inner);
+		async move {
+			inner
+				.write_internal(
+					inner_lock_ref,
+					locking_key,
+					version,
+					&primary_namespace,
+					&secondary_namespace,
+					&key,
+					buf,
+				)
+				.await
+		}
+	}
+
+	fn remove(
+		&self, primary_namespace: &str, secondary_namespace: &str, key: &str, _lazy: bool,
+	) -> impl Future<Output = Result<(), io::Error>> + 'static + Send {
+		let locking_key = self.build_locking_key(primary_namespace, secondary_namespace, key);
+		let (inner_lock_ref, version) = self.get_new_version_and_lock_ref(locking_key.clone());
+		let primary_namespace = primary_namespace.to_string();
+		let secondary_namespace = secondary_namespace.to_string();
+		let key = key.to_string();
+		let inner = Arc::clone(&self.inner);
+		async move {
+			inner
+				.remove_internal(
+					inner_lock_ref,
+					locking_key,
+					version,
+					&primary_namespace,
+					&secondary_namespace,
+					&key,
+				)
+				.await
+		}
+	}
+
+	fn list(
+		&self, primary_namespace: &str, secondary_namespace: &str,
+	) -> impl Future<Output = Result<Vec<String>, io::Error>> + 'static + Send {
+		let primary_namespace = primary_namespace.to_string();
+		let secondary_namespace = secondary_namespace.to_string();
+		let inner = Arc::clone(&self.inner);
+		async move { inner.list_internal(&primary_namespace, &secondary_namespace).await }
+	}
+}
+
+impl PostgresStore {
+	fn internal_runtime(&self) -> io::Result<&tokio::runtime::Runtime> {
+		self.internal_runtime.as_ref().ok_or_else(|| {
+			debug_assert!(false, "Failed to access internal PostgreSQL runtime");
+			io::Error::new(io::ErrorKind::Other, "Failed to access internal PostgreSQL runtime")
+		})
+	}
+
+	fn block_on<F: Future>(&self, fut: F) -> io::Result<F::Output> {
+		let internal_runtime = self.internal_runtime()?;
+		Ok(tokio::task::block_in_place(move || internal_runtime.block_on(fut)))
+	}
+}
+
+impl KVStoreSync for PostgresStore {
+	fn read(
+		&self, primary_namespace: &str, secondary_namespace: &str, key: &str,
+	) -> io::Result<Vec<u8>> {
+		self.block_on(KVStore::read(self, primary_namespace, secondary_namespace, key))?
+	}
+
+	fn write(
+		&self, primary_namespace: &str, secondary_namespace: &str, key: &str, buf: Vec<u8>,
+	) -> io::Result<()> {
+		self.block_on(KVStore::write(self, primary_namespace, secondary_namespace, key, buf))?
+	}
+
+	fn remove(
+		&self, primary_namespace: &str, secondary_namespace: &str, key: &str, lazy: bool,
+	) -> io::Result<()> {
+		self.block_on(KVStore::remove(self, primary_namespace, secondary_namespace, key, lazy))?
+	}
+
+	fn list(&self, primary_namespace: &str, secondary_namespace: &str) -> io::Result<Vec<String>> {
+		self.block_on(KVStore::list(self, primary_namespace, secondary_namespace))?
+	}
+}
+
+impl PaginatedKVStoreSync for PostgresStore {
+	fn list_paginated(
+		&self, primary_namespace: &str, secondary_namespace: &str, page_token: Option<PageToken>,
+	) -> io::Result<PaginatedListResponse> {
+		self.block_on(PaginatedKVStore::list_paginated(
+			self,
+			primary_namespace,
+			secondary_namespace,
+			page_token,
+		))?
+	}
+}
+
+impl PaginatedKVStore for PostgresStore {
+	fn list_paginated(
+		&self, primary_namespace: &str, secondary_namespace: &str, page_token: Option<PageToken>,
+	) -> impl Future<Output = Result<PaginatedListResponse, io::Error>> + 'static + Send {
+		let primary_namespace = primary_namespace.to_string();
+		let secondary_namespace = secondary_namespace.to_string();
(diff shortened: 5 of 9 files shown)


.github/workflows/postgres-integration.yml
```diff
@@ -0,0 +1,42 @@
...
+          POSTGRES_PASSWORD: postgres
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
```
tnull: I think there are newer action versions available? Cf #861.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,986 @@
...
+pub struct PostgresStore {
+	inner: Arc<PostgresStoreInner>,
+
+	// Version counter to ensure that writes are applied in the correct order. It is assumed that read and list
+	// operations aren't sensitive to the order of execution.
+	next_write_version: AtomicU64,
+
+	// An internal runtime we use to avoid any deadlocks we could hit when waiting on async
+	// operations to finish from a sync context.
+	internal_runtime: Option<tokio::runtime::Runtime>,
```
tnull: I think we could avoid using an internal runtime for this and rather just reuse `crate::runtime::Runtime`, if we wanted. Though, when upstreaming to LDK we'll need to rethink all of this anyways (cf. VssStore upstreaming PR).

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,986 @@
...
+}
+
+// tokio::sync::Mutex (used for the DB client) contains UnsafeCell which opts out of
+// RefUnwindSafe. std::sync::Mutex (used by SqliteStore) doesn't have this issue because
+// it poisons on panic. This impl is needed for do_read_write_remove_list_persist which
```
tnull, line 113: Okay, but then we should probably `cfg(test)` this, right?

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,986 @@
...
+				static ATOMIC_ID: AtomicUsize = AtomicUsize::new(0);
+				let id = ATOMIC_ID.fetch_add(1, Ordering::SeqCst);
+				format!("ldk-node-pg-runtime-{id}")
+			})
+			.worker_threads(INTERNAL_RUNTIME_WORKERS)
+			.max_blocking_threads(INTERNAL_RUNTIME_WORKERS)
+			.build()
+			.unwrap();
+
+		let connection_string = connection_string.to_string();
```
tnull: Why not take a `String` if we'd allocate anyways? It would at least avoid a re-allocation if users held a `String` already.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,986 @@
...
+	write_version_locks: Mutex<HashMap<String, Arc<tokio::sync::Mutex<u64>>>>,
+	next_sort_order: AtomicI64,
+}
+
+impl PostgresStoreInner {
+	async fn new(connection_string: &str, kv_table_name: Option<String>) -> io::Result<Self> {
+		let kv_table_name = kv_table_name.unwrap_or(DEFAULT_KV_TABLE_NAME.to_string());
+
+		let (client, connection) =
+			tokio_postgres::connect(connection_string, NoTls).await.map_err(|e| {
```
tnull: We probably should support TLS from the getgo?

Also, we'll want users to pick non-default databases and create them if they don't exist yet. Note for this to work you'll first need to create a connection with the default, to then create the database you'll have in mind (see https://github.com/lightningdevkit/vss-server/pull/55 and follow-ups for reference).

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,986 @@
...
+				let msg = format!("Failed to connect to PostgreSQL: {e}");
+				io::Error::new(io::ErrorKind::Other, msg)
+			})?;
+
+		// Spawn the connection task so it runs in the background.
+		tokio::spawn(async move {
+			if let Err(e) = connection.await {
```
tnull: How will we recover from connection errors? Should this somehow loop or similar?

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,986 @@
...
+		}
+	}
+
+	async fn write_internal(
+		&self, inner_lock_ref: Arc<tokio::sync::Mutex<u64>>, locking_key: String, version: u64,
+		primary_namespace: &str, secondary_namespace: &str, key: &str, buf: Vec<u8>,
+	) -> io::Result<()> {
+		check_namespace_key_validity(primary_namespace, secondary_namespace, Some(key), "write")?;
+
+		self.execute_locked_write(inner_lock_ref, locking_key, version, async move || {
```
tnull, line 642: Super annoying that we have that replicated three times (at least) by now. We should look to DRY this up, when upstreaming to `lightning-persister` at the very latest.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,986 @@
...
+		let mut outer_lock = self.write_version_locks.lock().unwrap();
+		Arc::clone(&outer_lock.entry(locking_key).or_default())
+	}
+
+	async fn read_internal(
+		&self, primary_namespace: &str, secondary_namespace: &str, key: &str,
+	) -> io::Result<Vec<u8>> {
+		check_namespace_key_validity(primary_namespace, secondary_namespace, Some(key), "read")?;
+
+		let locked_client = self.client.lock().await;
```
tnull: If all operations take a `MutexGuard` and `await` the operation anyways, do we really need the write-lock-versioning? Isn't this practically sync?

src/builder.rs
```diff
@@ -1083,10 +1105,27 @@ impl ArcedNodeBuilder {
 
 	/// Builds a [`Node`] instance with a [`SqliteStore`] backend and according to the options
 	/// previously configured.
+	#[cfg(feature = "sqlite")]
```
tnull: Why do we suddenly introduce features? We so far held off and meant to do that in a dedicated PR at some point for this or next release. Is it necessary to make Postgres work?

tnull (commented):
Took a first high-level look.

> Also gates sqlite behind a feature flag, since most tests relied on this, we ended up needing to gate a lot of them behind the flag. Curious on thoughts on how to handle this best

Not sure I'm following? Why do we want to feature-gate tests based on SQLite?

Also tagging @tankyleo as a secondary reviewer here as he has a lot of recent experience with Postgres on VSS Server.

src/builder.rs
benthecarman: okay removed for now

src/io/postgres_store/mod.rs
benthecarman: Technically we'd still need this if someone is gonna do a `join!` on two writes to the same key. But yeah, could probably be okay to remove?

src/io/postgres_store/mod.rs
benthecarman: Yeah I had same thought.

src/io/postgres_store/mod.rs
benthecarman: fixed

src/io/postgres_store/mod.rs
benthecarman: added both

src/io/postgres_store/mod.rs
benthecarman: Yeah I just copied VSS to be safe.

src/builder.rs
```diff
@@ -629,6 +629,40 @@ impl NodeBuilder {
...
+	/// certificate can be provided via
+	/// [`PostgresTlsConfig::certificate_pem`](io::postgres_store::PostgresTlsConfig::certificate_pem),
+	/// otherwise the system's default root certificates are used. If `tls_config` is `None`,
+	/// connections will be unencrypted.
+	///
+	/// [PostgreSQL]: https://www.postgresql.org
+	#[cfg(feature = "postgres")]
+	pub fn build_with_postgres_store(
+		&self, node_entropy: NodeEntropy, connection_string: String, db_name: Option<String>,
+		kv_table_name: Option<String>, tls_config: Option<io::postgres_store::PostgresTlsConfig>,
```
tankyleo: In case using a full `PostgresTlsConfig` is too onerous, we use an `Option<Option<String>>` for internal APIs in VSS. Outer option determines Tls vs Plain and inner option gives the option to add a certificate. The full `PostgresTlsConfig` struct is just there to specify how the toml file should look. Your call.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1184 @@
...
+			outer_lock.remove(&locking_key);
+		}
+	}
+}
+
+/// TLS configuration for PostgreSQL connections.
+#[derive(Debug, Clone)]
+pub struct PostgresTlsConfig {
+	/// PEM-encoded CA certificate. If `None`, the system's default root certificates are used.
+	pub certificate_pem: Option<String>,
```
tankyleo: Worth clarifying if the user sets a certificate here, it just adds to the system's root certificates, it does not replace them.

src/builder.rs
```diff
@@ -629,6 +629,40 @@ impl NodeBuilder {
 		self.build_with_store(node_entropy, kv_store)
 	}
 
+	/// Builds a [`Node`] instance with a [PostgreSQL] backend and according to the options
+	/// previously configured.
+	///
+	/// Connects to the PostgreSQL database at the given `connection_string`.
```
tankyleo: Let's give an example here on what connection string we expect. Should it include the database name ?

src/builder.rs
```diff
@@ -629,6 +629,40 @@ impl NodeBuilder {
...
+	/// If `tls_config` is `Some`, TLS will be used for database connections. A custom CA
+	/// certificate can be provided via
+	/// [`PostgresTlsConfig::certificate_pem`](io::postgres_store::PostgresTlsConfig::certificate_pem),
+	/// otherwise the system's default root certificates are used. If `tls_config` is `None`,
+	/// connections will be unencrypted.
+	///
+	/// [PostgreSQL]: https://www.postgresql.org
+	#[cfg(feature = "postgres")]
+	pub fn build_with_postgres_store(
+		&self, node_entropy: NodeEntropy, connection_string: String, db_name: Option<String>,
```
tankyleo, line 671: I'd prefer if `connection_string` and `db_name` did not potentially step on each other's toes ? Otherwise clarify if `db_name` is set, `connection_string` should contain no database.

Another option potentially is to have the `connection_string` point to the default database, see comment below.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1184 @@
...
+		let write_version_locks = Mutex::new(HashMap::new());
+		Ok(Self { client, config, kv_table_name, tls, write_version_locks, next_sort_order })
+	}
+
+	async fn create_database_if_not_exists(
+		config: &Config, db_name: &str, tls: &PgTlsConnector,
+	) -> io::Result<()> {
+		// Connect without a dbname (to the default database) so we can create the target.
+		let mut config = config.clone();
+		config.dbname("postgres");
```
tankyleo: We've had a few complaints, we can't really make the assumption that the default database is called "postgres". Should we ask for the default database in the connection string ?

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1184 @@
...
+		tls: PgTlsConnector,
+	) -> io::Result<Self> {
+		let kv_table_name = kv_table_name.unwrap_or(DEFAULT_KV_TABLE_NAME.to_string());
+
+		let mut config: Config = connection_string.parse().map_err(|e: PgError| {
+			let msg = format!("Failed to parse PostgreSQL connection string: {e}");
+			io::Error::new(io::ErrorKind::InvalidInput, msg)
+		})?;
+
+		if db_name.is_some() && config.get_dbname().is_some() {
```
tankyleo, line 386: Sounds like this could be hit quite easily. Clarify the docs and/or ask connection string to point to default database ?

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1184 @@
...
+	/// created automatically if it doesn't already exist.
+	///
+	/// The given `kv_table_name` will be used or default to [`DEFAULT_KV_TABLE_NAME`].
+	///
+	/// If `tls_config` is `Some`, TLS will be used for database connections. A custom CA
+	/// certificate can be provided via [`PostgresTlsConfig::certificate_pem`], otherwise the
+	/// system's default root certificates are used. If `tls_config` is `None`, connections
+	/// will be unencrypted.
+	pub fn new(
+		connection_string: String, db_name: Option<String>, kv_table_name: Option<String>,
```
tankyleo, line 137: VSS hardcodes the table name, is it worth the extra parameter ?

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1184 @@
...
+				io::ErrorKind::InvalidInput,
+				"db_name must not be set when the connection string already contains a dbname",
+			));
+		}
+
+		let db_name = db_name
+			.or_else(|| config.get_dbname().map(|s| s.to_string()))
+			.unwrap_or(DEFAULT_DB_NAME.to_string());
+		config.dbname(&db_name);
+		Self::create_database_if_not_exists(&config, &db_name, &tls).await?;
```
tankyleo: Found it a little awkward to pass `db_name` both in `config` and `db_name` here

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1184 @@
...
+		let client = Self::make_config_connection(&config, &tls).await?;
+
+		// Create the KV data table if it doesn't exist.
+		let sql = format!(
+			"CREATE TABLE IF NOT EXISTS {kv_table_name} (
+			primary_namespace TEXT NOT NULL,
+			secondary_namespace TEXT NOT NULL DEFAULT '',
+			key TEXT NOT NULL CHECK (key <> ''),
+			value BYTEA,
+			sort_order BIGINT NOT NULL DEFAULT 0,
```
tankyleo: Let's add a ">= 0" constraint here ? We really don't expect any negative values here.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1184 @@
...
+		})?;
+
+		// Read the schema version from the table comment (analogous to SQLite's PRAGMA user_version).
+		let sql = format!("SELECT obj_description('{kv_table_name}'::regclass, 'pg_class')");
+		let row = client.query_one(sql.as_str(), &[]).await.map_err(|e| {
+			let msg = format!("Failed to read schema version for {kv_table_name}: {e}");
+			io::Error::new(io::ErrorKind::Other, msg)
+		})?;
+		let version_res: u16 = match row.get::<_, Option<&str>>(0) {
+			Some(version_str) => version_str.parse().map_err(|_| {
```
tankyleo: Should we complain if we parse a `0` here ? Something is quite wrong if we do.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1184 @@
...
+				let msg = format!("Invalid schema version: {version_str}");
+				io::Error::new(io::ErrorKind::Other, msg)
+			})?,
+			None => 0,
+		};
+
+		if version_res == 0 {
```
tankyleo, line 468: Could do `version_res == None` here, then `is_some_and(|v| v < SCHEMA_VERSION)` below, instead of mapping `None` to 0 above.

src/io/postgres_store/mod.rs
tankyleo: Did you consider using timestamps for ordering ? At least on VSS they allow VSS instances against the database to scale horizontally as we don't need to keep any state in VSS.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1184 @@
...
+			io::Error::new(io::ErrorKind::Other, msg)
+		})?;
+
+		// Initialize next_sort_order from the max existing value.
+		let sql = format!("SELECT COALESCE(MAX(sort_order), 0) FROM {kv_table_name}");
+		let row = client.query_one(sql.as_str(), &[]).await.map_err(|e| {
+			let msg = format!("Failed to read max sort_order from {kv_table_name}: {e}");
+			io::Error::new(io::ErrorKind::Other, msg)
+		})?;
+		let max_sort_order: i64 = row.get(0);
```
tankyleo: An option is to assert this max_sort_order here is not negative.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1184 @@
...
+		self.execute_locked_write(inner_lock_ref, locking_key, version, async move || {
+			let mut locked_client = self.client.lock().await;
+			self.ensure_connected(&mut locked_client).await?;
+
+			let sort_order = self.next_sort_order.fetch_add(1, Ordering::Relaxed);
+
+			let sql = format!(
+				"INSERT INTO {} (primary_namespace, secondary_namespace, key, value, sort_order) \
+				 VALUES ($1, $2, $3, $4, $5) \
+				 ON CONFLICT (primary_namespace, secondary_namespace, key) DO UPDATE SET value = EXCLUDED.value",
```
tankyleo, line 646: `sort_order` should *not* be updated here on conflict because it tracks creation order. Let's rename `sort_order` to `creation_order` across this file ?

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1184 @@
...
+
+		let mut entries: Vec<(String, i64)> = match page_token {
+			Some(ref token) => {
+				let token_sort_order: i64 = token.as_str().parse().map_err(|_| {
+					let token_str = token.as_str();
+					let msg = format!("Invalid page token: {token_str}");
+					io::Error::new(io::ErrorKind::InvalidInput, msg)
+				})?;
+				let sql = format!(
+					"SELECT key, sort_order FROM {} \
```
tankyleo, line 782: Any cool SQL tricks available to grab the `sort_order` just from the last row, instead of grabbing all of them ?

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1184 @@
...
+			"list_paginated",
+		)?;
+
+		let mut locked_client = self.client.lock().await;
+		self.ensure_connected(&mut locked_client).await?;
+
+		// Fetch one extra row beyond PAGE_SIZE to determine whether a next page exists.
+		let fetch_limit = (PAGE_SIZE + 1) as i64;
+
+		let mut entries: Vec<(String, i64)> = match page_token {
```
tankyleo: I think we can delete this temporary vector here, and allocate from the rows directly to the `Vec<String>` vector, and grab the `last_sort_order` from the rows vector, not from this temporary vector.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1184 @@
...
+				})
+			});
+		}
+
+		// Read should auto-reconnect and return the previously written value.
+		let data = KVStoreSync::read(&store, ns, sub, "key_a").unwrap();
+		assert_eq!(data, vec![1u8; 8]);
+
+		// Write should also work after reconnect.
+		KVStoreSync::write(&store, ns, sub, "key_b", vec![2u8; 8]).unwrap();
```
tankyleo, line 943: Can we also try a write immediately after reconnection, with no reads preceding it ?

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1184 @@
...
+use native_tls::TlsConnector;
+use postgres_native_tls::MakeTlsConnector;
+use tokio_postgres::{Client, Config, Error as PgError, NoTls};
+
+use crate::io::utils::check_namespace_key_validity;
+
+mod migrations;
+
+/// The default database name used when none is specified.
+pub const DEFAULT_DB_NAME: &str = "ldk_node";
```
tankyleo: In ldk-server we'll override this to `ldk_server` ? How about just `ldk_db` so that we get a name that works well on both ldk-node and server ?

tankyleo (commented):
Verbose review, please dismiss aggressively :)

src/io/postgres_store/mod.rs
benthecarman: we ended up not using time for the sqlite db because of this:

https://github.com/lightningdevkit/ldk-node/pull/807#pullrequestreview-4007146376

src/io/postgres_store/mod.rs
benthecarman: This is what we called it in the sqlite store so I am going to keep the same name

src/io/postgres_store/mod.rs
benthecarman: isn't `0` what we expect here?  we dont have any migrations

src/io/postgres_store/mod.rs
benthecarman: Doing this because if we want to create the db when it doesn't exist, the connection string cant include the db name otherwise it'll fail to connect. So we use the default `postgres` when trying. Made this a little more fault tolerant and cleaned up the docs

src/io/postgres_store/mod.rs
benthecarman: sadly no

src/io/postgres_store/mod.rs
tankyleo: let me see `SCHEMA_VERSION` starts at 1, and we are only going to increase it right ? that's why im thinking if we do read a version, it should be 1 or greater.

src/io/postgres_store/mod.rs
tankyleo: > Doing this because if we want to create the db when it doesn't exist, the connection string cant include the db name otherwise it'll fail to connect.

Don't fully understand this one. If the connection string includes the default db name, we just use that to create the database with the name given by `db_name` ?

src/io/postgres_store/mod.rs
tankyleo: Wow ok should we fix this for VSS too ?

src/io/postgres_store/mod.rs
tankyleo: Will take a look at the current code a little later anyways

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1215 @@
...
+				tokio::spawn(async move {
+					if let Err(e) = connection.await {
+						log::error!("PostgreSQL connection error: {e}");
+					}
+				});
+				Ok(client)
+			},
+			PgTlsConnector::NativeTls(tls_connector) => {
+				let (client, connection) =
+					config.connect(tls_connector.clone()).await.map_err(err_map)?;
```
tankyleo, line 567: Sounds like the `Config` default for `sslmode` is `prefer`. Could be more conservative with `require`, do you have a preference here ?

src/builder.rs
```diff
@@ -1087,6 +1125,28 @@ impl ArcedNodeBuilder {
 		self.inner.read().unwrap().build(*node_entropy).map(Arc::new)
 	}
 
+	/// Builds a [`Node`] instance with a [PostgreSQL] backend and according to the options
+	/// previously configured.
+	///
+	/// [PostgreSQL]: https://www.postgresql.org
```
tankyleo, line 1150: Should we copy paste the other docs here?

src/io/postgres_store/mod.rs
tankyleo: following up here, pretty sure we never expect to parse 0 here.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1215 @@
...
+		})?;
+
+		// Initialize next_sort_order from the max existing value.
+		let sql = format!("SELECT COALESCE(MAX(sort_order), 0) FROM {kv_table_name}");
+		let row = client.query_one(sql.as_str(), &[]).await.map_err(|e| {
+			let msg = format!("Failed to read max sort_order from {kv_table_name}: {e}");
+			io::Error::new(io::ErrorKind::Other, msg)
+		})?;
+		let max_sort_order: i64 = row.get(0);
+		debug_assert!(max_sort_order >= 0, "Unexpected negative max sort_order: {max_sort_order}");
```
tankyleo, line 474: nit: since we've got the database constraint i think we can drop this. but it's just a debug assert your call.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1215 @@
...
+
+		// Try connecting to the target database directly — if it exists we're done.
+		if Self::make_config_connection(config, tls).await.is_ok() {
+			return Ok(());
+		}
+
+		// Target database doesn't exist (or isn't reachable). Connect to the
+		// default "postgres" database to create it.
+		let mut bootstrap_config = config.clone();
+		bootstrap_config.dbname("postgres");
```
tankyleo, line 498: I'm still quite certain we should handle the case where the default database is not called "postgres". We've had people complain quite fast when we made this assumption in VSS.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1215 @@
...
+						));
+					}
+				},
+			}
+		}
+
+		Ok(())
+	}
+
+	async fn make_config_connection(config: &Config, tls: &PgTlsConnector) -> io::Result<Client> {
```
tankyleo, line 544: At the top of `create_database_if_not_exists` we make sure that `config` has `db_name` set, should we do something similar here ? Or perhaps this is already handled by the `config.connect` call, you decide.

EDIT: ah if `config` has no `db_name`, it sets `db_name` to `username`, don't think we'd ever want this.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1215 @@
...
+			client.execute(sql.as_str(), &[]).await.map_err(|e| {
+				let msg = format!("Failed to set schema version: {e}");
+				io::Error::new(io::ErrorKind::Other, msg)
+			})?;
+		} else if version_res < SCHEMA_VERSION {
+			migrations::migrate_schema(&client, &kv_table_name, version_res, SCHEMA_VERSION)
+				.await?;
+		} else if version_res > SCHEMA_VERSION {
+			let msg = format!(
+				"Failed to open database: incompatible schema version {version_res}. Expected: {SCHEMA_VERSION}"
```
tankyleo, line 453: IIRC we had a message that was a little bit more user friendly along the lines of "we do not allow downgrades"

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1215 @@
...
+
+		let mut locked_client = self.client.lock().await;
+		self.ensure_connected(&mut locked_client).await?;
+
+		// Fetch one extra row beyond PAGE_SIZE to determine whether a next page exists.
+		let fetch_limit = (PAGE_SIZE + 1) as i64;
+
+		let rows = match page_token {
+			Some(ref token) => {
+				let token_sort_order: i64 = token.as_str().parse().map_err(|_| {
```
tankyleo, line 759: nit: If we parse a negative value here, seems to me that's an error ? Not a huge deal as we'd just return an empty list here, but perhaps the user would appreciate a heads up instead of us returning an empty list.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1215 @@
...
+		Ok(Self { client, config, kv_table_name, tls, write_version_locks, next_sort_order })
+	}
+
+	async fn create_database_if_not_exists(
+		config: &Config, tls: &PgTlsConnector,
+	) -> io::Result<()> {
+		let db_name = config.get_dbname().ok_or_else(|| {
+			io::Error::new(
+				io::ErrorKind::InvalidInput,
+				"Database name must be specified in connection string or as an argument",
```
tankyleo: I read this and I thought: if the `db_name` is not set here, that is a logic error on our part, not a user-error. The default for `db_name` is the username, which I don't think we'd ever want.

src/io/postgres_store/mod.rs
benthecarman: yeah that sounds better

src/builder.rs
benthecarman: which other docs, we already mimic the other `build_with_x_store` docs

src/io/postgres_store/mod.rs
benthecarman: will just keep, doesn't hurt

src/io/postgres_store/mod.rs
benthecarman: this is same error msg as sqlite https://github.com/lightningdevkit/ldk-node/blob/093eefa9fcee8ee2bb407800cea96bdd94fff3e6/src/io/sqlite_store/mod.rs#L315

src/builder.rs
tankyleo: the docs on the other `pub fn build_with_postgres_store` introduced in this PR

src/io/postgres_store/mod.rs
tankyleo: sounds good, good to resolve

src/builder.rs
benthecarman: oh woops i see, yes will do

src/io/postgres_store/mod.rs
benthecarman: Missed your last comment, fixed.

src/io/postgres_store/mod.rs
benthecarman: Yeah which is why we are requiring them to configure it here.

src/io/postgres_store/mod.rs
benthecarman: Based off our discussion offline you seemed to come to opposite conclusion. Do you still feel strongly about this? Can add if so

tankyleo: I confirm yes, for now we hardcode to "postgres" and wait for user demand. 

still wanted to make sure we had some good error message / docs to point people in the correct direction in case "postgres" does not work

src/io/postgres_store/mod.rs
tankyleo: I mentioned this because right before the callsite of `create_database_if_not_exists`, we call `config.dbname(&db_name);` So this error felt a little out of place here, it seems like it's validating user-input, but we already set the `db_name` ourselves before calling this function.

Overall I expect `assert` for logic errors on our side, and `io::Error` for user error ? Could also return `io::Error` for 500 type server side errors.

src/io/postgres_store/mod.rs
benthecarman: added a better error message

src/builder.rs
```diff
@@ -632,6 +629,44 @@ impl NodeBuilder {
...
+	///
+	/// Connects to the PostgreSQL database at the given `connection_string`, e.g.,
+	/// `"postgres://user:password@localhost/ldk_db"`.
+	///
+	/// The given `db_name` will be used or default to
+	/// [`DEFAULT_DB_NAME`](io::postgres_store::DEFAULT_DB_NAME). The database will be created
+	/// automatically if it doesn't already exist. The `connection_string` must not include a
+	/// `dbname` when `db_name` is set, providing both is an error. When auto-creating the
+	/// database, the initial connection is made using the database from the connection string,
+	/// or the server's default if none is specified.
```
tankyleo: Quick followup: I think this merits an update too. We don't quite use the database from the connection string, we always use the "postgres" database.

src/io/postgres_store/mod.rs
benthecarman: Ah i see!

src/builder.rs
```diff
@@ -632,6 +629,44 @@ impl NodeBuilder {
...
+	/// previously configured.
+	///
+	/// Connects to the PostgreSQL database at the given `connection_string`, e.g.,
+	/// `"postgres://user:password@localhost/ldk_db"`.
+	///
+	/// The given `db_name` will be used or default to
+	/// [`DEFAULT_DB_NAME`](io::postgres_store::DEFAULT_DB_NAME). The database will be created
+	/// automatically if it doesn't already exist. The `connection_string` must not include a
+	/// `dbname` when `db_name` is set, providing both is an error. When auto-creating the
+	/// database if id doesn't exist, the initial connection is made using the default database
```
tankyleo: nit: if it doesn't exist

src/builder.rs
```diff
@@ -1085,21 +1116,62 @@ impl ArcedNodeBuilder {
...
```
tankyleo: nit: if it doesn't exist

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1260 @@
...
+	}
+
+	async fn make_config_connection(config: &Config, tls: &PgTlsConnector) -> io::Result<Client> {
+		// make sure db_name is configured, otherwise it may attempt to use the username instead.
+		config.get_dbname().ok_or_else(|| {
+			io::Error::new(
+				io::ErrorKind::InvalidInput,
+				"Database name must be specified in connection string or as an argument",
+			)
+		})?;
```
tankyleo: nit: I was thinking we should do the same thing here as we do at the top of `create_database_if_not_exists` for the same reason: at all callsites of `make_config_connection` we expect `dbname` to be set already.

It might seem aggressive to crash the node, but I myself would prefer to crash earlier rather than later if we reach this point and `dbname` is not set.

tankyleo (commented):
two small nits thank you

benthecarman:
> two small nits thank you

fixed

src/io/postgres_store/mod.rs
tnull: Yeah, let's drop the extra runtime then.

src/builder.rs
tnull: Seems the `postgres` feature is still present in the current version?

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1257 @@
...
+
+		if db_name.is_some() && config.get_dbname().is_some() {
+			return Err(io::Error::new(
+				io::ErrorKind::InvalidInput,
+				"db_name must not be set when the connection string already contains a dbname",
+			));
+		}
+
+		// Default to sslmode=require when TLS is configured
+		if matches!(tls, PgTlsConnector::NativeTls(_)) && config.get_ssl_mode() == SslMode::Prefer {
```
tnull, line 385: Claude:

```
 - Silent TLS downgrade. new() only upgrades SslMode::Prefer → Require when tls_config is Some. If a caller passes tls_config=Some(...) but a connection string with sslmode=disable, the connection is plaintext despite having opted into TLS. Consider erroring (or forcing Require) when
  tls_config.is_some() and sslmode is Disable/Allow. src/io/postgres_store/mod.rs:385.
```

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1257 @@
...
+
+	async fn create_database_if_not_exists(
+		config: &Config, tls: &PgTlsConnector,
+	) -> io::Result<()> {
+		let db_name = config.get_dbname().expect(
+			"database name must be set on config before calling create_database_if_not_exists",
+		);
+
+		// Try connecting to the target database directly — if it exists we're done.
+		let initial_err = match Self::make_config_connection(config, tls).await {
```
tnull, line 490: Claude:
```
- Swallowed initial connect error. create_database_if_not_exists treats any connect failure to the target DB as "DB doesn't exist" (src/io/postgres_store/mod.rs:490). Auth failures, wrong port, etc. fall through to the bootstrap-connect path. The final error message does include initial_err,
   which is good, but logging it at debug! before swallowing would help diagnose the common "looks like it worked, actually didn't" case.
```

src/builder.rs
```diff
@@ -1085,21 +1116,62 @@ impl ArcedNodeBuilder {
...
+	/// previously configured.
+	///
+	/// Connects to the PostgreSQL database at the given `connection_string`, e.g.,
+	/// `"postgres://user:password@localhost/ldk_db"`.
+	///
+	/// The given `db_name` will be used or default to
+	/// [`DEFAULT_DB_NAME`](io::postgres_store::DEFAULT_DB_NAME). The database will be created
+	/// automatically if it doesn't already exist. The `connection_string` must not include a
+	/// `dbname` when `db_name` is set, providing both is an error. When auto-creating the
+	/// database if is doesn't exist, the initial connection is made using the default database
```
tnull, line 1138:
```suggestion
	/// database if it doesn't exist, the initial connection is made using the default database
```

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1257 @@
...
+		Arc::clone(&outer_lock.entry(locking_key).or_default())
+	}
+
+	async fn read_internal(
+		&self, primary_namespace: &str, secondary_namespace: &str, key: &str,
+	) -> io::Result<Vec<u8>> {
+		check_namespace_key_validity(primary_namespace, secondary_namespace, Some(key), "read")?;
+
+		let mut locked_client = self.client.lock().await;
+		self.ensure_connected(&mut locked_client).await?;
```
tnull, line 600: Calling this on every operation should cost us an extra RTT each time, no?

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1257 @@
...
+		async move {
+			inner
+				.list_paginated_internal(&primary_namespace, &secondary_namespace, page_token)
+				.await
+		}
+	}
+}
+
+struct PostgresStoreInner {
+	client: tokio::sync::Mutex<Client>,
```
tnull, line 357: Rather than just using a single client, can we steal `vss-server`'s [`SmallPool`](https://github.com/lightningdevkit/vss-server/blob/53d1fad98bf10b4181535843622d19e3bc0e57ba/rust/impls/src/postgres_store.rs#L53) to not make all operations fully sync and sequenced?

src/builder.rs
```diff
@@ -1085,21 +1116,62 @@ impl ArcedNodeBuilder {
...
+	///
+	/// If `tls_config` is `Some`, TLS will be used for database connections. A custom CA
+	/// certificate can be provided via
+	/// [`PostgresTlsConfig::certificate_pem`](io::postgres_store::PostgresTlsConfig::certificate_pem),
+	/// which will be added to the system's default root certificates (not replace them).
+	/// If `tls_config` is `None`, connections will be unencrypted.
+	///
+	/// [PostgreSQL]: https://www.postgresql.org
+	#[cfg(feature = "postgres")]
+	pub fn build_with_postgres_store(
```
tnull, line 1152: We'll need to add this to `ldk_node.udl` to actually make this useful in bindings.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1257 @@
...
+		let client = Self::make_config_connection(&config, &tls).await?;
+
+		// Create the KV data table if it doesn't exist.
+		let sql = format!(
+			"CREATE TABLE IF NOT EXISTS {kv_table_name} (
+			primary_namespace TEXT NOT NULL,
+			secondary_namespace TEXT NOT NULL DEFAULT '',
+			key TEXT NOT NULL CHECK (key <> ''),
+			value BYTEA,
+			sort_order BIGINT NOT NULL DEFAULT 0 CHECK (sort_order >= 0),
```
tnull, line 405: Should we now also follow what we did in https://github.com/lightningdevkit/vss-server/pull/96 and use `BIGSERIAL`?

src/io/postgres_store/mod.rs
tnull: Hmm, true, and IMO it would be good to actually switch to use a connection pool anyways, so probably we'd need it for that anyways.

src/builder.rs
```diff
@@ -632,6 +629,44 @@ impl NodeBuilder {
...
+	///
+	/// Connects to the PostgreSQL database at the given `connection_string`, e.g.,
+	/// `"postgres://user:password@localhost/ldk_db"`.
+	///
+	/// The given `db_name` will be used or default to
+	/// [`DEFAULT_DB_NAME`](io::postgres_store::DEFAULT_DB_NAME). The database will be created
+	/// automatically if it doesn't already exist. The `connection_string` must not include a
+	/// `dbname` when `db_name` is set, providing both is an error. When auto-creating the
+	/// database if it doesn't exist, the initial connection is made using the default database
+	/// `postgres`.
```
Camillarhi, line 643: Small doc nit, the "initial connection is made using the default database postgres" isn't quite accurate, as the initial connection actually goes to the target database, and `postgres` is only used as a fallback to create it. Could we reorder this so it reads in the order the code actually runs? The same applies to the `ArcedNodeBuilder` doc below and `PostgresStore::new` in `postgres_store/mod.rs`. Suggested wording: 

``` The given `db_name` will be used or default to
    [`DEFAULT_DB_NAME`](io::postgres_store::DEFAULT_DB_NAME). The `connection_string` must not include a `dbname` when `db_name` is set, providing both is an error. The database will be created automatically if it doesn't already exist. The initial connection is made to the target database, and if it fails we fall back to the default `postgres` database to create it.```

File: src/io/postgres_store/mod.rs (hunk 43 above)
Comment (Camillarhi) [inline src/io/postgres_store/mod.rs:357]: I noticed `Mutex<Client>` is used here. Since `Postgres` handles concurrent queries server-side (unlike SQLite), is this intentional for simplicity or to match `SqliteStore`?

**Comment by tnull** [comment]

**Comment by benthecarman** [comment]
Responded to review

File: src/io/postgres_store/mod.rs [hunk 47]
```diff
@@ -0,0 +1,1338 @@
...
+/// A fixed-size pool of mutex-guarded PostgreSQL clients.
+///
+/// `get` first tries each slot in round-robin order with [`try_lock`]; if every slot is
+/// contended it awaits on the round-robin-selected one. Up to [`POOL_SIZE`] operations can run
+/// concurrently against the database. Reconnection only happens when the chosen client's local
+/// mpsc sender is already closed — no per-acquire health-check round trip.
+///
+/// Round-robin (rather than `tokio::select!` over all slot locks) is preferred so [`POOL_SIZE`]
+/// can be `cfg`-conditional without rewriting select arms. The trade-off: under contention the
+/// fallback `lock().await` may park on a busy slot while another frees up a moment later.
```
tankyleo: Hmm not sure it's worth dropping the `tokio::select` to help the tests ? You probably saw this I pass `--test-threads 1` over at VSS when running tests to avoid flooding connections like you mention above. But perhaps there's a better path that doesn't require passing this flag (including maybe sticking to what you have here your call).

src/builder.rs
```diff
@@ -640,6 +640,44 @@ impl NodeBuilder {
...
+		&self, node_entropy: NodeEntropy, connection_string: String, db_name: Option<String>,
+		kv_table_name: Option<String>, tls_config: Option<io::postgres_store::PostgresTlsConfig>,
+	) -> Result<Node, BuildError> {
+		let kv_store = io::postgres_store::PostgresStore::new(
+			connection_string,
+			db_name,
+			kv_table_name,
+			tls_config,
+		)
+		.map_err(|_| BuildError::KVStoreSetupFailed)?;
```
tankyleo: Can we log the error returned here, like what I do in https://github.com/lightningdevkit/ldk-node/pull/893 ?

benthecarman: that needs `build_with_store_and_logger`. Maybe just need to have one of these depend on the other

src/io/postgres_store/mod.rs
benthecarman: Switched to be a select, just have a smaller one for tests

src/builder.rs
tnull: Landed #893, please rebase to use `build_with_store_and_logger`.

src/io/postgres_store/mod.rs
tnull: This seems to be unaddressed? (cc @benthecarman). Or I guess you interpreted 'then' as 'when we upstream'? IMO it's likely unnecessary in this case, so might as well do it now, but let me know if you'd rather keep it for now.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1336 @@
...
+impl std::panic::RefUnwindSafe for PostgresStore {}
+
+impl PostgresStore {
+	/// Constructs a new [`PostgresStore`].
+	///
+	/// Connects to the PostgreSQL database at the given `connection_string`, e.g.,
+	/// `"postgres://user:password@localhost/ldk_db"`.
+	///
+	/// The given `db_name` will be used or default to [`DEFAULT_DB_NAME`]. The
+	/// `connection_string` must not include a `dbname` when `db_name` is set, providing both
```
tnull, line 125: I believe in VSS Server we discovered that different providers have different names for default databases, and we'll need to connect to the default database to be able to initialize the configured database if it's not pre-existing. It seems we might need to add the same config option for 'default db' here, too, which will only be used for the initial connection? Or do we see a better option (given this is a quite ugly/confusing API)?

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1336 @@
...
+	}
+}
+
+/// A fixed-size pool of mutex-guarded PostgreSQL clients.
+///
+/// `get` awaits on whichever slot lock becomes available first via [`tokio::select!`]. Up to
+/// [`POOL_SIZE`] operations can run concurrently against the database. Reconnection only happens
+/// when the chosen client's local mpsc sender is already closed — no per-acquire health-check
+/// round trip.
+struct SmallPool {
```
tnull: We already have a folder for `postgres_store`, let's move this to a `utils.rs` or `pool.rs` or similar.

src/io/mod.rs
```diff
@@ -82,3 +84,13 @@ pub(crate) const STATIC_INVOICE_STORE_PRIMARY_NAMESPACE: &str = "static_invoices
 /// The pending payment information will be persisted under this prefix.
 pub(crate) const PENDING_PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE: &str = "pending_payments";
 pub(crate) const PENDING_PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE: &str = "";
+
+/// TLS configuration for PostgreSQL connections.
+#[cfg(any(feature = "postgres", feature = "uniffi"))]
+#[derive(Debug, Clone)]
+pub struct PostgresTlsConfig {
```
tnull: This doesn't belong into this file but should be part of the `postgres_store` sub-module *or* if we deem it part of the user-facing builder API, it should be in `builder.rs`.

src/builder.rs
```diff
@@ -640,6 +640,44 @@ impl NodeBuilder {
...
+	/// will be created automatically if it doesn't already exist. The initial connection is
+	/// made to the target database, and if it fails we fall back to the default `postgres`
+	/// database to create it.
+	///
+	/// The given `kv_table_name` will be used or default to
+	/// [`DEFAULT_KV_TABLE_NAME`](io::postgres_store::DEFAULT_KV_TABLE_NAME).
+	///
+	/// If `tls_config` is `Some`, TLS will be used for database connections. A custom CA
+	/// certificate can be provided via
+	/// [`PostgresTlsConfig::certificate_pem`](io::postgres_store::PostgresTlsConfig::certificate_pem),
```
tnull: Given the `PostgresTlsConfig` currently only holds one field, should `certificate_pem` just be an optional field directly on `build_with_postgres_store`? And then we could use `PostgresTlsConfig` internally only for now?

src/io/mod.rs
benthecarman: Put it here so the bindings had access to it. Was unsure if we wanted to have the bindings require postgres enabled or not. Currently its behind the feature flag but we still need the type exposed

src/io/postgres_store/mod.rs
benthecarman: @tankyleo discussed this and thought it might be better to punt on it for now. You can always just use whatever bespoke default db they use as your db instead of the `DEFAULT_DB_NAME` in your connection string. But yeah can add if you think it'll cause issues

src/builder.rs
benthecarman: yeah this solves the bindings issue above too

src/io/postgres_store/mod.rs
benthecarman: Ohh missed your other comment, fixed

src/io/mod.rs
tnull: We usually have the pattern of exposing any types required by the bindigns by re-exporting at crate level via the `use` statements in `ffi/types.rs`. That should work here, too? I.e., have the type live in the sub-module, and just expose it via that at crate level under `uniffi`?

We def. want to expose it also under bindings. While probably not super interesting for Swift/Kotlin mobile users, we also have some Go bindings users for which postgres could be interesting at some point, potentially.

Cargo.toml
```diff
@@ -25,6 +25,7 @@ panic = 'abort'     # Abort on panic
 
 [features]
 default = []
+postgres = ["dep:tokio-postgres", "dep:native-tls", "dep:postgres-native-tls"]
```
tnull, line 28: Hmm, we probably want to introduce features even before the next release, but still not a fan of *just* adding a one-off feature for this.

tnull: Reconsidered offline. Fine to leave for now, for the release we'll change it anyways as part of https://github.com/lightningdevkit/ldk-node/issues/900

tnull (commented):
This needs a minor rebase by now. Also let me know once you looked further into the blocking behavior you saw in the benchmarks (discussed offline)

benthecarman:
> Also let me know once you looked further into the blocking behavior you saw in the benchmarks (discussed offline)

Yeah so in the payment benchmark (see #915) it currently fails with a postgres backend because we’re hitting the blocking task limit in tokio by using `block_in_place` for lots of async writes (postgres writes).

We can fix this by adding a dedicated runtime to the postgres store similar to VSS or by taking the dep `postgres-sync` (it works by adding a runtime too). However we were removed that in this PR previously.

Otherwise the real fix is to have `DataStore` use the async KV Store, however, this would be a much larger refactor as this would likely require us to have the actual api of ldk-node be async

tnull:
> > Also let me know once you looked further into the blocking behavior you saw in the benchmarks (discussed offline)
> 
> Yeah so in the payment benchmark (see #915) it currently fails with a postgres backend because we’re hitting the blocking task limit in tokio by using `block_in_place` for lots of async writes (postgres writes).
> 
> We can fix this by adding a dedicated runtime to the postgres store similar to VSS or by taking the dep `postgres-sync` (it works by adding a runtime too). However we were removed that in this PR previously.
> 
> Otherwise the real fix is to have `DataStore` use the async KV Store, however, this would be a much larger refactor as this would likely require us to have the actual api of ldk-node be async

Okay, good to hear it was indeed the blocking worker limit. Mind opening an issue for this? We plan to make `DataStore` async anyways (though not in this PR ofc), I think then we should just be able to merge.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1125 @@
...
+		Self::create_database_if_not_exists(&config, &tls).await?;
+
+		let client = make_config_connection(&config, &tls).await?;
+
+		// Create the KV data table if it doesn't exist. `sort_order` uses BIGSERIAL so
+		// the database assigns a fresh, monotonically increasing value on each INSERT and
+		// keeps the previous value untouched on UPSERT-update; the sequence persists across
+		// restarts.
+		let sql = format!(
+			"CREATE TABLE IF NOT EXISTS {kv_table_name} (
```
tnull: Codex nit:    
>   - [P2] Quote or validate PostgreSQL identifiers — /home/tnull/workspace/ldk-node/src/io/postgres_store/mod.rs:348-348
>     Because kv_table_name is interpolated directly as an unquoted SQL identifier, a caller-provided table name such as tenant-1, a reserved word, or a schema-qualified name will fail during store setup even though the API accepts an arbitrary String. Quote identifiers safely or reject
>     unsupported names before formatting them into the table/comment/index SQL.

benthecarman: fixed

tnull: Seems now all Postgres tests are failing.

benthecarman: fixed

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1170 @@
...
+		// Try connecting to the target database directly — if it exists we're done.
+		let initial_err = match make_config_connection(config, tls).await {
+			Ok(_) => return Ok(()),
+			Err(e) => e,
+		};
+		// `initial_err` is only included in the final error if the bootstrap connect also
+		// fails; otherwise it gets dropped silently. Log it at debug! so a later failure
+		// (e.g. an auth error that surfaces only when the pool is built) can be traced
+		// back to the real cause.
+		log::debug!(
```
tnull: Huh, what's this? We're not (always) using `log` facade in LDK Node. You'll need to give `PostgresStore` an `Arc<Logger>` and log towards that using the LDK macros to make sure users will always see the outputs independently of chosen backend.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1170 @@
...
+	/// will be unencrypted.
+	pub async fn new(
+		connection_string: String, db_name: Option<String>, kv_table_name: Option<String>,
+		certificate_pem: Option<String>,
+	) -> io::Result<Self> {
+		let tls = Self::build_tls_connector(certificate_pem)?;
+		let inner = PostgresStoreInner::new(connection_string, db_name, kv_table_name, tls).await?;
+		let inner = Arc::new(inner);
+		let next_write_version = AtomicU64::new(1);
+		let handle = tokio::runtime::Handle::current();
```
tnull: No, please hand in the `Arc<Runtime>` use that for the blocking tasks. We should also drop the custom `block_on` helper as it makes things even more confusing. Also note there is a bug here:

>  - [P2] Build PostgresStore on the configured runtime — /home/tnull/workspace/ldk-node/src/builder.rs:676-677
>     When set_runtime is used and build_with_postgres_store is called from a different Tokio runtime, Runtime::block_on uses the caller's current runtime instead of the configured handle. PostgresStore::new captures Handle::current() and spawns its connection drivers there, so a node built from
>     a temporary caller runtime can later run persistence against a dropped runtime rather than the node runtime. Please force construction onto the configured runtime handle.

benthecarman: fixed

src/io/postgres_store/mod.rs
benthecarman: oops yeah we had a few of those. fixed them all

src/io/postgres_store/mod.rs
tnull: No, I meant to hand in the `Arc<crate::runtime::Runtime>` when setting up the store not to revert back to create it's own `Arc<tokio::runtime::Runtime>` (also cf. the original comment https://github.com/lightningdevkit/ldk-node/pull/863/changes#r3027338155)

tnull: Anyways, I'll just land this as we'll rip the `Sync` path out soon anyways.

src/builder.rs
```diff
@@ -644,6 +644,50 @@ impl NodeBuilder {
...
+	/// will be unencrypted.
+	///
+	/// [PostgreSQL]: https://www.postgresql.org
+	#[cfg(feature = "postgres")]
+	pub fn build_with_postgres_store(
+		&self, node_entropy: NodeEntropy, connection_string: String, db_name: Option<String>,
+		kv_table_name: Option<String>, certificate_pem: Option<String>,
+	) -> Result<Node, BuildError> {
+		let logger = setup_logger(&self.log_writer_config, &self.config)?;
+		let runtime = self.setup_runtime(&logger)?;
```
tnull, line 675: Can we should just use this `crate::runtime::Runtime` and hand it to `PostgresStore` below, at least when building via the builder? But, maybe also fine to leave as-is for now as we'll rip the `KVStoreSync` path out very soon anyways.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1227 @@
...
+		certificate_pem: Option<String>,
+	) -> io::Result<Self> {
+		Self::new_with_logger(connection_string, db_name, kv_table_name, certificate_pem, None)
+			.await
+	}
+
+	/// Like [`Self::new`], but lets crate-internal callers route setup logs through the node
+	/// logger. This stays separate because [`Logger`] is crate-private and cannot be part of the
+	/// public [`PostgresStore::new`] signature.
+	pub(crate) async fn new_with_logger(
```
tnull, line 147: Why is this `async` now that we call `block_on` internally? It should be either or, not both.

src/builder.rs
```diff
@@ -1116,6 +1169,58 @@ impl ArcedNodeBuilder {
...
+	/// Builds a [`Node`] instance with a [PostgreSQL] backend and according to the options
+	/// previously configured.
+	///
+	/// This requires the `postgres` crate feature.
+	#[cfg(not(feature = "postgres"))]
+	pub fn build_with_postgres_store(
+		&self, _node_entropy: Arc<NodeEntropy>, _connection_string: String,
+		_db_name: Option<String>, _kv_table_name: Option<String>, _certificate_pem: Option<String>,
+	) -> Result<Arc<Node>, BuildError> {
+		Err(BuildError::KVStoreSetupFailed)
```
tnull, line 1221: We really need to find a way to avoid this. Just failing here is an impossible API, so we need to fix it before the next release.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1227 @@
...
+			.worker_threads(INTERNAL_RUNTIME_WORKERS)
+			.max_blocking_threads(INTERNAL_RUNTIME_WORKERS)
+			.build()
+			.map_err(|e| {
+				io::Error::new(
+					io::ErrorKind::Other,
+					format!("Failed to build PostgreSQL runtime: {e}"),
+				)
+			})?;
+		let tls = Self::build_tls_connector(certificate_pem)?;
```
tnull, line 167: >   - [P1] Shut down the runtime on setup errors — /home/tnull/workspace/ldk-node/src/io/postgres_store/mod.rs:167-167
>     When PostgresStore::new(...).await is called from a Tokio runtime, any ? after internal_runtime is created (e.g. invalid certificate_pem, bad connection string, or PostgreSQL being unavailable) drops that Tokio runtime inside an async context. Tokio panics in that situation, so expected
>     setup errors can crash the caller instead of returning io::Error; validate before creating the runtime or call shutdown_background before returning errors.

src/io/postgres_store/mod.rs
```diff
@@ -0,0 +1,1227 @@
...
+		let runtime_handle = internal_runtime.handle();
+		let inner = tokio::task::block_in_place(|| {
+			runtime_handle.block_on(async {
```
tnull, line 170: >   - [P2] Avoid block_in_place in the async constructor — /home/tnull/workspace/ldk-node/src/io/postgres_store/mod.rs:169-169
>     When the public PostgresStore::new(...).await is polled on a Tokio current_thread runtime (including the default #[tokio::test] flavor), block_in_place panics before setup can complete. Since the store already creates its own multi-thread runtime for PostgreSQL work, use a construction
>     path that doesn't require the caller to be on a multi-thread runtime.

tnull (commented):
Hmm, I'll land this and then see to rip out the `KVStoreSync` path. But the current state is really messy and we can ship it.

tnull (approved):

PR #864, opened 2026-04-01
Add PaginatedKVStore support to VssStore

benthecarman:
Implement `PaginatedKVStore` and `PaginatedKVStoreSync` traits for `VssStore`, enabling paginated key listing via cursor-based pagination using `PageToken`.

At the moment VSS returns them in key order which is not ideal. We would need to add to the VSS server timestamps so they are returned in creation order. Ideally though after that change, it wouldn't require any changes on the client or in here because we have already hooked up the pagination. Curious on @tankyleo's thoughts here.

After this we'll have pagination for all the kv stores and can move forward with using it inside of ldk-node

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,7 +82,7 @@ async-trait = { version = "0.1", default-features = false }
 tokio-postgres = { version = "0.7", default-features = false, features = ["runtime"], optional = true }
 native-tls = { version = "0.2", default-features = false, optional = true }
 postgres-native-tls = { version = "0.5", default-features = false, features = ["runtime"], optional = true }
-vss-client = { package = "vss-client-ng", version = "0.5" }
+vss-client = { package = "vss-client-ng", version = "0.6" }
 prost = { version = "0.11.6", default-features = false}
 #bitcoin-payment-instructions = { version = "0.6" }
 bitcoin-payment-instructions = { git = "https://github.com/tnull/bitcoin-payment-instructions", rev = "ff09ce9401afa448549a8f101172700bcd14d7bb" }
--- a/src/io/vss_store.rs
+++ b/src/io/vss_store.rs
@@ -24,7 +24,7 @@ use bitcoin::Network;
 use lightning::impl_writeable_tlv_based_enum;
 use lightning::io::{self, Error, ErrorKind};
 use lightning::sign::{EntropySource as LdkEntropySource, RandomBytes};
-use lightning::util::persist::KVStore;
+use lightning::util::persist::{KVStore, PageToken, PaginatedKVStore, PaginatedListResponse};
 use lightning::util::ser::{Readable, Writeable};
 use prost::Message;
 use vss_client::client::VssClient;
@@ -69,6 +69,8 @@ impl_writeable_tlv_based_enum!(VssSchemaVersion,
 	(0, V0) => {},
 	(1, V1) => {},
 );
+
+const PAGE_SIZE: i32 = 50;
 
 const VSS_HARDENED_CHILD_INDEX: u32 = 877;
 const VSS_SIGS_AUTH_HARDENED_CHILD_INDEX: u32 = 139;
@@ -293,6 +295,32 @@ impl KVStore for VssStore {
 	}
 }
 
+impl PaginatedKVStore for VssStore {
+	fn list_paginated(
+		&self, primary_namespace: &str, secondary_namespace: &str, page_token: Option<PageToken>,
+	) -> impl Future<Output = Result<PaginatedListResponse, io::Error>> + 'static + Send {
+		let primary_namespace = primary_namespace.to_string();
+		let secondary_namespace = secondary_namespace.to_string();
+		let inner = Arc::clone(&self.inner);
+		let runtime = self.internal_runtime();
+		async move {
+			let task = runtime.spawn(async move {
+				inner
+					.list_paginated_internal(
+						&inner.async_client,
+						primary_namespace,
+						secondary_namespace,
+						page_token,
+					)
+					.await
+			});
+			task.await.map_err(|e| {
+				io::Error::new(io::ErrorKind::Other, format!("VSS runtime task failed: {}", e))
+			})?
+		}
+	}
+}
+
 impl Drop for VssStore {
 	fn drop(&mut self) {
 		if let Some(runtime) = self.internal_runtime.take() {
@@ -393,9 +421,8 @@ impl VssStoreInner {
 
 	async fn list_keys(
 		&self, client: &VssClient<CustomRetryPolicy>, primary_namespace: &str,
-		secondary_namespace: &str, page_token: Option<String>, page_size: Option<i32>,
+		secondary_namespace: &str, key_prefix: String, page_token: Option<String>, page_size: Option<i32>,
 	) -> io::Result<(Vec<String>, Option<String>)> {
-		let key_prefix = self.build_obfuscated_prefix(primary_namespace, secondary_namespace);
 		let request = ListKeyVersionsRequest {
 			store_id: self.store_id.clone(),
 			key_prefix: Some(key_prefix),
@@ -542,11 +569,12 @@ impl VssStoreInner {
 	) -> io::Result<Vec<String>> {
 		check_namespace_key_validity(&primary_namespace, &secondary_namespace, None, "list")?;
 
+		let key_prefix = self.build_obfuscated_prefix(&primary_namespace, &secondary_namespace);
 		let mut page_token: Option<String> = None;
 		let mut keys = vec![];
 		loop {
 			let (page_keys, next_page_token) = self
-				.list_keys(client, &primary_namespace, &secondary_namespace, page_token, None)
+				.list_keys(client, &primary_namespace, &secondary_namespace, key_prefix.clone(), page_token, None)
 				.await?;
 			keys.extend(page_keys);
 			match next_page_token {
@@ -555,6 +583,35 @@ impl VssStoreInner {
 			}
 		}
 		Ok(keys)
+	}
+
+	async fn list_paginated_internal(
+		&self, client: &VssClient<CustomRetryPolicy>, primary_namespace: String,
+		secondary_namespace: String, page_token: Option<PageToken>,
+	) -> io::Result<PaginatedListResponse> {
+		check_namespace_key_validity(
+			&primary_namespace,
+			&secondary_namespace,
+			None,
+			"list_paginated",
+		)?;
+
+		let key_prefix = self.build_obfuscated_prefix(&primary_namespace, &secondary_namespace);
+		let vss_page_token = page_token.map(|t| t.to_string());
+		let (keys, next_page_token) = self
+			.list_keys(
+				client,
+				&primary_namespace,
+				&secondary_namespace,
+				key_prefix,
+				vss_page_token,
+				Some(PAGE_SIZE),
+			)
+			.await?;
+
+		let next_page_token = next_page_token.map(PageToken::new);
+
+		Ok(PaginatedListResponse { keys, next_page_token })
 	}
 
 	async fn execute_locked_write<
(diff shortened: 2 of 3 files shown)


tnull (commented):
Ah, I had hoped that we could do this after https://github.com/lightningdevkit/rust-lightning/pull/4323 lands, i.e., add it upstream directly. However, that PR seems to be delayed now as reviewer brought up more requirements, so probably need to go ahead here.

I'll let @tankyleo take a first look here and on the corresponding `vss-server` PR.

src/io/vss_store.rs
```diff
@@ -638,6 +686,49 @@ impl VssStoreInner {
...
+		})?;
+
+		let mut keys = Vec::with_capacity(response.key_versions.len());
+		for kv in response.key_versions {
+			keys.push(self.extract_key(&kv.key)?);
+		}
+
+		// VSS uses empty string to signal the last page
+		let next_page_token =
+			response.next_page_token.filter(|t| !t.is_empty()).map(PageToken::new);
```
tankyleo: Sounds to me like we should line-up the VSS API and the `PaginatedKVStore` API better, I raised a similar point in the VSS PR.

Right now a VSS could return a non-empty keys list, but signal "no more data to give" with a `String::new()` for the page token.

A consumer of `PaginatedKVStore` does another roundtrip, but the VSS did not expect any further roundtrips.

src/io/vss_store.rs
```diff
@@ -1051,4 +1142,92 @@ mod tests {
...
+		let mut page_token = None;
+
+		loop {
+			let response =
+				PaginatedKVStoreSync::list_paginated(&store, ns, sub, page_token).unwrap();
+			all_keys.extend(response.keys);
+			match response.next_page_token {
+				Some(token) => page_token = Some(token),
+				None => break,
+			}
```
tankyleo, line 1056: hmm here we determine whether we are done with pagination based on whether the page token is `None`, and not whether the list is empty. I guess just looking for clarification on the `PaginatedKVStore` API :)

src/io/vss_store.rs
benthecarman: Yeah I feel like `page_token: None` should just be the signal that there is no more. That's what we do in the file system store and sqlite. Should fix that on the vss-server side but we can keep this is_empty check for safety.

benthecarman: Changed in https://github.com/lightningdevkit/vss-server/pull/96 to do this

tnull: > Yeah I feel like `page_token: None` should just be the signal that there is no more. That's what we do in the file system store and sqlite. Should fix that on the vss-server side but we can keep this is_empty check for safety.

I think we're following protobuf's/Google's best practices here. https://google.aip.dev/158 states:

> - Response messages for collections should define a string next_page_token field, providing the user with a page token that may be used to retrieve the next page.
>     - The field containing pagination results should be the first field in the message and have a field number of 1. It should be a repeated field containing a list of resources constituting a single page of results.
>     - If the end of the collection has been reached, the next_page_token field must be empty. This is the only way to communicate "end-of-collection" to users.
>     - If the end of the collection has not been reached (or if the API can not determine in time), the API must provide a next_page_token.

src/io/vss_store.rs
benthecarman: Update to handle empty string

benthecarman:
Okay updated to be compliant with the AIP 158 stuff

src/io/vss_store.rs
```diff
@@ -1051,4 +1142,92 @@ mod tests {
...
+
+		let mut all_keys = Vec::new();
+		let mut page_token = None;
+
+		loop {
+			let response =
+				PaginatedKVStoreSync::list_paginated(&store, ns, sub, page_token).unwrap();
+			all_keys.extend(response.keys);
+			match response.next_page_token {
+				Some(token) if !token.as_str().is_empty() => page_token = Some(token),
```
tankyleo: Sounds to me here we should break only if the page token field is `None`, and not if the token is `Some("")`, otherwise, some details of the VSS API are leaking past the `PaginatedKVStore` API ?

src/io/vss_store.rs
```diff
@@ -1051,4 +1142,92 @@ mod tests {
...
+	#[test]
+	fn vss_paginated_namespace_isolation() {
+		let store = build_vss_store();
+
+		KVStoreSync::write(&store, "ns_a", "sub", "key_1", vec![1u8; 8]).unwrap();
+		KVStoreSync::write(&store, "ns_a", "sub", "key_2", vec![2u8; 8]).unwrap();
+		KVStoreSync::write(&store, "ns_b", "sub", "key_3", vec![3u8; 8]).unwrap();
+
+		let response = PaginatedKVStoreSync::list_paginated(&store, "ns_a", "sub", None).unwrap();
+		assert_eq!(response.keys.len(), 2);
```
tankyleo, line 1196: let's also assert on the exact keys returned ? Just to make sure, here and below

src/io/vss_store.rs
```diff
@@ -1051,4 +1142,92 @@ mod tests {
...
+
+		KVStoreSync::write(&store, ns, sub, "a", vec![1u8; 8]).unwrap();
+		KVStoreSync::write(&store, ns, sub, "b", vec![2u8; 8]).unwrap();
+		KVStoreSync::write(&store, ns, sub, "c", vec![3u8; 8]).unwrap();
+
+		KVStoreSync::remove(&store, ns, sub, "b", false).unwrap();
+
+		let response = PaginatedKVStoreSync::list_paginated(&store, ns, sub, None).unwrap();
+		assert_eq!(response.keys.len(), 2);
+		assert!(!response.keys.contains(&"b".to_string()));
```
tankyleo, line 1183: likewise here, just one or two more asserts to assert on the exact keys listed.

src/io/vss_store.rs
```diff
@@ -638,6 +686,49 @@ impl VssStoreInner {
...
+			);
+			Error::new(ErrorKind::Other, msg)
+		})?;
+
+		let mut keys = Vec::with_capacity(response.key_versions.len());
+		for kv in response.key_versions {
+			keys.push(self.extract_key(&kv.key)?);
+		}
+
+		// VSS uses empty string to signal the last page
```
tankyleo: Should we update the docs on the `PaginatedKVStore` to let people know they should stop asking for pages once they get a `None` for the page token ?

Maybe this is implied in the current API docs over there. I want to make sure people don't wait for the "empty list" condition to stop asking for more pages, as VSS API does not quite fit this. They should look at their page_token they got back and stop asking for pages once it is none.

src/io/vss_store.rs
benthecarman: ah true, good catch

src/io/vss_store.rs
benthecarman: Yeah I think that's already implied

tankyleo (commented):
LGTM !

src/io/vss_store.rs
```diff
@@ -1057,4 +1142,97 @@ mod tests {
 		do_read_write_remove_list_persist(&vss_store);
 		drop(vss_store)
 	}
+
+	fn build_vss_store() -> VssStore {
```
tnull: nit: This seems to simply copy the init code from the tests above. If we want to introduce a helper, please do it in a prefactor commit, add it to the top of `mod tests` (after imports), and have the pre-existing tests use the util.

src/io/vss_store.rs
```diff
@@ -642,6 +686,49 @@ impl VssStoreInner {
 		Ok(keys)
 	}
 
+	async fn list_paginated_internal(
```
tnull, line 687: It seems this is duplicating a lot of the logic of `list_all_keys`. Rather than doing that, can we

1) move the logic to a new method called `list_keys` or similar that takes the paget oken
2) move the `	while page_token != Some("".to_string())` loop to `list_internal`, calling `list_keys`
3) hence have both `list_internal` and `list_internal_paginated` reuse the same `list_keys` code to avoid duplication

benthecarman: Done, also fixed a potential issue where if the VSS server returned `None` for the page token we could go into an infinite loop.

src/io/vss_store.rs
benthecarman: done

tnull (commented):
Looks good I think, but I do wonder if we now need protocol-level versioning for VSS, see https://github.com/lightningdevkit/vss-server/pull/96#pullrequestreview-4119541908.

src/io/vss_store.rs
```diff
@@ -476,35 +522,32 @@ impl VssStoreInner {
...
-				keys.push(self.extract_key(&kv.key)?);
-			}
-			page_token = response.next_page_token;
+		let mut keys = Vec::with_capacity(response.key_versions.len());
+		for kv in response.key_versions {
+			keys.push(self.extract_key(&kv.key)?);
 		}
-		Ok(keys)
+
+		Ok((keys, response.next_page_token))
```
Camillarhi, line 550: `list_keys` returns `response.next_page_token` raw, so both callers independently handle the `Some("")` case. Would be cleaner to normalize it in list_keys itself. What do you think?

tnull:

src/io/vss_store.rs
benthecarman: done

tnull (commented):
As mentioned above, I think we should wait to land this until we have the protocol versioning in VSS Server, so we can require it here (and error out if not met) before attempting to use pagination and fail at runtime.

benthecarman:
Rebased and updated vss-client to the 0.6

Cargo.toml
```diff
@@ -82,7 +82,7 @@ async-trait = { version = "0.1", default-features = false }
 tokio-postgres = { version = "0.7", default-features = false, features = ["runtime"], optional = true }
 native-tls = { version = "0.2", default-features = false, optional = true }
 postgres-native-tls = { version = "0.5", default-features = false, features = ["runtime"], optional = true }
-vss-client = { package = "vss-client-ng", version = "0.5" }
+vss-client = { package = "vss-client-ng", version = "0.6" }
```
tnull, line 85: Given this version introduced new error types (namely version mismatch, but can't recall if there are others), we'll need to update our retry policy to *not* retry if we hit a version mismatch, as it's unrecoverable.

src/io/vss_store.rs
```diff
@@ -416,7 +442,9 @@ impl VssStoreInner {
 			keys.push(self.extract_key(&kv.key)?);
 		}
 
-		Ok((keys, response.next_page_token))
+		// VSS may return an empty string instead of None to signal the last page.
+		let next_page_token = response.next_page_token.filter(|t| !t.is_empty());
```
tnull, line 447: Why is this necessary in this commit? Shouldn't this happen in the previous commit?

src/io/vss_store.rs
```diff
@@ -548,13 +576,42 @@ impl VssStoreInner {
 				.await?;
 			keys.extend(page_keys);
 			match next_page_token {
-				Some(t) if !t.is_empty() => page_token = Some(t),
-				_ => break,
+				Some(t) => page_token = Some(t),
```
tnull, line 581: Same here: why in this commit?

tnull (commented):
Mostly looks good, just some minor comments.

Please also include an entry in the compatibility section of the pending CHANGELOG.md to note that users *must* upgrade VSS server before upgrading LDK Node.

Cargo.toml
benthecarman: Fixed. Looks like it would have failed before but now has better and more explicit error message

src/io/vss_store.rs
benthecarman: moved back

src/io/vss_store.rs
benthecarman: moved back

Cargo.toml
tnull: No, not what I meant: we need to update

```rust
fn retry_policy() -> CustomRetryPolicy {
```

above to `skip_retry_on_error` for `VSSVersionMismatchError`.

benthecarman: Ah, added

src/io/vss_store.rs
```diff
@@ -391,35 +417,34 @@ impl VssStoreInner {
...
-	async fn list_all_keys(
+	async fn list_keys(
 		&self, client: &VssClient<CustomRetryPolicy>, primary_namespace: &str,
-		secondary_namespace: &str,
-	) -> io::Result<Vec<String>> {
-		let mut page_token = None;
-		let mut keys = vec![];
+		secondary_namespace: &str, page_token: Option<String>, page_size: Option<i32>,
+	) -> io::Result<(Vec<String>, Option<String>)> {
 		let key_prefix = self.build_obfuscated_prefix(primary_namespace, secondary_namespace);
```
tankyleo: We used to do this encryption once for the entire list, but now we do this once for every page. Encryption with chacha20 is fast so should be fine?

src/io/vss_store.rs
```diff
@@ -543,20 +568,50 @@ impl VssStoreInner {
...
+		secondary_namespace: String, page_token: Option<PageToken>,
+	) -> io::Result<PaginatedListResponse> {
+		check_namespace_key_validity(
+			&primary_namespace,
+			&secondary_namespace,
+			None,
+			"list_paginated",
+		)?;
+
+		const PAGE_SIZE: i32 = 50;
```
tankyleo: nit: let's put this at the top of this module here, similar to how we do it in other modules?

Also thought since all stores use 50, maybe that can me a global constant in the `io` module.

src/io/vss_store.rs
```diff
@@ -543,20 +568,50 @@ impl VssStoreInner {
...
+
+		let vss_page_token = page_token.map(|t| t.to_string());
```
tankyleo, line 600: Thought we could do a `into_inner` instead of allocating a new string to copy a string, but realized this is gated behind the upstream API so should be fine for now.

tankyleo (commented):
LGTM a few nits/questions

src/io/vss_store.rs
benthecarman: moved outside to be safe

tankyleo (approved):

tnull (approved):

tnull:
Oh, seems `cargo fmt` didn't pass on this branch. Now fixed in #942. I miss CI.

PR #865, opened 2026-04-02
Fix balance candidate selection during pending splices

joostjager:
Mirror LDK's sentinel logic for confirmed_balance_candidate_index in ClaimableOnChannelClose: when the index is 0 (no specific alternative funding confirmed), use the last balance candidate (most current splice/RBF attempt) instead of the first. This aligns with LDK's claimable_amount_satoshis() behavior and fixes a mismatch where total_lightning_balance_sats could differ from the sum of individual LightningBalance amounts during pending splices.

AI tools were used in preparing this commit.

### DIFF
--- a/src/balance.rs
+++ b/src/balance.rs
@@ -231,10 +231,16 @@ impl LightningBalance {
 				inbound_claiming_htlc_rounded_msat,
 				inbound_htlc_rounded_msat,
 			} => {
-				// unwrap safety: confirmed_balance_candidate_index is guaranteed to index into balance_candidates
-				let balance = balance_candidates
-					.get(confirmed_balance_candidate_index)
-					.expect("LDK should provide a valid confirmed balance candidate index");
+				// When confirmed_balance_candidate_index is 0, no specific alternative
+				// funding has been confirmed yet, so use the last candidate (most current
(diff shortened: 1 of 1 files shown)


src/balance.rs
```diff
@@ -231,8 +231,14 @@ impl LightningBalance {
...
 			} => {
-				// unwrap safety: confirmed_balance_candidate_index is guaranteed to index into balance_candidates
-				let balance = balance_candidates.get(confirmed_balance_candidate_index).unwrap();
+				// When confirmed_balance_candidate_index is 0, no specific alternative
+				// funding has been confirmed yet, so use the last candidate (most current
+				// splice/RBF attempt), matching LDK's claimable_amount_satoshis behavior.
+				let balance = if confirmed_balance_candidate_index != 0 {
+					&balance_candidates[confirmed_balance_candidate_index]
+				} else {
+					balance_candidates.last().unwrap()
```
tnull: Mind adding the unwrap-safety comment back?

tnull (commented):
One comment, otherwise LGTM

src/balance.rs
tnull: I guess a corresponding `expect` will now be necessary post-rebase

joostjager: Added

tnull (approved):

PR #867, opened 2026-04-02
Re-pin `idna_adapter` to for MSRV builds

tnull:
.. we previously dropped the pin when moving MSRV to 1.85, but it seems that is not sufficient anymore..


joostjager (approved):
Ack, assuming build passes

PR #868, opened 2026-04-03
Linting: Disallow `unwrap`s, replace with `expect` or error propagation

tnull:
For the longest time we wanted to enforce clear rules around `unwrap` use. Here we finally do that:

1) We switch all `Mutex`/`RwLock` locking to `Ext` helpers that avoid the `unwrap` and are shorter, reducing clutter in code
2) We switch over any other `unwrap`s to use `expect`s, providing clear rationale why they are deemed unreachable
3) For some cases we switch to error propagation. While they are not reachable, it seems simpler/cleaner that way.
4) We ban the introduction of new `unwrap`s through `clippy` in CI (in non-test/documentation code, that is)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -90,6 +90,21 @@ jobs:
         run: |
           RUSTFLAGS="--cfg no_download --cfg cycle_tests" cargo test --features uniffi
 
+  linting:
+    name: Linting
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@v6
+      - name: Install Rust and clippy
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --profile=minimal --default-toolchain stable
+          rustup component add clippy
+      - name: Ban `unwrap` in library code
+        run: |
+          cargo clippy --lib --verbose --color always -- -A warnings -D clippy::unwrap_used -A clippy::tabs_in_doc_comments
+          cargo clippy --lib --features uniffi --verbose --color always -- -A warnings -D clippy::unwrap_used -A clippy::tabs_in_doc_comments
+
   doc:
     name: Documentation
     runs-on: ubuntu-latest
--- a/src/chain/bitcoind.rs
+++ b/src/chain/bitcoind.rs
@@ -194,10 +194,11 @@ impl BitcoindChainSource {
 			{
 				Ok(chain_tip) => {
 					{
+						let elapsed_ms = now.elapsed().map(|d| d.as_millis()).unwrap_or(0);
 						log_info!(
 							self.logger,
 							"Finished synchronizing listeners in {}ms",
-							now.elapsed().expect("system time must not go backwards").as_millis()
+							elapsed_ms,
 						);
 						*self.latest_chain_tip.write().expect("lock") = Some(chain_tip);
 						let unix_time_secs_opt =
@@ -410,11 +411,8 @@ impl BitcoindChainSource {
 		let now = SystemTime::now();
 		match spv_client.poll_best_tip().await {
 			Ok((ChainTip::Better(tip), true)) => {
-				log_trace!(
-					self.logger,
-					"Finished polling best tip in {}ms",
-					now.elapsed().expect("system time must not go backwards").as_millis()
-				);
+				let elapsed_ms = now.elapsed().map(|d| d.as_millis()).unwrap_or(0);
+				log_trace!(self.logger, "Finished polling best tip in {}ms", elapsed_ms);
 				*self.latest_chain_tip.write().expect("lock") = Some(tip);
 			},
 			Ok(_) => {},
@@ -434,12 +432,13 @@ impl BitcoindChainSource {
 			.await
 		{
 			Ok((unconfirmed_txs, evicted_txids)) => {
+				let elapsed_ms = now.elapsed().map(|d| d.as_millis()).unwrap_or(0);
 				log_trace!(
 					self.logger,
 					"Finished polling mempool of size {} and {} evicted transactions in {}ms",
 					unconfirmed_txs.len(),
 					evicted_txids.len(),
-					now.elapsed().expect("system time must not go backwards").as_millis()
+					elapsed_ms,
 				);
 				onchain_wallet.apply_mempool_txs(unconfirmed_txs, evicted_txids).unwrap_or_else(
 					|e| {
--- a/src/chain/bitcoind.rs
+++ b/src/chain/bitcoind.rs
@@ -132,7 +132,7 @@ impl BitcoindChainSource {
 		// First register for the wallet polling status to make sure `Node::sync_wallets` calls
 		// wait on the result before proceeding.
 		{
-			let mut status_lock = self.wallet_polling_status.lock().unwrap();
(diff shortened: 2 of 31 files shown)


src/chain/bitcoind.rs
```diff
@@ -132,7 +133,7 @@ impl BitcoindChainSource {
 		// First register for the wallet polling status to make sure `Node::sync_wallets` calls
 		// wait on the result before proceeding.
 		{
-			let mut status_lock = self.wallet_polling_status.lock().unwrap();
+			let mut status_lock = self.wallet_polling_status.lck();
```
joostjager: I don't like this much, renaming methods on data types that are very famliar.

src/lib.rs
```diff
@@ -1194,7 +1205,9 @@ impl Node {
 
 		let push_msat = push_to_counterparty_msat.unwrap_or(0);
 		let user_channel_id: u128 = u128::from_ne_bytes(
-			self.keys_manager.get_secure_random_bytes()[..16].try_into().unwrap(),
+			self.keys_manager.get_secure_random_bytes()[..16]
+				.try_into()
+				.expect("a 16-byte slice should convert into a [u8; 16]"),
```
joostjager, line 1219: I think this belongs in a later commit?

src/types.rs
```diff
@@ -578,9 +577,9 @@ impl From<LdkChannelDetails> for ChannelDetails {
 			channel_value_sats: value.channel_value_satoshis,
 			unspendable_punishment_reserve: value.unspendable_punishment_reserve,
 			user_channel_id: UserChannelId(value.user_channel_id),
-			// unwrap safety: This value will be `None` for objects serialized with LDK versions
-			// prior to 0.0.115.
-			feerate_sat_per_1000_weight: value.feerate_sat_per_1000_weight.unwrap(),
+			feerate_sat_per_1000_weight: value
+				.feerate_sat_per_1000_weight
+				.expect("LDK objects that are serialized on v0.0.115+ (as ldk-node v0.1 already depended on v0.0.115)"),
```
joostjager: phrasing similar to the below might be clearer

"should always be present ... etc"

.github/workflows/rust.yml
```diff
@@ -81,6 +81,14 @@ jobs:
       - name: Check release build with UniFFI support on Rust ${{ matrix.toolchain }}
         if: matrix.build-uniffi
         run: cargo check --release --features uniffi --verbose --color always
+      - name: Ban unwrap in library code on Rust ${{ matrix.toolchain }}
+        if: matrix.build-uniffi
+        env:
+          RUSTFLAGS: ""
```
joostjager: Comment needed why this is cleared?

.github/workflows/rust.yml
```diff
@@ -81,6 +81,14 @@ jobs:
...
+        run: |
+          rustup component add clippy
+          cargo clippy --lib --verbose --color always -- -A warnings -D clippy::unwrap_used -A clippy::tabs_in_doc_comments
```
joostjager: It was unclear to me why this runs under if: matrix.build-uniffi? Perhaps it can be a separate step

joostjager (commented):
Small comments only.

src/chain/bitcoind.rs
tnull: Yeah, I usually follow the same train of thought, however, in this case it also leads us to the anti-pattern of sprinkling `unwrap` everywhere. We could replace them with `except("lock")` or something like this, but this would make it even more verbose everywhere. Hence why I opted to go this way, which even reduces clutter everywhere. 

Let me know if you deem this a blocker though, happy to revert that commit and go the cluttered route if you prefer.

src/types.rs
tnull: Hmm, not sure, for `except` it's usually preferable to keep it very brief and only state the expectation.

src/chain/bitcoind.rs
joostjager: Would it be possible to do add-on a lock-specific expect method, as a middle road?

Otherwise it might be good to ask a second reviewer for their opinion. This is quite a visible thing.

src/types.rs
joostjager: It doesn't need to be more or fewer words, just the structure of the sentence. I think you'd expect it structured indeed as the actual expectation? "LDK objects that are serialized on v0.0.115+" isn't an expectation?

src/chain/bitcoind.rs
tnull: > Would it be possible to do add-on a lock-specific expect method, as a middle road?

I'm not quite sure I understand what you mean by that? Anything different than `expect("lock")` I mentioned above?

src/types.rs
tnull: > "LDK objects that are serialized on v0.0.115+" isn't an expectation?

Expecting LDK objects that are serialized (..) is though? Anyways, now switched back to ~the old format to keep moving.

.github/workflows/rust.yml
tnull: Moved to `linting` job.

.github/workflows/rust.yml
tnull: It's gone.

src/lib.rs
tnull: These agents are really not good with git, are they? Fixed (and a few other places)

src/types.rs
joostjager: Ah damn, yes, you can indeed read it like so too. Sorry

src/chain/bitcoind.rs
joostjager: I was thinking something like

```
 trait LockResultExt<T> {
      fn expect_not_poisoned(self) -> T;
  }

  impl<T> LockResultExt<MutexGuard<'_, T>> for LockResult<MutexGuard<'_, T>> {
      fn expect_not_poisoned(self) -> MutexGuard<'_, T> {
          self.expect("mutex poisoned")
      }
  }

  impl<T> LockResultExt<RwLockReadGuard<'_, T>> for LockResult<RwLockReadGuard<'_, T>> {
      fn expect_not_poisoned(self) -> RwLockReadGuard<'_, T> {
          self.expect("rwlock poisoned")
      }
  }

  impl<T> LockResultExt<RwLockWriteGuard<'_, T>> for LockResult<RwLockWriteGuard<'_, T>> {
      fn expect_not_poisoned(self) -> RwLockWriteGuard<'_, T> {
          self.expect("rwlock poisoned")
      }
  }
```
  Usage: self.wallet_polling_status.lock().expect_not_poisoned()

tnull: Decidedly didn't go this way to keep it as brief as possible. Anyways, now switched to use `expect` for locks also.

joostjager: Don't mind the duplication of the panic string really, only wanted to suggest something more in the direction of what you had initially.

src/logger.rs
```diff
@@ -424,6 +424,6 @@ mod tests {
 
 		writer.log(record);
 
-		assert_eq!(*log.lock().unwrap(), "Test message (ch:abcdef p:02abcd)");
+		assert_eq!(*log.lock().expect("lock"), "Test message (ch:abcdef p:02abcd)");
```
joostjager, line 427: I don't think the test code is passed through the linter. It doesn't hurt, but also doesn't add much here.

joostjager (approved):

PR #872, opened 2026-04-10
Safely migrate to FileSystemStoreV2

benthecarman:
Before moving to PaginatedKVStore everywhere we need to use FileSystemStoreV2 instead of FileSystemStoreV1. This will safely migrate over to it on first start up. Also adds a test to make sure we handle it properly.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -39,7 +39,6 @@ use lightning::util::persist::{
 };
 use lightning::util::ser::ReadableArgs;
 use lightning::util::sweep::OutputSweeper;
-use lightning_persister::fs_store::v1::FilesystemStore;
 use vss_client::headers::VssHeaderProvider;
 
 use crate::chain::ChainSource;
@@ -55,9 +54,9 @@ use crate::fee_estimator::OnchainFeeEstimator;
 use crate::gossip::GossipSource;
 use crate::io::sqlite_store::SqliteStore;
 use crate::io::utils::{
-	read_event_queue, read_external_pathfinding_scores_from_cache, read_network_graph,
-	read_node_metrics, read_output_sweeper, read_payments, read_peer_info, read_pending_payments,
-	read_scorer, write_node_metrics,
+	open_or_migrate_fs_store, read_event_queue, read_external_pathfinding_scores_from_cache,
+	read_network_graph, read_node_metrics, read_output_sweeper, read_payments, read_peer_info,
+	read_pending_payments, read_scorer, write_node_metrics,
 };
 use crate::io::vss_store::VssStoreBuilder;
 use crate::io::{
@@ -629,15 +628,22 @@ impl NodeBuilder {
 		self.build_with_store(node_entropy, kv_store)
 	}
 
-	/// Builds a [`Node`] instance with a [`FilesystemStore`] backend and according to the options
+	/// Builds a [`Node`] instance with a [`FilesystemStoreV2`] backend and according to the options
 	/// previously configured.
+	///
+	/// If the storage directory contains data from a v1 filesystem store, it will be
+	/// automatically migrated to the v2 format.
+	///
+	/// [`FilesystemStoreV2`]: lightning_persister::fs_store::v2::FilesystemStoreV2
 	pub fn build_with_fs_store(&self, node_entropy: NodeEntropy) -> Result<Node, BuildError> {
 		let mut storage_dir_path: PathBuf = self.config.storage_dir_path.clone().into();
 		storage_dir_path.push("fs_store");
 
 		fs::create_dir_all(storage_dir_path.clone())
 			.map_err(|_| BuildError::StoragePathAccessFailed)?;
-		let kv_store = FilesystemStore::new(storage_dir_path);
+
+		let kv_store = open_or_migrate_fs_store(storage_dir_path)?;
+
 		self.build_with_store(node_entropy, kv_store)
 	}
 
@@ -1087,7 +1093,7 @@ impl ArcedNodeBuilder {
 		self.inner.read().unwrap().build(*node_entropy).map(Arc::new)
 	}
 
-	/// Builds a [`Node`] instance with a [`FilesystemStore`] backend and according to the options
+	/// Builds a [`Node`] instance with a [`FilesystemStoreV2`] backend and according to the options
 	/// previously configured.
 	pub fn build_with_fs_store(
 		&self, node_entropy: Arc<NodeEntropy>,
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -43,7 +43,6 @@ use lightning::util::persist::{
 use lightning::util::ser::ReadableArgs;
 use lightning::util::sweep::OutputSweeper;
 use lightning_dns_resolver::OMDomainResolver;
-use lightning_persister::fs_store::v1::FilesystemStore;
 use vss_client::headers::VssHeaderProvider;
 
 use crate::chain::ChainSource;
@@ -59,8 +58,9 @@ use crate::fee_estimator::OnchainFeeEstimator;
 use crate::gossip::GossipSource;
 use crate::io::sqlite_store::SqliteStore;
 use crate::io::utils::{
-	read_all_objects, read_event_queue, read_external_pathfinding_scores_from_cache,
-	read_network_graph, read_node_metrics, read_output_sweeper, read_peer_info, read_scorer,
+	open_or_migrate_fs_store, read_all_objects, read_event_queue,
+	read_external_pathfinding_scores_from_cache, read_network_graph, read_node_metrics,
+	read_output_sweeper, read_peer_info, read_scorer,
 };
 use crate::io::vss_store::VssStoreBuilder;
 use crate::io::{
@@ -644,18 +644,19 @@ impl NodeBuilder {
 		self.build_with_store_and_logger(node_entropy, kv_store, logger)
 	}
 
-	/// Builds a [`Node`] instance with a [`FilesystemStore`] backend and according to the options
+	/// Builds a [`Node`] instance with a [`FilesystemStoreV2`] backend and according to the options
 	/// previously configured.
+	///
+	/// If the storage directory contains data from a v1 filesystem store, it will be
+	/// automatically migrated to the v2 format.
+	///
+	/// [`FilesystemStoreV2`]: lightning_persister::fs_store::v2::FilesystemStoreV2
 	pub fn build_with_fs_store(&self, node_entropy: NodeEntropy) -> Result<Node, BuildError> {
 		let logger = setup_logger(&self.log_writer_config, &self.config)?;
 		let mut storage_dir_path: PathBuf = self.config.storage_dir_path.clone().into();
 		storage_dir_path.push("fs_store");
 
-		fs::create_dir_all(storage_dir_path.clone()).map_err(|e| {
-			log_error!(logger, "Failed to setup Filesystem store: {}", e);
-			BuildError::StoragePathAccessFailed
-		})?;
-		let kv_store = FilesystemStore::new(storage_dir_path);
+		let kv_store = open_or_migrate_fs_store(storage_dir_path)?;
 		self.build_with_store_and_logger(node_entropy, kv_store, logger)
 	}
 
@@ -1115,7 +1116,7 @@ impl ArcedNodeBuilder {
 		self.inner.read().expect("lock").build(*node_entropy).map(Arc::new)
 	}
 
-	/// Builds a [`Node`] instance with a [`FilesystemStore`] backend and according to the options
+	/// Builds a [`Node`] instance with a [`FilesystemStoreV2`] backend and according to the options
 	/// previously configured.
 	pub fn build_with_fs_store(
 		&self, node_entropy: Arc<NodeEntropy>,
--- a/src/io/utils.rs
+++ b/src/io/utils.rs
@@ -10,7 +10,7 @@ use std::io::Write;
 use std::ops::Deref;
 #[cfg(unix)]
 use std::os::unix::fs::OpenOptionsExt;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 use bdk_chain::indexer::keychain_txout::ChangeSet as BdkIndexerChangeSet;
@@ -26,14 +26,16 @@ use lightning::routing::scoring::{
 	ChannelLiquidities, ProbabilisticScorer, ProbabilisticScoringDecayParameters,
 };
 use lightning::util::persist::{
-	KVStore, KVStoreSync, KVSTORE_NAMESPACE_KEY_ALPHABET, KVSTORE_NAMESPACE_KEY_MAX_LEN,
-	NETWORK_GRAPH_PERSISTENCE_KEY, NETWORK_GRAPH_PERSISTENCE_PRIMARY_NAMESPACE,
-	NETWORK_GRAPH_PERSISTENCE_SECONDARY_NAMESPACE, OUTPUT_SWEEPER_PERSISTENCE_KEY,
-	OUTPUT_SWEEPER_PERSISTENCE_PRIMARY_NAMESPACE, OUTPUT_SWEEPER_PERSISTENCE_SECONDARY_NAMESPACE,
-	SCORER_PERSISTENCE_KEY, SCORER_PERSISTENCE_PRIMARY_NAMESPACE,
-	SCORER_PERSISTENCE_SECONDARY_NAMESPACE,
+	migrate_kv_store_data, KVStore, KVStoreSync, KVSTORE_NAMESPACE_KEY_ALPHABET,
+	KVSTORE_NAMESPACE_KEY_MAX_LEN, NETWORK_GRAPH_PERSISTENCE_KEY,
+	NETWORK_GRAPH_PERSISTENCE_PRIMARY_NAMESPACE, NETWORK_GRAPH_PERSISTENCE_SECONDARY_NAMESPACE,
+	OUTPUT_SWEEPER_PERSISTENCE_KEY, OUTPUT_SWEEPER_PERSISTENCE_PRIMARY_NAMESPACE,
+	OUTPUT_SWEEPER_PERSISTENCE_SECONDARY_NAMESPACE, SCORER_PERSISTENCE_KEY,
+	SCORER_PERSISTENCE_PRIMARY_NAMESPACE, SCORER_PERSISTENCE_SECONDARY_NAMESPACE,
 };
 use lightning::util::ser::{Readable, ReadableArgs, Writeable};
+use lightning_persister::fs_store::v1::FilesystemStore;
+use lightning_persister::fs_store::v2::FilesystemStoreV2;
 use lightning_types::string::PrintableString;
 
 use super::*;
@@ -48,7 +50,7 @@ use crate::payment::PendingPaymentDetails;
 use crate::peer_store::PeerStore;
 use crate::types::{Broadcaster, DynStore, KeysManager, Sweeper};
 use crate::wallet::ser::{ChangeSetDeserWrapper, ChangeSetSerWrapper};
-use crate::{Error, EventQueue, NodeMetrics, PaymentDetails};
+use crate::{BuildError, Error, EventQueue, NodeMetrics, PaymentDetails};
 
 pub const EXTERNAL_PATHFINDING_SCORES_CACHE_KEY: &str = "external_pathfinding_scores_cache";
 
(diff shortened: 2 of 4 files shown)


src/builder.rs
```diff
@@ -2007,6 +2004,40 @@ fn build_with_store_internal(
 	})
 }
 
+/// Opens a [`FilesystemStoreV2`], automatically migrating from v1 format if necessary.
+///
+/// If the directory contains v1 data (files at the top level), the data is migrated to v2 format
+/// in a temporary directory, the original is renamed to `fs_store_v1_backup`, and the migrated
+/// directory is moved into place.
+fn open_or_migrate_fs_store(storage_dir_path: PathBuf) -> Result<FilesystemStoreV2, BuildError> {
```
tnull: Let's move this to `io/utils.rs`

src/builder.rs
```diff
@@ -2007,6 +2004,40 @@ fn build_with_store_internal(
...
+	match FilesystemStoreV2::new(storage_dir_path.clone()) {
+		Ok(store) => Ok(store),
+		Err(e) if e.kind() == std::io::ErrorKind::InvalidData => {
+			// The directory contains v1 data, migrate to v2.
```
tnull: Huh, how do we know `io::ErrorKind::InvalidData` corresponds to `v1`? Couldn't it have other reasons (eg if users misconfigured something) and we might be screwing with users data here?

src/builder.rs
```diff
@@ -2007,6 +2004,40 @@ fn build_with_store_internal(
...
+				.map_err(|_| BuildError::KVStoreSetupFailed)?;
+
+			migrate_kv_store_data(&mut v1_store, &mut v2_store)
+				.map_err(|_| BuildError::KVStoreSetupFailed)?;
+
+			// Swap directories: rename v1 out of the way, move v2 into place.
+			let mut backup_dir = storage_dir_path.clone();
+			backup_dir.set_file_name("fs_store_v1_backup");
+			fs::rename(&storage_dir_path, &backup_dir)
+				.map_err(|_| BuildError::KVStoreSetupFailed)?;
```
tnull: Note that if we crash here the user can't recover. Can we somehow make the two renames part of the same `fsync`/metadata operation? Not sure?

tests/integration_tests_rust.rs
```diff
@@ -2461,14 +2458,18 @@ async fn build_0_6_2_node(
 
 async fn build_0_7_0_node(
 	bitcoind: &BitcoinD, electrsd: &ElectrsD, storage_path: String, esplora_url: String,
-	seed_bytes: [u8; 64],
+	seed_bytes: [u8; 64], use_fs_store: bool,
```
tnull: No, please have this take and use `TestConfig` for stuff like that, all the bool flags in test code are getting hard to read.

tnull (commented):
Hmm, I have to say for the file system store I am a bit hesitant to just kick-off auto-migration like that, given users might have messed with the store in some manual fashion (def. had users do that before)?

src/builder.rs
benthecarman: The only time we throw the error is when we detect a v1 fs store. We shouldn't ever encounter it otherwise, it only ever thrown by the std lib when trying to parse things like strings (checking for utf8 encoding), since we just read raw bytes we should be safe.

src/builder.rs
benthecarman: I think the only real way would be to implement some recovery logic. Basically if the dir doesn't exist but `fs_store_v2_migrating` does, we do the rename and continue.

tests/integration_tests_rust.rs
benthecarman: fixed

benthecarman:
> Hmm, I have to say for the file system store I am a bit hesitant to just kick-off auto-migration like that, given users might have messed with the store in some manual fashion (def. had users do that before)?

Yeah I understand your hesitancy, however I am not really sure what else would be the best way forward if we are going to require the `PaginatedKvStore`. Otherwise we would just have this function fail with old ones and the dev/user would have to figure out how to migrate themselves. Maybe that's preferable so people move to sqlite?

src/builder.rs
tnull: > The only time we throw the error is when we detect a v1 fs store.

Hmm, right, that's the current state, but it's still bad to lean on this, as we might very well change that behavior in the future? Do we need to change the `new` API to return a `FilesystemStore`-specific error type with a `V1DataFound` error variant?

src/builder.rs
tnull: Hmm, yeah, might be good to at least add a test that checks we can safely retry and finish the migration after restart if we somehow crash/abort. AFAIU, this is currently not the case, i.e., we wouldn't actually continue/retry.

tnull:
> Yeah I understand your hesitancy, however I am not really sure what else would be the best way forward if we are going to require the `PaginatedKvStore`. 

Well, as you know I was in favor of simply deprecating `FilesystemStore`, but it seems that ship has sailed. We could have also deprecated support in LDK Node, but given that we now want to ship this for the next release, just ripping it out for 0.8 seems like a really short notice for any users that might be out there, so likely we don't want to do that.

> Otherwise we would just have this function fail with old ones and the dev/user would have to figure out how to migrate themselves. Maybe that's preferable so people move to sqlite?

Yeah, not entirely sure what's preferable yet, but it might be an option.

src/builder.rs
tnull: @benthecarman Any update here? Seems like we might need to still land the API change (explicit error variant) before the LDK v0.3 cutoff?

benthecarman: I thought I wrote a comment but I guess not. I researched and looked, the only way `InvalidData` is thrown is where we throw it explicitly. The only way for the normal std lib to throw is it when its parsing data, not just reading bytes like how we do so we should be safe. But yeah if that guarantee isn't enough, can create an error type for this.

tnull: > I thought I wrote a comment but I guess not. I researched and looked, the only way InvalidData is thrown is where we throw it explicitly. 

Right, but that's not part of any API contract, nor documented. So we might very well change it at some point going forward. IMO it therefore would be preferable to at the very least clearly document this behavior or introduce a dedicated error type *before* we can safely lean on it here.

benthecarman: made https://github.com/lightningdevkit/rust-lightning/pull/4573

tnull:

benthecarman:
sorry busy with conference, found time to get to this one

tnull:
No worries, for this one we'll need to wait for the splice-related changes to go in first anyways, see failing CI.

tnull (commented):
This needs a rebase by now.

src/payment/unified.rs
```diff
@@ -256,6 +256,7 @@ impl UnifiedPayment {
 			PaymentMethod::LightningBolt12(_) => 0,
 			PaymentMethod::LightningBolt11(_) => 1,
 			PaymentMethod::OnChain(_) => 2,
+			PaymentMethod::Cashu(_) => 3,
```
tnull: Seems you also bumped `bitcoin-payment-instructions` base with this? That's great, but let's make it a separate commit please.

src/wallet/mod.rs
```diff
@@ -955,11 +955,41 @@ impl Wallet {
...
+		// the new funding output, every sat BDK over-reserves gets funneled into
+		// the channel instead of staying in change. Absorb the worst-case
+		// over-reservation by bumping the change output up here.
+		//
+		// Two sources of rounding contribute:
+		//
+		//  * Per-input ceilings — `bdk_wallet`'s `select_sorted_utxos` ceilings
+		//    `fee_rate * (TxIn::default().segwit_weight() + satisfaction_weight)`
+		//    once per input. With `N` inputs, the worst-case over-reservation is
+		//    `N - 1` sats. TODO: drop once
```
tnull: Please put this TODO on a dedicated line. (probably at the start or end of the comment).

Also, can we create an issue for this so we don't forget to drop this?

src/io/utils.rs
```diff
@@ -619,6 +621,42 @@ pub(crate) fn read_bdk_wallet_change_set(
...
+			let mut v2_store = FilesystemStoreV2::new(v2_dir.clone())
+				.map_err(|_| BuildError::KVStoreSetupFailed)?;
+
+			migrate_kv_store_data(&mut v1_store, &mut v2_store)
+				.map_err(|_| BuildError::KVStoreSetupFailed)?;
+
+			// Swap directories: rename v1 out of the way, move v2 into place.
+			let mut backup_dir = storage_dir_path.clone();
+			backup_dir.set_file_name("fs_store_v1_backup");
+			fs::rename(&storage_dir_path, &backup_dir)
```
tnull, line 656: Codex:

>   - P1 /tmp/ldk-node-pr872/src/io/utils.rs:650, /tmp/ldk-node-pr872/src/builder.rs:659: the v1 to v2 migration is not crash-recoverable. After fs_store is renamed to fs_store_v1_backup, a crash or error before fs_store_v2_migrating is renamed into place leaves no fs_store. On the next startup,
    build_with_fs_store recreates an empty fs_store, FilesystemStoreV2::new succeeds, and the node can start against empty storage while the real data sits in backup/temp dirs. The migration helper should detect and recover incomplete states before creating a fresh store.

Seems we still might want to double-check we survive an ill-timed crash?

src/wallet/mod.rs
```diff
@@ -955,11 +955,41 @@ impl Wallet {
...
+		//    https://github.com/bitcoindevkit/bdk_wallet/pull/479 lands in a
+		//    release.
+		//
+		//  * Initial-tx and change ceilings — `create_tx` ceilings
+		//    `fee_rate * tx.weight()` on the recipient-only tx, and `decide_change`
+		//    ceilings `fee_rate * Weight::from_vb(drain_output_len)`. Two more
+		//    independent ceilings, contributing up to 2 sats.
+		if let Some(ref mut change) = change_output {
+			let num_inputs = (confirmed_utxos.len() + must_spend.len()) as u64;
+			let per_input_rounding_surplus_sat = num_inputs.saturating_sub(1);
```
tnull: Codex:    
>  - P2 /tmp/ldk-node-pr872/src/wallet/mod.rs:986: the splice fee rounding adjustment adds the worst-case surplus to the change output instead of the actual BDK over-reservation. For example, at 1000 sat/kwu the per-component fee math is exact, so the actual rounding surplus is 0, but this
    still adds N + 1 sats to change. Since LDK derives value_added from inputs minus outputs/change/fees, splice_in(..., amount) can silently add less than the requested amount. This should compute the actual fee delta from the selected inputs/outputs, or cap the change bump to that delta.

tnull (commented):
Thanks, I'll ping @jkczyz as a secondary reviewer on the splice-related changes.

src/payment/unified.rs
jkczyz: I'd prefer we only bump that if needed. Just makes things more difficult for others updating the `lightning` crate in `bitcoin-payment-instructions`.

src/wallet/mod.rs
benthecarman: Ended up just removing this extra handling its only for an extra 2 sats, complex, and potentially error prone.

src/io/utils.rs
benthecarman: Added migration defense and tests for all scenarios

tests/integration_tests_rust.rs
```diff
@@ -2549,15 +2549,19 @@ async fn build_0_6_2_node(
...
 	builder_old.set_network(bitcoin::Network::Regtest);
-	builder_old.set_storage_dir_path(storage_path);
+	builder_old.set_storage_dir_path(config.node_config.storage_dir_path.clone());
 	builder_old.set_entropy_seed_bytes(seed_bytes);
 	builder_old.set_chain_source_esplora(esplora_url, None);
-	let node_old = builder_old.build().unwrap();
+	let node_old = match config.store_type {
+		TestStoreType::FilesystemStore => builder_old.build_with_fs_store().unwrap(),
+		TestStoreType::Sqlite => builder_old.build().unwrap(),
+		TestStoreType::TestSyncStore => panic!("TestSyncStore not supported in v0.7.0 builder"),
```
tnull, line 2570: Hmm, bit odd to switch to the `TestConfig` to then not support it. Can't we still setup `TestSyncStore` here?

src/io/utils.rs
```diff
@@ -619,10 +621,103 @@ pub(crate) fn read_bdk_wallet_change_set(
...
+) -> Result<FilesystemStoreV2, BuildError> {
+	let parent_dir = storage_dir_path.parent().ok_or(BuildError::StoragePathAccessFailed)?;
+	fs::create_dir_all(parent_dir).map_err(|_| BuildError::StoragePathAccessFailed)?;
+	recover_incomplete_fs_store_migration(&storage_dir_path)?;
+	if !storage_dir_path.exists() {
+		fs::create_dir_all(storage_dir_path.clone())
+			.map_err(|_| BuildError::StoragePathAccessFailed)?;
+	}
+
+	match FilesystemStoreV2::new(storage_dir_path.clone()) {
```
tnull, line 640: >   - [P2] Detect v1 empty-secondary namespace data — /home/tnull/workspace/ldk-node/src/io/utils.rs:640-641
>     When an old fs store only has v1 entries under a non-empty primary namespace with an empty secondary namespace, such as fs_store/bdk_wallet/descriptor before any top-level LDK files have been persisted, FilesystemStoreV2::new succeeds because it only rejects top-level files. Startup then
>     skips migration and V2 looks under bdk_wallet/[empty]/descriptor, silently ignoring the existing wallet state and potentially missing historical funds; please detect these direct namespace files and migrate them too.

src/io/utils.rs
```diff
@@ -619,10 +621,103 @@ pub(crate) fn read_bdk_wallet_change_set(
...
+		Err(FilesystemStoreV2Error::V1DataDetected(_)) => {
+			// The directory contains v1 data, migrate to v2.
+			let mut v1_store = FilesystemStore::new(storage_dir_path.clone());
+
+			let v2_dir = fs_store_sibling_path(&storage_dir_path, "fs_store_v2_migrating");
+			fs::create_dir_all(v2_dir.clone()).map_err(|_| BuildError::StoragePathAccessFailed)?;
+			let mut v2_store = FilesystemStoreV2::new(v2_dir.clone())
+				.map_err(|_| BuildError::KVStoreSetupFailed)?;
+
+			migrate_kv_store_data(&mut v1_store, &mut v2_store)
```
tnull, line 651: > - [P2] Ignore stale temp files before migrating — /home/tnull/workspace/ldk-node/src/io/utils.rs:651-652
>     If a v1 filesystem store contains a leftover *.tmp file from a crash during a prior write, this migration path will fail: migrate_kv_store_data calls the v1 store's exhaustive listing, which skips temp files as keys but then tries to recurse into them as directories. The old store would
>     otherwise ignore those files, so a crash-safe v1 store can become unable to start after upgrade unless temp/trash files are cleaned or ignored before migration.

src/lib.rs
```diff
@@ -1699,15 +1699,8 @@ impl Node {
...
-				.runtime
-				.block_on(funding_template.splice_out(
-					outputs,
-					min_feerate,
-					max_feerate,
-					Arc::clone(&self.wallet),
-				))
-				.map_err(|e| {
+			let contribution =
+				funding_template.splice_out(outputs, min_feerate, max_feerate).map_err(|e| {
```
tnull, line 1719: >  - [P2] Preserve exact splice amounts with prior contributions — /home/tnull/workspace/ldk-node/src/lib.rs:1702-1703
>     With the updated rust-lightning API, FundingTemplate::splice_out amends any prior_contribution by appending these outputs rather than replacing the pending request. If a user retries or replaces a pending splice-out, this can withdraw the previous output plus the new splice_amount_sats, so
>     the public Node API removes more than requested; opt out with without_prior_contribution or explicitly account for the prior contribution.

tests/integration_tests_rust.rs
benthecarman: We can but we have to impl the ldk-node 0.7 version of the kv store traits for `TestSyncStore`. We only use ldk-node 0.7 to test upgrades from prior storage versions so it wouldn't be very useful and it'd be pretty verbose so elected not to

src/io/utils.rs
benthecarman: `migrate_kv_store_data` does this already

src/io/utils.rs
benthecarman: https://github.com/lightningdevkit/rust-lightning/pull/4618

src/lib.rs
jkczyz: FWIW, we should do these updates in separate PRs, IMO, especially if there are behavioral changes to consider. This is also related to #888, though it can probably be done independently.

That said, we need to decide what API we want to expose to the user. If they already initiated a splice, then should we allow them to add / remove more in a subsequent call if the first hasn't locked? In LDK, doing so results in RBF'ing the splice with the added / remove contributions. However, even if `without_prior_contribution` is used (which was the previous behavior), the first one can still confirm which might be unexpected for the user.

Maybe best to disallow a splice for now if one was already negotiated to unblock you. Otherwise, we need to consider what to do about the portion that didn't lock.

benthecarman: Maybe stick with this for now and decide / fix in #888?

tests/integration_tests_rust.rs
```diff
@@ -1067,7 +1067,9 @@ async fn splice_channel() {
 	expect_channel_ready_event!(node_a, node_b.node_id());
 	expect_channel_ready_event!(node_b, node_a.node_id());
 
-	let expected_splice_in_fee_sat = 255;
+	let expected_splice_in_fee_sat = 251;
+	let expected_splice_in_onchain_cost_sat = 254;
+	let expected_splice_in_lightning_balance_sat = 4_000_003;
```
jkczyz, line 1079: IIUC, the extra three sats will disappear once BDK (and rust-bitcoin?) are upgraded?

src/lib.rs
jkczyz: Could you drop the first two commits and rebase onto https://github.com/jkczyz/ldk-node/tree/2026-05-splice-builder-api? That fails if we already have a splice and makes my life a little easier in #888 for the Cargo.toml changes.

benthecarman: done

tests/integration_tests_rust.rs
benthecarman: I don't think it'll get rid of all 3 but maybe 1 or 2. Some of them aren't really avoidable in bdk with how things are structured

jkczyz: Was there something to do with the segwit marker that would need a `rust-bitcoin` update, too?

benthecarman: No that one was a Claude hallucination sadly. rust-bitcoin already treats empty inputs as segwit

tests/integration_tests_rust.rs
tnull: Okay, yeah, is probably fine for now.

benthecarman:
rebased on to latest `main` with the ldk updates now merged in

src/io/utils.rs
tnull: Well, no, the first files we create are the ones persisted by BDK, which all go in `bdk_wallet`. The `FilesystemStoreV2::new` logic only detects v1 if there are files on the top level, so wouldn't trigger migration of a wallet that was created but never started, i.e., didn't hit a first persistence of channelmanager, AFAIU.

src/io/utils.rs
```diff
@@ -619,10 +621,103 @@ pub(crate) fn read_bdk_wallet_change_set(
...
+				.map_err(|_| BuildError::KVStoreSetupFailed)?;
+
+			// Swap directories: rename v1 out of the way, move v2 into place.
+			let backup_dir = fs_store_sibling_path(&storage_dir_path, "fs_store_v1_backup");
+			fs::rename(&storage_dir_path, &backup_dir)
+				.map_err(|_| BuildError::KVStoreSetupFailed)?;
+			fs::rename(&v2_dir, &storage_dir_path).map_err(|_| BuildError::KVStoreSetupFailed)?;
```
tnull, line 658: nit: Can we add an `fsync` here?

tnull (commented):
Looks good, mod https://github.com/lightningdevkit/ldk-node/pull/872#discussion_r3328348134 mostly, which we probably still want to address.

src/io/utils.rs
benthecarman: Okay will fix upstream

benthecarman: https://github.com/lightningdevkit/rust-lightning/pull/4649

src/io/utils.rs
benthecarman: added

tnull (approved):

PR #873, opened 2026-04-11
Add tests and improve documentation for zero reserve channels

tankyleo:
Follow-up to #853

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1449,8 +1449,8 @@ impl Node {
 	/// Connect to a node and open a new unannounced channel, in which the target node can
 	/// spend its entire balance.
 	///
-	/// This channel allows the target node to try to steal your funds with no financial
-	/// penalty, so this channel should only be opened to nodes you trust.
+	/// This channel allows the target node to try to steal your channel balance with no
+	/// financial penalty, so this channel should only be opened to nodes you trust.
 	///
 	/// Disconnects and reconnects are handled automatically.
 	///
@@ -1484,8 +1484,8 @@ impl Node {
 	/// minus fees and anchor reserves. The target node will be able to spend its entire channel
 	/// balance.
 	///
-	/// This channel allows the target node to try to steal your funds with no financial
-	/// penalty, so this channel should only be opened to nodes you trust.
+	/// This channel allows the target node to try to steal your channel balance with no
+	/// financial penalty, so this channel should only be opened to nodes you trust.
 	///
 	/// Disconnects and reconnects are handled automatically.
 	///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1149,7 +1149,7 @@ impl Node {
 	fn open_channel_inner(
 		&self, node_id: PublicKey, address: SocketAddress, channel_amount_sats: FundingAmount,
 		push_to_counterparty_msat: Option<u64>, channel_config: Option<ChannelConfig>,
-		announce_for_forwarding: bool, set_0reserve: bool,
+		announce_for_forwarding: bool, disable_counterparty_reserve: bool,
 	) -> Result<UserChannelId, Error> {
 		if !*self.is_running.read().expect("lock") {
 			return Err(Error::NotRunning);
@@ -1219,7 +1219,7 @@ impl Node {
 				.expect("a 16-byte slice should convert into a [u8; 16]"),
 		);
 
-		let result = if set_0reserve {
+		let result = if disable_counterparty_reserve {
 			self.channel_manager.create_channel_to_trusted_peer_0reserve(
 				peer_info.node_id,
 				channel_amount_sats,
@@ -1239,7 +1239,7 @@ impl Node {
 			)
 		};
 
-		let zero_reserve_string = if set_0reserve { "0reserve " } else { "" };
+		let zero_reserve_string = if disable_counterparty_reserve { "0reserve " } else { "" };
 
 		match result {
 			Ok(_) => {
--- a/src/liquidity.rs
+++ b/src/liquidity.rs
@@ -142,9 +142,13 @@ pub struct LSPS2ServiceConfig {
 	///
 	/// [`bLIP-52`]: https://github.com/lightning/blips/blob/master/blip-0052.md#trust-models
 	pub client_trusts_lsp: bool,
-	/// When set, clients that we open channels to will be allowed to spend their entire channel
-	/// balance. This allows clients to try to steal your funds with no financial penalty, so
-	/// this should only be set if you trust your clients.
+	/// When set, we will allow clients to spend their entire channel balance in the channels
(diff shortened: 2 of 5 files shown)


tankyleo:
Let me know how this sounds, is that the testing you had in mind ?

benthecarman:
vss tests don't compile but yeah lgtm

tankyleo:
Fixed the VSS tests, and reworked the documentation wording

tests/common/mod.rs
```diff
@@ -846,15 +846,27 @@ pub(crate) async fn do_channel_full_cycle<E: ElectrumApi>(
...
-			node_b.node_id(),
-			node_b.listening_addresses().unwrap().first().unwrap().clone(),
-			funding_amount_sat,
-			Some(push_msat),
-			None,
-		)
-		.unwrap();
+	if allow_0reserve {
+		node_a
+			.open_0reserve_channel(
```
tnull, line 851: That's fine, but should we also assert that we can spend the full channel balance now that both coutnerparties set 0reserve for each other?

src/lib.rs
```diff
@@ -1449,8 +1449,8 @@ impl Node {
 	/// Connect to a node and open a new unannounced channel, in which the target node can
 	/// spend its entire balance.
 	///
-	/// This channel allows the target node to try to steal your funds with no financial
-	/// penalty, so this channel should only be opened to nodes you trust.
+	/// The target node may have nothing at stake if it tries to steal your funds in this
+	/// channel, so only open this channel to nodes you trust.
```
tnull: That 'only open this channel to nodes you trust' part is odd. There will only be one channel, there isn't the potential to open this one channel to multiple nodes/other nodes, just the one we're opening to...

FWIW, I preferred the previous version honestly.

tests/common/mod.rs
tankyleo: thank you see below. A caveat is that we don't have an API yet for the opener of the channel to get 0-reserve.

tankyleo:
Reverted comment to the previous version, and added some coverage on full channel balance sends.

src/lib.rs
```diff
@@ -1449,8 +1449,8 @@ impl Node {
 	/// Connect to a node and open a new unannounced channel, in which the target node can
 	/// spend its entire balance.
 	///
-	/// This channel allows the target node to try to steal your funds with no financial
-	/// penalty, so this channel should only be opened to nodes you trust.
+	/// This channel allows the target node to try to steal your funds in that channel with no
```
tnull: nit: Please avoid the 'this channel (..) that channel' duplication (here and below).

src/liquidity.rs
```diff
@@ -142,9 +142,9 @@ pub struct LSPS2ServiceConfig {
 	///
 	/// [`bLIP-52`]: https://github.com/lightning/blips/blob/master/blip-0052.md#trust-models
 	pub client_trusts_lsp: bool,
-	/// When set, clients that we open channels to will be allowed to spend their entire channel
-	/// balance. This allows clients to try to steal your funds with no financial penalty, so
-	/// this should only be set if you trust your clients.
+	/// When set, we will allow clients to spend their entire channel balance in the channels
+	/// we open to them. This allows clients to try to steal your funds in those channels with
```
tnull: Maybe just say 'your channel balance' rather than funds, which should allow to drop the 'in those channels' part? (also note it should be 'which', not 'with' in the current version, but probably better to drop that altogether).

tests/common/mod.rs
```diff
@@ -1267,6 +1295,37 @@ pub(crate) async fn do_channel_full_cycle<E: ElectrumApi>(
 		2
 	);
 
+	if allow_0reserve {
+		let node_a_outbound_capacity_msat = node_a.list_channels()[0].outbound_capacity_msat;
+		let node_a_reserve_msat =
+			node_a.list_channels()[0].unspendable_punishment_reserve.unwrap() * 1000;
+		// TODO: Update this for zero-fee commitment channels
```
tnull: Mind opening an issue for this TODO, so we don't forget to actually address it. Could also be a bit more verbose, i.e., what are we waiting for exactly/what do we intend to update?

tnull (commented):
Thanks!

Some more comments. Not quite sure about the doc changes yet. Tbh current version on `main` would be fine by me, but if we want to change them further, it should be a clear improvement.

src/liquidity.rs
tankyleo: you mean "which no financial penalty" ? not sure about that :)

src/lib.rs
tankyleo: right i think we wanted to clarify the target node cannot steal *all* your funds, just those in the channel you opened to that node.

src/liquidity.rs
tankyleo: Right hopefully with "channel balance", people understand this only applies to the funds they have in that particular channel, not the sum of all your channel balances across all your channels.

tests/common/mod.rs
tankyleo: Done here thank you 

https://github.com/lightningdevkit/ldk-node/issues/875

tankyleo:
Took your feedback tnull. If you'd rather drop the documentation commit let me know, I agree we should aim for clear improvements here.

tankyleo:
I renamed the `LSPS2ServiceConfig` field to make it consistent with the current field in https://github.com/lightningdevkit/ldk-server/pull/187

benthecarman (dismissed):

src/liquidity.rs
```diff
@@ -142,10 +142,10 @@ pub struct LSPS2ServiceConfig {
...
 	/// [`bLIP-52`]: https://github.com/lightning/blips/blob/master/blip-0052.md#trust-models
 	pub client_trusts_lsp: bool,
-	/// When set, clients that we open channels to will be allowed to spend their entire channel
-	/// balance. This allows clients to try to steal your funds with no financial penalty, so
-	/// this should only be set if you trust your clients.
-	pub allow_client_0reserve: bool,
+	/// When set, we will allow clients to spend their entire channel balance in the channels
+	/// we open to them. This allows clients to try to steal your channel balance with
+	/// no financial penalty, so this should only be set if you trust your clients.
+	pub disable_client_reserve: bool,
```
tnull, line 152: Hmm, if we really want to rename this field, it would still be good to mention '0-reserve' somewhere, as this is what we use elsewhere in the API to refer to this feature.

tnull (commented):
One comment, otherwise LGTM

src/liquidity.rs
tankyleo: Thank you I added a link to the `Node::open_0reserve_channel` API below. Also made a few small renames to make things consistent.

tnull (approved):

PR #876, opened 2026-04-15
DRY up batched `KVStore` reads utility methods

tnull:
```
DRY up batched KVStore reads with `read_all_objects` helper

The parallel `JoinSet`-based batching loop was duplicated across
`read_payments` and `read_pending_payments`. Extract it into a generic
`read_all_objects<T: Readable>` helper that callers invoke directly
with the relevant namespace constants. Per-type log messages are
preserved via `std::any::type_name::<T>()`.

Co-Authored-By: HAL 9000
```

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -59,9 +59,8 @@ use crate::fee_estimator::OnchainFeeEstimator;
 use crate::gossip::GossipSource;
 use crate::io::sqlite_store::SqliteStore;
 use crate::io::utils::{
-	read_event_queue, read_external_pathfinding_scores_from_cache, read_network_graph,
-	read_node_metrics, read_output_sweeper, read_payments, read_peer_info, read_pending_payments,
-	read_scorer,
+	read_all_objects, read_event_queue, read_external_pathfinding_scores_from_cache,
+	read_network_graph, read_node_metrics, read_output_sweeper, read_peer_info, read_scorer,
 };
 use crate::io::vss_store::VssStoreBuilder;
 use crate::io::{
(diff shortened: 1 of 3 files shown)


Camillarhi:
LGTM

Camillarhi:
This now needs a rebase, there's a merge conflict.

src/builder.rs
joostjager, line 1524: Doesn't this one need to go through the batch store too?

src/io/utils.rs
```diff
@@ -221,81 +221,62 @@ where
...
-				&*kv_store,
-				PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
-				PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
-				&next_key,
-			);
-			set.spawn(fut);
-			debug_assert!(set.len() <= BATCH_SIZE);
-		}
+	for key in keys {
+		set.spawn(KVStore::read(kv_store, primary_namespace, secondary_namespace, &key));
```
joostjager: It is possible that this now creates a large amount of (blocked) tasks?

joostjager (commented):
I am not sure a generic client-side read throttle is the right abstraction here.

I think the motivation is mainly VSS-specific? For local stores such as SQLite, this mostly limits task fan-out rather than real backend concurrency. Perhaps it should be implemented only in the VSS client. And perhaps longer term, a multi-key read would be helpful in the VSS protocol?

tnull:
> I am not sure a generic client-side read throttle is the right abstraction here.

Hmm, I tend to agree, though pre-existing.

> I think the motivation is mainly VSS-specific? 

Well, yes, presumably postgres could also benefit from it, though the store probably should switch to connection pooling anyways?

> For local stores such as SQLite, this mostly limits task fan-out rather than real backend concurrency. Perhaps it should be implemented only in the VSS client. And perhaps longer term, a multi-key read would be helpful in the VSS protocol?

Yeah, maybe for now I should just switch to wrap `VssStore` in `BatchingStore` in this PR?  WDYT?

src/builder.rs
tnull: Ah, yes, I had that change at some point, not sure why it was reverted.

joostjager (commented):
>Yeah, maybe for now I should just switch to wrap VssStore in BatchingStore in this PR? WDYT?

I think this is a good move, and later expand either towards protocol-level multi reads for vss, or perhaps generalize for other backends if the need arises.

tnull:
Discussed offline, decided we're not going with the `BatchingStore` abstraction right now, but only DRY up the `read_` utility methods in `io/utils.rs` in this PR.

src/io/utils.rs
```diff
@@ -221,21 +220,19 @@ where
 	})
 }
 
-/// Read previously persisted payments information from the store.
-pub(crate) async fn read_payments<L: Deref>(
-	kv_store: &DynStore, logger: L,
-) -> Result<Vec<PaymentDetails>, std::io::Error>
+/// Read all objects of type `T` from the given namespace, spawning reads in parallel.
+pub(crate) async fn read_all_objects<T, L>(
```
joostjager, line 224: I briefly had to think of making this an extension method, but don't think it is a great idea.

joostjager (dismissed):
Looks good, uncontentious. Rebase needed.

tnull:
> Looks good, uncontentious. Rebase needed.

Rebased, sorry, forgot about that before.

joostjager (approved):

PR #882, opened 2026-04-20
Add persistent closed channel history and `list_closed_channels()`

### LINKED ISSUE #851: Persist historical channels
When a channel is closed, all metadata is lost. `list_channels()` only returns open channels, and would require the user to do their own after persistence after a `ChannelClosed` event. Applications have no way to answer basic questions after closure: how large was the channel, final balance, closing reason,  funding txo, etc. This makes it difficult to do proper accounting without out-of-band tracking.

Would be good to persist channels so we have a historical record of everything we can.

f3r10:
## Summary

  Closes #851.

Adds persistent historical closed channel records so applications can query all channels that have ever closed on this node, surviving restarts.

  ### New public API

  - Node::list_closed_channels() -> Vec<ClosedChannelDetails>
  - ClosedChannelDetails struct: channel_id, user_channel_id, counterparty_node_id (non-optional — always set for channels closed on LDK Node v0.2+), funding_txo, channel_capacity_sats, last_local_balance_msat, is_outbound, is_announced, closure_reason, closed_at
  - Event::ChannelClosed gains three new fields: channel_capacity_sats, channel_funding_txo, last_local_balance_msat

  ### Implementation

Closed channel persistence (src/closed_channel.rs, src/io/): ClosedChannelDetails implements StorableObject with insert_or_update semantics (closed records are immutable — update() always returns false). Records are stored under the closed_channels KV namespace keyed by UserChannelId. insert_or_update makes the ChannelClosed handler idempotent: if the event is replayed after a crash between persistence and add_event, the second write is a no-op and the event goes throuh cleanly. Failure to persist returns ReplayEvent so LDK retries.

Per-channel state store (src/channel/store.rs, new): Introduces a ChannelRecord enum with a single Funded variant that durably captures counterparty_node_id, channel_id, is_outbound, and is_announced at ChannelPending time (before funding confirms). Records are stored under the channel_records KV namespace and removed once ChannelClosed is fully processed. This is more robust than a purely in-memory approach: if the node crashes between ChannelPending and ChannelClosed, the flags survive the restart. The enum design is intentionally extensible — a future pending_splice field can be added as a new optional TLV without breking existing records (motivated by the planned splice-retry support).

ChannelClosed lookup chain: is_outbound and is_announced are resolved with a priority fallback: (1) channel_record_store.get() — the durable record written at ChannelPending; (2) closed_channel_store.get() — handles the replay case where insert_or_update already succeeded but add_event failed and the channel record was alreaddy removed; (3) in-memory outbound_channel_ids / announced_channel_ids sets — fallback for channels opened before this upgrade.

Startup (src/builder.rs): Loading closed channel records and channel records joins the existing tokio::join! block alongside payments, pending payments, and node metrics — no added sequential startup cost as history grows.

Serialization compatibility: counterparty_node_id on ChannelReady and ChannelClosed events is now a required field. Events persisted by LDK Node v0.1.0 and prior that are missing this field will fail to deserialize. closed_at in ClosedChannelDetails is also required (no default_value fallback) since there are no pre-existing records to migrate.

### DIFF
--- a/src/closed_channel.rs
+++ b/src/closed_channel.rs
@@ -1,0 +1,131 @@
+// This file is Copyright its original authors, visible in version control history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
+// accordance with one or both of these licenses.
+
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+use bitcoin::secp256k1::PublicKey;
+use bitcoin::OutPoint;
+use lightning::events::ClosureReason;
+use lightning::ln::msgs::DecodeError;
+use lightning::ln::types::ChannelId;
+use lightning::util::ser::{Readable, Writeable, Writer};
+use lightning::{_init_and_read_len_prefixed_tlv_fields, write_tlv_fields};
+
+use crate::data_store::{StorableObject, StorableObjectId, StorableObjectUpdate};
+use crate::hex_utils;
+use crate::types::UserChannelId;
+
+/// Details of a closed channel.
+///
+/// Returned by [`Node::list_closed_channels`].
+///
+/// [`Node::list_closed_channels`]: crate::Node::list_closed_channels
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct ClosedChannelDetails {
+	/// The channel's ID at the time it was closed.
+	pub channel_id: ChannelId,
+	/// The local identifier of the channel.
+	pub user_channel_id: UserChannelId,
+	/// The node ID of the channel's counterparty.
+	pub counterparty_node_id: Option<PublicKey>,
+	/// The channel's funding transaction outpoint.
+	pub funding_txo: Option<OutPoint>,
+	/// The channel's capacity in satoshis.
+	pub channel_capacity_sats: Option<u64>,
+	/// Our local balance in millisatoshis at the time of channel closure.
+	pub last_local_balance_msat: Option<u64>,
+	/// Indicates whether we initiated the channel opening.
+	///
+	/// `true` if the channel was opened by us (outbound), `false` if opened by the counterparty
+	/// (inbound). This will be `false` for channels opened prior to this field being tracked.
+	pub is_outbound: bool,
+	/// The reason for the channel closure.
+	pub closure_reason: Option<ClosureReason>,
+	/// The timestamp, in seconds since start of the UNIX epoch, when the channel was closed.
+	pub closed_at: u64,
+}
+
+impl Writeable for ClosedChannelDetails {
+	fn write<W: Writer>(&self, writer: &mut W) -> Result<(), lightning::io::Error> {
+		write_tlv_fields!(writer, {
+			(0, self.channel_id, required),
+			(2, self.user_channel_id, required),
+			(4, self.counterparty_node_id, option),
+			(6, self.funding_txo, option),
+			(8, self.channel_capacity_sats, option),
+			(10, self.last_local_balance_msat, option),
+			(12, self.is_outbound, required),
+			(14, self.closure_reason, upgradable_option),
+			(16, self.closed_at, required),
+		});
+		Ok(())
+	}
+}
+
+impl Readable for ClosedChannelDetails {
+	fn read<R: lightning::io::Read>(reader: &mut R) -> Result<Self, DecodeError> {
+		let unix_time_secs = SystemTime::now()
+			.duration_since(UNIX_EPOCH)
+			.unwrap_or(Duration::from_secs(0))
+			.as_secs();
+		_init_and_read_len_prefixed_tlv_fields!(reader, {
+			(0, channel_id, required),
+			(2, user_channel_id, required),
+			(4, counterparty_node_id, option),
+			(6, funding_txo, option),
+			(8, channel_capacity_sats, option),
+			(10, last_local_balance_msat, option),
+			(12, is_outbound, required),
+			(14, closure_reason, upgradable_option),
+			(16, closed_at, (default_value, unix_time_secs)),
+		});
+		Ok(ClosedChannelDetails {
+			channel_id: channel_id.0.ok_or(DecodeError::InvalidValue)?,
+			user_channel_id: user_channel_id.0.ok_or(DecodeError::InvalidValue)?,
+			counterparty_node_id,
+			funding_txo,
+			channel_capacity_sats,
+			last_local_balance_msat,
+			is_outbound: is_outbound.0.ok_or(DecodeError::InvalidValue)?,
+			closure_reason,
+			closed_at: closed_at.0.ok_or(DecodeError::InvalidValue)?,
+		})
+	}
+}
+
+pub(crate) struct ClosedChannelDetailsUpdate(pub UserChannelId);
+
+impl StorableObjectUpdate<ClosedChannelDetails> for ClosedChannelDetailsUpdate {
+	fn id(&self) -> UserChannelId {
+		self.0
+	}
+}
+
+impl StorableObject for ClosedChannelDetails {
+	type Id = UserChannelId;
+	type Update = ClosedChannelDetailsUpdate;
+
+	fn id(&self) -> UserChannelId {
+		self.user_channel_id
+	}
+
+	fn update(&mut self, _update: Self::Update) -> bool {
+		// Closed channel records are immutable once written.
+		false
+	}
+
+	fn to_update(&self) -> Self::Update {
+		ClosedChannelDetailsUpdate(self.user_channel_id)
+	}
+}
+
+impl StorableObjectId for UserChannelId {
+	fn encode_to_hex_str(&self) -> String {
+		hex_utils::to_string(&self.0.to_be_bytes())
+	}
+}
--- a/src/closed_channel.rs
+++ b/src/closed_channel.rs
@@ -1,0 +1,89 @@
+// This file is Copyright its original authors, visible in version control history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
+// accordance with one or both of these licenses.
+
+use bitcoin::secp256k1::PublicKey;
+use bitcoin::OutPoint;
+use lightning::events::ClosureReason;
+use lightning::impl_writeable_tlv_based;
+use lightning::ln::types::ChannelId;
+
+use crate::data_store::{StorableObject, StorableObjectUpdate};
+use crate::types::UserChannelId;
+
+/// Details of a closed channel.
+///
+/// Returned by [`Node::list_closed_channels`].
+///
+/// [`Node::list_closed_channels`]: crate::Node::list_closed_channels
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct ClosedChannelDetails {
+	/// The channel's ID at the time it was closed.
+	pub channel_id: ChannelId,
+	/// The local identifier of the channel.
+	pub user_channel_id: UserChannelId,
+	/// The node ID of the channel's counterparty.
+	pub counterparty_node_id: PublicKey,
+	/// The channel's funding transaction outpoint.
+	pub funding_txo: Option<OutPoint>,
+	/// The channel's capacity in satoshis.
+	pub channel_capacity_sats: Option<u64>,
+	/// Our local balance in millisatoshis at the time of channel closure.
+	pub last_local_balance_msat: Option<u64>,
+	/// Indicates whether we initiated the channel opening.
+	///
+	/// `true` if the channel was opened by us (outbound), `false` if opened by the counterparty
+	/// (inbound). This will be `false` for channels opened prior to this field being tracked.
+	pub is_outbound: bool,
+	/// Indicates whether the channel was publicly announced.
+	///
+	/// This will be `false` for channels opened prior to this field being tracked.
+	pub is_announced: bool,
+	/// The reason for the channel closure.
+	pub closure_reason: Option<ClosureReason>,
+	/// The timestamp, in seconds since start of the UNIX epoch, when the channel was closed.
+	pub closed_at: u64,
+}
+
+impl_writeable_tlv_based!(ClosedChannelDetails, {
+	(0, channel_id, required),
+	(2, user_channel_id, required),
+	(4, counterparty_node_id, required),
+	(6, funding_txo, option),
+	(8, channel_capacity_sats, option),
+	(10, last_local_balance_msat, option),
+	(12, is_outbound, required),
+	(14, closure_reason, upgradable_option),
+	(16, closed_at, required),
+	(18, is_announced, required),
+});
+
+pub(crate) struct ClosedChannelDetailsUpdate(pub UserChannelId);
+
+impl StorableObjectUpdate<ClosedChannelDetails> for ClosedChannelDetailsUpdate {
+	fn id(&self) -> UserChannelId {
+		self.0
+	}
+}
+
+impl StorableObject for ClosedChannelDetails {
+	type Id = UserChannelId;
+	type Update = ClosedChannelDetailsUpdate;
+
+	fn id(&self) -> UserChannelId {
+		self.user_channel_id
+	}
+
+	fn update(&mut self, _update: Self::Update) -> bool {
+		// Closed channel records are immutable once written.
+		false
+	}
+
+	fn to_update(&self) -> Self::Update {
+		ClosedChannelDetailsUpdate(self.user_channel_id)
+	}
+}
--- a/src/closed_channel.rs
+++ b/src/closed_channel.rs
@@ -29,10 +29,16 @@ pub struct ClosedChannelDetails {
 	/// The node ID of the channel's counterparty.
 	pub counterparty_node_id: PublicKey,
 	/// The channel's funding transaction outpoint.
+	///
+	/// Will be `None` if the channel was closed before a funding transaction was established.
 	pub funding_txo: Option<OutPoint>,
 	/// The channel's capacity in satoshis.
+	///
+	/// Will be `None` if the channel was closed before the capacity was known.
 	pub channel_capacity_sats: Option<u64>,
 	/// Our local balance in millisatoshis at the time of channel closure.
+	///
+	/// Will be `None` if the local balance was not available at the time of closure.
 	pub last_local_balance_msat: Option<u64>,
 	/// Indicates whether we initiated the channel opening.
 	///
@@ -44,6 +50,9 @@ pub struct ClosedChannelDetails {
 	/// This will be `false` for channels opened prior to this field being tracked.
 	pub is_announced: bool,
 	/// The reason for the channel closure.
+	///
+	/// Will be `None` if the closure reason could not be decoded, e.g., if it was written by a
+	/// future version of LDK Node using a closure reason variant not yet known to this version.
 	pub closure_reason: Option<ClosureReason>,
 	/// The timestamp, in seconds since start of the UNIX epoch, when the channel was closed.
 	pub closed_at: u64,
--- a/src/event.rs
+++ b/src/event.rs
@@ -7,9 +7,10 @@
 
 use core::future::Future;
 use core::task::{Poll, Waker};
-use std::collections::VecDeque;
+use std::collections::{HashSet, VecDeque};
 use std::ops::Deref;
 use std::sync::{Arc, Mutex};
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use bitcoin::blockdata::locktime::absolute::LockTime;
 use bitcoin::secp256k1::PublicKey;
@@ -35,6 +36,7 @@ use lightning::util::ser::{Readable, ReadableArgs, Writeable, Writer};
 use lightning_liquidity::lsps2::utils::compute_opening_fee;
 use lightning_types::payment::{PaymentHash, PaymentPreimage};
 
+use crate::closed_channel::ClosedChannelDetails;
 use crate::config::{may_announce_channel, Config};
 use crate::connection::ConnectionManager;
 use crate::data_store::DataStoreUpdateResult;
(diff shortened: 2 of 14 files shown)


src/closed_channel.rs
```diff
@@ -0,0 +1,131 @@
...
+			counterparty_node_id,
+			funding_txo,
+			channel_capacity_sats,
+			last_local_balance_msat,
+			is_outbound: is_outbound.0.ok_or(DecodeError::InvalidValue)?,
+			closure_reason,
+			closed_at: closed_at.0.ok_or(DecodeError::InvalidValue)?,
+		})
+	}
+}
```
Camillarhi, line 99: Can we use `impl_writeable_tlv_based!` here instead? Matches the rest of the `StorableObject` types in the codebase

src/lib.rs
```diff
@@ -1093,6 +1097,13 @@ impl Node {
 		self.channel_manager.list_channels().into_iter().map(|c| c.into()).collect()
 	}
 
+	/// Retrieve a list of closed channels.
+	///
+	/// Channels are added to this list when they are closed and will be persisted across restarts.
+	pub fn list_closed_channels(&self) -> Vec<ClosedChannelDetails> {
```
Camillarhi, line 1103: The new `list_closed_channels()` API isn't exposed in `bindings`.

src/closed_channel.rs
```diff
@@ -0,0 +1,131 @@
...
+use crate::types::UserChannelId;
+
+/// Details of a closed channel.
+///
+/// Returned by [`Node::list_closed_channels`].
+///
+/// [`Node::list_closed_channels`]: crate::Node::list_closed_channels
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct ClosedChannelDetails {
```
Camillarhi, line 29: `ClosedChannelDetails` isn't exposed in bindings

Camillarhi (changes requested):
Thanks for looking into this. I dropped some comments. Also, CI seems to be failing due to a formatting issue. Kindly run `cargo fmt --all` to resolve this

src/io/utils.rs
```diff
@@ -298,6 +301,82 @@ where
...
+		CLOSED_CHANNEL_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
+	)
+	.await?;
+
+	const BATCH_SIZE: usize = 50;
+
+	let mut set = tokio::task::JoinSet::new();
+
+	// Fill JoinSet with tasks if possible
+	while set.len() < BATCH_SIZE && !stored_keys.is_empty() {
```
tnull, line 244: Makes sense to DRY this up for now, but depending on when #876 lands we might need to adapt this to use the new `BatchingStore`.

tests/integration_tests_rust.rs
```diff
@@ -2957,3 +2957,98 @@ async fn splice_in_with_all_balance() {
...
+		assert_eq!(closed_a.len(), 1);
+		let record = &closed_a[0];
+		assert_eq!(record.channel_capacity_sats, Some(channel_amount_sat));
+		assert!(record.is_outbound);
+		assert_eq!(record.counterparty_node_id, Some(node_b.node_id()));
+		assert!(record.funding_txo.is_some());
+		assert_eq!(record.funding_txo.unwrap().txid, funding_txo.txid);
+		assert!(record.closure_reason.is_some());
+
+		closed_channel_before = record.clone();
```
tnull: nit: Why do we clone here?

tests/integration_tests_rust.rs
```diff
@@ -2957,3 +2957,98 @@ async fn splice_in_with_all_balance() {
...
+		node_a.close_channel(&user_channel_id, node_b.node_id()).unwrap();
+		expect_event!(node_a, ChannelClosed);
+		expect_event!(node_b, ChannelClosed);
+
+		generate_blocks_and_wait(&bitcoind.client, &electrsd.client, 1).await;
+		node_a.sync_wallets().unwrap();
+		node_b.sync_wallets().unwrap();
+
+		// Verify the record is present before restart.
+		let closed_a = node_a.list_closed_channels();
```
tnull, line 4169: Should we also check that b also persisted the closed channel?

src/event.rs
tnull: I think we should be good breaking backwards compat with LDK node v0.1, so let's make this a required field. Then we can also avoid having the `Option` propagate. Please add a not regarding serialization compatibility as part of a `# Pending` section in `CHANGELOG.md`

src/closed_channel.rs
tnull, line 24: Might be good also recording whether this was an announced channel.

tnull (commented):
Thanks, already looks pretty good. Some comments.

src/io/utils.rs
f3r10: rebased to use the latest `read_all_objects`

src/closed_channel.rs
```diff
@@ -0,0 +1,98 @@
...
+	///
+	/// `true` if the channel was opened by us (outbound), `false` if opened by the counterparty
+	/// (inbound). This will be `false` for channels opened prior to this field being tracked.
+	pub is_outbound: bool,
+	/// Indicates whether the channel was publicly announced.
+	///
+	/// This will be `false` for channels opened prior to this field being tracked.
+	pub is_announced: bool,
+	/// The reason for the channel closure.
+	pub closure_reason: Option<ClosureReason>,
```
Camillarhi, line 56: Is there a reason `closure_reason` is `Option<ClosureReason>`? The LDK reason on `ChannelClosed` isn't optional and we always wrap it in `Some` at write time, so I don't think we can ever persist a `None`. Could it just be `ClosureReason`?

src/event.rs
```diff
@@ -1581,15 +1636,63 @@ where
...
+						channel_id,
+						e
+					);
+				}
+
+				let event = Event::ChannelClosed {
+					channel_id,
+					user_channel_id,
+					counterparty_node_id: counterparty_node_id
+						.expect("counterparty_node_id must be set for closed channels"),
```
Camillarhi: This could panic as the `counterparty_node_id` from LDK is `Option<Publickey>`

Camillarhi (changes requested):
Thanks! I left a few more comments.

src/event.rs
```diff
@@ -1691,7 +1687,8 @@ where
 				let event = Event::ChannelClosed {
 					channel_id,
 					user_channel_id,
-					counterparty_node_id,
+					counterparty_node_id: counterparty_node_id
+						.expect("counterparty_node_id must be set for closed channels"),
```
tnull: Please provide a reason why this expectation always holds (i.e. since which version of LDK or LDK node this is the case).

src/closed_channel.rs
tnull: This seems valid. More generally, for all optional fields here, please add a note to the docs describing when/from which version they can be expected to be set.

tnull (commented):
Feel free to squash. Two minor comments, and then we should move forward with this.

src/event.rs
```diff
@@ -1581,15 +1636,63 @@ where
...
-				..
+				channel_capacity_sats,
+				channel_funding_txo,
+				last_local_balance_msat,
 			} => {
 				log_info!(self.logger, "Channel {} closed due to: {}", channel_id, reason);
 
-				let event = Event::ChannelClosed {
+				let user_channel_id = UserChannelId(user_channel_id);
+				let is_outbound = self
```

>  - [P2] Persist channel flags for replayed closures — /home/tnull/workspace/ldk-node/src/event.rs:1646-1655
>     When a ChannelClosed LDK event is replayed after a restart or after handle_event returns ReplayEvent, the channel can already be gone from list_channels() and these process-local sets are not reconstructed. In that case .remove() returns false, so an outbound/public channel is persisted in
>     ClosedChannelDetails as inbound/unannounced; a failed first attempt also consumes the entries before replay. Store these flags durably before closure or make this path idempotent.

src/event.rs
```diff
@@ -1581,15 +1636,63 @@ where
...
+					funding_txo,
+					channel_capacity_sats,
+					last_local_balance_msat,
+					is_outbound,
+					is_announced,
+					closure_reason: Some(reason.clone()),
+					closed_at,
+				};
+
+				if let Err(e) = self.closed_channel_store.insert(record) {
```

>  - [P2] Replay ChannelClosed when history persistence fails — /home/tnull/workspace/ldk-node/src/event.rs:1678-1685
>     If closed_channel_store.insert(record) fails but event_queue.add_event succeeds, this only logs the error and eventually returns Ok(()), so LDK considers the ChannelClosed handled and will not replay it. The user still sees the close event, but list_closed_channels() permanently misses the
>     record; propagate this failure as ReplayEvent like other persisted event side effects.

src/closed_channel.rs
f3r10: As I understand, `ClosureReason` is an evolving enum — LDK could add new variants over time.
If we used required + a plain ClosureReason field, an unknown future variant would cause `read_all_objects` to return an error, and the node would fail to start 🤔 
wdyt @Camillarhi ?

src/event.rs
f3r10: @tnull I added a changelog entry about this change https://github.com/lightningdevkit/ldk-node/pull/882/changes#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR3 
```The `counterparty_node_id` field of the `ChannelReady` and `ChannelClosed` events is now
  required. Events persisted by LDK Node v0.1.0 and prior that are missing this field will
  fail to deserialize.```
But since LDK can return None for early-lifecycle close, maybe the right approach would be: revert Event::ChannelClosed::counterparty_node_id back to Option<PublicKey>, remove the .expect(), and remove the CHANGELOG entry. Should I make those changes?

Comment (tnull) [reply src/event.rs]: No, breaking backwards compat with LDK Node v0.1 is fine, we should just (very briefly) also mention in the except that that is what's happening (i.e., that >v0.1 this is expected to hold true).

File: src/closed_channel.rs (hunk 7 above)
Comment (tnull) [reply src/closed_channel.rs:56]: > As I understand, `ClosureReason` is an evolving enum — LDK could add new variants over time. If we used required + a plain ClosureReason field, an unknown future variant would cause `read_all_objects` to return an error, and the node would fail to start 🤔 wdyt @Camillarhi ?

Hmm, no, LDK's compatibility guarantees will protect us from that. Upstream we'd add new fields only as optional/odd usually, to ensure we never break serialization compatibility.

Comment (f3r10) [reply src/closed_channel.rs:56]: oh ok thanks @Camillarhi. I am going to remove the `Option` there

File: src/event.rs (hunk 9 above)
Comment (f3r10) [reply src/event.rs]: Great, I am going to add that note

File: src/closed_channel.rs (hunk 7 above)
Comment (f3r10) [reply src/closed_channel.rs:56]: @Camillarhi I tried to make the change, but it looks like `ClosureReason` intentionally does not implement Readable only MaybeReadable, and `impl_writeable_tlv_based!` returns an error 🤔

Comment (Camillarhi) [reply src/closed_channel.rs:56]: Checked this out on my end as well, and you are right. Let's keep it as is

Comment (tnull) [reply src/closed_channel.rs:56]: > [@Camillarhi](https://github.com/Camillarhi) I tried to make the change, but it looks like `ClosureReason` intentionally does not implement Readable only MaybeReadable, and `impl_writeable_tlv_based!` returns an error 🤔

Hmm, ah, good point, didn't realize it only implemented `MaybeReadable`.

File: src/closed_channel.rs [hunk 12]
```diff
@@ -0,0 +1,110 @@
...
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct ClosedChannelDetails {
+	/// The channel's ID at the time it was closed.
+	pub channel_id: ChannelId,
+	/// The local identifier of the channel.
+	pub user_channel_id: UserChannelId,
+	/// The node ID of the channel's counterparty.
+	///
+	/// Will be `None` if the channel was closed before the counterparty's node ID could be
+	/// determined (e.g., very early in the channel negotiation process).
```
tnull: Hmm, is this correct? Wouldn't we always get a `ChannelClosed` event that by now has the required field?

src/closed_channel.rs
```diff
@@ -0,0 +1,110 @@
...
+impl_writeable_tlv_based!(ClosedChannelDetails, {
+	(0, channel_id, required),
+	(2, user_channel_id, required),
+	(4, counterparty_node_id, option),
+	(6, funding_txo, option),
+	(8, channel_capacity_sats, option),
+	(10, last_local_balance_msat, option),
+	(12, is_outbound, required),
+	(14, closure_reason, upgradable_option),
+	(16, closed_at, (default_value, SystemTime::now().duration_since(UNIX_EPOCH).unwrap_or(Duration::from_secs(0)).as_secs())),
```
tnull: Hmm, given we don't have any backwards compat. issues, let's just make this a `required` field and set the value upon initialization rather than via `default_value`?

src/event.rs
```diff
@@ -1581,15 +1636,76 @@ where
...
+					.map(|existing| (existing.is_outbound, existing.is_announced))
+					.unwrap_or((is_outbound_from_set, is_announced_from_set));
+
+				let closed_at = SystemTime::now()
+					.duration_since(UNIX_EPOCH)
+					.unwrap_or(Duration::ZERO)
+					.as_secs();
+
+				let funding_txo =
+					channel_funding_txo.map(|op| OutPoint { txid: op.txid, vout: op.index as u32 });
```
tnull: nit: Should be able to use `into_bitcoin_outpoint` here.

src/event.rs
```diff
@@ -1581,15 +1636,76 @@ where
...
+					.announced_channel_ids
+					.lock()
+					.expect("Lock poisoned")
+					.remove(&user_channel_id);
+
+				// On replay (after a restart or after handle_event returns ReplayEvent),
+				// the channel is no longer in list_channels() and the in-memory sets are
+				// not repopulated for it, so .remove() returns false. Fall back to any
+				// already-persisted record so we don't overwrite correct values with false.
+				let (is_outbound, is_announced) = self
```
tnull, line 1717: IIUC, this will fall back to `(false, false)` in case of a replayed event, leading to inaccurate information. This makes we wonder if we don't need to start storing channel metadata generally (i.e., prior to close), not only once it's already gone and we struggle to maintain access to all necessary information?

tnull (commented):
Excuse the delay here! Generally looks good mod some minor comments.

The only more fundamental question is https://github.com/lightningdevkit/ldk-node/pull/882/changes#r3340832531 which we probably should clarify before moving on.

src/closed_channel.rs
Camillarhi: Yeah, `ClosureReason` uses `impl_writeable_tlv_based_enum_upgradable!`, so it's `MaybeReadable` only. An unrecognized closure reason from a newer LDK gets skipped as `Ok(None)` rather than erroring.

tnull: Right, we technically still can never hit this right now as we only support upgrades, not downgrades, but hopefully we eventually also support the latter, so better safe than sorry.

src/event.rs
f3r10: 🤔 I have been thinking about this problem, and you're right, and the correct fix is the durable approach: persisting the flags at ChannelPending time.

> * ClosedChannelDetails as inbound/unannounced; a failed first attempt also consumes the entries before replay. Store these flags durably before closure or make this path idempotent.

From the previous review of codex, I chose the idempotent path because it was quicker to implement.  The insert_or_update + fallback change covers the restart replay case where a prior insert did succeed, but it doesn't fully solve the problem.

I can try the `durable approach` but it looks more difficult and would take more time. Should I make this change on this PR or open a follow-up issue? wdyt @tnull ?

tnull: We're not super pressed for time here, so it would be good to do it in one PR. Given it's basically a pretty different approach you could even consider whether to open it as an alternative PR.

f3r10: ok cool, I am going to start with the `durable approach` and let you know when is ready

jkczyz: During Monday's dev sync, @tnull suggested that this storage could be generalized further to hold channel splice intents.

Essentially, when a user initiates a splice, LDK needs to reach quiescence with the counterparty, negotiated and sign the new funding transaction, and wait for the splice to lock. If we restart before signatures are exchanged, we need to retry the splice. Similar for v2 opens, but we don't support those yet.

I had Claude plan this out to work with this PR. We'd store a `ChannelRecord` that has three variants (`Pending`, `Funded` and `Closed`) where the `Funded` variant can hold a `SpliceIntent`. Something like this:

```rust
enum ChannelRecord {                     // src/...; StorableObject, Id = UserChannelId
    /// Channel being established — holds the open/funding intent.        [opens — later]
    Unfunded { counterparty_node_id: PublicKey, intent: OpenIntent },
    /// Channel live — holds an in-flight splice intent, if any.
    Funded {
        counterparty_node_id: PublicKey,
        channel_id: ChannelId,
        pending_splice: Option<SpliceIntent>,                            // [this PR]
        // is_outbound / funding_txo / capacity_sats / is_announced fold in with #882
    },
    /// Channel closed — #882's immutable snapshot.                        [#882 — later]
    Closed { /* channel_id, funding_txo, capacity_sats, is_outbound, is_announced, reason, closed_at */ },
}

struct SpliceIntent {
    pre_splice_funding_txo: OutPoint,    // completion discriminator — changes only when the splice locks
    contribution: FundingContribution,   // re-fed → funding_contributed(.., None)
    kind: SpliceKind,                    // user params — rebuilds a fresh contribution when the stored one is stale
    attempts: u8,
}
enum SpliceKind { In { amount_sats: u64 }, Out { outputs: Vec<TxOut> }, Rbf }
// `kind` serves two purposes: Rbf changes the probe interpretation (prior `Some` is *expected*),
// and In/Out carry the params needed to rebuild a contribution via a fresh splice_channel template.
// enum OpenIntent { V1 { amount_sats, push_msat, config, announce, zero_reserve },  // → create_channel  [later]
//                   V2 { contribution: FundingContribution } }                       // → funding_contributed  [later]
```

@f3r10 Would this be sufficient to cover your use cases?

jkczyz: Here's the draft implementation: https://github.com/lightningdevkit/ldk-node/pull/930/changes/492bd32461d512ffd2418b4ffa3f8a5aa8e1d044

f3r10: oh ok great thanks @jkczyz I am going to check the draft and push a commit

f3r10: ready @jkczyz https://github.com/lightningdevkit/ldk-node/pull/882/changes/2c7e16f33acc270ef3040fc6586a073d6083f13d
Let me know if something is missing

tnull: FTR, concluded offline that this should now probably based on #930.

src/event.rs
```diff
@@ -1346,12 +1390,23 @@ where
...
 				let res = if allow_0conf {
+					let trusted_features = if is_lsps2_channel {
+						TrustedChannelFeatures::ZeroConfZeroReserve
+					} else {
+						TrustedChannelFeatures::ZeroConf
+					};
 					self.channel_manager.accept_inbound_channel_from_trusted_peer(
 						&temporary_channel_id,
 						&counterparty_node_id,
 						user_channel_id,
```
Jolah1: This `ZeroConfZeroReserve` handling for LSPS2 JIT channels looks unrelated to closed-channel history. is it intentional? 
Either way, it's a distinct behavioral change (dropping the counterparty reserve for LSP-accepted channels) with its own correctness implications, and it isn't mentioned in the PR description or CHANGELOG. Could it be split into its own PR with a dedicated test and #Pending note? That keeps the reserve change reviewable on its own and keeps this PR's diff focused on the persistence/serialization work.

src/event.rs
```diff
@@ -1609,19 +1702,100 @@ where
...
+				let is_outbound_from_set = self
+					.outbound_channel_ids
+					.lock()
+					.expect("Lock poisoned")
+					.remove(&user_channel_id);
+				let is_announced_from_set = self
+					.announced_channel_ids
+					.lock()
+					.expect("Lock poisoned")
+					.remove(&user_channel_id);
```
Jolah1: Following up on the durable-ChannelRecord work that resolved your earlier Codex [P2] note: this covers the post-upgrade case nicely (the ChannelRecord lookup at 1723 wins, and it's only removed after add_event succeeds). But the in-memory sets are drained here, before the closed_channel_store.insert_or_update().await at 1768. 
For a pre-upgrade channel, no durable ChannelRecord, and no ClosedChannelDetails yet.  If that persist fails and the event is replayed, both fallbacks miss and we settle on the drained (false, false), persisting the channel as inbound/unannounced.

 It's a narrow window (KV-write failure on a pre-upgrade channel), but it's the same replay-correctness class as [P2]. Reading the sets with .contains() and only .remove()  after the closed-channel persist succeeds would close it.

Jolah1 (commented):
Took a careful pass over this, nice work! 

The durable ChannelRecord written at ChannelPending is a clean answer to the replay concern from the earlier rounds. A few things, mostly small, plus one scope question. I've left two inline comments and have a couple of discussion points below. (Noting you've concluded offline to rebase this onto #930, so I've skipped anything tied to the exact ChannelRecord shape, TLV type numbers, or the builder join! wiring, since those will move.)

  - Scope (event.rs ~1393): an LSPS2 ZeroConfZeroReserve change that looks unrelated to closed-channel history. flagged inline, suggesting it be split out.
  
  - Replay correctness (event.rs ~1711): the in-memory sets are drained before the closed-channel persist, so a pre-upgrade channel can settle on (false, false) on a persist-failure replay. Residual of the [P2] thread; details inline.

  Happy to re-review after the #930 rebase.

src/event.rs
jkczyz: Since #930 is based on #888, I opened #946 with just the channel storage code. You can base your work off of that instead.

joostjager (commented):
Close-channel history seems common enough that maybe LDK should optionally record or retain it itself. I think that is the main question to answer here, because if the history can live at the LDK layer, we avoid introducing another store in ldk-node and avoid reasoning about possible desync with LDK’s persisted `ChannelManager`/`ChannelMonitor` state.

And if there is a convincing reason this does not make sense in LDK at all, then I’d still expect LDK to emit all close-relevant channel facts on `ChannelClosed`? That would let ldk-node persist a terminal closed-channel state directly, instead of keeping a live mirror store only to reconstruct the snapshot later.

src/event.rs
f3r10: @tnull @jkczyz I finally was able to complete the rebase. Let me know if there is something wrong.

f3r10:
> Close-channel history seems common enough that maybe LDK should optionally record or retain it itself. I think that is the main question to answer here, because if the history can live at the LDK layer, we avoid introducing another store in ldk-node and avoid reasoning about possible desync with LDK’s persisted `ChannelManager`/`ChannelMonitor` state.
> 
> And if there is a convincing reason this does not make sense in LDK at all, then I’d still expect LDK to emit all close-relevant channel facts on `ChannelClosed`? That would let ldk-node persist a terminal closed-channel state directly, instead of keeping a live mirror store only to reconstruct the snapshot later.

I agreed with the comment. If LDK tracked close history natively, we'd need nothing here. Short of that, if ChannelClosed carried everything needed to write a terminal record in one step, we could drop the channel_record_store mirror entirely. 
 
I could open an LDK issue to discuss both. @tnull, would you prefer to wait on that before merging, or is this acceptable as an interim approach?

src/event.rs
f3r10: thanks @Jolah1

PR #884, opened 2026-04-21
Allow wallet imports from arbitrary heights, forcing full scans

### LINKED ISSUE #818: Wallet birthday height for seed recovery on pruned nodes

tnull:
Fixes #818.

Previously, we would always start syncing from the latest block height *or* were able to force a full-resync from genesis via `recovery_mode = true`. Here we add more fine-grained control and allow an initial wallet import from a known birthday height for `bitcoind` RPC/REST chain sources. Relatedly, we also fix a minor bug where, if we failed to retrieve the tip hash on first startup, we'd always resync from genesis, which can be very costly.

Moreover, we also add a flag that allows the user to force a full scan for Esplora/Electrum wallets.

### DIFF
--- a/src/chain/electrum.rs
+++ b/src/chain/electrum.rs
@@ -125,9 +125,11 @@ impl ElectrumChainSource {
 			return Err(Error::FeerateEstimationUpdateFailed);
 		};
 		// If this is our first sync, do a full scan with the configured gap limit.
-		// Otherwise just do an incremental sync.
-		let incremental_sync =
+		// Otherwise just do an incremental sync. Also take one-shot priority over an incremental
+		// sync if the user opted into recovery mode via `NodeBuilder::set_wallet_recovery_mode`.
+		let has_prior_sync =
 			self.node_metrics.read().expect("lock").latest_onchain_wallet_sync_timestamp.is_some();
+		let incremental_sync = has_prior_sync && !onchain_wallet.take_force_full_scan();
 
 		let apply_wallet_update =
 			|update_res: Result<BdkUpdate, Error>, now: Instant| match update_res {
--- a/src/chain/electrum.rs
+++ b/src/chain/electrum.rs
@@ -6,6 +6,7 @@
 // accordance with one or both of these licenses.
 
 use std::collections::HashMap;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex, RwLock};
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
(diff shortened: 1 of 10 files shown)


tnull:
Rebased & refactored a bit, ready for review.

tnull:
Rebased to resolve minor conflict.

src/chain/electrum.rs
```diff
@@ -50,6 +51,7 @@ pub(super) struct ElectrumChainSource {
 	config: Arc<Config>,
 	logger: Arc<Logger>,
 	node_metrics: Arc<PersistedNodeMetrics>,
+	force_wallet_full_scan: AtomicBool,
```
wpaulino, line 54: Should we expose a block height as well to let the user optionally decide whether they want to start from a specific point?

wpaulino (approved):
LGTM, just one comment

src/chain/electrum.rs
tnull: You mean the current block height the node is synced to? That's already exposed via `NodeStatus::current_best_block` (cf. https://docs.rs/ldk-node/latest/ldk_node/struct.NodeStatus.html).

PR #889, opened 2026-04-24
Expose configurable routing scorer parameters

pwltr:
### Summary

Adds configurable pathfinding scorer parameters so apps can tune routing behavior instead of always using LDK defaults.

It introduces two new config surfaces:

- `ScoringFeeParameters` (maps to `ProbabilisticScoringFeeParameters`)
- `ScoringDecayParameters` (maps to `ProbabilisticScoringDecayParameters`)

These can now be set through:

- `Config` (`scoring_fee_params`, `scoring_decay_params`)
- `Builder` APIs:
  - `set_scoring_fee_params(...)`
  - `set_scoring_decay_params(...)`

At build time, `ldk-node` now uses configured scorer params when constructing a new `ProbabilisticScorer` (when no persisted scorer exists) and router fee params, falling back to LDK defaults when unset.

_Changes and PR description created with agentic assistance_

### DIFF
--- a/bindings/ldk_node.udl
+++ b/bindings/ldk_node.udl
@@ -11,6 +11,10 @@ typedef dictionary ElectrumSyncConfig;
 
 typedef dictionary TorConfig;
 
+typedef dictionary ScoringFeeParameters;
+
+typedef dictionary ScoringDecayParameters;
+
 typedef interface NodeEntropy;
 
 typedef enum WordCount;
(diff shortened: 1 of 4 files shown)


tnull (commented):
Hmm, so far we intentionally were holding back exposing all of these fields, as many are pretty cryptic to most users. Rather it would be preferable if we could provide sane defaults.

Could you expand on what you're actually looking to change? Also note that some of these changes might conflict/be impacted by the work over at #815.

pwltr:
> Could you expand on what you're actually looking to change?

Yeah so we are in the situation that we are seeing really bad payment/probing success rate at the moment and relying on a good set of scores from a remote prober alone does not resolve the issue. From testing it seems to be necessary to also tune scoring params on the client itself, predominately bumping `base_penalty_msat` and `base_penalty_amount_multiplier_msat` showed immediate improvement.

If we can find good defaults that would be preferable. For now we are grossly increasing `basePenaltyMsat` to default * 500 and `basePenaltyAmountMultiplierMsat` to default * 3, maybe unnecessarily high, but defaults really didn't give good results for any medium to large amounts.

PR #893, opened 2026-04-30
Log the error returned from `SqliteStore::new` and `fs::create_dir_all`

tankyleo:
This matches the logging done when setting up a VSS store.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -641,7 +641,7 @@ impl NodeBuilder {
 			log_error!(logger, "Failed to setup Sqlite store: {}", e);
 			BuildError::KVStoreSetupFailed
 		})?;
-		self.build_with_store(node_entropy, kv_store)
+		self.build_with_store_and_logger(node_entropy, kv_store, logger)
 	}
 
 	/// Builds a [`Node`] instance with a [`FilesystemStore`] backend and according to the options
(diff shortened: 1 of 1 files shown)


tnull (commented):
Hmm, it' a bit unfortunate that we now doubly-instantiate the logger. The pattern is pre-existing for VSS, but if we do this, maybe we want to introduce a `build_with_store_and_logger` internal method that all other `build_with` methods (including `build_with_store`) use, to avoid having to doubly setting up the logger every time?

tankyleo:
Thanks for the review, I added a commit to make this change, let me know if you would prefer to squash it with the previous one.

src/builder.rs
```diff
@@ -636,20 +637,26 @@ impl NodeBuilder {
 			Some(io::sqlite_store::SQLITE_DB_FILE_NAME.to_string()),
 			Some(io::sqlite_store::KV_TABLE_NAME.to_string()),
 		)
-		.map_err(|_| BuildError::KVStoreSetupFailed)?;
-		self.build_with_store(node_entropy, kv_store)
+		.map_err(|e| {
+			log_error!(logger, "Failed to setup Sqlite store: {}", e);
```
benthecarman, line 641: nit: here and below
```suggestion
			log_error!(logger, "Failed to set up Sqlite store: {}", e);
```

tnull: I'll land this shortly to keep moving, but feel free to add a commit addressing the nit in #863.

tnull (commented):
Thanks, LGTM! Landing this to unblock #863.

tnull (approved):

PR #894, opened 2026-05-01
Make sure the mutual close gets confirmed in `do_channel_full_cycle`

tankyleo:
In the mutual close case, before moving to check the expected balances of each node, we now assert that the mutual close transaction actually made it into a block. If the mutual close transaction got rejected for some reason, we now stop the test right there and fail instead of continuing onto balance checks.


tnull (commented):
~~It seems the CLN tests are failing now, mind looking into that?~~
Nvm, seems like unrelated flakiness.

tnull (approved):

Issue #745, opened 2026-01-08
Consider not removing peers on closure if trusted


tnull:
If our counterparty is LND they might not handle the force-closure error message correctly. We therefore might want to keep remembering the peer to reconnect automatically to initiate the channel-reestablish-recovery-flow after a channel had been force closed.

Relatedly, we might also want to remove a peer if the counterparty initiates the closure of the last remaining channel.

Jolah1:
Hi @tnull can i work on this?

tnull:
> Hi [@tnull](https://github.com/tnull) can i work on this?

Yes, please go ahead!

tnull:
Done in #895.

PR #895, opened 2026-05-01
Fix peer store management on channel closure

Jolah1:
## Summary

This PR resolves both issues described in #745 related to peer store management on channel closure.

### Fixed behaviors

- **Local force-close** → peer is **retained** to allow `channel_reestablish` recovery
- **Counterparty force-closes last channel** → peer is **removed** to avoid unnecessary reconnections

---

### 1. Local force-close removes peer too early

When we force-closed a channel in `close_channel_internal`, we immediately removed the peer from the store regardless of closure type:

```rust
// Before — peer removed for BOTH cooperative and force-close
if open_channels.len() == 1 {
    self.peer_store.remove_peer(&counterparty_node_id)?;
}
```
This broke recovery. The background reconnection task only reconnects to peers in the store. Removing the peer immediately guaranteed we would never reconnect, so `channel_reestablish` could never complete. This is especially problematic for LND peers, which may not handle force-closure error messages correctly and rely on us reconnecting to recover.

### 2. Counterparty force-close never cleans up peer

When a counterparty force-closed their last channel with us, the `ChannelClosed` event handler did nothing with the peer store. The peer remained persisted indefinitely and the node kept attempting to reconnect on every background tick. wasted resources and misleading peer state.

---

### 1. Retain peer on local force-close (`src/lib.rs`)

In `close_channel_internal`, peers are now only removed for cooperative
closes:

```rust
if open_channels.len() == 1 && !force {
    self.peer_store.remove_peer(&counterparty_node_id)?;
}
```

For force-closes, the peer is retained in the store so the background reconnection task can fire, `channel_reestablish` can complete, and peer cleanup is deferred to the `ChannelClosed` event handler.

### 2. Remove peer on counterparty-initiated closure (`src/event.rs`)

In the `LdkEvent::ChannelClosed` handler, we now check:

- Is this the last channel with the peer?
- Did the counterparty initiate the closure?

If both conditions are true, the peer is removed from the store.

---

### LDK timing behavior

LDK emits `ChannelClosed` before removing the channel from its internal state. A naive call to `list_channels_with_counterparty()` inside the handler would still return the closing channel, making the "last channel" check always wrong. We fix this by excluding the closing channel explicitly:

```rust
let has_other_channels = self
    .channel_manager
    .list_channels_with_counterparty(&counterparty_node_id)
    .iter()
    .any(|c| c.channel_id != channel_id); // exclude the channel that just closed
```

### Ordering to avoid race conditions

Peer removal is performed **before** calling `add_event`. This ensures consumers of `next_event_async()` always observe a consistent peer store when they react to `ChannelClosed` — avoiding a race where the event is visible before the removal completes.

---

## Tests

Added two integration tests in `tests/integration_tests_rust.rs`:

**`test_peer_removed_on_counterparty_force_close`**
Verifies that after the counterparty force-closes the last channel, the peer is no longer persisted in the local peer store.

**`test_peer_retained_on_local_force_close`**
Verifies that after a locally initiated force-close, the peer remains persisted in the local peer store to allow background reconnection and `channel_reestablish` recovery.

Both tests assert on `is_persisted` rather than peer presence to correctly distinguish stored peers from transient TCP connections that may still appear in `list_peers()` during teardown.

---

## Results
test test_peer_removed_on_counterparty_force_close ... ok
test test_peer_retained_on_local_force_close       ... ok
test result: ok. 41 passed; 0 failed; 0 ignored; 0 measured
cargo build  → no errors
cargo clippy → no new warnings
cargo fmt    → applied

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -1613,10 +1613,45 @@ where
 			} => {
 				log_info!(self.logger, "Channel {} closed due to: {}", channel_id, reason);
 
+				// `counterparty_node_id` has been set on every `ChannelClosed` since LDK 0.0.117.
+				let counterparty_node_id = counterparty_node_id
+					.expect("counterparty_node_id is always set since LDK 0.0.117");
+
+				// Drop the peer once its last channel with us has reached a terminal state
+				// that reconnection cannot recover. Every closure reason is terminal except
+				// `HolderForceClosed`: when *we* force-close, we keep reconnecting so that
+				// `channel_reestablish` can drive recovery (see `Node::close_channel_internal`).
+				// This also cleans up peers persisted for a channel that closed before funding
+				// (e.g. `CounterpartyCoopClosedUnfundedChannel`), which would otherwise be
+				// retried forever.
+				// We exclude `channel_id` from the count because LDK emits `ChannelClosed`
+				// before removing it from its internal list.
+				let dont_reconnect = !matches!(reason, ClosureReason::HolderForceClosed { .. });
+
+				if dont_reconnect {
+					let has_other_channels = self
+						.channel_manager
+						.list_channels_with_counterparty(&counterparty_node_id)
+						.iter()
+						.any(|c| c.channel_id != channel_id);
+
+					if !has_other_channels {
+						if let Err(e) = self.peer_store.remove_peer(&counterparty_node_id).await {
+							log_error!(
+								self.logger,
+								"Failed to remove peer {} from peer store: {}",
+								counterparty_node_id,
+								e
+							);
+							return Err(ReplayEvent());
+						}
+					}
+				}
+
 				let event = Event::ChannelClosed {
 					channel_id,
 					user_channel_id: UserChannelId(user_channel_id),
-					counterparty_node_id,
+					counterparty_node_id: Some(counterparty_node_id),
 					reason: Some(reason),
 				};
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1919,10 +1919,13 @@ impl Node {
 					})?;
 			}
 
-			// Check if this was the last open channel, if so, forget the peer.
-			if open_channels.len() == 1 {
-				self.runtime.block_on(self.peer_store.remove_peer(&counterparty_node_id))?;
-			}
+			// Peer store cleanup is handled centrally in the `ChannelClosed` event handler,
+			// which drops the peer once its last channel reaches a terminal state that
+			// reconnection cannot recover. We intentionally do nothing here so that a
+			// force-closed peer is retained, letting the background reconnection task keep
+			// firing and drive the `channel_reestablish` recovery flow. This is especially
+			// important against LND peers, which don't always handle force-closure error
+			// messages correctly.
 		}
 
 		Ok(())
--- a/tests/integration_tests_rust.rs
+++ b/tests/integration_tests_rust.rs
@@ -98,6 +98,56 @@ async fn channel_full_cycle_force_close_trusted_no_reserve() {
 }
 
 #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn peer_removed_when_counterparty_force_closes_last_channel() {
+	// When we open a channel outbound, we persist the counterparty so the background
+	// reconnection task can reach them. If the counterparty then force-closes what turns out
+	// to be their last channel with us, the channel is terminal and there is nothing left for
+	// `channel_reestablish` to recover, so the peer should be dropped from the store rather
+	// than reconnected to forever.
+	let (bitcoind, electrsd) = setup_bitcoind_and_electrsd();
+	let chain_source = random_chain_source(&bitcoind, &electrsd);
+	let (node_a, node_b) = setup_two_nodes(&chain_source, false, true, false);
+
+	let address_a = node_a.onchain_payment().new_address().unwrap();
+	let premine_amount_sat = 5_000_000;
+	premine_and_distribute_funds(
+		&bitcoind.client,
+		&electrsd.client,
+		vec![address_a],
+		Amount::from_sat(premine_amount_sat),
+	)
+	.await;
+	node_a.sync_wallets().unwrap();
+
+	// node_a opens the channel, so node_a persists node_b in its peer store.
+	open_channel(&node_a, &node_b, 4_000_000, false, &electrsd).await;
+	generate_blocks_and_wait(&bitcoind.client, &electrsd.client, 6).await;
+	node_a.sync_wallets().unwrap();
+	node_b.sync_wallets().unwrap();
+
+	let _user_channel_id_a = expect_channel_ready_event!(node_a, node_b.node_id());
+	let user_channel_id_b = expect_channel_ready_event!(node_b, node_a.node_id());
+
+	assert!(
+		node_a.list_peers().iter().any(|p| p.node_id == node_b.node_id() && p.is_persisted),
+		"node_a should persist node_b after opening a channel to it"
+	);
+
+	// The counterparty force-closes their last channel with us.
+	node_b.force_close_channel(&user_channel_id_b, node_a.node_id(), None).unwrap();
+
+	expect_event!(node_a, ChannelClosed);
+	expect_event!(node_b, ChannelClosed);
+
+	// node_a should have dropped node_b from its peer store. We assert on `is_persisted` rather
+	// than peer presence so a lingering transient TCP connection doesn't mask the removal.
+	assert!(
+		!node_a.list_peers().iter().any(|p| p.node_id == node_b.node_id() && p.is_persisted),
+		"node_a should drop node_b from its peer store after node_b force-closed the last channel"
+	);
+}
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
 async fn channel_full_cycle_0conf() {
 	let (bitcoind, electrsd) = setup_bitcoind_and_electrsd();
 	let chain_source = random_chain_source(&bitcoind, &electrsd);
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -1597,6 +1597,20 @@ pub(crate) async fn do_channel_full_cycle<E: ElectrumApi>(
 		assert!(node_b.list_balances().pending_balances_from_channel_closures.is_empty());
 	}
 
+	if force_close {
+		// Peer retained after local force-close to allow channel_reestablish recovery.
+		assert!(
+			node_a.list_peers().iter().any(|p| p.node_id == node_b.node_id() && p.is_persisted),
+			"node_b should remain persisted in node_a peer store after locally-initiated force-close"
+		);
+	} else {
+		// Peer removed after cooperative close — no further reason to reconnect.
+		assert!(
+			!node_a.list_peers().iter().any(|p| p.node_id == node_b.node_id() && p.is_persisted),
+			"node_b should be removed from node_a peer store after cooperative close"
+		);
+	}
+
 	let sum_of_all_payments_sat = (push_msat
 		+ invoice_amount_1_msat
 		+ overpaid_amount_msat


src/event.rs
```diff
@@ -1585,11 +1585,47 @@ where
...
+						}
+					}
+				}
+
 				let event = Event::ChannelClosed {
 					channel_id,
 					user_channel_id: UserChannelId(user_channel_id),
 					counterparty_node_id,
-					reason: Some(reason),
+					reason: Some(reason.clone()),
```
Camillarhi: `reason.clone()` is unnecessary as `reason` was not moved by the `match`

src/event.rs
```diff
@@ -1585,11 +1585,47 @@ where
...
+						self.channel_manager.list_channels_with_counterparty(&counterparty_node_id);
+
+					let has_other_channels =
+						all_channels.iter().any(|c| c.channel_id != channel_id);
+
+					let counterparty_initiated = matches!(
+						reason,
+						ClosureReason::CounterpartyForceClosed { .. }
+							| ClosureReason::CounterpartyInitiatedCooperativeClosure
+					);
```
Camillarhi: We should flip the order here and check the closure reason first, so the read on the channel manager only happens when it's actually relevant

src/lib.rs
```diff
@@ -1842,7 +1842,13 @@ impl Node {
 			}
 
 			// Check if this was the last open channel, if so, forget the peer.
-			if open_channels.len() == 1 {
+			//For force-closes we keep the peer
+			// in the store so the background reconnection task can fire and
+			// complete the channel_reestablish recovery flow. This is especially
+			// important for LND peers
+			// The peer will be removed by the ChannelClosed event handler once
+			// the counterparty reconnects and confirms closure.
```
Camillarhi: This comment should flow as one paragraph rather than fragmented lines, and the last sentence is misleading, the `ChannelClosed` event handler only removes the peer on counterparty-initiated closures, so after a local force-close the peer isn't actually cleaned up there. It's retained indefinitely until the user calls `disconnect`.

Something like:

`// If this was the last open channel and we're closing cooperatively, forget
// the peer, since we have no further reason to reconnect. For force-closes we
// intentionally keep the peer in the store so the background reconnection
// task keeps firing and can drive the channel_reestablish recovery flow.
// This is especially important against LND peers, which don't always handle
// force-closure error messages correctly. Note that this means a force-closed
// peer is retained until the user explicitly calls Node::disconnect.`

tests/integration_tests_rust.rs
```diff
@@ -2957,3 +2957,124 @@ async fn splice_in_with_all_balance() {
...
+		!node_a.list_peers().iter().any(|p| p.node_id == node_b_id && p.is_persisted),
+		"node_b should not be persisted in node_a peer store after counterparty force-close"
+	);
+
+	node_a.stop().unwrap();
+	node_b.stop().unwrap();
+}
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn test_peer_retained_on_local_force_close() {
```
Camillarhi: This test can be folded into `do_channel_full_cycle` as it already runs both `force_close == true and false` paths, so the assertion can be done there

Camillarhi (changes requested):
Thanks for working on this. I left some comments. Also, can you squash the commits?

src/event.rs
Jolah1: Good catch, removed.
Thanks!

src/event.rs
Jolah1: That's true, no point querying the channel manager if the closure reason rules out action. 
It has been reordered.

src/lib.rs
Jolah1: You're right, the last sentence was inaccurate.
The handler only removes on counterparty-initiated closures, not local force-closes. 
Replaced with your suggested comment.

tests/integration_tests_rust.rs
Jolah1: Thanks for pointing this out. I have  folded the assertions into `do_channel_full_cycle` alongside the existing force-close paths rather than keeping them as separate tests.

Jolah1:
> Thanks for working on this. I left some comments. Also, can you squash the commits?

Thanks for the thorough review. All four points have been addressed . Will squash the commits once the changes are in.

src/event.rs
```diff
@@ -1585,6 +1585,44 @@ where
...
+				// with us, remove them from the peer store so we stop trying to reconnect.
+				//
+				// If we initiated the closure, keep them in the peer store so the
+				// background reconnection task fires and we can complete the
+				// channel_reestablish recovery flow. This matters especially for LND
+				// peers, which need us to reconnect to recover from force-closures.
+				//
+				// We exclude `channel_id` from the remaining-channel check because LDK
+				// fires ChannelClosed before removing the channel from its internal list,
+				// so list_channels_with_counterparty still includes the closing channel.
```
Camillarhi: This comment partially duplicates the rationale in the `lib.rs` change.  I suggest trimming it to just what this block does, and let the `lib.rs` comment carry the LND recovery rationale

Jolah1: Thank you!
I have  trimmed the comment here to focus on the behavior of this block.

src/event.rs
```diff
@@ -1585,6 +1585,39 @@ where
...
+
+					if counterparty_initiated {
+						let has_other_channels = self
+							.channel_manager
+							.list_channels_with_counterparty(&counterparty_node_id)
+							.iter()
+							.any(|c| c.channel_id != channel_id);
+
+						if !has_other_channels {
+							if let Err(e) = self.peer_store.remove_peer(&counterparty_node_id) {
```
tnull: We usually replay the event in case of persistence failures. While this path is not critical, we should probably do the same here to keep it consistent.

src/lib.rs
```diff
@@ -1840,8 +1841,16 @@ impl Node {
...
-			// Check if this was the last open channel, if so, forget the peer.
-			if open_channels.len() == 1 {
+			// If this was the last open channel and we're closing cooperatively, forget the peer
+			// since we have no further reason to reconnect.
+
+			// For force-closes we intentionally keep the peer in the store so the background reconnection
+			// task keeps firing and can drive the channel_reestablish recovery flow.
+			// This is especially important against LND peers, which don't always handle force-closure error messages correctly.
+
+			//Note that this means a force-closed peer is retained until the user explicitly calls Node::disconnect.
```
tnull: nit: 
```suggestion
			// Note that this means a force-closed peer is retained until the user explicitly calls Node::disconnect.
```

tnull (commented):
Thanks for looking into this. I think technically we could consider cleaning up the stored peer once all funds of the last channel were swept successfully, though I'm not sure if it's worth the complexity adding explicit tracking for it (but let me know if you see an easy/minimally invasive approach to do it).

src/lib.rs
Jolah1: Fixed, Thanks!

src/event.rs
Jolah1: I agree, I've updated this to return ReplayEvent() on failure to keep it consistent with other persistence paths.

Jolah1:
> Thanks for looking into this. I think technically we could consider cleaning up the stored peer once all funds of the last channel were swept successfully, though I'm not sure if it's worth the complexity adding explicit tracking for it (but let me know if you see an easy/minimally invasive approach to do it).

That makes sense, and would be the ideal cleanup point. I don't see a minimally invasive way to track it right now without additional tracking of channel resolution state across events, so I'd lean toward leaving it as a follow-up issue rather than adding complexity here. Happy to open one if that's useful.

Camillarhi:
CI seems to be failing.

Jolah1:
> CI seems to be failing.

Yes, the CI failure was a flaky async_payment test on 1.85.0 — connection refused on the mock Esplora server before it was ready, unrelated to this change. Triggered a fresh run and all checks are now passing.
Thank youuuu.

src/event.rs
```diff
@@ -1585,6 +1585,40 @@ where
 			} => {
 				log_info!(self.logger, "Channel {} closed due to: {}", channel_id, reason);
 
+				// If the counterparty initiated closure of their last remaining channel,
+				// remove them from the peer store so we stop reconnect attempts.
+
+				// We exclude `channel_id` from the check because LDK emits
+				// ChannelClosed before removing the channel from its internal list.
+
+				if let Some(counterparty_node_id) = counterparty_node_id {
```
tnull: nit: The `counterparty_node_id` field is always set post LDK v0.0.117 and LDK Node v0.2 was already based on LDK v0.0.118. We should be good to just `expect`ing it, reducing indetation.

tnull (commented):
Excuse the delay here - LGTM!

Please squash and clean up the commit history while you're at it. All commits should feature a description that (briefly!) outlines the motivation and approach for the change. Feel free to check out https://cbea.ms/git-commit/ for guidance.

src/event.rs
```diff
@@ -1585,6 +1585,40 @@ where
...
+					let counterparty_initiated = matches!(
```

>   - [P2] Include on-chain force-closes in peer cleanup — /home/tnull/workspace/ldk-node/src/event.rs:1595-1599
>     When a peer unilaterally broadcasts its commitment while we are disconnected (or their error message is never processed), LDK closes our channel with ClosureReason::CommitmentTxConfirmed, not CounterpartyForceClosed. Because this match only treats error-message force closes as counterparty
>     initiated, the last-channel peer remains in peer_store and the reconnection loop continues indefinitely after the channel is gone; handle the confirmed commitment/closing-tx case or otherwise clean it up as well.

> • Yes, the retention is intentional, but only for locally initiated force-closes.
> 
>   The review is pointing at a different case: the peer force-closes while we are disconnected, and LDK later reports the terminal close as ClosureReason::CommitmentTxConfirmed rather than CounterpartyForceClosed. At that point the channel is already closed on-chain, so keeping the peer
>   persisted no longer helps channel_reestablish; it just keeps the reconnect loop alive for a peer with no remaining channel.
> 
>   So I’d treat the review as valid. The fix should preserve:
> 
>   - HolderForceClosed { .. }: keep persisted peer, per the comment in src/lib.rs:1847
>   - CounterpartyForceClosed { .. }: remove if it was their last channel
>   - CommitmentTxConfirmed: also remove if it was their last channel, because this is the disconnected/on-chain remote force-close path
> 
>   Caveat: CommitmentTxConfirmed is not a perfect “counterparty initiated” signal; LDK documents that it can also come from our own monitor copy. But for peer-store cleanup after a terminal ChannelClosed, it still seems closer to “no reconnect needed” than to the local HolderForceClosed
>   recovery case.

Jolah1: Thanks, and no worries on the delay!

Squashed down to two commits, each with a description covering the motivation and approach. Also addressed both comments:

Switched to expecting `counterparty_node_id` and dedented the block, since it's always set on LDK ≥ 0.0.117.
Added `CommitmentTxConfirmed` to the cleanup set so a remote force-close observed on-chain while disconnected no longer leaves the peer spinning in the reconnect loop. Renamed the boolean to `reconnect_unneeded` since the set is no longer strictly counterparty-initiated, kept `HolderForceClosed` excluded per the local-recovery rationale, and left a comment noting `CommitmentTxConfirmed` isn't strictly counterparty-initiated but is terminal on-chain either way.

I didn't add a dedicated test for the `CommitmentTxConfirmed `path. it routes through the same removal logic already covered by do_channel_full_cycle, and a faithful test would need a remote commitment broadcast while disconnected, which would be flaky. Happy to add one if thats better.
One CI note: the two VSS integration jobs are failing at a setup step `(cd vss-server/rust: No such file or directory)` — looks like a workflow/environment issue. The Rust build, lint, docs, SemVer, and the CLN/LND/Eclair suites are passing. Let me know if there's a re-trigger you'd like.

Jolah1:
CI is currently failing on `persistence_backwards_compatibility`,  
I reproduced it locally with the CI flags (RUSTFLAGS="--cfg cycle_tests --cfg tokio_unstable") on `main` (`00dba45`) without any of this PR's commits applied.

The failure is a runtime teardown panic:

```text
thread 'persistence_backwards_compatibility' panicked at tokio-1.52.3/src/runtime/blocking/shutdown.rs:51: Cannot drop a runtime in a context where blocking is not allowed.
```

This appears to be a base-branch issue. 
Would appreciate a review of the PR changes when you have a chance.
@tnull

tnull:
> CI is currently failing on `persistence_backwards_compatibility`, I reproduced it locally with the CI flags (RUSTFLAGS="--cfg cycle_tests --cfg tokio_unstable") on `main` (`00dba45`) without any of this PR's commits applied.
> 
> The failure is a runtime teardown panic:
> 
> ```
> thread 'persistence_backwards_compatibility' panicked at tokio-1.52.3/src/runtime/blocking/shutdown.rs:51: Cannot drop a runtime in a context where blocking is not allowed.
> ```
> 
> This appears to be a base-branch issue. Would appreciate a review of the PR changes when you have a chance. @tnull

Yes, this was fixed in https://github.com/lightningdevkit/ldk-node/pull/947. Please rebase now that it landed to get CI to pass.

Jolah1:
> > CI is currently failing on `persistence_backwards_compatibility`, I reproduced it locally with the CI flags (RUSTFLAGS="--cfg cycle_tests --cfg tokio_unstable") on `main` (`00dba45`) without any of this PR's commits applied.
> > The failure is a runtime teardown panic:
> > ```
> > thread 'persistence_backwards_compatibility' panicked at tokio-1.52.3/src/runtime/blocking/shutdown.rs:51: Cannot drop a runtime in a context where blocking is not allowed.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This appears to be a base-branch issue. Would appreciate a review of the PR changes when you have a chance. @tnull
> 
> Yes, this was fixed in #947. Please rebase now that it landed to get CI to pass.

Rebased onto main with #947 - the persistence_backwards_compatibility panic is resolved. 

CI still having flaky integration/infra failures, across re-runs I've seen VSS No space left on device, a BrokenPipe subscribing to Electrum, and now FeerateEstimationUpdateFailed fetching fees from Esplora, three different backend hiccups in tests this PR doesn't touch (the force-close tests pass consistently). 

could you re-run the failed build (ubuntu-latest, beta) job? 
Happy to rebase again if you'd prefer.

tnull:
> CI still having flaky integration/infra failures, across re-runs I've seen VSS No space left on device, a BrokenPipe subscribing to Electrum, and now FeerateEstimationUpdateFailed fetching fees from Esplora, three different backend hiccups in tests this PR doesn't touch (the force-close tests pass consistently).
>
> could you re-run the failed build (ubuntu-latest, beta) job? Happy to rebase again if you'd prefer.

Yeah, most of these are known flakes that happen when Github is too busy/github runners are buggy. I re-ran tests, though note that Kotlin tests were also failing with a 'failed to close channel' error, that seemed legit.

Jolah1:
> > CI still having flaky integration/infra failures, across re-runs I've seen VSS No space left on device, a BrokenPipe subscribing to Electrum, and now FeerateEstimationUpdateFailed fetching fees from Esplora, three different backend hiccups in tests this PR doesn't touch (the force-close tests pass consistently).
> > could you re-run the failed build (ubuntu-latest, beta) job? Happy to rebase again if you'd prefer.
> 
> Yeah, most of these are known flakes that happen when Github is too busy/github runners are buggy. I re-ran tests, though note that Kotlin tests were also failing with a 'failed to close channel' error, that seemed legit.

Thanks for re-running! @tnull 

I Looked into the Kotlin failed to close channel: it's the monitor-update race in fullCycle(), not this PR. 
The log shows channel monitor update is in progress but not yet complete right before fullCycle does a 
cooperative close immediately after a payment (LibraryTest.kt:307), so the close fires before the update from that payment settles, and LDK returns the "waiting on a monitor update" error.

This PR's changes runs after a close, not on the close-initiation path that failed: the lib.rs cleanup (&& !force) runs after close_channel returns Ok, and the event.rs change runs in the ChannelClosed handler (after 
the channel has already closed). Neither touches the live connection. The re-run passed on the exact same commit (1af7e7c) with no changes.
All 23 checks are green now.

Jolah1:
> > > CI still having flaky integration/infra failures, across re-runs I've seen VSS No space left on device, a BrokenPipe subscribing to Electrum, and now FeerateEstimationUpdateFailed fetching fees from Esplora, three different backend hiccups in tests this PR doesn't touch (the force-close tests pass consistently).
> > > could you re-run the failed build (ubuntu-latest, beta) job? Happy to rebase again if you'd prefer.
> > 
> > 
> > Yeah, most of these are known flakes that happen when Github is too busy/github runners are buggy. I re-ran tests, though note that Kotlin tests were also failing with a 'failed to close channel' error, that seemed legit.
> 
> Thanks for re-running! @tnull
> 
> I Looked into the Kotlin failed to close channel: it's the monitor-update race in fullCycle(), not this PR. The log shows channel monitor update is in progress but not yet complete right before fullCycle does a cooperative close immediately after a payment (LibraryTest.kt:307), so the close fires before the update from that payment settles, and LDK returns the "waiting on a monitor update" error.
> 
> This PR's changes runs after a close, not on the close-initiation path that failed: the lib.rs cleanup (&& !force) runs after close_channel returns Ok, and the event.rs change runs in the ChannelClosed handler (after the channel has already closed). Neither touches the live connection. The re-run passed on the exact same commit ([1af7e7c](https://github.com/lightningdevkit/ldk-node/commit/1af7e7c47ec8eaa8735b5efeb23fe6598b53bd68)) with no changes. All 23 checks are green now.

@tnull

src/lib.rs
```diff
@@ -1919,8 +1919,16 @@ impl Node {
...
+			// since we have no further reason to reconnect.
+
+			// For force-closes we intentionally keep the peer in the store so the background reconnection
+			// task keeps firing and can drive the channel_reestablish recovery flow.
+			// This is especially important against LND peers, which don't always handle force-closure error messages correctly.
+
+			// Note that this means a force-closed peer is retained until the user explicitly calls Node::disconnect.
+
+			if open_channels.len() == 1 && !force {
 				self.runtime.block_on(self.peer_store.remove_peer(&counterparty_node_id))?;
```
tnull: I do wonder if we actually still want to remove the peer here at all, or if we should always leave it to `ChannelClosed` handling?

src/event.rs
```diff
@@ -1613,10 +1613,49 @@ where
...
+
+				// Drop the peer once their last channel with us has reached a terminal
+				// state reconnection cannot recover. `CommitmentTxConfirmed` is included
+				// because LDK reports a remote (or our own) commitment confirming on-chain
+				// via this variant, leaving nothing for `channel_reestablish` to recover.
+				// `HolderForceClosed` is deliberately excluded so reconnection can still
+				// drive recovery (see `Node::close_channel_internal`).
+				// We exclude `channel_id` from the count because LDK emits `ChannelClosed`
+				// before removing it from its internal list.
+				let reconnect_unneeded = matches!(
```

>   - P2: /home/tnull/workspace/ldk-node-pr-895/src/event.rs:1628 misses ClosureReason::CounterpartyCoopClosedUnfundedChannel. Since outbound channel creation persists the peer immediately at /home/tnull/workspace/ldk-node-pr-895/src/lib.rs:1316, a counterparty close before funding leaves the
>     last peer in PeerStore, so the reconnection task keeps retrying even though the channel is terminal. This looks like the same class of cleanup the PR is trying to add for counterparty-driven closes.

(also, nit: `dont_reconnect` would be a bit more straightforward naming IMO.

tnull (commented):
Excuse the delay here! Two minor comments. 

Maybe we could also add some small test coverage to assert the behavior now matches our expectation?

src/lib.rs
Jolah1: Thats right, left it all to ChannelClosed. close_channel_internal no longer touches the peer store; the handler is now the single owner of cleanup.

src/event.rs
Jolah1: Inverted the check instead of extending the allowlist: let dont_reconnect = !matches!(reason, ClosureReason::HolderForceClosed { .. });. HolderForceClosed is the only reason we keep reconnecting (for channel_reestablish recovery); every other terminal reason drops the peer once it has no other channel so CounterpartyCoopClosedUnfundedChannel and the rest of the pre-funding/terminal class are covered in one go. Also took the `dont_reconnect` rename.

Jolah1:
> Excuse the delay here! Two minor comments.
> 
> Maybe we could also add some small test coverage to assert the behavior now matches our expectation?

It's Fine, Thanks for the Follow Up!

I Added peer_removed_when_counterparty_force_closes_last_channel: opens (and persists) the counterparty, has them force-close the last channel, and asserts the peer is dropped. 
do_channel_full_cycle still covers the local force-close/coop paths.

src/event.rs
```diff
@@ -1613,10 +1613,45 @@ where
 			} => {
 				log_info!(self.logger, "Channel {} closed due to: {}", channel_id, reason);
 
+				// `counterparty_node_id` has been set on every `ChannelClosed` since LDK 0.0.117.
+				let counterparty_node_id = counterparty_node_id
+					.expect("counterparty_node_id is always set since LDK 0.0.117");
```
tnull, line 1618: So this should be fine, but if we do that we probably also want to make our field `required` and document it in the changelog. I'll do that in a follow-up.

tnull (approved):
Thanks!

PR #898, opened 2026-05-04
Support LNURL resolution via optional HTTP fallback

### LINKED ISSUE #870: Unified send doesn't support lnurl
We use ldk's `LDKOnionMessageDNSSECHrnResolver` to resolve HRNs which doesn't support lnurl. 

Some options to fix this:

- Replace `LDKOnionMessageDNSSECHrnResolver` with `HttpResolver`: simplest but loses privacy benefits
- Update `LDKOnionMessageDNSSECHrnResolver` to optionally support lnurl through providing a http resolver
- Update bitcoin-payment-instructions to support providing multiple resolvers.

benthecarman:
Closes #870

Add an `enable_lnurl_resolution` flag to `HumanReadableNamesConfig` that, when set, attaches an `HTTPHrnResolver` as a fallback to the configured BIP 353 / bLIP-32 resolver. The fallback is consulted only when the primary resolver returns an error, so it covers LN-Address recipients that don't publish a BIP 353 record and direct LNURL-pay links, while never shadowing a successful primary resolution.

Enabling this opts in to revealing our IP and payment target to the LNURL endpoint, so it defaults to off.

Co-developed with Claude Code (claude-opus-4-7).

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -266,6 +266,18 @@ pub struct HumanReadableNamesConfig {
 	/// * **DNS Server**: `8.8.8.8:53` (Google Public DNS)
 	/// * **Resolution Service**: Disabled (`false`)
 	pub resolution_config: HRNResolverConfig,
+	/// If set to true, enables resolving [LNURL-pay] links and the LN-Address LNURL fallback
+	/// via HTTP, in addition to the BIP 353 / bLIP-32 resolution method configured in
+	/// [`resolution_config`].
+	///
+	/// **Default:** `false`
+	///
+	/// **Note:** Enabling this may reveal our IP address to the recipient and information about
+	/// who we're paying to the LNURL endpoint we're querying.
+	///
+	/// [LNURL-pay]: https://github.com/lnurl/luds/blob/luds/06.md
+	/// [`resolution_config`]: Self::resolution_config
+	pub enable_lnurl_resolution: bool,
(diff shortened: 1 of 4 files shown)


TheBlueMatt:
This feels like it belongs in bitcoin-payment-instructions.

benthecarman:
> This feels like it belongs in bitcoin-payment-instructions.

opened https://github.com/rust-bitcoin/bitcoin-payment-instructions/pull/23 however I still think we'd still want some of this unless we want to add a fallback to the dns resolver too

src/config.rs
```diff
@@ -266,6 +266,18 @@ pub struct HumanReadableNamesConfig {
...
+	/// [`resolution_config`].
+	///
+	/// **Default:** `false`
+	///
+	/// **Note:** Enabling this may reveal our IP address to the recipient and information about
+	/// who we're paying to the LNURL endpoint we're querying.
+	///
+	/// [LNURL-pay]: https://github.com/lnurl/luds/blob/luds/06.md
+	/// [`resolution_config`]: Self::resolution_config
+	pub enable_lnurl_resolution: bool,
```
tnull, line 280: Hmm, I'm not sure we want this to be a config option to begin with? If this is just due to the possible privacy leak, it should probably be part of the respective variant on the resolution config?

PR #899, opened 2026-05-06
Move BOLT11 JIT params to payment metadata

tnull:
## Context 

We recently found that for the intended BOLT12-JIT flow we'll have to encode the LSPS2 parameters in the payment metadata (based on https://github.com/lightningdevkit/rust-lightning/pull/4584). As a prefactor to that (and to simplify things for #811, rather than add yet another store type there, just to revert once we get to the BOLT12-JIT changes), we here switch to store the LSPS2 parameters in the BOLT11 `payment_metadata` rather than persisting them on-disk. To make this safe we'll also need https://github.com/lightningdevkit/rust-lightning/pull/4528, as otherwise the payer could collude with the LSP to rob the payee.

## Summary

  This moves LSPS2/JIT-channel receive parameters out of dedicated payment-store
  state and into the BOLT11 invoice `payment_metadata`, so the payment store no
  longer needs a `PaymentKind::Bolt11Jit` variant for new payments.

## Changes

  - Rename `LSPFeeLimits` to `LSPS2Parameters`.
  - Add `Bolt11PaymentMetadata` in `bolt11.rs` with TLV-based encoding.
  - Encode `LSPS2Parameters` into BOLT11 `payment_metadata` when creating JIT
    invoices.
  - Store new JIT receives as `PaymentKind::Bolt11`.
  - Decode legacy `PaymentKind::Bolt11Jit` records as `PaymentKind::Bolt11`.
  - Reject skimmed inbound payments unless valid BOLT11 `payment_metadata` proves
    the LSPS2 fee is within the negotiated limit.
  - Drop the stale top-level `PaymentDetails` field-1 JIT metadata reader, while
    keeping the released legacy `PaymentKind::Bolt11Jit` decoder.
  - Add a changelog note that pending JIT payments may fail after upgrade because
    prior JIT state is not migrated.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -50,6 +50,7 @@ use crate::payment::asynchronous::static_invoice_store::StaticInvoiceStore;
 use crate::payment::store::{
 	PaymentDetails, PaymentDetailsUpdate, PaymentDirection, PaymentKind, PaymentStatus,
 };
+use crate::payment::PaymentMetadata;
 use crate::runtime::Runtime;
 use crate::types::{
 	CustomTlvRecord, DynStore, KeysManager, OnionMessenger, PaymentStore, Sweeper, Wallet,
@@ -580,6 +581,36 @@ where
 		}
 	}
 
+	fn fail_claimable_payment(
+		&self, payment_id: PaymentId, payment_hash: &PaymentHash,
+	) -> Result<(), ReplayEvent> {
+		self.channel_manager.fail_htlc_backwards(payment_hash);
+
+		let update = PaymentDetailsUpdate {
+			hash: Some(Some(*payment_hash)),
+			status: Some(PaymentStatus::Failed),
+			..PaymentDetailsUpdate::new(payment_id)
+		};
+		match self.payment_store.update(update) {
+			Ok(_) => Ok(()),
+			Err(e) => {
+				log_error!(self.logger, "Failed to access payment store: {}", e);
+				Err(ReplayEvent())
+			},
+		}
+	}
+
+	fn lsps2_max_total_opening_fee_msat(payment_metadata: &[u8], amount_msat: u64) -> Option<u64> {
+		let metadata = PaymentMetadata::read(&mut &payment_metadata[..]).ok()?;
+		let lsps2_parameters = metadata.lsps2_parameters?;
+		lsps2_parameters.max_total_opening_fee_msat.or_else(|| {
+			lsps2_parameters.max_proportional_opening_fee_ppm_msat.and_then(|max_prop_fee| {
+				// If it's a variable amount payment, compute the actual fee.
+				compute_opening_fee(amount_msat, 0, max_prop_fee)
+			})
+		})
+	}
+
 	pub async fn handle_event(&self, event: LdkEvent) -> Result<(), ReplayEvent> {
 		match event {
 			LdkEvent::FundingGenerationReady {
@@ -693,7 +724,8 @@ where
 				..
 			} => {
 				let payment_id = PaymentId(payment_hash.0);
-				if let Some(info) = self.payment_store.get(&payment_id) {
+				let payment_info = self.payment_store.get(&payment_id);
+				if let Some(info) = payment_info.as_ref() {
 					if info.direction == PaymentDirection::Outbound {
 						log_info!(
 							self.logger,
@@ -716,14 +748,13 @@ where
 					}
 
 					if info.status == PaymentStatus::Succeeded
-						|| matches!(info.kind, PaymentKind::Spontaneous { .. })
+						|| matches!(&info.kind, PaymentKind::Spontaneous { .. })
 					{
-						let stored_preimage = match info.kind {
+						let stored_preimage = match &info.kind {
 							PaymentKind::Bolt11 { preimage, .. }
-							| PaymentKind::Bolt11Jit { preimage, .. }
 							| PaymentKind::Bolt12Offer { preimage, .. }
 							| PaymentKind::Bolt12Refund { preimage, .. }
-							| PaymentKind::Spontaneous { preimage, .. } => preimage,
+							| PaymentKind::Spontaneous { preimage, .. } => *preimage,
 							_ => None,
 						};
 
@@ -758,22 +789,28 @@ where
 							},
 						};
 					}
+				}
 
-					let max_total_opening_fee_msat = match info.kind {
-						PaymentKind::Bolt11Jit { lsp_fee_limits, .. } => {
-							lsp_fee_limits
-								.max_total_opening_fee_msat
-								.or_else(|| {
-									lsp_fee_limits.max_proportional_opening_fee_ppm_msat.and_then(
-										|max_prop_fee| {
-											// If it's a variable amount payment, compute the actual fee.
-											compute_opening_fee(amount_msat, 0, max_prop_fee)
-										},
-									)
-								})
-								.unwrap_or(0)
-						},
-						_ => 0,
+				if counterparty_skimmed_fee_msat > 0 {
+					let max_total_opening_fee_msat = match &purpose {
+						PaymentPurpose::Bolt11InvoicePayment { .. } => onion_fields
+							.as_ref()
+							.and_then(|fields| fields.payment_metadata.as_ref())
+							.and_then(|metadata| {
+								Self::lsps2_max_total_opening_fee_msat(metadata, amount_msat)
+							}),
+						_ => None,
+					};
+
+					let Some(max_total_opening_fee_msat) = max_total_opening_fee_msat else {
+						log_info!(
+							self.logger,
+							"Refusing inbound payment with hash {} as the counterparty withheld {}msat without valid BOLT11 LSPS2 payment metadata",
+							hex_utils::to_string(&payment_hash.0),
+							counterparty_skimmed_fee_msat,
+						);
+						self.fail_claimable_payment(payment_id, &payment_hash)?;
+						return Ok(());
 					};
 
 					if counterparty_skimmed_fee_msat > max_total_opening_fee_msat {
@@ -784,26 +821,13 @@ where
 							counterparty_skimmed_fee_msat,
 							max_total_opening_fee_msat,
 						);
-						self.channel_manager.fail_htlc_backwards(&payment_hash);
-
-						let update = PaymentDetailsUpdate {
-							hash: Some(Some(payment_hash)),
-							status: Some(PaymentStatus::Failed),
-							..PaymentDetailsUpdate::new(payment_id)
-						};
-						match self.payment_store.update(update) {
-							Ok(_) => return Ok(()),
-							Err(e) => {
-								log_error!(self.logger, "Failed to access payment store: {}", e);
-								return Err(ReplayEvent());
-							},
-						};
+						self.fail_claimable_payment(payment_id, &payment_hash)?;
+						return Ok(());
 					}
 
-					// If the LSP skimmed anything, update our stored payment.
-					if counterparty_skimmed_fee_msat > 0 {
-						match info.kind {
-							PaymentKind::Bolt11Jit { .. } => {
+					if let Some(info) = payment_info.as_ref() {
+						match &info.kind {
+							PaymentKind::Bolt11 { .. } => {
 								let update = PaymentDetailsUpdate {
 									counterparty_skimmed_fee_msat: Some(Some(counterparty_skimmed_fee_msat)),
 									..PaymentDetailsUpdate::new(payment_id)
@@ -816,16 +840,17 @@ where
 									},
 								};
 							}
-							_ => debug_assert!(false, "We only expect the counterparty to get away with withholding fees for JIT payments."),
+							_ => debug_assert!(false, "We only expect the counterparty to get away with withholding fees for BOLT11 payments."),
 						}
 					}
+				}
 
+				if let Some(info) = payment_info {
 					// If this is known by the store but ChannelManager doesn't know the preimage,
 					// the payment has been registered via `_for_hash` variants and needs to be manually claimed via
 					// user interaction.
 					match info.kind {
-						PaymentKind::Bolt11 { preimage, .. }
-						| PaymentKind::Bolt11Jit { preimage, .. } => {
+						PaymentKind::Bolt11 { preimage, .. } => {
 							if purpose.preimage().is_none() {
 								debug_assert!(
 									preimage.is_none(),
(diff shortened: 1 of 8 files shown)


src/payment/bolt11.rs
```diff
@@ -201,7 +201,7 @@ impl Bolt11Payment {
 		// Register payment in payment store.
 		let payment_hash = invoice.payment_hash();
 		let payment_secret = invoice.payment_secret();
-		let lsp_fee_limits = LSPFeeLimits {
+		let lsp_fee_limits = LSPS2Parameters {
```
joostjager: Not sure if this rename is strictly better. Parameters sounds broader than what it is.

src/liquidity.rs
```diff
@@ -1346,7 +1359,11 @@ where
 			.current_timestamp()
 			.min_final_cltv_expiry_delta(min_final_cltv_expiry_delta.into())
 			.expiry_time(Duration::from_secs(expiry_secs.into()))
-			.private_route(route_hint);
+			.private_route(route_hint)
+			.payment_metadata(
+				Bolt11PaymentMetadata { lsps2_parameters: Some(lsps2_parameters) }.encode(),
+			)
+			.require_payment_metadata();
```
joostjager: Is it safe to assume that this is universally supported by senders?

tests/integration_tests_rust.rs
```diff
@@ -1940,7 +1940,7 @@ async fn do_lsps2_client_service_integration(client_trusts_lsp: bool) {
 		expect_payment_received_event!(client_node, expected_received_amount_msat).unwrap();
 	let client_payment = client_node.payment(&client_payment_id).unwrap();
 	match client_payment.kind {
-		PaymentKind::Bolt11Jit { counterparty_skimmed_fee_msat, .. } => {
+		PaymentKind::Bolt11 { counterparty_skimmed_fee_msat, .. } => {
```
joostjager, line 1958: Would it be worth to add some coverage for "skimmed fee with missing/malformed metadata" failing?

joostjager (commented):
PR looks great. I like the offloading of state to clients. And ofc happy to see usage of https://github.com/lightning/bolts/pull/912 😄

src/payment/bolt11.rs
tnull: Yes, that's intentional as we might add more fields in the BOLT12 context that are not 'fee limits'. Sorry, maybe should have given that rationale in the commit description.

joostjager: Which parameters are that?

tnull: > Which parameters are that?

See https://github.com/lightningdevkit/rust-lightning/pull/4463/changes#diff-dd4e4cab42ecc909f892759965a0ef44080f865dcae8173f084d960f296406b5R32-R39

src/liquidity.rs
tnull: Asked claude, also discussed elsewhere. It seems we should be good by now. Wallets are supporting your stuff!

tests/integration_tests_rust.rs
tnull: I added a small fixup that basically asserts that, let me know if you deem this sufficient. Note that asserting `handle_event` behavior overall is a bit tricky, and we don't have a good way to create integration tests that feature LSPs maliciously withholding funds.

tnull:
Now updated and included a commit that allows us to de/encrypt `PaymentMetadata`, as we should do that given that otherwise the payer will see whatever we put into that. Currently the LSP parameters aren't *super* sensitive, but especially given we're adding the general struct for expandibility, we want to make sure encrypt from the start to avoid any privacy leakage.

src/payment/metadata.rs
```diff
@@ -0,0 +1,204 @@
...
+
+impl PaymentMetadataKeys {
+	pub(crate) fn new(base_secret: [u8; 32]) -> Self {
+		Self {
+			encryption_key: hmac_sha256(&base_secret, b"ldk_node_payment_metadata_encryption_key"),
+			nonce_key: hmac_sha256(&base_secret, b"ldk_node_payment_metadata_nonce_key"),
+		}
+	}
+
+	fn nonce(&self, payment_hash: &PaymentHash, payment_secret: &PaymentSecret) -> [u8; 12] {
```
joostjager: Is it totally ruled out that payment hash and secret are never reused, also not in some manual flow for example? Perhaps defensively picking a random nonce and storing it in the metadata is safer?

src/payment/metadata.rs
```diff
@@ -0,0 +1,204 @@
...
+#[derive(Clone, Copy)]
+pub(crate) struct PaymentMetadataKeys {
+	encryption_key: [u8; 32],
+	nonce_key: [u8; 32],
+}
+
+impl PaymentMetadataKeys {
+	pub(crate) fn new(base_secret: [u8; 32]) -> Self {
+		Self {
+			encryption_key: hmac_sha256(&base_secret, b"ldk_node_payment_metadata_encryption_key"),
```
joostjager: With random nonces, this derivation isn't necessary anymore because base_secret is just the encryption key, I think

src/payment/metadata.rs
```diff
@@ -0,0 +1,204 @@
...
+
+impl PaymentMetadata {
+	pub(crate) fn encrypt(
+		&self, keys: &PaymentMetadataKeys, payment_hash: &PaymentHash,
+		payment_secret: &PaymentSecret,
+	) -> EncryptedPaymentMetadata {
+		let nonce = keys.nonce(payment_hash, payment_secret);
+		let mut ciphertext = sealed::PaymentMetadataTlv::from(self.clone()).encode();
+		let cipher = ChaCha20Poly1305::new(Key::new(keys.encryption_key), Nonce::new(nonce));
+		let tag = cipher.encrypt(&mut ciphertext, Some(PAYMENT_METADATA_AAD));
```
joostjager: I would really check to see if anything useful can be exposed from `lightning/src/crypto` rather than doing it again here. And maybe also just drop encryption from this PR until there is an answer to that.

src/liquidity.rs
joostjager: I'd expected it by now indeed. Good!

src/liquidity.rs
```diff
@@ -1338,6 +1355,10 @@ where
 			htlc_maximum_msat: None,
 		}]);
 
+		let payment_metadata = PaymentMetadata { lsps2_parameters: Some(lsps2_parameters) }
+			.encrypt(&payment_metadata_keys, &payment_hash, &payment_secret)
```
joostjager: For bolt12, I think the encryption happens on the blinded path level? Is there anything in the design of this PR that would then double-encrypt when bolt12 is added?

src/payment/metadata.rs
tnull: Yes, pretty much. If we reuse the payment hash for example we have worse issues than just privacy leakage at our hands. Happy to store the nonce if you insist, but note that we try to keep invoices as small as possible, in particular for QR encoding.

src/payment/metadata.rs
tnull: Hmm, doing one static derivation step is still best practice. Would leave it in.

src/payment/metadata.rs
tnull: Well, upstream we use the same library ([chacha20poly1305](https://crates.io/crates/chacha20poly1305)), which is why adding the direct dependency is also fine, IMO.

src/liquidity.rs
tnull: Yes, for BOLT12 we could leak `PaymentMetadataTlvs` and use that directly, potentially, but I'd like to get into those details when we get to the BOLT12 part.

joostjager: Just checking that there are no obvious rewrites necessary, but seems good then

src/payment/metadata.rs
joostjager: Still I think it is better to not rely on that, but might be worth getting a 2nd opinion.

src/payment/metadata.rs
joostjager: What I was trying to question is whether RL has a reusable layer above the raw ChaCha call, something that already standardizes the envelope, key separation, nonce handling, etc.

src/payment/metadata.rs
joostjager: Related question: if size is important, is the current scheme minimal? Perhaps the double tlv wrapper and/or u64 can be shaved down too.

src/payment/metadata.rs
joostjager: Ok

src/payment/metadata.rs
joostjager: And with https://github.com/lightningdevkit/rust-lightning/pull/4528, perhaps the tag isn't needed anymore for auth, because auth is already on the rust-lightning level?

src/payment/metadata.rs
tnull: No, nothing reusable. We now also moved forward with https://github.com/lightningdevkit/rust-lightning/pull/4528, without introducing any utilities there (see https://github.com/lightningdevkit/rust-lightning/pull/4528#discussion_r3215476972). So will need the code to live here, which I'm fine with.

src/payment/metadata.rs
tnull: > Related question: if size is important, is the current scheme minimal? Perhaps the double tlv wrapper and/or u64 can be shaved down too.

We want the wrapper for clean extensibility, and I'd rather not go down the road of using something besides `u64` for `msat` values, as we're very consistent about that, and we soon want to switch to a `LightningAmount` type that will be serialized as `u64`, so hopefully we can just use that everywhere in-place.

tnull:
@joostjager Let me know if you have any other questions, otherwise it would be good to land this soon to keep moving.

joostjager (commented):
Only open item for me is a [2nd reviewer opinion on the nonce generation](https://github.com/lightningdevkit/ldk-node/pull/899#discussion_r3209192068)

tnull:
> Only open item for me is a [2nd reviewer opinion on the nonce generation](https://github.com/lightningdevkit/ldk-node/pull/899#discussion_r3209192068)

Alright, don't have a too strong opinion.

@TheBlueMatt Do you have an opinion whether to deterministically derive the nonces (from payment hash and secret) or whether to append them to the encrypted payment metadata blob that lands in `m`?

src/payment/metadata.rs
TheBlueMatt: Why are we HMAC'ing this rather than just encrypting? Is that required anywhere we're storing this? AFAIU this ends up in the onion and every byte counts.

src/payment/metadata.rs
TheBlueMatt: Not quite sure how this is supposed to work? As of https://github.com/lightningdevkit/rust-lightning/pull/4528 we need the encrypted data first to calculate the payment secret, so we can't encrypt using the payment secret has a key.

Not quite sure what the solution is, though, it does seem like we should be able to just encrypt without an extra IV. For auto-generated-payment_hashes, the payment secret includes a random IV, so we should be able to use that, it needs an API tweak though. For manually-provided payment_hashes I think using the payment_hash as the encryption key is fine - we're ultimately storing information about a specific payment in the invoice. If someone generates a duplicate bolt 11 invoice with the same payment hash and the same amount and expiry, that's the same payment - an attacker being able to swap the payment metadata between them isn't an interesting attack imo.

src/payment/metadata.rs
tnull: Right, see https://github.com/lightningdevkit/ldk-node/pull/899#discussion_r3217979562, Joost also brough up if we could save some bytes by dropping the tag and assuming it empty. Probably makes sense.

src/payment/metadata.rs
tnull: > Not quite sure how this is supposed to work? As of [lightningdevkit/rust-lightning#4528](https://github.com/lightningdevkit/rust-lightning/pull/4528) we need the encrypted data first to calculate the payment secret, so we can't encrypt using the payment secret has a key.

Yeah, lol, I see your point. Not sure I missed that after reviewing #4528. 

> Not quite sure what the solution is, though, it does seem like we should be able to just encrypt without an extra IV. For auto-generated-payment_hashes, the payment secret includes a random IV, so we should be able to use that, it needs an API tweak though. For manually-provided payment_hashes I think using the payment_hash as the encryption key is fine - we're ultimately storing information about a specific payment in the invoice. If someone generates a duplicate bolt 11 invoice with the same payment hash and the same amount and expiry, that's the same payment - an attacker being able to swap the payment metadata between them isn't an interesting attack imo.

Well, if we don't want to randomize and store nonces, this brings up the question again I had raised on https://github.com/lightningdevkit/rust-lightning/pull/4528#discussion_r3208402018: Should we just add the entire payment-metadata encryption wrapper upstream (maybe even as a simple `.encrypt_metadata()` call on the `InvoiceBuilder`?), as we'd also have access to all required data there, without leaking them in the API?

tnull:
Alright, now reverted back to just encoding the `PaymentMetadata` struct as we encrypt the payment metadata in LDK. Also bumped the dependency to include these changes.

src/payment/bolt11.rs
```diff
@@ -119,9 +119,14 @@ impl Bolt11Payment {
...
 			// If the user hasn't registered a custom payment hash, we're positive ChannelManager
 			// will know the preimage at this point.
+			let mut payment_metadata = invoice.payment_metadata().cloned();
 			let res = self
 				.channel_manager
-				.get_payment_preimage(payment_hash, payment_secret.clone(), None)
+				.get_payment_preimage_decrypt_metadata(
+					payment_hash,
+					payment_secret.clone(),
+					payment_metadata.as_deref_mut(),
```
joostjager, line 139: Purely as a bump commit, I think this commit does a bit too much such as decrypting and discarding the metadata?

src/event.rs
```diff
@@ -580,6 +581,35 @@ where
 		}
 	}
 
+	fn fail_claimable_payment(
```
joostjager, line 584: Could have been independent extraction refactor. Might have been easier to spot that it now no longer sets `hash: Some(Some(payment_hash))`.

joostjager (dismissed):
I think it turned out well with the encryption in rust-lightning.

src/payment/bolt11.rs
tnull: Well, it does essenntially only adjust to the new APIs.

src/event.rs
tnull: Hmm, actually good catch. Dropping that wasn't intentional in this PR and while it's not a behavioral change right now, it might be in the future when we implement BOLT12<>LSPS2. So added it back defensively for now, even if it's technically not necessary.

src/payment/bolt11.rs
joostjager: Strictly speaking I think `None` could have been passed here because the decrypted value isn't used in this commit?

src/event.rs
joostjager: I thought I'd be rewarded with a separate extraction commit for this discovery, but no 😅

joostjager (approved):

PR #904, opened 2026-05-14
Derive `PartialEq` and `Eq` for `AsyncPaymentsRole`

benthecarman:
In https://github.com/lightningdevkit/ldk-server/pull/208 I needed to use `matches!` instead of an `assert_eq!`


Camillarhi (approved):
LGTM

tnull (approved):
Sure, why not.

PR #905, opened 2026-05-17
tests: Add interoperability tests for BOLT12 (Offers) with CLN

### LINKED ISSUE #856: Add BOLT12 scenario coverage to interop test harness
## Summary

Follow-up from #839. Add BOLT12 (offers) test scenarios to the interop test harness.

## Context

The current interop test harness covers BOLT11 payments, keysend, and on-chain operations. BOLT12 offer-based payments should also be tested against external implementations that support it.

Ref: https://github.com/lightningdevkit/ldk-node/pull/839#issuecomment-4148400285

alexgrad42:
Activates scenarios for BOLT12 offer payments between ldk-node and Core Lightning (CLN) inside integration tests

Fix #856

### DIFF
--- a/tests/common/cln.rs
+++ b/tests/common/cln.rs
@@ -181,12 +181,81 @@ impl ExternalNode for TestClnNode {
 		Ok(invoice.bolt11)
 	}
 
+	async fn create_offer(
+		&self, amount_msat: u64, description: &str,
+	) -> Result<String, TestFailure> {
+		let desc = description.to_string();
+		let label = format!(
+			"{}-{}",
+			desc,
+			std::time::SystemTime::now()
+				.duration_since(std::time::UNIX_EPOCH)
+				.unwrap_or_default()
+				.as_nanos()
+		);
+
+		let params = serde_json::json!({
+			"amount": format!("{}msat", amount_msat),
+			"description": desc,
+			"label": label,
+			"single_use": true,
+		});
+
+		let response: serde_json::Value = self
+			.rpc(move |c| c.call("offer", &params))
+			.await
+			.map_err(|e| self.make_error(format!("offer RPC call failed: {}", e)))?;
+
+		let offer_str = response["bolt12"]
+			.as_str()
+			.ok_or_else(|| {
+				self.make_error("Failed to parse 'bolt12' from CLN response".to_string())
+			})?
+			.to_string();
+
+		Ok(offer_str)
+	}
+
 	async fn pay_invoice(&self, invoice: &str) -> Result<String, TestFailure> {
 		let inv = invoice.to_string();
 		let result = self
 			.rpc(move |c| c.pay(&inv, PayOptions::default()))
 			.await
 			.map_err(|e| self.make_error(format!("pay: {}", e)))?;
+		Ok(result.payment_preimage)
+	}
+
+	async fn pay_offer(
+		&self, offer_str: &str, amount_msat: Option<u64>,
+	) -> Result<String, TestFailure> {
+		let offer = offer_str.to_string();
+
+		let mut fetch_params = serde_json::json!({
+			"offer": offer,
+			"quantity": 1,
+		});
+
+		if let Some(msat) = amount_msat {
+			fetch_params["amount_msat"] = serde_json::json!(format!("{}msat", msat));
+		}
+
+		let fetch_response: serde_json::Value =
+			self.rpc(move |c| c.call("fetchinvoice", fetch_params)).await.map_err(|e| {
+				self.make_error(format!("fetchinvoice RPC call failed for BOLT12: {}", e))
+			})?;
+
+		let inv = fetch_response["invoice"]
+			.as_str()
+			.ok_or_else(|| {
+				self.make_error("Failed to parse 'invoice' from fetchinvoice response".to_string())
+			})?
+			.to_string();
+
+		let result = self
+			.rpc(move |c| c.pay(&inv, PayOptions::default()))
+			.await
+			.map_err(|e| self.make_error(format!("pay: {}", e)))?;
+
 		Ok(result.payment_preimage)
 	}
 
--- a/tests/common/eclair.rs
+++ b/tests/common/eclair.rs
@@ -211,6 +211,12 @@ impl ExternalNode for TestEclairNode {
 		Ok(invoice.to_string())
 	}
 
+	async fn create_offer(
+		&self, amount_msat: u64, description: &str,
+	) -> Result<String, TestFailure> {
+		Err(self.make_error("create_offer is not supported on Eclair".to_string()))
+	}
+
 	async fn pay_invoice(&self, invoice: &str) -> Result<String, TestFailure> {
 		let result = self.post("/payinvoice", &[("invoice", invoice)]).await?;
 		let payment_id = result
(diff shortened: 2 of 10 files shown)


tests/common/scenarios/mod.rs
```diff
@@ -87,6 +88,29 @@ pub(crate) async fn wait_for_htlcs_settled(
 	panic!("HTLCs did not settle on {} channel {} within 15s", peer.name(), ext_channel_id);
 }
 
+/// Blocks execution until the channel with the specified peer becomes active (`is_usable == true`).
+/// Maximum wait time is 15 seconds (30 attempts with a 500ms interval).
+pub(crate) async fn wait_for_channel_usable(node: &Node, counterparty_node_id: PublicKey) {
```
tnull: We shouldn't use this manual sleep, we have events for that.

tests/common/scenarios/mod.rs
```diff
@@ -201,11 +256,38 @@ pub(crate) async fn force_close_after_payment_scenario<E: ElectrumApi>(
...
+		},
+	}
+
 	wait_for_htlcs_settled(peer, &ext_ch).await;
 	channel::force_close(node, peer, bitcoind, electrs, &user_ch, &ext_ch, Side::Ldk).await;
 }
 
+/// Specialized version of `force_close_after_payment_scenario` for BOLT11 payments.
+/// See [`force_close_after_payment_scenario`] for details.
+pub(crate) async fn force_close_after_payment_bolt11_scenario<E: ElectrumApi>(
```
tnull: I don't understand why we need all this additional boilerplate.

tests/common/scenarios/mod.rs
```diff
@@ -245,7 +328,33 @@ pub(crate) async fn splice_in_scenario<E: ElectrumApi>(
 	sync_wallets_with_retry(node).await;
 	expect_channel_ready_event!(node, ext_node_id);
 
-	payment::send_bolt11_to_peer(node, peer, 5_000_000, "post-splice").await;
+	match protocol {
+		PaymentProtocol::Bolt11 => {
+			payment::send_bolt11_to_peer(node, peer, 5_000_000, "post-splice-bolt11").await;
+		},
+		PaymentProtocol::Bolt12 => {
+			// Wait for Onion Message router updates the channel to usable state.
```
tnull: I don't think I understand what "Wait for Onion Message router updates the channel to usable state." means. Can you expand?

tests/common/scenarios/mod.rs
```diff
@@ -245,7 +328,33 @@ pub(crate) async fn splice_in_scenario<E: ElectrumApi>(
...
+			payment::send_bolt12_to_peer(node, peer, 5_000_000, "post-splice-bolt12").await;
+		},
+	}
 
 	channel::cooperative_close(node, peer, bitcoind, electrs, &user_ch, &ext_ch, Side::Ldk).await;
 }
+
+/// Specialized version of `splice_in_scenario` for BOLT11 payments.
+/// See [`splice_in_scenario`] for details.
+pub(crate) async fn splice_in_bolt11_scenario<E: ElectrumApi>(
```
tnull: Please avoid adding helper methods that are less than 5 lines of code.

tnull (commented):
Thanks for looking into this! Some comments.

tests/common/scenarios/mod.rs
alexgrad42: Removed. Note that BOLT11/BOLT12 scenarios are almost identical now. We can extract 11/12 step and run as prepayment->payment(this one id different)->postpayment chain on  integration_tests_cln if you want.

tests/common/scenarios/mod.rs
alexgrad42: Removed

tests/common/scenarios/mod.rs
alexgrad42: Sorry, I think I've confused both of you and myself. My test were unstable, so I decided this is timing issue due to asynchronous nature of the peer-to-payer invoice delivery via messaging.  In fact this seems to be my local WSL setup issue, and if it breaks splicing test, it does it for bolt11 too. Removed, please ignore

tests/common/scenarios/mod.rs
alexgrad42: Removed

tests/common/eclair.rs
```diff
@@ -211,6 +211,12 @@ impl ExternalNode for TestEclairNode {
 		Ok(invoice.to_string())
 	}
 
+	async fn create_offer(
+		&self, amount_msat: u64, description: &str,
+	) -> Result<String, TestFailure> {
+		Err(self.make_error("create_offer is not supported on Eclair".to_string()))
```
tnull, line 217: Hmm, I think this shouldn't be true anymore, no? Can you double-check that the `latest` docker image really doesn't support BOLT12? If it does we should def. un-ignore the Eclair BOLT12 test coverage in this PR.

tests/integration_tests_eclair.rs
```diff
@@ -69,6 +83,12 @@ async fn test_disconnect_during_payment() {
 
 #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
 #[ignore = "Eclair advertises splicing via custom bit 154 instead of BOLT bit 62/63; disjoint from LDK until Eclair migrates"]
-async fn test_splice_in() {
-	run_interop_scenario(setup_clients(), splice_in_scenario).await;
+async fn test_splice_in_bolt11() {
+	run_interop_scenario(setup_clients(), splice_in_bolt12_scenario).await;
```
tnull: Codex:

> - [P3] Run the BOLT11 splice scenario — /home/tnull/workspace/ldk-node/tests/integration_tests_eclair.rs:87-87
>     When the ignored Eclair splice test is run, this *_bolt11 test executes the BOLT12 splice scenario instead, so it never covers the BOLT11 path and will fail for Eclair's current lack of BOLT12 support. This should call splice_in_bolt11_scenario.

tests/common/cln.rs
```diff
@@ -181,6 +181,41 @@ impl ExternalNode for TestClnNode {
...
+
+		let params = serde_json::json!({
+			"amount": format!("{}msat", amount_msat),
+			"description": desc,
+			"label": label,
+			"single_use": true,
+		});
+
+		let response: serde_json::Value = self
+			.rpc(move |c| c.call("offer", params))
```
tnull: Codex:

> - [P1] Borrow CLN RPC parameter values — /home/tnull/workspace/ldk-node/tests/common/cln.rs:205-205
>     Under the CLN integration build, LightningRPC::call takes the params by reference, matching the existing &json!(...)/&params call sites. Passing the new serde_json::Value by value here, and again for fetchinvoice, will fail to type-check under --cfg cln_test; pass a borrow instead.

tests/common/eclair.rs
alexgrad42: I was going to add Eclair and/or LND in separate PRs honestly, but let me check

tests/integration_tests_eclair.rs
alexgrad42: Fixed

tests/common/cln.rs
alexgrad42: Fixed

PR #906, opened 2026-05-17
fix(payment): skip store insertion on routing failures to allow retry

### LINKED ISSUE #903: Should we update payment store if we fail finding a route?
Currently, when we fail to find a route for a given payment, we update the payment store to `PaymentFailed` and return an error. 

However, this means that the user won't be able to retry paying the same invoice as we'll of course avoid duplicate payments if we already have the payment hash in the store. This is maybe somewhat unnecessary in this case, as, when we fail to find a route, we never get to the point where the payment is sent out, hence we should be able to safely retry that payment.

alexgrad42:
Adjust the error handling for outbound payments so that when a payment fails at the pathfinding stage (e.g., `RetryableSendFailure::RouteNotFound`), we bypass inserting a `PaymentStatus::Failed` entry into the database.

Fix #903

### DIFF
--- a/src/payment/bolt11.rs
+++ b/src/payment/bolt11.rs
@@ -307,6 +307,7 @@ impl Bolt11Payment {
 				log_error!(self.logger, "Failed to send payment: {:?}", e);
 				match e {
 					RetryableSendFailure::DuplicatePayment => Err(Error::DuplicatePayment),
+					RetryableSendFailure::RouteNotFound => Err(Error::PaymentSendingFailed),
 					_ => {
 						let kind = PaymentKind::Bolt11 {
 							hash: payment_hash,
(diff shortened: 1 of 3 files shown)


benthecarman (changes requested):
I think we already handle this, the test passes for me without your changes or at least the test doesn't actually cover the issue you're trying to fix

alexgrad42:
> **[benthecarman](/benthecarman) **

Thanks for the review. Although the issue explicitly mentions DuplicatePayment, in the current implementation the send() method actually allows a retry if the payment status in the store is set to Failed. The underlying logic was that a RouteNotFound error would persist a redundant Failed payment record to the database, even though the HTLC was never sent out. I've updated the test to ensure that the payment store remains clean after a routing failure.

I'll leave it up to you to decide whether we should keep this exact check or modify it (e.g., make it configurable).

tnull (commented):
Thanks for looking into this, but it seems the original issue was based on a wrong/outdated assumption I made, see https://github.com/lightningdevkit/ldk-node/issues/903#issuecomment-4601708285

It seems we already do the right thing and simply didn't remember we did already fix this at some point. Please excuse the noise! Going ahead closing this.

alexgrad42:
No worries! Thanks for giving me the opportunity to dive into LDK internals.

PR #907, opened 2026-05-17
Update bdk_wallet to v3.0.0 and refactor test helpers

### LINKED ISSUE #776: Drop copied-over `wallet_events` when possible
In #769 we discovered that `bdk_wallet` currently doesn't support retrieving `WalletEvent`s when applying mempool transactions. We therefore had to resort to copy over the `wallet_events` utility method. Once this is resolved (cf. https://github.com/bitcoindevkit/bdk_wallet/issues/374), we should drop that copied code again.

HARSHVARANDANI:
Closes #776 
Updated to using bdk_wallet v3.0.0 which supports the api for wallet events and removed the copied code for the same being used earlier

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -13,10 +13,9 @@ use std::sync::{Arc, Mutex};
 use bdk_chain::spk_client::{FullScanRequest, SyncRequest};
 use bdk_wallet::descriptor::ExtendedDescriptor;
 use bdk_wallet::error::{BuildFeeBumpError, CreateTxError};
-use bdk_wallet::event::WalletEvent;
 #[allow(deprecated)]
 use bdk_wallet::SignOptions;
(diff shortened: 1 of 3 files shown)


src/wallet/mod.rs
Camillarhi, line 485: Perhaps `cancel_tx` should be renamed here as well, BDK removed it because the name was misleading since it doesn't actually cancel anything, it just unmarks unused addresses, and the same reasoning should apply to this.

Camillarhi (commented):
Thanks for looking into this. Left a comment, and also a minor typo in the second commit message ("teh" → "the").

tnull (commented):
Excuse the delay here! 

It seems this unfortunately would already need a rebase, but honestly I'd probably prefer to do the BDK upgrade myself, as there might be other associated tasks and/or new API changes that we might want to account for.

tnull:
Superseded by https://github.com/lightningdevkit/ldk-node/pull/937.

PR #910, opened 2026-05-19
Refine async payment cache admission

tnull:
Keep existing onion-message mailbox queues when the peer map is full, and recycle rate limiter bucket state when admitting a new user at capacity.

### DIFF
--- a/src/payment/asynchronous/om_mailbox.rs
+++ b/src/payment/asynchronous/om_mailbox.rs
@@ -19,22 +19,15 @@ impl OnionMessageMailbox {
 	pub(crate) fn onion_message_intercepted(&self, peer_node_id: PublicKey, message: OnionMessage) {
 		let mut map = self.map.lock().expect("lock");
 
+		if !map.contains_key(&peer_node_id) && map.len() >= Self::MAX_PEERS {
+			return;
+		}
+
 		let queue = map.entry(peer_node_id).or_insert_with(VecDeque::new);
 		if queue.len() >= Self::MAX_MESSAGES_PER_PEER {
 			queue.pop_front();
 		}
 		queue.push_back(message);
-
-		// Enforce a peers limit. If exceeded, evict the peer with the longest queue.
-		if map.len() > Self::MAX_PEERS {
-			let peer_to_remove = map
-				.iter()
-				.max_by_key(|(_, queue)| queue.len())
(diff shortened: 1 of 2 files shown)


src/payment/asynchronous/om_mailbox.rs
```diff
@@ -19,22 +19,15 @@ impl OnionMessageMailbox {
 	pub(crate) fn onion_message_intercepted(&self, peer_node_id: PublicKey, message: OnionMessage) {
 		let mut map = self.map.lock().expect("lock");
 
+		if !map.contains_key(&peer_node_id) && map.len() >= Self::MAX_PEERS {
```
joostjager, line 22: I think we talked about this before. Is it really better to allow an attacker to fill up all slots and DoS legit users?

src/payment/asynchronous/rate_limiter.rs
```diff
@@ -43,20 +44,22 @@ impl RateLimiter {
 		if is_new_user {
 			self.garbage_collect(self.max_idle);
 			if self.users.len() >= MAX_USERS {
-				return false;
+				self.evict_least_recently_seen();
```
joostjager, line 47: This can now be a rate-limiter bypass for attackers?

joostjager (approved):
I’m fine with these changes, but I’m not sure they make the system strictly better so much as change the failure mode.

The more fundamental issue is of course that onion-message handling is exposed to Sybil-style pressure. Not sure if it was a good idea to rely on it for async payments.

src/payment/asynchronous/rate_limiter.rs
```diff
@@ -72,7 +75,18 @@ impl RateLimiter {
 
 	fn garbage_collect(&mut self, max_idle: Duration) {
 		let now = Instant::now();
-		self.users.retain(|_, bucket| now.duration_since(bucket.last_refill) < max_idle);
+		self.users.retain(|_, bucket| now.duration_since(bucket.last_seen) < max_idle);
```
Jolah1, line 78: The doc comment correctly now says "remained unused for longer than the max idle duration", but no test exercises the new last_seen-based GC distinct from the old last_refill behavior — e.g. a bucket whose last_refill is stale (full of tokens, never refilled) but recently touched via allow() should not be GC'd. Worth a unit test?

PR #817, opened 2026-03-05
Add BOLT12 support to LSPS2 flow

tnull:
Based on https://github.com/lightningdevkit/rust-lightning/pull/4463

We implement the LSPS2 flow for BOLT12 and async payments.

PR #914, opened 2026-05-30
Introduce receive_via_jit_channel to UnifiedPayment

### LINKED ISSUE #507: Unified QR: Enable request of JIT invoice
It would be helpful if `ldkNode.unifiedQrPayment().receive` had a parameter that could return a JIT version of the invoice.

Use case:
- User has insufficient receive capacity
- Developer can still use `ldkNode.unifiedQrPayment().receive`, but can get a JIT version back

This saves the developer creating individual addresses and then baking them into a unified QR themselves.

chuksys:
Closes #507. 

This PR introduces `receive_via_jit_channel` to `UnifiedPayment`, exposing a clean way for downstream mobile applications to generate unified BIP 21 URIs that leverage JIT inbound liquidity. It leaves capacity detection up to the developer's discretion while providing a robust, graceful fallback to standard local invoices if the configured LSP is unavailable.

### DIFF
--- a/src/payment/unified.rs
+++ b/src/payment/unified.rs
@@ -32,7 +32,7 @@ use lightning_invoice::{Bolt11Invoice, Bolt11InvoiceDescription, Description};
 use crate::config::HRN_RESOLUTION_TIMEOUT_SECS;
 use crate::error::Error;
 use crate::ffi::maybe_wrap;
-use crate::logger::{log_error, LdkLogger, Logger};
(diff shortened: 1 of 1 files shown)


tnull (commented):
Hmm, I think this could make sense, but we should probably wait until after #817 (which we intend to ship for the next release).

chuksys:
> Hmm, I think this could make sense, but we should probably wait until after #817 (which we intend to ship for the next release).

Sure, that sounds good to me. Thanks!

PR #919, opened 2026-06-03
Fully switch to async `KVStore` persistence

tnull:
We planned to make the full switch to the async `KVStore` for a while. Here we do just that: first step-by-step moving the remaining dependencies over to use `KVStore`, then finally dropping `KVStoreSync` and all implementations.

Depends on https://github.com/lightningdevkit/rust-lightning/pull/4658 (will drop the `DROPME` commit once that's merged).

### DIFF
--- a/src/chain/electrum.rs
+++ b/src/chain/electrum.rs
@@ -34,7 +34,7 @@ use crate::io::utils::update_and_persist_node_metrics;
 use crate::logger::{log_bytes, log_debug, log_error, log_trace, LdkLogger, Logger};
 use crate::runtime::Runtime;
 use crate::types::{ChainMonitor, ChannelManager, DynStore, Sweeper, Wallet};
-use crate::NodeMetrics;
+use crate::PersistedNodeMetrics;
 
 const BDK_ELECTRUM_CLIENT_BATCH_SIZE: usize = 5;
 const ELECTRUM_CLIENT_NUM_RETRIES: u8 = 3;
@@ -49,14 +49,14 @@ pub(super) struct ElectrumChainSource {
 	kv_store: Arc<DynStore>,
 	config: Arc<Config>,
 	logger: Arc<Logger>,
-	node_metrics: Arc<RwLock<NodeMetrics>>,
+	node_metrics: Arc<PersistedNodeMetrics>,
 }
 
 impl ElectrumChainSource {
 	pub(super) fn new(
 		server_url: String, sync_config: ElectrumSyncConfig,
 		fee_estimator: Arc<OnchainFeeEstimator>, kv_store: Arc<DynStore>, config: Arc<Config>,
-		logger: Arc<Logger>, node_metrics: Arc<RwLock<NodeMetrics>>,
+		logger: Arc<Logger>, node_metrics: Arc<PersistedNodeMetrics>,
 	) -> Self {
 		let electrum_runtime_status = RwLock::new(ElectrumRuntimeStatus::new());
 		let onchain_wallet_sync_status = Mutex::new(WalletSyncStatus::Completed);
@@ -129,31 +129,6 @@ impl ElectrumChainSource {
 		let incremental_sync =
 			self.node_metrics.read().expect("lock").latest_onchain_wallet_sync_timestamp.is_some();
 
-		let apply_wallet_update =
-			|update_res: Result<BdkUpdate, Error>, now: Instant| match update_res {
-				Ok(update) => match onchain_wallet.apply_update(update) {
-					Ok(()) => {
-						log_debug!(
-							self.logger,
-							"{} of on-chain wallet finished in {}ms.",
-							if incremental_sync { "Incremental sync" } else { "Sync" },
-							now.elapsed().as_millis()
-						);
-						let unix_time_secs_opt =
-							SystemTime::now().duration_since(UNIX_EPOCH).ok().map(|d| d.as_secs());
-						update_and_persist_node_metrics(
-							&self.node_metrics,
-							&*self.kv_store,
-							&*self.logger,
-							|m| m.latest_onchain_wallet_sync_timestamp = unix_time_secs_opt,
-						)?;
-						Ok(())
-					},
-					Err(e) => Err(e),
-				},
-				Err(e) => Err(e),
-			};
-
 		let cached_txs = onchain_wallet.get_cached_txs();
 
 		let res = if incremental_sync {
@@ -162,18 +137,60 @@ impl ElectrumChainSource {
 				.get_incremental_sync_wallet_update(incremental_sync_request, cached_txs);
 
 			let now = Instant::now();
-			let update_res = incremental_sync_fut.await.map(|u| u.into());
-			apply_wallet_update(update_res, now)
+			let update_res: Result<BdkUpdate, Error> = incremental_sync_fut.await.map(|u| u.into());
+			self.apply_onchain_wallet_update(
+				onchain_wallet.as_ref(),
+				incremental_sync,
+				update_res,
+				now,
+			)
+			.await
 		} else {
 			let full_scan_request = onchain_wallet.get_full_scan_request();
 			let full_scan_fut =
 				electrum_client.get_full_scan_wallet_update(full_scan_request, cached_txs);
 			let now = Instant::now();
-			let update_res = full_scan_fut.await.map(|u| u.into());
-			apply_wallet_update(update_res, now)
+			let update_res: Result<BdkUpdate, Error> = full_scan_fut.await.map(|u| u.into());
+			self.apply_onchain_wallet_update(
+				onchain_wallet.as_ref(),
+				incremental_sync,
+				update_res,
+				now,
+			)
+			.await
 		};
 
 		res
+	}
+
+	async fn apply_onchain_wallet_update(
+		&self, onchain_wallet: &Wallet, incremental_sync: bool,
+		update_res: Result<BdkUpdate, Error>, now: Instant,
+	) -> Result<(), Error> {
+		match update_res {
+			Ok(update) => match onchain_wallet.apply_update(update) {
+				Ok(()) => {
+					log_debug!(
+						self.logger,
+						"{} of on-chain wallet finished in {}ms.",
+						if incremental_sync { "Incremental sync" } else { "Sync" },
+						now.elapsed().as_millis()
+					);
+					let unix_time_secs_opt =
+						SystemTime::now().duration_since(UNIX_EPOCH).ok().map(|d| d.as_secs());
+					update_and_persist_node_metrics(
+						&self.node_metrics,
+						&*self.kv_store,
+						&*self.logger,
+						|m| m.latest_onchain_wallet_sync_timestamp = unix_time_secs_opt,
+					)
+					.await?;
+					Ok(())
+				},
+				Err(e) => Err(e),
+			},
+			Err(e) => Err(e),
+		}
 	}
 
 	pub(crate) async fn sync_lightning_wallet(
@@ -239,7 +256,8 @@ impl ElectrumChainSource {
 				&*self.kv_store,
 				&*self.logger,
 				|m| m.latest_lightning_wallet_sync_timestamp = unix_time_secs_opt,
-			)?;
+			)
+			.await?;
 		}
 
 		res
@@ -270,7 +288,8 @@ impl ElectrumChainSource {
 			SystemTime::now().duration_since(UNIX_EPOCH).ok().map(|d| d.as_secs());
 		update_and_persist_node_metrics(&self.node_metrics, &*self.kv_store, &*self.logger, |m| {
 			m.latest_fee_rate_cache_update_timestamp = unix_time_secs_opt
-		})?;
+		})
+		.await?;
 
 		Ok(())
 	}
--- a/src/data_store.rs
+++ b/src/data_store.rs
@@ -9,7 +9,7 @@ use std::collections::{hash_map, HashMap};
 use std::ops::Deref;
 use std::sync::{Arc, Mutex};
 
-use lightning::util::persist::KVStoreSync;
+use lightning::util::persist::KVStore;
 use lightning::util::ser::{Readable, Writeable};
 
 use crate::logger::{log_error, LdkLogger};
@@ -45,6 +45,7 @@ where
 	L::Target: LdkLogger,
 {
 	objects: Mutex<HashMap<SO::Id, SO>>,
+	mutation_lock: tokio::sync::Mutex<()>,
 	primary_namespace: String,
 	secondary_namespace: String,
 	kv_store: Arc<DynStore>,
@@ -61,50 +62,64 @@ where
 	) -> Self {
 		let objects =
 			Mutex::new(HashMap::from_iter(objects.into_iter().map(|obj| (obj.id(), obj))));
-		Self { objects, primary_namespace, secondary_namespace, kv_store, logger }
+		Self {
+			objects,
+			mutation_lock: tokio::sync::Mutex::new(()),
+			primary_namespace,
+			secondary_namespace,
+			kv_store,
+			logger,
+		}
 	}
 
-	pub(crate) fn insert(&self, object: SO) -> Result<bool, Error> {
-		let mut locked_objects = self.objects.lock().expect("lock");
+	pub(crate) async fn insert(&self, object: SO) -> Result<bool, Error> {
+		let _guard = self.mutation_lock.lock().await;
 
-		self.persist(&object)?;
+		self.persist(&object).await?;
+		let mut locked_objects = self.objects.lock().expect("lock");
 		let updated = locked_objects.insert(object.id(), object).is_some();
 		Ok(updated)
 	}
 
-	pub(crate) fn insert_or_update(&self, object: SO) -> Result<bool, Error> {
-		let mut locked_objects = self.objects.lock().expect("lock");
+	pub(crate) async fn insert_or_update(&self, object: SO) -> Result<bool, Error> {
+		let _guard = self.mutation_lock.lock().await;
+		let (updated, data_to_persist) = {
+			let mut locked_objects = self.objects.lock().expect("lock");
+			match locked_objects.entry(object.id()) {
+				hash_map::Entry::Occupied(mut e) => {
+					let update = object.to_update();
+					let updated = e.get_mut().update(update);
+					let data_to_persist =
+						if updated { Some(Self::encode_object(e.get())) } else { None };
+					(updated, data_to_persist)
+				},
+				hash_map::Entry::Vacant(e) => {
+					let data_to_persist = Self::encode_object(&object);
+					e.insert(object);
+					(true, Some(data_to_persist))
+				},
+			}
+		};
 
-		let updated;
-		match locked_objects.entry(object.id()) {
-			hash_map::Entry::Occupied(mut e) => {
-				let update = object.to_update();
-				updated = e.get_mut().update(update);
-				if updated {
-					self.persist(&e.get())?;
-				}
-			},
-			hash_map::Entry::Vacant(e) => {
-				e.insert(object.clone());
-				self.persist(&object)?;
-				updated = true;
-			},
+		if let Some((store_key, data)) = data_to_persist {
+			self.persist_encoded(store_key, data).await?;
 		}
-
 		Ok(updated)
 	}
 
-	pub(crate) fn remove(&self, id: &SO::Id) -> Result<(), Error> {
-		let removed = self.objects.lock().expect("lock").remove(id).is_some();
+	pub(crate) async fn remove(&self, id: &SO::Id) -> Result<(), Error> {
+		let _guard = self.mutation_lock.lock().await;
+		let removed = { self.objects.lock().expect("lock").remove(id).is_some() };
 		if removed {
 			let store_key = id.encode_to_hex_str();
-			KVStoreSync::remove(
+			KVStore::remove(
 				&*self.kv_store,
 				&self.primary_namespace,
 				&self.secondary_namespace,
 				&store_key,
 				false,
 			)
+			.await
 			.map_err(|e| {
 				log_error!(
 					self.logger,
@@ -120,40 +135,63 @@ where
 		Ok(())
 	}
 
+	/// Returns the current in-memory object for `id`.
+	///
+	/// The async mutation lock serializes writers, but this synchronous reader cannot wait on it.
+	/// Until store reads are async, callers may temporarily see in-memory state that is either
+	/// still being persisted or has not yet caught up to a write in progress.
 	pub(crate) fn get(&self, id: &SO::Id) -> Option<SO> {
 		self.objects.lock().expect("lock").get(id).cloned()
 	}
 
-	pub(crate) fn update(&self, update: SO::Update) -> Result<DataStoreUpdateResult, Error> {
-		let mut locked_objects = self.objects.lock().expect("lock");
-
-		if let Some(object) = locked_objects.get_mut(&update.id()) {
-			let updated = object.update(update);
-			if updated {
-				self.persist(&object)?;
-				Ok(DataStoreUpdateResult::Updated)
+	pub(crate) async fn update(&self, update: SO::Update) -> Result<DataStoreUpdateResult, Error> {
+		let _guard = self.mutation_lock.lock().await;
+		let (res, data_to_persist) = {
+			let mut locked_objects = self.objects.lock().expect("lock");
+			if let Some(object) = locked_objects.get_mut(&update.id()) {
+				let updated = object.update(update);
+				if updated {
+					(DataStoreUpdateResult::Updated, Some(Self::encode_object(object)))
+				} else {
+					(DataStoreUpdateResult::Unchanged, None)
+				}
 			} else {
-				Ok(DataStoreUpdateResult::Unchanged)
+				(DataStoreUpdateResult::NotFound, None)
 			}
-		} else {
-			Ok(DataStoreUpdateResult::NotFound)
+		};
+		if let Some((store_key, data)) = data_to_persist {
+			self.persist_encoded(store_key, data).await?;
 		}
+		Ok(res)
 	}
 
+	/// Returns in-memory objects matching `f`.
+	///
+	/// The async mutation lock serializes writers, but this synchronous reader cannot wait on it.
+	/// Until store reads are async, callers may temporarily see in-memory state that is either
+	/// still being persisted or has not yet caught up to a write in progress.
 	pub(crate) fn list_filter<F: FnMut(&&SO) -> bool>(&self, f: F) -> Vec<SO> {
 		self.objects.lock().expect("lock").values().filter(f).cloned().collect::<Vec<SO>>()
 	}
 
-	fn persist(&self, object: &SO) -> Result<(), Error> {
-		let store_key = object.id().encode_to_hex_str();
-		let data = object.encode();
-		KVStoreSync::write(
+	async fn persist(&self, object: &SO) -> Result<(), Error> {
+		let (store_key, data) = Self::encode_object(object);
+		self.persist_encoded(store_key, data).await
+	}
+
+	fn encode_object(object: &SO) -> (String, Vec<u8>) {
+		(object.id().encode_to_hex_str(), object.encode())
+	}
+
+	async fn persist_encoded(&self, store_key: String, data: Vec<u8>) -> Result<(), Error> {
+		KVStore::write(
 			&*self.kv_store,
 			&self.primary_namespace,
 			&self.secondary_namespace,
 			&store_key,
 			data,
 		)
+		.await
 		.map_err(|e| {
 			log_error!(
 				self.logger,
@@ -168,6 +206,11 @@ where
 		Ok(())
 	}
 
+	/// Returns whether the in-memory store contains `id`.
+	///
+	/// The async mutation lock serializes writers, but this synchronous reader cannot wait on it.
+	/// Until store reads are async, callers may temporarily see in-memory state that is either
+	/// still being persisted or has not yet caught up to a write in progress.
 	pub(crate) fn contains_key(&self, id: &SO::Id) -> bool {
 		self.objects.lock().expect("lock").contains_key(id)
 	}
@@ -238,8 +281,8 @@ mod tests {
 		(2, data, required),
 	});
 
-	#[test]
-	fn data_is_persisted() {
+	#[tokio::test]
+	async fn data_is_persisted() {
 		let store: Arc<DynStore> = Arc::new(DynStoreWrapper(InMemoryStore::new()));
 		let logger = Arc::new(TestLogger::new());
 		let primary_namespace = "datastore_test_primary".to_string();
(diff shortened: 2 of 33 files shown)


src/payment/asynchronous/static_invoice_store.rs
```diff
@@ -78,33 +78,28 @@ impl StaticInvoiceStore {
...
-			|e| {
-				if e.kind() == lightning::io::ErrorKind::NotFound {
-					Ok(None)
-				} else {
-					Err(e)
-				}
-			},
-		)
+		.await
+		{
```
joostjager: This looks like an unnecessary style change?

src/peer_store.rs
```diff
@@ -40,22 +40,25 @@ where
...
+	pub(crate) async fn add_peer(&self, peer_info: PeerInfo) -> Result<(), Error> {
+		let data = {
+			let mut locked_peers = self.peers.write().expect("lock");
+			if locked_peers.contains_key(&peer_info.node_id) {
+				return Ok(());
+			}
+			locked_peers.insert(peer_info.node_id, peer_info);
+			PeerStoreSerWrapper(&locked_peers).encode()
+		};
+		self.persist_peers(data).await
```
joostjager, line 55: Is it safe to have already dropped the lock? If two `add_peer` calls run interleaved, outdated state may be written?

src/event.rs
```diff
@@ -1524,36 +1524,38 @@ where
...
+					let channels =
+						self.channel_manager.list_channels_with_counterparty(&counterparty_node_id);
+					channels
+						.into_iter()
+						.find(|c| c.channel_id == channel_id)
+						.filter(|pending_channel| {
+							!pending_channel.is_outbound
+								&& self.peer_store.get_peer(&counterparty_node_id).is_none()
+						})
+						.and_then(|_| {
```
joostjager, line 1538: Is this full style rewrite necessary?

src/data_store.rs
```diff
@@ -45,6 +45,7 @@ where
 	L::Target: LdkLogger,
 {
 	objects: Mutex<HashMap<SO::Id, SO>>,
+	mutation_lock: tokio::sync::Mutex<()>,
```
joostjager, line 48: Can you help me remember why we have a `Runtime` abstraction, but also reference tokio directly?

src/data_store.rs
```diff
@@ -61,50 +62,64 @@ where
...
+		}
 	}
 
-	pub(crate) fn insert(&self, object: SO) -> Result<bool, Error> {
-		let mut locked_objects = self.objects.lock().expect("lock");
+	pub(crate) async fn insert(&self, object: SO) -> Result<bool, Error> {
+		let _guard = self.mutation_lock.lock().await;
 
-		self.persist(&object)?;
+		self.persist(&object).await?;
```
joostjager, line 78: Now that this no longer holds self.objects, doesn't that, isn't this creating a window where readers get interleaved in an undesirable way?

src/chain/electrum.rs
```diff
@@ -129,30 +129,36 @@ impl ElectrumChainSource {
...
-						let unix_time_secs_opt =
-							SystemTime::now().duration_since(UNIX_EPOCH).ok().map(|d| d.as_secs());
-						update_and_persist_node_metrics(
-							&self.node_metrics,
-							&*self.kv_store,
-							&*self.logger,
-							|m| m.latest_onchain_wallet_sync_timestamp = unix_time_secs_opt,
-						)?;
-						Ok(())
+		macro_rules! apply_wallet_update {
```
joostjager: Is this macro absolutely necessary? They've caused so much pain elsewhere that I think the bar should be really high to use them. Perhaps just share the match.

src/io/utils.rs
```diff
@@ -336,26 +336,26 @@ where
...
-	let mut locked_node_metrics = node_metrics.write().expect("lock");
-	update(&mut *locked_node_metrics);
-	let data = locked_node_metrics.encode();
-	KVStoreSync::write(
+	let data = {
+		let mut locked_node_metrics = node_metrics.write().expect("lock");
+		update(&mut *locked_node_metrics);
+		locked_node_metrics.encode()
+	};
+	KVStore::write(
```
joostjager, line 352: Outside the lock now, can it cause out of order writes?

src/wallet/persist.rs
```diff
@@ -5,45 +5,39 @@
...
 	latest_change_set: Option<ChangeSet>,
 	kv_store: Arc<DynStore>,
-	runtime: Arc<Runtime>,
 	logger: Arc<Logger>,
 }
 
 impl KVStoreWalletPersister {
-	pub(crate) fn new(kv_store: Arc<DynStore>, runtime: Arc<Runtime>, logger: Arc<Logger>) -> Self {
-		Self { latest_change_set: None, kv_store, runtime, logger }
+	pub(crate) fn new(kv_store: Arc<DynStore>, logger: Arc<Logger>) -> Self {
```
joostjager, line 30: This is undoing earlier changes. Might be better to do it all in one commit and avoid introducing runtime here?

src/io/utils.rs
joostjager: Also use the async version here?

src/io/in_memory_store.rs
```diff
@@ -0,0 +1,223 @@
...
+	fn write_internal(
+		&self, primary_namespace: &str, secondary_namespace: &str, key: &str, buf: Vec<u8>,
+	) -> io::Result<()> {
+		let mut persisted_lock = self.persisted_bytes.lock().unwrap();
+
+		let prefixed = format!("{primary_namespace}/{secondary_namespace}");
+		let outer_e = persisted_lock.entry(prefixed.clone()).or_insert(HashMap::new());
+		outer_e.insert(key.to_string(), buf);
+
+		let mut ct_lock = self.creation_times.lock().unwrap();
```
joostjager, line 60: Is this supposed to be a pure move? Some comments may have dropped out such as 

`		// Only assign creation time on first write (not on update)`

src/io/utils.rs
joostjager: Ah it's done later

src/lib.rs
```diff
@@ -2062,20 +2062,21 @@ impl Node {
...
-		)
-		.map_err(|e| {
-			log_error!(
-				self.logger,
-				"Failed to access store while exporting pathfinding scores: {}",
-				e
-			);
-			Error::PersistenceFailed
-		})
+		self.runtime
```
joostjager, line 2065: Commit message says test changes, but this is a prod change?

src/chain/electrum.rs
```diff
@@ -129,37 +129,6 @@ impl ElectrumChainSource {
 		let incremental_sync =
 			self.node_metrics.read().expect("lock").latest_onchain_wallet_sync_timestamp.is_some();
 
-		macro_rules! apply_wallet_update {
```
joostjager: Hmm and now it is dropped again. Perhaps the commit stack can be shaped differently to avoid this?

src/io/vss_store.rs
joostjager: No await here? Can that go undetected? Perhaps elsewhere too.

src/io/test_utils.rs
```diff
@@ -5,34 +5,157 @@
...
+use crate::runtime::Runtime;
+
+pub(crate) struct TestMonitorUpdatePersister<'a, K> {
+	store: &'a K,
+	runtime: Runtime,
+	entropy_source: &'a test_utils::TestKeysInterface,
+	signer_provider: &'a test_utils::TestKeysInterface,
+}
+
+impl<K: KVStore + Sync> TestMonitorUpdatePersister<'_, K> {
```
joostjager, line 48: Commit message says dropping the trait, but this looks more substantial?

src/io/test_utils.rs
```diff
@@ -46,71 +169,96 @@ pub(crate) fn random_storage_path() -> PathBuf {
...
 
-	let listed_keys = kv_store.list(&max_chars, &max_chars).unwrap();
+	let listed_keys = KVStore::list(kv_store, &max_chars, &max_chars).await.unwrap();
 	assert_eq!(listed_keys.len(), 0);
 }
 
-pub(crate) fn create_persister<'a, K: KVStoreSync + Sync>(
-	store: &'a K, chanmon_cfg: &'a TestChanMonCfg, max_pending_updates: u64,
+pub(crate) fn create_persister<'a, K: KVStore + Sync>(
+	store: &'a K, chanmon_cfg: &'a TestChanMonCfg, _max_pending_updates: u64,
```
joostjager, line 249: Wasn't max pending updates exercised in a test?

src/io/vss_store.rs
```diff
@@ -122,42 +98,25 @@ impl VssStore {
...
 
-		let sync_retry_policy = retry_policy();
-		let blocking_client = VssClient::new_with_headers(
-			base_url.clone(),
-			sync_retry_policy,
-			header_provider.clone(),
-		);
-
-		let runtime_handle = internal_runtime.handle();
-		let schema_version = tokio::task::block_in_place(|| {
```
joostjager, line 133: Isn't it a problem that setup can't fail fast here anymore?

src/io/vss_store.rs
```diff
@@ -449,11 +428,22 @@ impl VssStoreInner {
 			"write",
 		)?;
 
-		let store_key = self.build_obfuscated_key(&primary_namespace, &secondary_namespace, &key);
+		let schema_version = match self.schema_version().await {
```
joostjager: This repetitive code doesn't look great. Maybe an explicit setup call is better?

joostjager (commented):
Overall happy with this simplification. Big changeset though, even though not all that much is happening. Risk nonetheless.

Bigger open question for me is what the latest is on runtime abstraction and how that is going to work in combination with potential wasm support.

Also, are there sync public api calls that can now be made async such as `export_pathfinding_scores`?

src/io/utils.rs
tnull: Yup.

src/wallet/persist.rs
tnull: No, this is intentional as otherwise we'd need to do all in one commit basically, as everything that requires an onchain address for example would need to be in the same async call path, i.e., same commit. Here we just migrate some over, then go fix the callers, then come back to do full async persistence in BDK.

src/payment/asynchronous/static_invoice_store.rs
tnull: Hmm, right. I have to admit I like the changed version better (which is why I left it in), but now added a fixup to keep it minimally invasive.

src/data_store.rs
tnull: That's somewhat orthogonal, right now `Runtime` is an abstraction for the `Runtime` that ensures we do the right thing (detect outer runtimes, always wrap `block_on` in `block_in_place` etc). This is simply using the mutex type.

src/data_store.rs
tnull: That's why we introduced the `mutation_lock` for the write paths. Or maybe I'm misunderstanding the question?

src/event.rs
tnull: Well, a refactor was somewhat necessary anyways as we needed to move out the `add_peer` call, and it's way cleaner to do it this way as otherwise we'd need to add at least 2 more else clauses with `None` anyways.

src/io/in_memory_store.rs
tnull: It seems you can't trust an agent.

src/peer_store.rs
tnull: Fair, would you prefer we take a mutation guard as we do in `DataStore` in instances like this?

src/chain/electrum.rs
tnull: I asked the same thing, seems Codex fixed it in the wrong commit. Sorry, will try to find a better solution.

src/io/utils.rs
tnull: See above, seems like a more general question. Maybe we should take temporary mutation locks for all of these?

src/io/vss_store.rs
tnull: We fail fast elsewhere via `setup_schema_version`.

src/io/vss_store.rs
tnull: Good catch. Fixed.

src/peer_store.rs
```diff
@@ -40,22 +40,25 @@ where
...
-
-		if locked_peers.contains_key(&peer_info.node_id) {
-			return Ok(());
-		}
-
-		locked_peers.insert(peer_info.node_id, peer_info);
-		self.persist_peers(&*locked_peers)
+	pub(crate) async fn add_peer(&self, peer_info: PeerInfo) -> Result<(), Error> {
+		let data = {
+			let mut locked_peers = self.peers.write().expect("lock");
```
benthecarman, line 48: should we use an async lock for peers now?

src/io/utils.rs
```diff
@@ -336,26 +336,26 @@ where
...
 ) -> Result<(), Error>
 where
 	L::Target: LdkLogger,
 {
-	let mut locked_node_metrics = node_metrics.write().expect("lock");
-	update(&mut *locked_node_metrics);
-	let data = locked_node_metrics.encode();
-	KVStoreSync::write(
+	let data = {
+		let mut locked_node_metrics = node_metrics.write().expect("lock");
```
benthecarman, line 348: should this be async lock?

benthecarman (commented):
the vss ci job hung for 2 hours hopefully a transient issue

benthecarman:
codex seems to agree on the locks

  - [P2] Serialize peer-store persistence updates — /home/ben/projects/ldk-node/src/peer_store.rs:43-52
    When two peer-store mutations overlap, this now releases the peer map lock before the async write completes, so an
    older snapshot can persist after a newer one. For example, concurrent add_peer calls can leave memory with both
    peers but the stored peer list with only the first one, causing reconnect peers to be missing after restart. Please
    serialize the mutation plus persistence, as DataStore now does with a separate async mutation lock.

  - [P2] Preserve node-metrics update ordering — /home/ben/projects/ldk-node/src/io/utils.rs:346-351
    If two node-metrics updates run concurrently, the lock is released before the async KV write, so an earlier encoded
    snapshot can finish after a later one and overwrite it. This can drop one of the timestamp fields on disk and make
    the node believe a sync/broadcast task has not run after restart. The previous code explicitly held the write lock
    through persistence to avoid this, so this path needs equivalent serialization for the async write.

src/data_store.rs
joostjager: Ah yes, that's right.

src/peer_store.rs
joostjager: I think so? Not sure if there are other ways to do it?

src/data_store.rs
joostjager: Yes it works for the writers, but readers can still see old in-memory state while a write is being persisted. `insert_or_update` is the other way around. A reader can see in-memory state that isn't on disk yet. Perhaps it is acceptable, but a comment is justified then.

src/event.rs
joostjager: I tried the simple refactor, and I actually think it is cleaner. It's fine, but with LLM-generated code in combination with a large PR, I think it is best to keep the style changes to an absolute minimum.

Image

src/io/utils.rs
joostjager: Yes not sure what the common pattern is for it, but allowing these interleaves seems like asking for trouble.

src/wallet/persist.rs
joostjager: Ok, I see the trade-off

src/io/in_memory_store.rs
joostjager: Can't trust them, can't live without them 😅

src/io/vss_store.rs
joostjager: Can't the schema_version then be stored as a normal field? It seems the code is currently halfway between lazy init and explicit setup.

tnull: Well, no, we need to discover it based on the server-side state for backwards compatibility, as unfortunately schema version wasn't always there.

src/peer_store.rs
tnull: Yeah, now did add safeguards for node metrics and peer store persistence, too, but once we switch to be fully async we'll be able to switch to `tokio::sync::{Mutex,RwLock}`s and hold them across `await`.

src/io/utils.rs
tnull: Right, now added `PersistedNodeMetrics` helper, will be able to simplify again once we can switch to `tokio::sync` types.

src/peer_store.rs
tnull: Yeah, fixed by adding a mutation lock for now, but we'll be able to simplify more once we switch to `tokio::sync`, once we go fully-async.

src/io/utils.rs
tnull: Can't do that right now as we have non-`async` callers (readers).

tnull:
As discussed offline, now updated:

1) Re-added the internal runtimes for Postgres / VSS to avoid `tokio` getting stuck if we end up using `block_on` while the same thread holds the IO driver. We'll be able to revert this/drop the runtimes once we go fully async API in a following PR.
2) Added mutation locks for node metrics and peer store persistence for now, but we'll be able to simplify more once we switch to `tokio::sync`, once we go fully-async.
3) Added a commit that enables eager handoff of the IO driver if we're built under `cfg(tokio_unstable)` and enables that for bindings builds.

tnull:
Hmm, seems Github CI might have an outage, at least it's not running right now..

src/runtime.rs
```diff
@@ -32,7 +32,11 @@ impl Runtime {
 		let mode = match tokio::runtime::Handle::try_current() {
 			Ok(handle) => RuntimeMode::Handle(handle),
 			Err(_) => {
-				let rt = tokio::runtime::Builder::new_multi_thread().enable_all().build()?;
+				let mut runtime_builder = tokio::runtime::Builder::new_multi_thread();
+				runtime_builder.enable_all();
+				#[cfg(tokio_unstable)]
+				runtime_builder.enable_eager_driver_handoff();
```
joostjager, line 52: I'd definitely add docs why this is needed and/or link to the tokio issue.

src/peer_store.rs
```diff
@@ -37,10 +38,12 @@ where
 {
 	pub(crate) fn new(kv_store: Arc<DynStore>, logger: L) -> Self {
 		let peers = RwLock::new(HashMap::new());
-		Self { peers, kv_store, logger }
+		let mutation_lock = tokio::sync::Mutex::new(());
+		Self { peers, mutation_lock, kv_store, logger }
 	}
 
 	pub(crate) async fn add_peer(&self, peer_info: PeerInfo) -> Result<(), Error> {
+		let _guard = self.mutation_lock.lock().await;
```
joostjager, line 46: Do you think that reading doesn't need to be in this lock too? Similar to what we discussed in the other store discussion threads? https://github.com/lightningdevkit/ldk-node/pull/919#discussion_r3354088301

src/runtime.rs
```diff
@@ -223,6 +225,60 @@ enum RuntimeMode {
 	Handle(tokio::runtime::Handle),
 }
 
+pub(crate) struct StoreRuntime {
```
joostjager, line 276: This can also use some serious documentation on why it is needed, even if temporary.

src/runtime.rs
```diff
@@ -223,6 +225,60 @@ enum RuntimeMode {
...
+	) -> io::Result<Self> {
+		let runtime = tokio::runtime::Builder::new_multi_thread()
+			.enable_all()
+			.thread_name_fn(move || {
+				static ATOMIC_ID: AtomicUsize = AtomicUsize::new(0);
+				let id = ATOMIC_ID.fetch_add(1, Ordering::SeqCst);
+				format!("{}-{}", thread_name_prefix, id)
+			})
+			.worker_threads(worker_threads)
+			.max_blocking_threads(worker_threads)
```
joostjager, line 292: What is the reason for this setting?

src/runtime.rs
```diff
@@ -223,6 +225,60 @@ enum RuntimeMode {
...
+	runtime: Option<tokio::runtime::Runtime>,
+}
+
+impl StoreRuntime {
+	pub(crate) fn new(
+		thread_name_prefix: &'static str, worker_threads: usize, runtime_name: &'static str,
+	) -> io::Result<Self> {
+		let runtime = tokio::runtime::Builder::new_multi_thread()
+			.enable_all()
+			.thread_name_fn(move || {
```
joostjager, line 286: Maybe add this debugging/naming setting also to other places where a runtime is created?

src/runtime.rs
joostjager: Also wondering if it isn't possible to unify this with the existing `Runtime` wrapper? That might also make the various modes more clear: current runtime, passed runtime, new runtime, or some combo.

src/io/vss_store.rs
```diff
@@ -267,16 +284,25 @@ impl KVStore for VssStore {
...
 	}
 }
 
 impl Drop for VssStore {
 	fn drop(&mut self) {
-		let internal_runtime = self.internal_runtime.take();
-		tokio::task::block_in_place(move || drop(internal_runtime));
+		if let Some(runtime) = self.internal_runtime.take() {
+			if let Ok(runtime) = Arc::try_unwrap(runtime) {
+				runtime.shutdown_background();
```
joostjager, line 300: What does it mean if we skip shutdown here?

src/io/postgres_store/mod.rs
```diff
@@ -41,6 +42,18 @@ const PAGE_SIZE: usize = 50;
 // Keep this small while still allowing progress if one runtime worker blocks on sync store access.
 const INTERNAL_RUNTIME_WORKERS: usize = 2;
 
+async fn run_on_internal_runtime<T>(
```
joostjager: Can this be a method on StoreRuntime?

src/runtime.rs
```diff
@@ -223,6 +225,60 @@ enum RuntimeMode {
 	Handle(tokio::runtime::Handle),
 }
 
+pub(crate) struct StoreRuntime {
+	runtime: Option<tokio::runtime::Runtime>,
+}
+
+impl StoreRuntime {
```
joostjager, line 280: Name describes usage, not really what it does. IsolatedRuntime?

src/runtime.rs
tnull: > Also wondering if it isn't possible to unify this with the existing `Runtime` wrapper? That might also make the various modes more clear: current runtime, passed runtime, new runtime, or some combo.

Considered that, but given the plan is to remove it ASAP, I decided to keep it separate which will make the cleanup much more straightforward.

joostjager: I don't think clean up is really much different, but in the mean time reusing the `Runtime` does make it more clear that we have different variations of the same concept. Maybe it takes longer than we'd like to get rid of it. And can reuse the thread naming thing too.

benthecarman (commented):
this mostly looks good to me, some is hard to review because unsure what is going to move with the full async switch too. Agree with @joostjager that more comments would make it easier to parse, at least adding todos on what is planned for removal as well

src/data_store.rs
tnull: Done.

src/lib.rs
tnull: Moved it to the next commit.

src/io/vss_store.rs
tnull: Outdated by now.

src/runtime.rs
tnull: Hmm, the thing itself also doesn't isolate the runtime, no? It's just used in an isolated manner?

src/io/postgres_store/mod.rs
tnull: Will drop the helpers, inlining will be more explicit anyways.

src/runtime.rs
tnull: Not sure this improves anything? Note it's not the object that isolates anything really, but its usage.

src/runtime.rs
tnull: Done.

src/runtime.rs
tnull: It shouldn't matter as we shouldn't hit blocking calls in the persistence pipeline anyways now.

src/peer_store.rs
tnull: Well, we can't without making everything async. Added docs as you requested.

src/runtime.rs
tnull: Done.

tnull:
Responded to / addressed open feedback (mostly had Codex write a novel on why tokio is broken).

src/io/vss_store.rs
tnull: I'm not sure I'm following what you mean with 'skip' in this context?

src/io/test_utils.rs
tnull: Ah right, move it to another commit for now, but will also put up a minor change in LDK so we can reuse the test utils wihtout rewriting this.

src/io/test_utils.rs
tnull: Yes, but we can't use it right now as LDK seems not to expose the right APIs to use async path with the test utils, will open  a fix in a bit.

src/io/test_utils.rs
tnull: That should allow us to avoid it: https://github.com/lightningdevkit/rust-lightning/pull/4665

tnull:
FYI, here's the draft for the full-async'ified API follow-up: https://github.com/lightningdevkit/ldk-node/pull/926

src/io/test_utils.rs
tnull: Will allow us to simply do:
```rust
  type TestMonitorPersister<K> = MonitorUpdatingPersisterAsync<
        TestStoreRef<K>,
        TestFutureSpawner,
        &'static test_utils::TestLogger,
        &'static test_utils::TestKeysInterface,
        &'static test_utils::TestKeysInterface,
        &'static test_utils::TestBroadcaster,
        &'static test_utils::TestFeeEstimator,
  >;

  type TestAsyncPersister<K> = lightning::chain::chainmonitor::AsyncPersister<
        TestStoreRef<K>,
        TestFutureSpawner,
        &'static test_utils::TestLogger,
        &'static test_utils::TestKeysInterface,
        &'static test_utils::TestKeysInterface,
        &'static test_utils::TestBroadcaster,
        &'static test_utils::TestFeeEstimator,
  >;
(..)
  let monitor_persister = TestMonitorPersister::new(
        store,
        TestFutureSpawner::new(Arc::clone(&runtime)),
        &chanmon_cfg.logger,
        max_pending_updates,
        &chanmon_cfg.keys_manager,
        &chanmon_cfg.keys_manager,
        &chanmon_cfg.tx_broadcaster,
        &chanmon_cfg.fee_estimator,
  );

  let event_notifier = Arc::new(Notifier::new());
  let persister = lightning::chain::chainmonitor::AsyncPersister::new_test(
        monitor_persister,
        Arc::clone(&event_notifier),
  );

(...)

  test_utils::TestChainMonitor::new_with_event_notifier(
        Some(&chanmon_cfg.chain_source),
        &chanmon_cfg.tx_broadcaster,
        &chanmon_cfg.logger,
        &chanmon_cfg.fee_estimator,
        &persister.persister,
        &chanmon_cfg.keys_manager,
        Arc::clone(&persister.event_notifier),
  )
```
.. rather than the custom `TestMonitorPersister`

src/io/vss_store.rs
joostjager: I mean the Drop path conditionally does not call shutdown_background_tasks, so the question is what state/resources/tasks may remain when that branch is taken?

src/runtime.rs
joostjager: I just mean that if you want to use this generally-usable object for something else, you need to rename it even though the code stays the same.

src/runtime.rs
joostjager: Still curious why it is needed

src/io/postgres_store/mod.rs
```diff
@@ -271,17 +283,26 @@ impl KVStore for PostgresStore {
...
+						&primary_namespace,
+						&secondary_namespace,
+						&key,
+					)
+					.await
+			});
+			task.await.map_err(|e| {
+				io::Error::new(
+					io::ErrorKind::Other,
+					format!("PostgreSQL runtime task failed: {}", e),
```
benthecarman, line 303: we have the same message for all of these, would be better if it had the function name in it for easier debugging. however this is removed in the next pr so not very important

src/io/vss_store.rs
```diff
@@ -306,10 +194,16 @@ impl KVStore for VssStore {
...
-			inner
-				.read_internal(&inner.async_client, primary_namespace, secondary_namespace, key)
-				.await
+			let task = runtime.spawn(async move {
+				inner
+					.read_internal(&inner.async_client, primary_namespace, secondary_namespace, key)
+					.await
+			});
+			task.await.map_err(|e| {
+				io::Error::new(io::ErrorKind::Other, format!("VSS runtime task failed: {}", e))
```
benthecarman, line 205: same here

tests/common/mod.rs
```diff
@@ -1660,104 +1663,53 @@ impl KVStore for TestSyncStore {
...
-
-	fn list(
-		&self, primary_namespace: &str, secondary_namespace: &str,
-	) -> lightning::io::Result<Vec<String>> {
-		self.inner.list_internal(primary_namespace, secondary_namespace)
+		async move { inner.list_internal_async(&primary_namespace, &secondary_namespace).await }
 	}
 }
 
 struct TestSyncStoreInner {
```
benthecarman, line 1703: Can this just be called `TestStoreInner`/`TestStore` since its all async now? Also noticed I missed adding postgres to this

src/io/postgres_store/mod.rs
joostjager: Agreed on the duplication

src/io/test_utils.rs
joostjager: Will await the discussion there. Would be nice to use the incremental persister in tests.

src/io/test_utils.rs
joostjager: Is it tracked somewhere?

src/io/vss_store.rs
joostjager: I see you removed the lazy init and made it a field? That was the suggestion indeed

src/runtime.rs
joostjager: I assume my last comment didn't change your view on it. It's temporary, so it's fine.

joostjager (dismissed):
Generally happy with where this landed. No blockers.

One process point: the PR is quite large and that makes review and re-review inefficient. Also managing the many discussing threads becomes unwieldy and risks loosing track of something that might be important. In my opinion better to avoid in the future.

I think the extensive comments that you and/or AI added are good context for future humans or AIs reading it.

I’ll take a look at the follow-up PR too, to see how the final async end state takes shape.

src/io/postgres_store/mod.rs
tnull: Well we had helper previously, but Joost asked to drop it and move it to `StoreRuntime`, which however doesn't work well either (given the error messages are store-specific), so decided to simply inline them, esp. given it's just 5 lines, and hiding them in a helper just obfuscates what happens concretely, i.e., what runtime exactly is used to await what etc..

src/io/vss_store.rs
tnull: See https://github.com/lightningdevkit/ldk-node/pull/919#discussion_r3378788909

tests/common/mod.rs
tnull: > Can this just be called `TestStoreInner`/`TestStore` since its all async now?

No, this would be pretty confusing, as the `Sync` part of it is important (i.e., the purpose is to assert all stores are kept in-sync). Note we also have `TestStore` in LDK test utils which we used to use here, but which unfortunately is now unusable for us as we decided that `KVStore` path it requires some additional per-operation confirmations that we won't be able to run in LDK Node, hence why we introduced `InMemoryStore`.

> Also noticed I missed adding postgres to this

Ah, right, feels a bit orthogonal to this PR, feel free to open a quick follow-up.

src/io/vss_store.rs
tnull: > I see you removed the lazy init and made it a field? That was the suggestion indeed

Well, we just returned to the previous behavior using the internal runtime to `block_on` on the version negotiation. We'll need to return to the lazy init likely in the next step though.

src/io/test_utils.rs
tnull: Yes https://github.com/lightningdevkit/rust-lightning/pull/4665

tnull:
> Generally happy with where this landed. No blockers.
> 
> One process point: the PR is quite large and that makes review and re-review inefficient. Also managing the many discussing threads becomes unwieldy and risks loosing track of something that might be important. In my opinion better to avoid in the future.
> 
> I think the extensive comments that you and/or AI added are good context for future humans or AIs reading it.
> 
> I’ll take a look at the follow-up PR too, to see how the final async end state takes shape.

Squashed fixups without further changes. Given how invasive the changeset is I'd like to move on this one rather fast, any remaining comments / improvements can happen in follow-ups IMO.

tnull:
Ah, grr, still blocked on https://github.com/lightningdevkit/rust-lightning/pull/4661.

tnull:
> Ah, grr, still blocked on [lightningdevkit/rust-lightning#4661](https://github.com/lightningdevkit/rust-lightning/pull/4661).

Okay, now converted the `DROPME` commit to an actual bump commit for the `0.3` revision post-merge of that PR.

src/runtime.rs
tnull: > Still curious why it is needed

Well, I don't think it is (given we should never use blocking threads on these runtimes), but also doesn't hurt to bound the number to be more in-line with the API-configured limit.

joostjager (approved):
No CI. I assume the tests pass locally?

PR #922, opened 2026-06-05
Add 60 minute timeout to CI jobs

benthecarman:
We ran out of our CI limit largely from ldk-node. We had a few jobs this week run for multiple hours because of a hanging test. Add 60 minute timeout to all our jobs to prevent this in the future.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -9,25 +9,26 @@ concurrency:
 jobs:
   benchmark:
     timeout-minutes: 60
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
     env:
       TOOLCHAIN: stable
     steps:
       - name: Checkout source code
-        uses: actions/checkout@v6
+        uses: actions/checkout@v4
       - name: Install Rust toolchain
         run: |
           curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --profile=minimal --default-toolchain stable
-          rustup override set stable
+      - name: Set Rust override
+        run: rustup override set stable
       - name: Enable caching for bitcoind
         id: cache-bitcoind
-        uses: actions/cache@v5
+        uses: actions/cache@v4
         with:
           path: bin/bitcoind-${{ runner.os }}-${{ runner.arch }}
           key: bitcoind-29.0-${{ runner.os }}-${{ runner.arch }}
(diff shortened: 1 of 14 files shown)


tnull (commented):
> We ran out of our CI limit largely from ldk-node. 

Not opposed to this change, but note that it's mainly `ldk-node` as `rust-lightning` has switched to a hosted runner for a while. The more important fix here is to also switch LDK Node over for all jobs where we can. I need to finally do that.

benthecarman:
> > We ran out of our CI limit largely from ldk-node.
> 
> Not opposed to this change, but note that it's mainly `ldk-node` as `rust-lightning` has switched to a hosted runner for a while. The more important fix here is to also switch LDK Node over for all jobs where we can. I need to finally do that.

Added a commit for that

benthecarman:
rip

benthecarman:
we're back?

.github/workflows/benchmarks.yml
```diff
@@ -9,25 +9,26 @@ concurrency:
...
   benchmark:
     timeout-minutes: 60
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
     env:
       TOOLCHAIN: stable
     steps:
       - name: Checkout source code
-        uses: actions/checkout@v6
+        uses: actions/checkout@v4
```
tnull, line 17: Doing all these downgrades makes no sense to me. @TheBlueMatt any reason we can't update the runners to support `node24`?

.github/workflows/rust.yml
```diff
@@ -94,14 +94,15 @@ jobs:
...
     steps:
       - name: Checkout source code
-        uses: actions/checkout@v6
-      - name: Install Rust and clippy
+        uses: actions/checkout@v4
+      - name: Install Rust stable toolchain
         run: |
           curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --profile=minimal --default-toolchain stable
-          rustup component add clippy
+      - name: Add clippy component
```
tnull, line 104: Why does this need to be a separate step now? How is this related to the change at hand?

benthecarman: On the self hosted runner it doesn't update our PATH from the rustup install until we do another step

tnull: Interesting. Alright!

tnull (approved):
Landing this

PR #927, opened 2026-06-10
Reject oversized splice-out amounts

tnull:
Validate splice-out requests against outbound capacity after converting the requested satoshi amount to millisatoshis with overflow handling. This prevents values above the spendable channel balance from slipping past the guard due to a unit mismatch.

Keep splice integration coverage aligned with the corrected capacity semantics by rejecting an amount one satoshi above outbound capacity and deriving the full-cycle splice-out amount from the channel's current spendable capacity.

AI-Assisted-By: OpenAI Codex

Co-Authored-By: HAL 9000

This finding was discovered by Project Loupe

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1700,7 +1700,9 @@ impl Node {
 		if let Some(channel_details) =
 			open_channels.iter().find(|c| c.user_channel_id == user_channel_id.0)
 		{
-			if splice_amount_sats > channel_details.outbound_capacity_msat {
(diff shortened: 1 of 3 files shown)


jkczyz (approved):
LGTM though commit message needs to be wrapped.

tnull:
> LGTM though commit message needs to be wrapped.

Yeah, it seems Codex keeps not doing that if I dont explictly prompt for it :(

PR #929, opened 2026-06-10
Prevent path traversal in sqlite file name

benthecarman:
Previously someone could set their db file name to something like `../ldk-node.sqlite` and we would accept this as a file path and write the databse in the parent directory. This can cause issues so we now only allow for normal file names. This still does allow for sub dirs, ie setting your db name to `mysubdir/ldk-node.sqlite` and it'll go into the sub directory `mysubdir`. I didn't want to disallow this people may be using this today and doesn't seem explicity bad like the parent dir traversal.

This finding was discovered by Project Loupe.

Codex was used to help write this.

### DIFF
--- a/src/io/sqlite_store/mod.rs
+++ b/src/io/sqlite_store/mod.rs
@@ -9,7 +9,7 @@
 use std::collections::HashMap;
 use std::fs;
 use std::future::Future;
-use std::path::PathBuf;
+use std::path::{Component, Path, PathBuf};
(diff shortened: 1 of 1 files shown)


tnull (commented):
I mean configuring SQLite should never be a public-facing API, and we can't guard against all cases of the operator messing up. Closing this as a wontfix, but feel free to reopen if you think this is worth the noise.

PR #931, opened 2026-06-11
Document compat. guarantees, monitor serialization compat

### LINKED ISSUE #74: Document compat. guarantees
Besides comments in the code, we'll add a dedicated doc that tracks (in particular forward) compat changes in question and at which point in time we'll consider breaking compat for them. This allows us to clean up our code base when possible.

tnull:
Fixes #74.

We briefly document our compat guarantees in `README.md`. 

We also add a canary test that checks whether we're forwards compatible with v0.7.0 on the serialization layer. Note that due to the recent schema upgrades of `SqliteStore`, `FilesystemStore`, `VssStore` we don't actually test full downgrades to prior versions. However, this canary is meant to be extended going forward, so that hopefully at some point we can be comfortable to guarantee forward compatibility guarantees also.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ LDK Node currently comes with a decidedly opinionated set of design choices:
 
 - On-chain data is handled by the integrated [BDK][bdk] wallet.
 - Chain data may currently be sourced from the Bitcoin Core RPC interface, or from an [Electrum][electrum] or [Esplora][esplora] server.
-- Wallet and channel state may be persisted to an [SQLite][sqlite] database, to file system, or to a custom back-end to be implemented by the user.
+- Wallet and channel state may be persisted to an [SQLite][sqlite] or [PostgreSQL][postgresql] database, to file system, or to a custom back-end to be implemented by the user.
 - Gossip data may be sourced via Lightning's peer-to-peer network or the [Rapid Gossip Sync](https://docs.rs/lightning-rapid-gossip-sync/*/lightning_rapid_gossip_sync/) protocol.
 - Entropy for the Lightning and on-chain wallets may be sourced from raw bytes or a [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) mnemonic. In addition, LDK Node offers the means to generate and persist the entropy bytes to disk.
 
@@ -85,6 +85,7 @@ The Minimum Supported Rust Version (MSRV) is currently 1.85.0.
 [electrum]: https://github.com/spesmilo/electrum-protocol
 [esplora]: https://github.com/Blockstream/esplora
 [sqlite]: https://sqlite.org/
+[postgresql]: https://www.postgresql.org/
 [rust]: https://www.rust-lang.org/
 [swift]: https://www.swift.org/
 [kotlin]: https://kotlinlang.org/
--- a/README.md
+++ b/README.md
@@ -66,6 +66,10 @@ LDK Node currently comes with a decidedly opinionated set of design choices:
 - Gossip data may be sourced via Lightning's peer-to-peer network or the [Rapid Gossip Sync](https://docs.rs/lightning-rapid-gossip-sync/*/lightning_rapid_gossip_sync/) protocol.
 - Entropy for the Lightning and on-chain wallets may be sourced from raw bytes or a [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) mnemonic. In addition, LDK Node offers the means to generate and persist the entropy bytes to disk.
 
+## Compatibility
+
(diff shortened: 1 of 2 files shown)


tests/upgrade_downgrade_tests.rs
```diff
@@ -0,0 +1,382 @@
...
+	);
+	let node_b_v070 = build_v070_node(
+		storage_path_b,
+		NODE_B_SEED_BYTES,
+		v070_addresses_b.clone(),
+		"downgrade-b",
+		&esplora_url,
+	);
+
+	assert_eq!(node_a_v070.node_id(), node_id_a);
```
joostjager, line 137: Maybe recheck to see if the payment is still there?

README.md
```diff
@@ -66,6 +66,10 @@ LDK Node currently comes with a decidedly opinionated set of design choices:
 - Gossip data may be sourced via Lightning's peer-to-peer network or the [Rapid Gossip Sync](https://docs.rs/lightning-rapid-gossip-sync/*/lightning_rapid_gossip_sync/) protocol.
 - Entropy for the Lightning and on-chain wallets may be sourced from raw bytes or a [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) mnemonic. In addition, LDK Node offers the means to generate and persist the entropy bytes to disk.
 
+## Compatibility
+
+LDK Node does not provide a stable public API until v1.0. We do aim to keep persisted node state backwards compatible, so newer releases are guaranteed to be able to load state written by older releases.
```
joostjager: This reads like an aspiration (aim) mixed with a guarantee. Is it a guarantee?

Perhaps also state explicitly that downgrades are not supported.

tests/upgrade_downgrade_tests.rs
```diff
@@ -0,0 +1,381 @@
...
+
+async fn next_v070_event(node: &ldk_node_070::Node) -> ldk_node_070::Event {
+	tokio::time::timeout(Duration::from_secs(common::INTEROP_TIMEOUT_SECS), node.next_event_async())
+		.await
+		.unwrap_or_else(|_| panic!("{} timed out waiting for event", node.node_id()))
+}
+
+async fn drain_v070_events(node: &ldk_node_070::Node) {
+	while tokio::time::timeout(Duration::from_millis(250), node.next_event_async()).await.is_ok() {
+		node.event_handled().unwrap();
```
joostjager, line 361: Not sure if you want to do any kind of matching on types here. Maybe an error surfaces here?

tnull: Not sure I follow, can you reformulate your question?

README.md
tnull: Yes, good point. Codex first take was too lax and I only amended the second part of the sentence. Now added a fixup.

tests/upgrade_downgrade_tests.rs
joostjager: I mean events are drained without looking at them, and I was wondering if a downgrade problem could surface there too (and is currently ignored).

tnull: Hmm, well, I think we have `expect` calls for the event types we want to check already, and for the rest we just ignore? Do you have any particular checks in mind that we'd should still be doing?

joostjager: No, nothing in mind. Just posing the question whether this is a way to detect problems. If not, also fine.

joostjager (dismissed):

tnull:
Squashed fixups, and included minor change to account for `uniffi` API differences in tests.

tnull:
Accidentally rebased before (now reverted). Net diff since last push is:

```diff
> git diff-tree -U2 6741c a3a76061
diff --git a/tests/upgrade_downgrade_tests.rs b/tests/upgrade_downgrade_tests.rs
index dbfcf7c0..b30b5a33 100644
--- a/tests/upgrade_downgrade_tests.rs
+++ b/tests/upgrade_downgrade_tests.rs
@@ -186,4 +186,5 @@ fn build_current_node(
 	let mut fs_store_path = PathBuf::from(&config.storage_dir_path);
 	fs_store_path.push("fs_store");
+	#[allow(unused_mut)]
 	let mut builder = ldk_node::Builder::from_config(config);
 	builder.set_node_alias(alias.to_string()).unwrap();
@@ -244,5 +245,8 @@ async fn send_current_bolt11_payment(
 		CurrentDescription::new(description.to_owned()).unwrap(),
 	);
-	let invoice = payee.bolt11_payment().receive(amount_msat, &invoice_description, 3600).unwrap();
+	let invoice = payee
+		.bolt11_payment()
+		.receive(amount_msat, &invoice_description.clone().into(), 3600)
+		.unwrap();
 	let payment_id = payer.bolt11_payment().send(&invoice, None).unwrap();
 	expect_current_payment_successful(payer, &payment_id).await;
```

joostjager (approved):
Didn't verify whether uniffi tests pass locally now.

PR #933, opened 2026-06-11
Persist datastore changes before caching

benthecarman:
Keep the in-memory datastore aligned with the persistent store when writes fail. Updates, insert-or-update calls, and removals now leave the cached state unchanged unless the backing write succeeds.

Add regression coverage for failed datastore persistence across insert, update, insert-or-update, and remove paths.

Issue found by Project Loupe

### DIFF
--- a/src/data_store.rs
+++ b/src/data_store.rs
@@ -150,23 +150,23 @@ where
 
 	pub(crate) async fn update(&self, update: SO::Update) -> Result<DataStoreUpdateResult, Error> {
 		let _guard = self.mutation_lock.lock().await;
-		let (res, data_to_persist) = {
-			let mut locked_objects = self.objects.lock().expect("lock");
-			if let Some(object) = locked_objects.get_mut(&update.id()) {
-				let updated = object.update(update);
-				if updated {
-					(DataStoreUpdateResult::Updated, Some(Self::encode_object(object)))
-				} else {
-					(DataStoreUpdateResult::Unchanged, None)
-				}
-			} else {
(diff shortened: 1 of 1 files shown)


tnull (commented):
Please break this into at least 3 commits, one per fix/method touched.

tnull (commented):

tnull (commented):
Thanks! I think at the time we actually preferred to avoid all the re-allocations as we expected to panic on persistence failures anyways. I think that might still be the case, but this PR is also fine.

Might need a rebase after https://github.com/lightningdevkit/ldk-node/pull/943 landed. This has me realizing that #943 maybe should have been less broad, i.e., maybe it should have only switched to `PaginatedKVStore` where we actually require pagination.

benthecarman:
> maybe it should have only switched to PaginatedKVStore where we actually require pagination.

How would we do that? If we want it when doing `list_forwarded_payments` it would have to be implemented in the node's kv store so it needs to be the builder. I guess we could do it so the user passes in the kv store but seems a bit weird

tnull:
> How would we do that? If we want it when doing `list_forwarded_payments` it would have to be implemented in the node's kv store so it needs to be the builder. I guess we could do it so the user passes in the kv store but seems a bit weird

Fair, conceptually it's now just a bit weird that we require the supertrait everywhere even when we don't require it. But yeah, probably we'll want `DataStore` to offer pagination support anyways.

tnull (approved):

PR #937, opened 2026-06-12
Bump BDK, address pending payment store follow ups

### LINKED ISSUE #770: Follow-ups to PR #658
In #658 we switched to use BDK wallet events to update the payment store and introduced a new 'pending payement store'. We merged the PR, though here are some follow-ups I want to address before the next release:

- [ ] `io/mod.rs`: Move persistence `const`s up to other payment `const`s
- [ ] DRY up `read_` method
- [ ] Don't expose `PendingPaymentDetails` in public API
- [ ] Move `PendingPaymentStore` to `payment_store.rs`
- [ ] Make `conflicting_txids` a `Vec` rather than `Option<Vec<>>`
- [ ] Not convinced we need to track entire `PaymentDetails` in the pending store, plus semantically it doens't make sense to track conflicts for BOLT11 payments for example. Maybe there is another way?
- [ ] Do we really need to track pending payments in both stores side-by-side?

### LINKED ISSUE #776: Drop copied-over `wallet_events` when possible

### LINKED ISSUE #901: Drop rounding correction once we update to BDK 3.1.0
Because of a small bug in bdk that causes us to estimate fees slightly differently than them. We have some logic to correct it. Once we update to bdk 3.1.0 we can drop this logic

https://github.com/lightningdevkit/ldk-node/pull/872#discussion_r3193905359_

tnull:
Fixes #770.
Fixes #776.
Fixes #901.

We bump our BDK dependency and address some follow-ups to #658.

(cc @Camillarhi)

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -185,29 +185,13 @@ impl Wallet {
 
 		let mut locked_wallet = self.inner.lock().expect("lock");
 
-		let chain_tip1 = locked_wallet.latest_checkpoint().block_id();
-		let wallet_txs1 = locked_wallet
-			.transactions()
-			.map(|wtx| (wtx.tx_node.txid, (wtx.tx_node.tx.clone(), wtx.chain_position)))
-			.collect::<std::collections::BTreeMap<
-				Txid,
-				(Arc<Transaction>, bdk_chain::ChainPosition<bdk_chain::ConfirmationBlockTime>),
-			>>();
-
-		locked_wallet.apply_unconfirmed_txs(unconfirmed_txs);
-		locked_wallet.apply_evicted_txs(evicted_txids);
-
-		let chain_tip2 = locked_wallet.latest_checkpoint().block_id();
-		let wallet_txs2 = locked_wallet
-			.transactions()
-			.map(|wtx| (wtx.tx_node.txid, (wtx.tx_node.tx.clone(), wtx.chain_position)))
-			.collect::<std::collections::BTreeMap<
-				Txid,
-				(Arc<Transaction>, bdk_chain::ChainPosition<bdk_chain::ConfirmationBlockTime>),
-			>>();
-
-		let events =
-			wallet_events(&mut *locked_wallet, chain_tip1, chain_tip2, wallet_txs1, wallet_txs2);
+		let events = locked_wallet
+			.events_helper(|wallet| -> Result<(), std::convert::Infallible> {
+				wallet.apply_unconfirmed_txs(unconfirmed_txs);
+				wallet.apply_evicted_txs(evicted_txids);
+				Ok(())
+			})
+			.expect("applying mempool updates cannot fail");
 
 		self.update_payment_store(&mut *locked_wallet, events).map_err(|e| {
 			log_error!(self.logger, "Failed to update payment store: {}", e);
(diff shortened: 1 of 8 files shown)


tnull:
Now that `bdk_wallet` 3.1 shipped over the weekend, I now directly bump to that and include a commit addressing https://github.com/lightningdevkit/ldk-node/issues/901. I also added a commit that has us respect anchor reserves when fee bumping.

src/wallet/mod.rs
```diff
@@ -526,6 +525,19 @@ impl Wallet {
 		Ok(())
 	}
 
+	fn cancel_tx_inner(
+		locked_wallet: &mut PersistedWallet<KVStoreWalletPersister>, tx: &Transaction,
+	) {
+		for txout in &tx.output {
+			if let Some((keychain, index)) =
+				locked_wallet.derivation_of_spk(txout.script_pubkey.clone())
```
jkczyz: Might be able to get rid of `clone` by passing the `tx` by value.

src/payment/store.rs
```diff
@@ -589,49 +589,56 @@ impl StorableObjectUpdate<PaymentDetails> for PaymentDetailsUpdate {
...
 	fn update(&mut self, update: Self::Update) -> bool {
 		let mut updated = false;
 
-		// Update the underlying payment details if present
-		if let Some(payment_update) = update.payment_update {
-			updated |= self.details.update(payment_update);
+		if self.txid != update.txid {
+			let old_txid = self.txid;
+			self.txid = update.txid;
+			if old_txid != self.txid && !self.conflicting_txids.contains(&old_txid) {
```
jkczyz: Isn't `old_txid != self.txid` redundant?

src/wallet/mod.rs
```diff
@@ -389,8 +403,22 @@ impl Wallet {
...
+								payment_id,
+							);
+							continue;
+						},
+					};
+					let pending_payment_details = self.create_pending_payment_from_tx(
+						payment_id,
+						payment_txid,
+						conflict_txids,
+					);
```
jkczyz: Regarding the approach from #888, it's probably better to take this approach rather than using `ChannelReady` as you suggest in that PR. IIUC, it will remove the race condition that needs to be protected yet. Gonna give that a try by rebasing on this PR and will report back there.

tnull: Alright, sounds good, let me know.

jkczyz: The one concern about this approach is the amount of data in `TransactionType::InteractiveFunding`. Following #791's approach, we can store minimal information like `Vec<Channel>`. But we do need to keep other information until confirmation in order to set the correct amount and fee in case an RBF candidate other than the latest confirms.

We can use the pending payment store for this, storing either everything or a minimal fields like this:

```rust
 pub(crate) struct FundingTxCandidate {
     pub txid: Txid,
     pub amount_msat: Option<u64>,
     pub fee_paid_msat: Option<u64>,
 }
```

Since we don't support mixed-mode splicing, we don't need to store the direction. But supporting that or batched splices / v2 opens would require including that here. Or including some other data instead of a direction if we think that it is no longer meaningful in the presence of those features.

jkczyz: Another confusing point about direction, if we splice-out we may say this is outbound, but should it be if it goes to our own address?

tnull: > The one concern about this approach is the amount of data in `TransactionType::InteractiveFunding`. Following #791's approach, we can store minimal information like `Vec<Channel>`. But we do need to keep other information until confirmation in order to set the correct amount and fee in case an RBF candidate other than the latest confirms.
> 
> We can use the pending payment store for this, storing either everything or a minimal fields like this:
> 
> ```rust
>  pub(crate) struct FundingTxCandidate {
>      pub txid: Txid,
>      pub amount_msat: Option<u64>,
>      pub fee_paid_msat: Option<u64>,
>  }
> ```
> 
> Since we don't support mixed-mode splicing, we don't need to store the direction. But supporting that or batched splices / v2 opens would require including that here. Or including some other data instead of a direction if we think that it is no longer meaningful in the presence of those features.

Okay, but that seems reasonable? It's fine to add additional fields to the pending payment store, as long as we're fine with them getting pruned eventually once the transactions confirm?

> Another confusing point about direction, if we splice-out we may say this is outbound, but should it be if it goes to our own address?

That seems also fine? Potentially we could consider a third variant that indicates 'internal balance transfer' or  similar, but not sure it's worth?

src/wallet/mod.rs
tnull: Done in fixup.

src/payment/store.rs
tnull: Done in fixup.

tests/integration_tests_rust.rs
```diff
@@ -1150,15 +1150,13 @@ async fn splice_channel() {
...
-	// change output.
-	//
-	// TODO: Some of the discrepancy is addressed upstream, so this number should be adjusted when
-	// updating the BDK wallet dependency. See: https://github.com/bitcoindevkit/bdk_wallet/pull/479
-	let expected_splice_in_lightning_balance_sat = 4_000_003;
+	// BDK 3.1.0 avoids the previous per-UTXO fee rounding during coin selection. Keep the
+	// remaining 2-sat LDK/BDK fee-accounting drift explicit so a dependency change cannot silently
+	// reintroduce the larger surplus. Rather than giving the extra sats to the miner, LDK sends
+	// them to the channel balance since there may not be a change output.
+	let expected_splice_in_lightning_balance_sat = 4_000_002;
```
jkczyz, line 1254: Should this go in the first commit? Otherwise, the test fails on all the commits until this one.

src/payment/store.rs
```diff
@@ -589,49 +589,56 @@ impl StorableObjectUpdate<PaymentDetails> for PaymentDetailsUpdate {
...
 
-		if let Some(new_conflicting_txids) = update.conflicting_txids {
-			if self.conflicting_txids != new_conflicting_txids {
-				self.conflicting_txids = new_conflicting_txids;
+		for txid in update.conflicting_txids {
+			if txid != self.txid && !self.conflicting_txids.contains(&txid) {
+				self.conflicting_txids.push(txid);
 				updated = true;
 			}
 		}
```
jkczyz: Claude found this and claims that a unit test would catch it:

In `PendingPaymentDetails::update`, the txid-rotation branch only de-dupes `old_txid` before pushing it — it doesn't exclude the *new* `self.txid` from `conflicting_txids`. The merge loop right below it does guard with `txid != self.txid`, so the two paths are inconsistent:

```rust
if self.txid != update.txid {
    let old_txid = self.txid;
    self.txid = update.txid;
    if !self.conflicting_txids.contains(&old_txid) {   // doesn't exclude the new self.txid
        self.conflicting_txids.push(old_txid);
    }
    updated = true;
}
for txid in update.conflicting_txids {
    if txid != self.txid && !self.conflicting_txids.contains(&txid) {  // this one does
        ...
    }
}
```

This lets the canonical txid end up listed among its own conflicts. Scenario: after an RBF `T1 -> T2`, the record is `{ payment_id: P(T1), txid: T2, conflicting_txids: [T1] }`. If `T2` is later evicted and the original `T1` confirms (shallower than `ANTI_REORG_DELAY`), `TxConfirmed { txid: T1 }` resolves back to `P(T1)` via the direct branch in `find_payment_by_txid`, and the pending arm calls `create_pending_payment_from_tx(P(T1), T1, vec![])`. `update()` then rotates `self.txid` back to `T1` and pushes `T2`, leaving `{ txid: T1, conflicting_txids: [T1, T2] }` — `T1` now conflicts with itself.

The impact is small (lookups still resolve correctly), but the persisted record becomes self-inconsistent, which also means the "`conflicting_txids` never contains the current txid" invariant isn't actually maintained on this path. Could fix it by normalizing after the rotation, e.g.:

```rust
self.txid = update.txid;
self.conflicting_txids.retain(|c| *c != self.txid);
```

Reachability depends on BDK re-emitting a `TxConfirmed` for a previously-replaced tx after the replacement is dropped — does that match what you'd expect here?

src/payment/store.rs
```diff
@@ -589,49 +589,56 @@ impl StorableObjectUpdate<PaymentDetails> for PaymentDetailsUpdate {
...
 		let mut updated = false;
 
-		// Update the underlying payment details if present
-		if let Some(payment_update) = update.payment_update {
-			updated |= self.details.update(payment_update);
+		if self.txid != update.txid {
+			let old_txid = self.txid;
+			self.txid = update.txid;
+			if old_txid != self.txid && !self.conflicting_txids.contains(&old_txid) {
+				self.conflicting_txids.push(old_txid);
```
jkczyz: IIUC, this means an `update` can add but never remove.

jkczyz (commented):
Looks like the fixups aren't interleaved

tests/integration_tests_rust.rs
tnull: Squashed into the first commit.

tnull:
Now decided to drop two commits (consolidating in one file, compacting pending payment store) to defer the decision how to simplify/refactor pending payment store post #948, as there we'll likely start to track user-registered payments (manual claim flow) in the pending payment store, i.e., it could be simpler to just keep the `PaymentDetails` around for that, though I'm not quite sure yet.

src/payment/store.rs
tnull: Addressed in a fixup.

src/payment/store.rs
tnull: Outdated by now, might revisit in the future. Now opened https://github.com/lightningdevkit/ldk-node/issues/950 to track that.

Camillarhi:
Thanks for this @tnull. Just a few commit nits, there's an `f` before two commit messages `f Build RBF test with UniFFI` and `f Keep pending tx conflicts canonical`

tnull:
> Thanks for this @tnull. Just a few commit nits, there's an `f` before two commit messages `f Build RBF test with UniFFI` and `f Keep pending tx conflicts canonical`

Yes, this is intentional as they are fixups to-be-squashed after review.

jkczyz (commented):
LGTM. Please squash.

tnull:
> LGTM. Please squash.

jkczyz (approved):

PR #939, opened 2026-06-12
Expose BOLT11 underpayment sends

benthecarman:
Needed for orange-sdk

Add a BOLT11 payment API for sending less than the invoice amount while
using the invoice amount as the declared total MPP value. Cover the path
with an integration test where two nodes each pay half of one invoice and
the receiver claims the full amount.

Did a refactor before to clean up and unify the bolt11 send functions.

### DIFF
--- a/src/payment/bolt11.rs
+++ b/src/payment/bolt11.rs
@@ -282,7 +282,8 @@ mod tests {
 impl Bolt11Payment {
 	fn send_internal(
 		&self, invoice: &LdkBolt11Invoice, amount_msat: Option<u64>,
-		route_parameters: Option<RouteParametersConfig>, invalid_amount_log: &'static str,
+		route_parameters: Option<RouteParametersConfig>,
+		declared_total_mpp_value_msat_override: Option<u64>, invalid_amount_log: &'static str,
 	) -> Result<PaymentId, Error> {
 		if !*self.is_running.read().expect("lock") {
 			return Err(Error::NotRunning);
@@ -314,6 +315,7 @@ impl Bolt11Payment {
 		let optional_params = OptionalBolt11PaymentParams {
 			retry_strategy,
 			route_params_config,
+			declared_total_mpp_value_msat_override,
 			..Default::default()
 		};
 		match self.channel_manager.pay_for_bolt11_invoice(
@@ -401,6 +403,7 @@ impl Bolt11Payment {
 			invoice,
 			None,
 			route_parameters,
+			None,
 			"Failed to send payment due to the given invoice being \"zero-amount\". Please use send_using_amount instead.",
 		)
 	}
@@ -436,6 +439,52 @@ impl Bolt11Payment {
 			invoice,
 			Some(amount_msat),
 			route_parameters,
+			None,
+			"Failed to send payment due to amount given being insufficient.",
+		)
+	}
+
+	/// Send a payment given an invoice and an amount lower than the invoice amount.
+	///
+	/// This uses LDK's partial MPP support by declaring the invoice amount as the total MPP value
+	/// while only sending `amount_msat` from this node. The receiving node must be willing to
+	/// accept underpaying HTLCs for the payment to complete.
+	///
+	/// This will fail if the invoice is a zero-amount invoice, or if the amount given is greater
+	/// than or equal to the value required by the invoice. Use [`Self::send_using_amount`] instead
+	/// when paying a zero-amount invoice or paying at least the invoice amount.
+	///
+	/// If `route_parameters` are provided they will override the default as well as the
+	/// node-wide parameters configured via [`Config::route_parameters`] on a per-field basis.
+	pub fn send_using_amount_underpaying(
+		&self, invoice: &Bolt11Invoice, amount_msat: u64,
+		route_parameters: Option<RouteParametersConfig>,
+	) -> Result<PaymentId, Error> {
+		if !*self.is_running.read().expect("lock") {
+			return Err(Error::NotRunning);
+		}
+
+		let invoice = maybe_deref(invoice);
+		let invoice_amount_msat = invoice.amount_milli_satoshis().ok_or_else(|| {
+			log_error!(self.logger, "Failed to underpay as the given invoice is \"zero-amount\".");
+			Error::InvalidInvoice
+		})?;
+
+		if amount_msat >= invoice_amount_msat {
+			log_error!(
+				self.logger,
+				"Failed to underpay as the given amount needs to be less than the invoice amount: required less than {}msat, gave {}msat.",
+				invoice_amount_msat,
+				amount_msat
+			);
+			return Err(Error::InvalidAmount);
+		}
+
+		self.send_internal(
+			invoice,
+			Some(amount_msat),
+			route_parameters,
+			Some(invoice_amount_msat),
 			"Failed to send payment due to amount given being insufficient.",
 		)
 	}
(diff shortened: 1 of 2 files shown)


src/payment/bolt11.rs
```diff
@@ -264,20 +279,20 @@ mod tests {
...
-#[cfg_attr(feature = "uniffi", uniffi::export)]
 impl Bolt11Payment {
-	/// Send a payment given an invoice.
-	///
-	/// If `route_parameters` are provided they will override the default as well as the
-	/// node-wide parameters configured via [`Config::route_parameters`] on a per-field basis.
-	pub fn send(
-		&self, invoice: &Bolt11Invoice, route_parameters: Option<RouteParametersConfig>,
-	) -> Result<PaymentId, Error> {
+	fn ensure_running(&self) -> Result<(), Error> {
```
tnull: No, this is a pattern we have across the codebase, please don't just arbitrarily start extracting it in a helper here.

tests/integration_tests_rust.rs
```diff
@@ -304,6 +304,100 @@ async fn multi_hop_sending() {
...
+		node_c.sync_wallets().unwrap();
+	}
+
+	generate_blocks_and_wait(&bitcoind.client, &electrsd.client, 6).await;
+
+	for node in [&node_a, &node_b, &node_c] {
+		node.sync_wallets().unwrap();
+	}
+
+	expect_channel_ready_event!(node_c, node_a.node_id());
```
tnull: This is flaky:

> thread 'split_underpaid_bolt11_payment' (63174) panicked at tests/integration_tests_rust.rs:358:5:
> assertion `left == right` failed
>   left: Some(PublicKey(188a7ccb0c94bfc5d135a2fcc02a37cca78d4c2c98f72912990ed0e594f5b765547537a41c32c32a0c87302c2d6ced61f6033d2b7c2e4109bb45c29b41439326))
>  right: Some(PublicKey(8cfa601d1be90d36097200793e66c50ac90b4836998149d5b04fc13caaadad08a5e6fa0651778c0eaa5826c52112c8aef2dd211b8fd92efff371063335858396))
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

tests/integration_tests_rust.rs
```diff
@@ -304,6 +304,100 @@ async fn multi_hop_sending() {
...
+	let chain_source = random_chain_source(&bitcoind, &electrsd);
+	let (node_a, node_b) = setup_two_nodes(&chain_source, false, true, false);
+	let node_c = setup_node(&chain_source, random_config(true));
+
+	let addr_c = node_c.onchain_payment().new_address().unwrap();
+	let premine_amount_sat = 5_000_000;
+	premine_and_distribute_funds(
+		&bitcoind.client,
+		&electrsd.client,
+		vec![addr_c],
```
tnull: If you only fund the one node, the others should reject incoming channels due to lack of reserves. This will have you fall back to staticremotekey, likely also fueling the flake below.

tnull (commented):
Few comments, but generally looks good. What will be the BOLT12 story for partial payments?

benthecarman:
> What will be the BOLT12 story for partial payments?

https://github.com/lightningdevkit/rust-lightning/issues/4380

tests/integration_tests_rust.rs
benthecarman: fixed

tests/integration_tests_rust.rs
benthecarman: fixed, needed to use `expect_channel_ready_events` because the ordering for the two channel ready events is ~random

comwanga:
Not assigned to this PR, but wanted to share a few observations;

The refactor into `send_internal` is a clean move — it removes the duplicated routing and payment-store logic across `send` and `send_using_amount`, and the new `send_using_amount_underpaying` builds on that foundation in a straightforward way. The integration test covers the core scenario well, and the fix for the channel-ready event ordering flake (`expect_channel_ready_events!` plural) is the right call.

A few things worth discussing:

**1. `amount_msat` vs `payment_amount_msat` passed to `pay_for_bolt11_invoice`**

In `send_internal`, you compute `payment_amount_msat` to resolve the actual amount to send, but the call to `channel_manager.pay_for_bolt11_invoice` still passes the original `amount_msat` (the `Option<u64>` parameter), not `Some(payment_amount_msat)`. For `send()`, this passes `None`, which is intentional — LDK reads the amount from the invoice. But it is worth double-checking whether this is also the intended behavior when `send_using_amount` calls `send_internal` with `Some(amount_msat)`. The two values are the same in that path so there is no bug, but it would be clearer to pass `Some(payment_amount_msat)` consistently to make the intent explicit and avoid confusion if `send_internal` is extended later.

**2. `invalid_amount_log` is reused for two distinct error conditions**

The same `invalid_amount_log` string is logged for both the case where no amount can be resolved (zero-amount invoice with no override) and the `Bolt11PaymentError::InvalidAmount` from LDK. These are different failures — one is a precondition check on our side, the other is a protocol-level rejection from LDK. Sharing one log string means the log message may be misleading depending on which path fires. Consider whether it is worth using separate log strings for each arm, or at least adding a note in the doc comment about what constitutes an "insufficient" amount.

**3. `PaymentDetails` stores the sent amount, not the invoice amount — verify this is documented**

For `send_using_amount_underpaying`, `PaymentDetails` records `Some(half_amount_msat)` while the invoice is for `full_amount_msat`. This is technically correct since it reflects what this node sent, and the integration test asserts this. But callers querying `list_payments` may be surprised to see an amount that does not match the invoice they paid. It may be worth adding a note to the doc comment on `send_using_amount_underpaying` explaining that the stored amount reflects the amount sent by this node, not the declared MPP total.

**4. Error type for `amount >= invoice_amount` check**

In `send_using_amount_underpaying`, when `amount_msat >= invoice_amount_msat`, you return `Err(Error::InvalidAmount)`. The validation in `send_using_amount` for an over-amount scenario returns `Err(Error::InvalidAmount)` too, so this is consistent. Just confirming this is intentional rather than `Error::InvalidInvoice`.

**5. Test: missing assertion that payment fails if amount equals invoice amount**

The integration test verifies the happy path where both halves add up to the full amount. There is no test for the boundary case where a caller passes `amount_msat == invoice_amount_msat` to `send_using_amount_underpaying`, which should return `InvalidAmount` based on the `>=` check. A small negative test for this edge would strengthen confidence in that guard.

src/payment/bolt11.rs
```diff
@@ -391,94 +435,58 @@ impl Bolt11Payment {
...
 
-						self.runtime.block_on(self.payment_store.insert(payment))?;
-						Err(Error::PaymentSendingFailed)
-					},
-				}
-			},
+		if amount_msat >= invoice_amount_msat {
+			log_error!(
+				self.logger,
+				"Failed to underpay as the given amount needs to be less than the invoice amount: required less than {}msat, gave {}msat.",
```
tnull, line 476: Should this even be an error condition? Yes, they use the wrong API, but we technically still can pay with 0 underpayment? But maybe failing here is preferred? Just checking..

tnull (commented):
Basically looks good, just one non-blocking question regarding the intended API behavior.

src/payment/bolt11.rs
benthecarman: Yeah I think failing here is correct. Maybe now it's technically the same api, but don't want someone to rely on this when in the future that could be harmful.

tnull (approved):

PR #940, opened 2026-06-12
Fix trivial underflow/overflow issues

benthecarman:
A bunch of small issues found by project-loupe. Can drop any commits that we don't think are worth it (the electrum fee one might be a bit much).

### DIFF
--- a/src/payment/unified.rs
+++ b/src/payment/unified.rs
@@ -129,9 +129,9 @@ impl UnifiedPayment {
 	pub fn receive(
 		&self, amount_sats: u64, description: &str, expiry_sec: u32,
 	) -> Result<String, Error> {
-		let onchain_address = self.onchain_payment.new_address()?;
+		let amount_msats = amount_sats.checked_mul(1_000).ok_or(Error::InvalidAmount)?;
 
-		let amount_msats = amount_sats * 1_000;
+		let onchain_address = self.onchain_payment.new_address()?;
 
 		let bolt12_offer =
 			match self.bolt12_payment.receive_inner(amount_msats, description, None, None) {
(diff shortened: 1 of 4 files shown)


src/payment/unified.rs
```diff
@@ -129,9 +129,9 @@ impl UnifiedPayment {
 	pub fn receive(
 		&self, amount_sats: u64, description: &str, expiry_sec: u32,
 	) -> Result<String, Error> {
-		let onchain_address = self.onchain_payment.new_address()?;
+		let amount_msats = amount_sats.checked_mul(1_000).ok_or(Error::InvalidAmount)?;
```
tnull, line 132: Rather than doing this here, shouldn't we rather do the check in BOLT11/BOLT12 receive methods, as this method is just meant to forward to them, with as little custom logic on top as possible. Apart from that we might also be okay to punt on it as we intend to switch to an `LightningAmount` type soon everywhere, that should enforce such checks at conversion time across the board.

src/chain/electrum.rs
tnull, line 1: Let's drop this commit for now, as chain sources are fundamentally considered trusted to a degree. Yes, eventually we might, but I fear that with some arbitrary limit we'll just end up shooting ourselves in the foot.

tnull (commented):
Please also include `This finding was discovered by Project Loupe` in the commit messages, just so we can keep track of how many reported things were fixed.

src/payment/unified.rs
benthecarman: If we wanted to do this check in the bolt11/12 functions then we'd have to have those take sats rather than msats. I think this is the right location for this. `LightningAmount` is the correct choice eventually.

src/chain/electrum.rs
benthecarman: removed

tnull (approved):

PR #941, opened 2026-06-17
Replace runtime's deprecated active_tasks_count with num_alive_tasks

enigbe:


tnull (commented):
When we do this we'll also need to bump our tokio dependency to the version where `num_alive_tasks` was introduced (I think 1.39 or so?)

enigbe:
> When we do this we'll also need to bump our tokio dependency to the version where `num_alive_tasks` was introduced (I think 1.39 or so?)

I've bumped `tokio` to 1.39

tnull (approved):

PR #943, opened 2026-06-18
Require PaginatedKVStore for all stores

benthecarman:
Now that all of our KVStore impl's also impl PaginatedKVStore we can require it when building a node. One caveat being that our upgrade downgrade test no longer works given that these new PaginatedKVStores don't support downgrading to their old, non paginated versions. Because of this we are commenting it out and will bring it back once we tag 0.8

### DIFF
--- a/tests/upgrade_downgrade_tests.rs
+++ b/tests/upgrade_downgrade_tests.rs
@@ -11,409 +11,411 @@
 // reads filesystem-store v1 data. There is no supported v2-to-v1 IO-layer downgrade:
 // v2 stores empty namespaces under `[empty]`, which v1 readers do not look up.
 //
+// TODO(@benthecarman) Bring back after 0.8 is cut.
+
 // To keep monitoring whether the serialized node/channel/payment state remains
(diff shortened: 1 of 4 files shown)


tests/upgrade_downgrade_tests.rs
```diff
@@ -15,405 +15,407 @@
...
-					addr: *addr,
-					port: *port,
-				}
-			},
-			_ => panic!("unexpected non-IPv4 test address: {:?}", address),
-		})
-		.collect()
-}
+//
+// TODO(@benthecarman) Bring back after 0.8 is cut
```
tnull: Mind opening an issue and putting it on the 0.8 milestone for this?

tnull (commented):
This needs a rebase.

tests/upgrade_downgrade_tests.rs
benthecarman: #944

tnull (commented):

tnull (approved):

PR #915, opened 2026-05-31
Add database persistence benchmarks

benthecarman:
Closes #908 

Benchmark realistic payment and pending-payment persistence workloads
across filesystem, SQLite, and optional PostgreSQL stores. Use the async
KV-store APIs so the measured paths match the database interfaces used by
async persistence.

Also adds support for different db backends to the current payment benchmark.

# Benchmark Results

Results on my computer with a Ryzen 7 5800XT 8-Core, 16-Thread CPU

## Operations

| Bench | filesystem | sqlite | postgres |
|---|---:|---:|---:|
| `channel_open` | 4.99 ms | 5.38 ms | 14.74 ms |
| `forwarding` | 5.13 s | 5.16 s | 5.70 s |
| `channel_open_close` | 5.07 s | 5.20 s | 5.08 s |
| `payments` | 5.69 s | 6.26 s | 11.73 s |

## Startup

| Scenario | filesystem | sqlite | postgres |
|---|---:|---:|---:|
| `channels_1_payments_2` | 10.37 ms | 12.04 ms | 57.57 ms |
| `channels_10_payments_2` | 12.02 ms | 16.68 ms | 56.77 ms |
| `channels_100_payments_2` | 18.37 ms | 50.61 ms | 62.03 ms |
| `channels_100_payments_1000` | 17.77 ms | 70.26 ms | 61.86 ms |

## Database Hot Path

| Bench | filesystem | sqlite | postgres |
|---|---:|---:|---:|
| `channel_open_like` | 0.26 ms | 0.46 ms | 7.54 ms |
| `forwarding_25_like` | 5.00 ms | 10.56 ms | 124.89 ms |

PR #945, opened 2026-06-23
Implement `MigratableKVStore` for all KV Stores

benthecarman:
Pulled out of #915

These were unimplemented for our kv stores. Useful if the user wants to migrate to a different database and also in tests so we don't have to re-init and setup a node.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,6 +92,7 @@ winapi = { version = "0.3", features = ["winbase"] }
 
 [dev-dependencies]
 lightning = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "3dfcc4cca1866c5e5d4d4eaf3b82e09584e2ce5c", features = ["std", "_test_utils"] }
+lightning-persister = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "3dfcc4cca1866c5e5d4d4eaf3b82e09584e2ce5c", features = ["tokio"] }
 rand = { version = "0.9.2", default-features = false, features = ["std", "thread_rng", "os_rng"] }
 proptest = "1.0.0"
 regex = "1.5.6"
--- a/src/io/in_memory_store.rs
+++ b/src/io/in_memory_store.rs
@@ -11,7 +11,9 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Mutex;
 
 use lightning::io;
-use lightning::util::persist::{KVStore, PageToken, PaginatedKVStore, PaginatedListResponse};
+use lightning::util::persist::{
+	KVStore, MigratableKVStore, PageToken, PaginatedKVStore, PaginatedListResponse,
+};
 
 const IN_MEMORY_PAGE_SIZE: usize = 50;
 
@@ -95,6 +97,28 @@ impl InMemoryStore {
 			hash_map::Entry::Occupied(e) => Ok(e.get().keys().cloned().collect()),
 			hash_map::Entry::Vacant(_) => Ok(Vec::new()),
 		}
+	}
+
+	fn list_all_keys_internal(&self) -> io::Result<Vec<(String, String, String)>> {
+		let persisted_lock = self.persisted_bytes.lock().unwrap();
+		let capacity = persisted_lock.values().map(|entries| entries.len()).sum();
+		let mut keys = Vec::with_capacity(capacity);
+
+		for (prefixed_namespace, namespace_entries) in persisted_lock.iter() {
+			let (primary_namespace, secondary_namespace) =
+				prefixed_namespace.split_once('/').ok_or_else(|| {
+					io::Error::new(io::ErrorKind::InvalidData, "Invalid namespace format")
+				})?;
+			for key in namespace_entries.keys() {
+				keys.push((
+					primary_namespace.to_string(),
+					secondary_namespace.to_string(),
+					key.clone(),
+				));
+			}
+		}
+
+		Ok(keys)
 	}
 }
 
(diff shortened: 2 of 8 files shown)


tnull (commented):

src/io/sqlite_store/mod.rs
```diff
@@ -486,6 +503,35 @@ impl SqliteStoreInner {
...
+		let sql = format!(
+			"SELECT primary_namespace, secondary_namespace, key FROM {}",
+			self.kv_table_name
+		);
+		let mut stmt = locked_conn.prepare_cached(&sql).map_err(|e| {
+			let msg = format!("Failed to prepare statement: {}", e);
+			io::Error::new(io::ErrorKind::Other, msg)
+		})?;
+
+		let mut keys = Vec::new();
```
tnull: Let's preallocate via `with_capacity` here.

src/io/vss_store.rs
```diff
@@ -625,6 +678,52 @@ impl VssStoreInner {
...
+		&self, client: &VssClient<CustomRetryPolicy>,
+	) -> io::Result<Vec<(String, String, String)>> {
+		let mut page_token: Option<String> = None;
+		let mut keys = vec![];
+		loop {
+			let request = ListKeyVersionsRequest {
+				store_id: self.store_id.clone(),
+				key_prefix: None,
+				page_token,
+				page_size: None,
```
tnull: Do we need to use pagination in case we're migrating with many keys?

src/io/vss_store.rs
```diff
@@ -625,6 +678,52 @@ impl VssStoreInner {
...
+					{
+						self.key_obfuscator.deobfuscate(&kv.key)?
+					},
+					Err(e) => return Err(e),
+				};
+				if primary_namespace.is_empty()
+					&& secondary_namespace.is_empty()
+					&& key == VSS_SCHEMA_VERSION_KEY
+				{
+					continue;
```
tnull, line 714: Hmm, maybe we should attempt to always migrate to the latest schema version, allowing nodes to upgrade?

src/io/in_memory_store.rs
```diff
@@ -96,6 +98,27 @@ impl InMemoryStore {
 			hash_map::Entry::Vacant(_) => Ok(Vec::new()),
 		}
 	}
+
+	fn list_all_keys_internal(&self) -> io::Result<Vec<(String, String, String)>> {
+		let persisted_lock = self.persisted_bytes.lock().unwrap();
+		let mut keys = Vec::new();
```
tnull: nit: Use `with_capacity` here.

tnull (commented):
Thanks, first pass.

Would be cool to add a commit that does a full end-to-end test of migration. I.e., pick a random backend, setup a node with it, open a channel to another node, send some payments, then drop everything and migrate to a random new backend, then ensure the node is still fully functional.

src/io/vss_store.rs
benthecarman: Here we are just listing all keys so we need to support both. The actual migration will just do whatever happens on `.write()`. Assuming it'll be a fresh VSS store it should persist as the latest schema version

https://github.com/lightningdevkit/rust-lightning/blob/9df5c9dc458fecd91c5018150dd2cec1e4ff04c3/lightning/src/util/persist.rs#L679C10-L679C37

src/io/sqlite_store/mod.rs
benthecarman: Added, this one was a little weird because need to do a query first. Maybe worth it though for very large dbs

tnull (commented):
Will need another rebase after https://github.com/lightningdevkit/ldk-node/pull/952 lands.

tests/integration_tests_migration.rs
```diff
@@ -0,0 +1,290 @@
...
+use ldk_node::entropy::NodeEntropy;
+use ldk_node::io::postgres_store::PostgresStore;
+use ldk_node::io::sqlite_store::{SqliteStore, KV_TABLE_NAME, SQLITE_DB_FILE_NAME};
+use ldk_node::{Builder, Event};
+use lightning::util::persist::{migrate_kv_store_data_async, MigratableKVStore};
+use lightning_invoice::{Bolt11InvoiceDescription, Description};
+use lightning_persister::fs_store::v2::FilesystemStoreV2;
+use rand::seq::SliceRandom;
+
+async fn drop_tables<'a>(table_names: impl IntoIterator<Item = &'a String>) {
```
tnull, line 29: Do we really need all this verbose boilerplate for the single test? Seems just choosing two backends at random could be only a few LoC inline?

Cargo.toml
```diff
@@ -92,6 +92,7 @@ winapi = { version = "0.3", features = ["winbase"] }
 
 [dev-dependencies]
 lightning = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "3dfcc4cca1866c5e5d4d4eaf3b82e09584e2ce5c", features = ["std", "_test_utils"] }
+lightning-persister = { git = "https://github.com/lightningdevkit/rust-lightning", rev = "3dfcc4cca1866c5e5d4d4eaf3b82e09584e2ce5c", features = ["tokio"] }
```
tnull, line 95: Why do we need an extra dev-dependency now?

tests/integration_tests_migration.rs
benthecarman: For this and on the dev-dep. We need to init the db outside the node instead of doing like `build_with_fs_store` because we need to be able to pass the db to `migrate_kv_store_data_async`.

Cargo.toml
benthecarman: see above

tests/integration_tests_migration.rs
benthecarman: Cleaned it up a bit with a macro

tnull (approved):

PR #947, opened 2026-06-24
Fix 0.6.2 compatibility test shutdown

tnull:
Stop and drop the old compatibility node from a blocking region.

This avoids a Tokio runtime-drop panic in the async test.

Honestly not quite sure what changed that this only surfaces now that CI is back  (likely an unrelated refactoring while CI was gone is to blame). But we need to fix it so CI finally is green again.

### DIFF
--- a/tests/integration_tests_rust.rs
+++ b/tests/integration_tests_rust.rs
@@ -2580,7 +2580,11 @@ async fn build_0_6_2_node(
 	assert!(balance > 0);
 	let node_id = node_old.node_id();
 
-	node_old.stop().unwrap();
+	// Workaround necessary as v0.6.2's runtime wasn't dropsafe in a tokio context.
+	tokio::task::block_in_place(move || {
+		node_old.stop().unwrap();
+		drop(node_old);
+	});
 
 	(balance, node_id)
 }


tests/integration_tests_rust.rs
```diff
@@ -2675,7 +2580,11 @@ async fn build_0_6_2_node(
 	assert!(balance > 0);
 	let node_id = node_old.node_id();
 
-	node_old.stop().unwrap();
+	// Workaround necessary as v0.6.2's runtime wasn't dropsafe in a tokio context.
+	tokio::task::block_in_place(move || {
```
joostjager, line 2584: The vss_no_auth integration test uses `spawn_blocking` for a similar code block.

joostjager:
AI analysis regarding why now:

What is happening: v0.6.2's `Node::start()` creates and stores its own `tokio::runtime::Runtime`. In the backwards-compatibility test we call `node_old.stop()` from inside `#[tokio::test]`, so shutdown runs while the test is already executing in a Tokio async context.

The subtle part is the `tokio_unstable` cfg. In v0.6.2, `Node::stop()` clones the runtime into `metrics_runtime` under `#[cfg(tokio_unstable)]` to log active task metrics. The shutdown body uses `block_in_place` around its internal `runtime.block_on(...)` calls, but that extra clone remains alive until the end of `stop()`. When `stop()` returns to the async test, the final runtime drop can happen in a context where Tokio does not allow blocking, and Tokio panics with:

```text
Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.

joostjager (approved):

PR #948, opened 2026-06-24
Track inbound payments by `PaymentId`

### LINKED ISSUE #298: Track inbound payments by also unique payment IDs

tnull:
Closes #298.

This finally switches our inbound payments over to be fully tracked by payment id, and decouples payment IDs from payment hashes.


src/payment/bolt11.rs
```diff
@@ -275,15 +213,7 @@ impl Bolt11Payment {
 		}
 
 		let payment_hash = invoice.payment_hash();
-		let payment_id = PaymentId(invoice.payment_hash().0);
-		if let Some(payment) = self.payment_store.get(&payment_id) {
```
tnull, line 279: Dropped this for now, bit on the fence on whether it's worth maintaining a secondary index for this is worth it.

tnull: (Although will probably be necessary for the next commit, too)

src/payment/bolt11.rs
```diff
@@ -69,9 +72,11 @@ impl_writeable_tlv_based!(PaymentMetadata, {
 pub struct Bolt11Payment {
 	runtime: Arc<Runtime>,
 	channel_manager: Arc<ChannelManager>,
+	keys_manager: Arc<KeysManager>,
 	connection_manager: Arc<ConnectionManager<Arc<Logger>>>,
 	liquidity_source: Arc<LiquiditySource<Arc<Logger>>>,
 	payment_store: Arc<PaymentStore>,
+	pending_payment_store: Arc<PendingPaymentStore>,
```
tankyleo: Here and elsewhere, can we avoid adding a separate store to `Bolt11Payment` and `EventHandler`, and move the `prune_expired_pending_payments` to an internal method that gets called internally inside the public methods of the store?

src/payment/bolt11.rs
```diff
@@ -582,6 +627,7 @@ impl Bolt11Payment {
 		}
 
 		self.channel_manager.fail_htlc_backwards(&payment_hash);
+		self.runtime.block_on(self.pending_payment_store.remove(&payment_id))?;
```
tankyleo: Similar vein as above comment, can we do this inside `payment_store.update` a few lines above?

src/event.rs
```diff
@@ -892,7 +1025,55 @@ where
 					amount_msat,
 				);
 				let payment_preimage = match purpose {
-					PaymentPurpose::Bolt11InvoicePayment { payment_preimage, .. } => {
+					PaymentPurpose::Bolt11InvoicePayment {
+						payment_preimage,
+						payment_secret,
+						..
+					} => {
+						if payment_info.is_none() {
```
tankyleo: nit: we can DRY this branch for each `PaymentPurpose` variant

src/payment/bolt11.rs
```diff
@@ -275,15 +301,7 @@ impl Bolt11Payment {
 		}
 
 		let payment_hash = invoice.payment_hash();
-		let payment_id = PaymentId(invoice.payment_hash().0);
-		if let Some(payment) = self.payment_store.get(&payment_id) {
-			if payment.status == PaymentStatus::Pending
-				|| payment.status == PaymentStatus::Succeeded
-			{
-				log_error!(self.logger, "Payment error: an invoice must not be paid twice.");
-				return Err(Error::DuplicatePayment);
```
tankyleo, line 284: Hmm this is for outbound bolt11s, are we good to delete this ?

src/payment/bolt11.rs
tnull: You mean conceptually merging pending payment store and payment store? Not sure we'd want to go there as it would limit the degrees of freedom we have in each.

src/payment/bolt11.rs
tnull: See https://github.com/lightningdevkit/ldk-node/pull/948#discussion_r3535391526, but not sure if I understand your idea correctly.

src/payment/bolt11.rs
tnull: So, LDK itself will error if we recently paid the same invoice. The issue is that going forward we can't necessarily easily keep a secondary index on the payment store, which would be necessary to look up entries by payment hash. We might need to re-add that at some point but I tried to avoid it here.

Note that a) the more important case is the receive case b) checking for duplication based on the payment store state was already a somewhat leaky concept as users ofc. might simply `remove` entries, at which point we don't know anything about them anymore. 

That said, all ears if you have an alternative suggestion.

tnull:
Rebased without further changes for now.

tnull:
Force-pushed some fixes:

```diff
> git diff-tree -U2  d1b5ffb3 b1f65649
diff --git a/bindings/python/src/ldk_node/test_ldk_node.py b/bindings/python/src/ldk_node/test_ldk_node.py
index 304caf9c..063c95f7 100644
--- a/bindings/python/src/ldk_node/test_ldk_node.py
+++ b/bindings/python/src/ldk_node/test_ldk_node.py
@@ -236,5 +236,5 @@ class TestLdkNode(unittest.TestCase):

         sender_payment = node_1.payment(keysend_payment_id)
-        receiver_payment = node_2.payment(keysend_payment_id)
+        receiver_payment = node_2.payment(received_event.payment_id)

         self.assertIsNotNone(sender_payment)
diff --git a/tests/integration_tests_migration.rs b/tests/integration_tests_migration.rs
index c4e63451..59621476 100644
--- a/tests/integration_tests_migration.rs
+++ b/tests/integration_tests_migration.rs
@@ -204,5 +204,5 @@ async fn migrate_node_across_all_backends() {
 	let invoice = node_b.bolt11_payment().receive(10_000, &description.into(), 3600).unwrap();
 	let ln_send_id = node.bolt11_payment().send(&invoice, None).unwrap();
-	expect_payment_successful_event!(node, Some(ln_send_id), None);
+	expect_payment_successful_event!(node, ln_send_id, None);
 	expect_payment_received_event!(node_b, 10_000);

@@ -212,5 +212,5 @@ async fn migrate_node_across_all_backends() {
 	let invoice = node.bolt11_payment().receive(5_000, &description.into(), 3600).unwrap();
 	let ln_receive_id = node_b.bolt11_payment().send(&invoice, None).unwrap();
-	expect_payment_successful_event!(node_b, Some(ln_receive_id), None);
+	expect_payment_successful_event!(node_b, ln_receive_id, None);
 	expect_payment_received_event!(node, 5_000);

```

src/payment/bolt11.rs
```diff
@@ -81,28 +86,88 @@ pub struct Bolt11Payment {
...
+		let now = Self::current_time_secs();
+		let expired_payment_ids = self
+			.pending_payment_store
+			.list_filter(|payment| payment.has_expired(now))
+			.into_iter()
+			.map(|payment| payment.details.id)
+			.collect::<Vec<_>>();
+
+		for payment_id in expired_payment_ids {
+			self.runtime.block_on(self.pending_payment_store.remove(&payment_id))?;
```
tankyleo: Can we do a `batch_remove` here so we avoid repeatedly locking and unlocking multiple mutexes ?

src/payment/pending_payment_store.rs
```diff
@@ -138,7 +167,151 @@ impl From<&PendingPaymentDetails> for PendingPaymentDetailsUpdate {
...
+
+	pub(crate) fn get_pending_manual_bolt11_by_payment_hash(
+		&self, payment_hash: &PaymentHash,
+	) -> Option<PendingPaymentDetails> {
+		let ids = self
+			.manual_bolt11_payment_hash_index
+			.lock()
+			.expect("lock")
+			.get(payment_hash)
+			.cloned()
```
tankyleo: nit: this clone is not necessary we can rework this method to remove it

src/payment/pending_payment_store.rs
```diff
@@ -138,7 +167,151 @@ impl From<&PendingPaymentDetails> for PendingPaymentDetailsUpdate {
...
+		for payment in pending_payments {
+			Self::insert_into_manual_bolt11_hash_index(&mut index, payment);
+		}
+		index
+	}
+
+	fn replace_in_index(
+		&self, before: Option<&PendingPaymentDetails>, after: Option<&PendingPaymentDetails>,
+	) {
+		let mut index = self.manual_bolt11_payment_hash_index.lock().expect("lock");
```
tankyleo: Could we return early here before locking if the payment is not a manual bolt11 ? It seems wasteful to lock the index even though we have enough information here to know whether we'll actually modify the index.

src/event.rs
```diff
@@ -611,6 +609,58 @@ where
...
+		let now = Self::current_time_secs();
+		let expired_payment_ids = self
+			.pending_payment_store
+			.list_filter(|payment| payment.has_expired(now))
+			.into_iter()
+			.map(|payment| payment.details.id)
+			.collect::<Vec<_>>();
+
+		for payment_id in expired_payment_ids {
+			if let Err(e) = self.pending_payment_store.remove(&payment_id).await {
```
tankyleo: As above, `batch_remove` here would be useful

src/event.rs
```diff
@@ -722,8 +773,17 @@ where
 				counterparty_skimmed_fee_msat,
 				..
 			} => {
-				let payment_id = PaymentId(payment_hash.0);
-				let payment_info = self.payment_store.get(&payment_id);
+				let event_payment_id = payment_id.unwrap_or(PaymentId(payment_hash.0));
+				let (mut payment_id, payment_info) =
+					self.resolve_inbound_payment_id(event_payment_id, &payment_hash);
+				let pending_payment = if payment_info.is_none() {
+					self.find_pending_inbound_payment(&payment_hash).await?
```
tankyleo: Should we add a debug assert somewhere in the call chain here that there is a single payment id for that hash ?

src/payment/bolt11.rs
```diff
@@ -124,41 +189,14 @@ impl Bolt11Payment {
...
-			id,
-			kind,
-			amount_msat,
-			None,
-			PaymentDirection::Inbound,
-			PaymentStatus::Pending,
-		);
-		self.runtime.block_on(self.payment_store.insert(payment))?;
+		if let Some(payment_hash) = manual_claim_payment_hash {
+			self.register_manual_claim_invoice(
```
tankyleo: found with codex: is it worth making the "check duplicates" step and the "register invoice" step atomic? Ie try to insert the invoice, and the `pending_payment_store` returns `DuplicatePayment` if there is a collision in the hash.

src/event.rs
```diff
@@ -611,6 +609,58 @@ where
...
+	}
+
+	async fn find_pending_inbound_payment(
+		&self, payment_hash: &PaymentHash,
+	) -> Result<Option<PendingPaymentDetails>, ReplayEvent> {
+		self.prune_expired_pending_payments().await?;
+		Ok(self.pending_payment_store.get_pending_manual_bolt11_by_payment_hash(payment_hash))
+	}
+
+	fn resolve_inbound_payment_id(
```
tankyleo, line 694: From codex: 
>  Manual-claim payment handling can lose the mapping from LDK payment ids to the user-facing random id across replay, causing duplicate/misidentified payment records.

>Review comment:

> - [P2] Resolve manual claims by stored hash after replay _ /home/ubuntu/ldk-node/src/event.rs:654-658
  For manual-claim payments the durable payment record is stored under the random id emitted in `PaymentClaimable`, while a replayed LDK `PaymentClaimed` event carries the LDK-derived id (or the hash-as-id fallback). If the pending row has already been removed, e.g. after a crash or `event_queue.add_
event` failure later in this handler, this resolver misses the existing payment and the replay inserts/emits a duplicate payment under the LDK id. Please also resolve existing Bolt11 records by their stored payment hash, or keep the pending mapping until the user event is durably queued.

src/payment/bolt11.rs
tnull: Yeah, makes sense, will do.

src/payment/pending_payment_store.rs
tnull: Fixed.

src/event.rs

src/event.rs

src/event.rs
tnull: Hmm, good point. Now going for an approach that only drops the entry after `add_event` succeeded, however, that could lead to cases where we end up with dangling entries. Therefore now also refactored the `expires_at` field to use an enum that we can also populate from `PaymentClaimable::claim_deadline`.

src/payment/pending_payment_store.rs
tnull: First had added a fixup that I now  dropped again. The modest gain didn't seem worth the added complexity/churn. Let me know if you feel strongly though.

src/payment/bolt11.rs
tnull: Should be addressed.

src/payment/bolt11.rs
joostjager: Is it really true that LDK errors if the payment id is different?

I was reading https://github.com/lightningdevkit/rust-lightning/blob/506cb91f2e0fb87906188b79777bcf42595d3623/lightning/src/ln/channelmanager.rs#L5653-L5665

tnull: > Is it really true that LDK errors if the payment id is different?
> 
> I was reading https://github.com/lightningdevkit/rust-lightning/blob/506cb91f2e0fb87906188b79777bcf42595d3623/lightning/src/ln/channelmanager.rs#L5653-L5665

Hmm, good catch. Then I was either misremembering or it changed at some point, but I thought `DuplicatePayment` was going off the payment hash, not the id. Still need to double-check how to avoid this case then, although I'm still a bit dubious how to avoid it.

joostjager:
Looking at this through the lens of self-imposed versus intrinsic complexity, I think this PR again makes the cost of our current architecture clear. LDK and ldk-node persist related state independently, without a consistent commit boundary, so we have to recover consistency through careful ordering, replay, cleanup, and expiry logic.

src/payment/pending_payment_store.rs
```diff
@@ -138,17 +189,200 @@ impl From<&PendingPaymentDetails> for PendingPaymentDetailsUpdate {
...
+
+	pub(crate) async fn remove(&self, id: &PaymentId) -> Result<(), Error> {
+		self.remove_batch(std::slice::from_ref(id)).await
+	}
+
+	pub(crate) async fn remove_batch(&self, ids: &[PaymentId]) -> Result<(), Error> {
+		let _guard = self.mutation_lock.lock().await;
+		let (removed_payments, result) = self.inner.remove_batch_with_partial_result(ids).await;
+		for payment in removed_payments {
+			self.remove_from_index(&payment);
```
tankyleo: nit: we can move the for loop to after the lock inside `remove_from_index`, and perhaps rename `remove_from_index` with something along the lines of `remove_batch_from_index`

src/data_store.rs
```diff
@@ -112,11 +112,29 @@ where
 	}
 
 	pub(crate) async fn remove(&self, id: &SO::Id) -> Result<(), Error> {
+		self.remove_batch(std::slice::from_ref(id)).await?;
+		Ok(())
+	}
+
+	pub(crate) async fn remove_batch(&self, ids: &[SO::Id]) -> Result<Vec<SO>, Error> {
```
tankyleo: `remove_batch` is really only used in tests at the moment, and `fn remove` above, should we delete it and have just `fn remove` and `fn remove_batch_with_partial_result` ?

src/data_store.rs
```diff
@@ -134,10 +152,16 @@ where
...
 				);
 				Error::PersistenceFailed
-			})?;
-			self.objects.lock().expect("lock").remove(id);
+			});
+			if let Err(e) = remove_result {
+				return (removed_objects, Err(e));
+			}
+
+			if let Some(object) = self.objects.lock().expect("lock").remove(id) {
```
tankyleo: If we are here can we expect this to be `Some`, so is that worth a `debug_assert` ?

tankyleo (commented):
A few nits feel free to squash will take a better pass tomorrow

src/event.rs
```diff
@@ -846,11 +922,63 @@ where
 					}
 				}
 
-				if let Some(info) = payment_info {
+				let payment_info = if let Some(pending_payment) = pending_payment.as_ref() {
```
tankyleo: I have a question about the approach, take a look at this vibed commit on top of this branch https://github.com/tankyleo/ldk-node/commit/c8c6947c41271dac50b70c85d82fc8739b8d7d6c

I wonder if we could simplify things by only keeping a simple in-memory list of all the manual hashes we've already received, and double check for any duplicates against that list. If the node restarts, that list gets dropped; that's ok, the user is already fully responsible for not issuing another invoice with the same payment hash *after* the first one got paid, since we prune the duplicate hash store on `PaymentClaimed`.

Said another way, I wonder if the additional complexity is worth it given that we don't protect users if they issue an invoice with the same payment hash after the first one got paid.

Then we could delete the separate cases here for the manual hash invoices, and manual hash invoices could also use the LDK payment ID like regular invoices do.

src/data_store.rs
```diff
@@ -112,11 +112,29 @@ where
...
+		Ok(removed_objects)
+	}
+
+	pub(crate) async fn remove_batch_with_partial_result(
+		&self, ids: &[SO::Id],
+	) -> (Vec<SO>, Result<(), Error>) {
 		let _guard = self.mutation_lock.lock().await;
-		let should_remove = { self.objects.lock().expect("lock").contains_key(id) };
-		if should_remove {
+		let mut removed_objects = Vec::new();
```
tankyleo: let's do a `with_capacity(ids.len())` here

src/data_store.rs
```diff
@@ -422,6 +446,45 @@ mod tests {
 		assert!(data_store.get(&new_id).is_none());
 	}
 
+	#[tokio::test]
+	async fn batch_remove_removes_persisted_objects() {
```
tankyleo: Can codex add coverage for the case where an `Err` is returned, with some successful removes ?

src/event.rs
```diff
@@ -1015,16 +1237,26 @@ where
...
-				payment_id: _,
 			} => {
-				let payment_id = PaymentId(payment_hash.0);
+				let event_payment_id = payment_id.unwrap_or(PaymentId(payment_hash.0));
+				let (mut payment_id, payment_info) =
+					self.resolve_inbound_payment_id(event_payment_id, &payment_hash);
+				let pending_payment = if payment_info.is_none() {
+					self.find_pending_inbound_payment(&payment_hash).await?
+				} else {
+					None
```
tankyleo: Is that a bug if we land here ?

src/event.rs
```diff
@@ -1079,11 +1351,23 @@ where
 						// be the result of a replayed event.
 					),
 					Ok(DataStoreUpdateResult::NotFound) => {
-						log_error!(
-							self.logger,
-							"Claimed payment with ID {} couldn't be found in store",
+						let payment = PaymentDetails::new(
```
tankyleo, line 1386: Similar here is that a bug if we land here ?

src/data_store.rs
tnull: It's called in `prune_expired_pendign_payments` though?

src/payment/pending_payment_store.rs
tnull: Done.

src/data_store.rs
tnull: Done.

src/data_store.rs
tnull: Done.

src/event.rs
tnull: Yeah, the issue with the sketched approach is that it would result in us having to emit `PaymentClaimable` whenever we receive a payment for an unknown preimage. I so far considered and avoided that because to a degree it introduces a DoS risk as it allows senders to fill up our event queue / payment store and pending HTLCs slots if receivers aren't handling that event (fast enough). 

That said, I'm coming around to re-exploring it, as indeed it would considerably simplify things in this PR. FWIW, I'm not even sure we need the 'recent hashes' cache (esp. if it would be in-memory only).

tnull: Now switched to this approach, requiring a new `Config::manually_claim_unknown_bolt11_payments` flag. Def. less invasive. Let me know what you think.

src/event.rs
tnull: Outdated.

src/data_store.rs
tnull: Outdated.

src/event.rs
tnull: Outdated.

CHANGELOG.md
```diff
@@ -7,6 +7,16 @@
...
+- The `payment_id` field on the `PaymentSuccessful`, `PaymentFailed`, and
+  `PaymentReceived` events is now a required (non-optional) `PaymentId`. Events
+  persisted by LDK Node v0.2.1 or earlier (which stored `payment_id` as
+  optional) will fail to deserialize on read; users upgrading from those
+  versions need to drain pending events before the upgrade.
+- Applications using the BOLT 11 manual-claiming receive APIs
+  (`receive_*_for_hash`) now need to set
+  `Config::manually_claim_unknown_bolt11_payments` to `true`. Otherwise
+  matching inbound HTLCs will be failed back instead of emitting
+  `Event::PaymentClaimable`.
```
tankyleo, line 19: We should clarify this is only true for newly created `for_hash` invoices. `for_hash` invoices created prior to this commit still need to be manually claimed or failed back, regardless of `manually_claim_unknown_bolt11_payments`

src/payment/bolt11.rs
```diff
@@ -486,17 +413,33 @@ impl Bolt11Payment {
 	///
 	/// When claiming the payment has succeeded, a [`PaymentReceived`] event will be emitted.
 	///
-	/// [`receive_for_hash`]: Self::receive_for_hash
-	/// [`receive_variable_amount_for_hash`]: Self::receive_variable_amount_for_hash
 	/// [`PaymentClaimable`]: crate::Event::PaymentClaimable
 	/// [`PaymentReceived`]: crate::Event::PaymentReceived
-	pub fn claim_for_hash(
-		&self, payment_hash: PaymentHash, claimable_amount_msat: u64, preimage: PaymentPreimage,
+	pub fn claim_for_id(
```
tankyleo, line 423: Can we add an upgrade test to make sure that this method will correctly claim a payment that was created using a `for_hash` variant on a previous release? Similar for the `fail_for_id` too.

src/config.rs
```diff
@@ -219,6 +220,17 @@ pub struct Config {
...
 	/// [BIP 353]: https://github.com/bitcoin/bips/blob/master/bip-0353.mediawiki
 	pub hrn_config: HumanReadableNamesConfig,
+	/// Whether to emit [`Event::PaymentClaimable`] for unknown BOLT 11 payments that were created
+	/// with a user-provided payment hash and therefore need to be manually claimed.
+	///
+	/// If disabled, such payments are failed back without being added to the payment store.
+	///
+	/// **Warning:** Enabling this may let any holder of a valid invoice generated by this node tie
+	/// up inbound HTLC slots and grow the persisted event queue until the payment is claimed,
+	/// failed, or times out.
```
tankyleo, line 230: I think it's worth highlighting that this could happen *after* the payment got claimed. In that case, the payment should be failed back instead of claimed twice.

Perhaps even suggest that the user should keep state tracking all pending payment hashes, and fail any payment hashes not in the pending list ?

src/config.rs
```diff
@@ -219,6 +220,17 @@ pub struct Config {
...
+	///
+	/// [`Event::PaymentClaimable`]: crate::Event::PaymentClaimable
+	pub manually_claim_unknown_bolt11_payments: bool,
```
tankyleo: nit: `manually_handle_unknown_bolt11_payments` would better convey that the user must fail back unknown bolt11 payments in addition to claiming them.

src/event.rs
```diff
@@ -200,17 +194,23 @@ pub enum Event {
...
+	/// retrieved in time, should be failed-back via [`fail_for_id`].
 	///
 	/// Note claiming will necessarily fail after the `claim_deadline` has been reached.
 	///
-	/// [`claim_for_hash`]: crate::payment::Bolt11Payment::claim_for_hash
-	/// [`fail_for_hash`]: crate::payment::Bolt11Payment::fail_for_hash
+	/// This event is only emitted for unknown BOLT 11 payments if
+	/// [`Config::manually_claim_unknown_bolt11_payments`] is enabled. Enabling this may let any
+	/// holder of a valid invoice generated by this node tie up inbound HTLC slots and grow the
+	/// persisted event queue until the payment is claimed, failed, or times out.
```
tankyleo: I'd add some language similar to this after this section:
"This event could be emitted multiple times for a single hash, including well after the invoice got claimed, in case this invoice gets paid multiple times. Users should claim the first, and fail the rest."

src/event.rs
tankyleo, line 1038: Also noted in the `CHANGELOG.md`, we could still reach this point with `manually_claim_unknown_bolt11_payments = false` if the user used the for_hash variants prior to upgrading past this commit.

src/event.rs
```diff
@@ -1145,11 +1364,23 @@ where
 						// be the result of a replayed event.
 					),
 					Ok(DataStoreUpdateResult::NotFound) => {
```
tankyleo, line 1385: Is it a bug if we land here ? Or are we guarding against a possible crash before the persistence completes on `PaymentClaimable` ?

src/payment/bolt11.rs
joostjager: @tnull any new insights here?

tnull: > @tnull any new insights here?

Not really. We could consider tracking paid-to invoices as part of the upcoming payment store, or, begrudgingly, restore the `PaymentId`/`PaymentHash` coupling for the special case of outbound-BOLT11 payments. While I'd still love to eventually drop that all `PaymentId` semantics, doing the latter is probably least painful for now. Will add a fixup.

src/payment/bolt11.rs
tnull: Done.

src/config.rs
tnull: Done.

src/event.rs
tnull: Note we also expose `Node::remove_payment` so we could always hit a case where the user for one reason or another prunes payment entries from their payment store.

CHANGELOG.md
tnull: Adjusted docs in fixup.

src/config.rs
tnull: Adjusted docs in fixup.

src/event.rs
tnull: Adjusted docs in fixup.

src/event.rs
```diff
@@ -675,6 +678,25 @@ where
...
+		&self, event_payment_id: Option<PaymentId>, payment_hash: &PaymentHash,
+	) -> (PaymentId, Option<PaymentDetails>) {
+		let legacy_id = PaymentId(payment_hash.0);
+		let payment_id = event_payment_id.unwrap_or(legacy_id);
+
+		if let Some(info) = self.payment_store.get(&payment_id) {
+			return (payment_id, Some(info));
+		}
+
+		if legacy_id != payment_id {
```
tankyleo, line 704: Codex:

>   - [P2] Migrate current refund records to the event payment ID _ /home/ubuntu/ldk-node/src/event.rs:691-694
    When request_refund_payment is used, it still inserts the inbound refund at PaymentId(payment_hash.0) (src/payment/bolt12.rs:444-464). This fallback therefore treats a current record as legacy and returns the hash-derived ID, so refund PaymentReceived events and store entries never use LDK's
    unique inbound ID; repeated refund payments also collapse onto the same record. Restrict the fallback to genuinely legacy rows or re-key this record when the event arrives.

Should we drop inserting the payment into the store on `request_refund_for_payment` ?

Cargo.toml
```diff
@@ -99,6 +99,7 @@ regex = "1.5.6"
 criterion = { version = "0.7.0", features = ["async_tokio"] }
 ldk-node-062 = { package = "ldk-node", version = "=0.6.2" }
 ldk-node-070 = { package = "ldk-node", version = "=0.7.0" }
+lightning-types-0-3 = { package = "lightning-types", version = "0.3.2" }
```
tankyleo: Codex:

>   - [P2] Reuse the v0.7 re-export instead of adding a dependency _ /home/ubuntu/ldk-node/Cargo.toml:102-102
    This dependency is unnecessary because ldk-node v0.7 publicly re-exports its exact lightning_types crate, allowing the test to import ldk_node_070::lightning_types::payment::PaymentHash. Remove the direct dependency and use that re-export, as the repository explicitly prohibits unrequested new
    dependencies (AGENTS.md:9).

CHANGELOG.md
tankyleo: Is a note in the `CHANGELOG.md` not needed ? Let me know could do with just the docs on `PaymentClaimable` yes.

src/config.rs
tankyleo: Don't yet see a note about this here do you think it's not worth it ?

src/event.rs
tankyleo: Likewise here, do you find it's not necessary ?

CHANGELOG.md
tnull: Hmm, given that the behavior doesn't actually change for legacy users (they'll still get the event and still have to handle the event and use the APIs) I'm not sure it's worth mentioning in changelog more explicitly?

src/config.rs
tnull: > Don't yet see a note about this here do you think it's not worth it ?

Hmm, we generally have the guidance that users that enable this need to handle `PaymentClaimable` and fail anything they're not going to claim? Or maybe I'm confused on what specifically you want mentioned?

src/event.rs
tnull: See above: users should generally fail anything they're not going to claim. Whether or not they claim doubly-sent payments is a bit up to them, though it's of course nicer not to do so.

Note that if we give this guidance we probably also want to implement this for LDK Server, which would mean yet-another persisted store, meaning we should then probably still do it in LDK Node if we deem it that important. But we just switched away from making this PR overly complex with all the secondary indexes etc. :grimacing: 

Any case, I now added a note.

src/event.rs
tnull: Hmm, good point. Issue is that we can't easily drop the insertion in `request_refund_for_payment` as otherwise we lose `payer_note` and `quantity` for inbound payments (the case where at least the former matters). Let's discuss offline whether we could carry them in `Bolt12RefundContext`.

Cargo.toml
tnull: Will fix.

CHANGELOG.md
tankyleo: sounds good

src/config.rs
tankyleo: See thread below

src/event.rs
tankyleo: Hmm ok to confirm the plan in ldk-server would be to claim all the hashes we now the preimage for, regardless of duplicates ?

My intuition there says the responsibility is on the sender of the payment to not pay an invoice twice, so seems good to me.

tankyleo: > Note that if we give this guidance we probably also want to implement this for LDK Server, which would mean yet-another persisted store, meaning we should then probably still do it in LDK Node if we deem it that important. But we just switched away from making this PR overly complex with all the secondary indexes etc. 😬

Perhaps then we should drop the "should be failed back" note, and just clarify things like this:

"If the same payment hash is paid more than once, claiming each event accepts each payment"

tankyleo:
Thanks feel free to squash the existing fixups

src/event.rs
tnull: > Hmm ok to confirm the plan in ldk-server would be to claim all the hashes we now the preimage for, regardless of duplicates ?

No, the plan would be to either not expose `PaymentClaimable` at all, or, if we do, still have it be the user's responsibility to handle the events.

> My intuition there says the responsibility is on the sender of the payment to not pay an invoice twice, so seems good to me.

Yeah, it is, but it's also not great to knowingly steal money, even if the other party offers it through a mistake. So preferably we would have default behavior that avoids it, though in this case I'm not sure if it's worth the complication.

src/event.rs
tnull: Discussed offline. Given that barely anybody is currently using `Refunds` we concluded it's fine to 'pause' setting the `payer_note`/`quantity` fields for now, and will start doing so after https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/issues/4859 lands with LDK v0.4.

tankyleo (commented):
LGTM ! Please squash thank you sorry for the delay.

tnull:
> LGTM ! Please squash thank you sorry for the delay.

Squashed without further changes!

src/payment/bolt11.rs
tankyleo, line 1: Sounds like we might need to correct the commit message ?

> Generate outbound BOLT11 IDs from KeysManager entropy instead of deriving them from the payment hash.

We no longer do this as far as I see, we reverted back to using the payment hash.

tnull: Ah, updated.

tankyleo (approved):

PR #955, opened 2026-07-01
Expose pathfinding scoring fee config

benthecarman:
Applications need to tune route selection for different fee and reliability preferences while continuing to use ldk-node's default router. This makes the `ProbabilisticScoringFeeParameters` public so users can customize it to their needs.

### DIFF
--- a/bindings/ldk_node.udl
+++ b/bindings/ldk_node.udl
@@ -11,6 +11,8 @@ typedef dictionary ElectrumSyncConfig;
 
 typedef dictionary TorConfig;
 
+typedef dictionary ProbabilisticScoringFeeParameters;
+
 typedef interface NodeEntropy;
 
 typedef enum WordCount;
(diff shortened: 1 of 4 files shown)


tnull:
> Applications need to tune route selection for different fee and reliability preferences while continuing to use ldk-node's default router.

No, I disagree with this. LDK's scorer parameters are very obscure and super confusing to 99% of users. I don't think we should expose them as an API and leave it up to the users to understand and attempt to 'improve their routing' as they very likely to mess things up even further or at least don't actually improve things.

If we think the parameters could be improved, we should rather adjust the defaults (in LDK or LDK Node) on the basis of actual benchmarks, A:B testing, simulations, or other data-driven models.

benthecarman:
> If we think the parameters could be improved, we should rather adjust the defaults (in LDK or LDK Node) on the basis of actual benchmarks, A:B testing, simulations, or other data-driven models.

I agree but feel users should be able to do the same for themselves.

Issue #936, opened 2026-06-12
Follow-ups to #792


tnull:
- [x] Codex review:

>   - [P1] trust_peer_0conf is only honored for LSPS2-discovered LSPs: /home/tnull/worktrees/ldk-node/pr-792/src/event.rs:1323. add_liquidity_source(..., true) documents generic 0-conf trust for the LSP (/home/tnull/worktrees/ldk-node/pr-792/src/builder.rs:448, /home/tnull/worktrees/ldk-node/pr-
>     792/src/liquidity/mod.rs:117), but inbound acceptance only looks up protocol 2. LSPS1-only LSPs, or trusted LSPs before/after failed LSPS0 discovery, now fall through to accept_inbound_channel instead of accept_inbound_channel_from_trusted_peer. The trust lookup should be by node id
>     independent of protocol; keep the LSPS2 protocol check only for accept_underpaying_htlcs.
- [x] Codex nit:  
>  - [P3] Three new files have CRLF line endings, so git diff --check upstream/main...HEAD reports trailing whitespace on every line: /home/tnull/worktrees/ldk-node/pr-792/src/liquidity/client/mod.rs:1, /home/tnull/worktrees/ldk-node/pr-792/src/liquidity/service/mod.rs:1, /home/tnull/worktrees/
>     ldk-node/pr-792/src/liquidity/service/lsps2.rs:1. cargo fmt tolerates it, but diff whitespace checks will fail. Normalize those files to LF.
- [ ] The PR made LSP functionalities always enabled (dropped the `Ignoring` CMH default). We might want to revisit that decision in conjuction with #900, as we might want to add features to enable LSP {client,service} functionalities. It might prove to be hard to do this from an API-design perspective, we'll see.

tnull:
(Mostly) done in #956.

Issue #900, opened 2026-05-07
Introduce Rust features


tnull:
.. we punted on this until now, but it seems we should do it for 0.8.

tnull:
When we do this we should drop the 

```rust
#[cfg(not(feature = "postgres"))]
	pub fn build_with_postgres_store(...)
```
from the builder.

PR #956, opened 2026-07-01
 Follow-up to #792 (multi-LSP support)

Camillarhi:
Some minor follow-ups after #792 landed.

- Retry LSP protocol discovery when it fails at startup, so a transient connection failure doesn't leave a configured LSP permanently  unusable. Adds a background retry with backoff and a `Liquidity::retry_discovery(node_id)` method for on-demand re-discovery (also useful when an LSP rolls out a new protocol).

-  Honor `trust_peer_0conf` independent of the LSP's supported protocols, It was previously only applied to LSPS2 peers, so LSPS1-only or undiscovered LSPs silently lost their configured 0-conf trust.

Leaving the LSP feature-gating for #900

Fixes - #936

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -189,6 +189,7 @@ pub use types::{
 };
 pub use vss_client;
 
+use crate::config::{LIQUIDITY_DISCOVERY_RETRY_INITIAL_DELAY, LIQUIDITY_DISCOVERY_RETRY_MAX_DELAY};
 use crate::ffi::maybe_wrap;
 use crate::liquidity::Liquidity;
 use crate::scoring::setup_background_pathfinding_scores_sync;
@@ -706,33 +707,7 @@ impl Node {
 				let logger = Arc::clone(&discovery_logger);
 				let ls = Arc::clone(&liquidity_handler);
 				discovery_set.spawn(async move {
-					if let Err(e) = cm.connect_peer_if_necessary(node_id, address.clone()).await {
-						log_error!(
-							logger,
-							"Failed to connect to LSP {} for protocol discovery: {}",
-							node_id,
-							e
-						);
-						return;
-					}
-					match ls.discover_lsp_protocols(&node_id).await {
-						Ok(protocols) => {
-							log_info!(
-								logger,
-								"Discovered protocols for LSP {}: {:?}",
-								node_id,
-								protocols
-							);
-						},
-						Err(e) => {
-							log_error!(
-								logger,
-								"Failed to discover protocols for LSP {}: {:?}",
-								node_id,
-								e
-							);
-						},
-					}
+					connect_and_discover_lsp(&cm, &ls, &logger, node_id, address).await;
 				});
 			}
 
@@ -759,6 +734,39 @@ impl Node {
 						}
 					}
 				}
+			}
+		});
+
+		// Retry protocol discovery for any LSPs that failed the startup batch, backing off up to a
+		// cap and then retrying at that cap until every configured LSP has been discovered.
+		let mut stop_retry = self.stop_sender.subscribe();
+		let retry_ls = Arc::clone(&self.liquidity_source);
+		let retry_logger = Arc::clone(&self.logger);
+		let retry_cm = Arc::clone(&self.connection_manager);
+		self.runtime.spawn_cancellable_background_task(async move {
+			let mut backoff = LIQUIDITY_DISCOVERY_RETRY_INITIAL_DELAY;
+			loop {
+				tokio::select! {
+						_ = stop_retry.changed() => return,
+						_ = tokio::time::sleep(backoff) => {},
+				}
+
+				let undiscovered_lsps = retry_ls.get_undiscovered_lsps();
+				if undiscovered_lsps.is_empty() {
+					break;
+				}
+
+				let mut discovery_set = tokio::task::JoinSet::new();
+				for (node_id, address) in undiscovered_lsps {
+					let cm = Arc::clone(&retry_cm);
+					let ls = Arc::clone(&retry_ls);
+					let logger = Arc::clone(&retry_logger);
+					discovery_set.spawn(async move {
+						connect_and_discover_lsp(&cm, &ls, &logger, node_id, address).await;
+					});
+				}
+				discovery_set.join_all().await;
+				backoff = (backoff * 2).min(LIQUIDITY_DISCOVERY_RETRY_MAX_DELAY);
 			}
 		});
 
@@ -2433,6 +2441,23 @@ pub(crate) fn new_channel_anchor_reserve_sats(
 			c.per_channel_reserve_sats
 		}
 	})
+}
+
+async fn connect_and_discover_lsp(
+	connection_manager: &ConnectionManager<Arc<Logger>>,
+	liquidity_source: &LiquiditySource<Arc<Logger>>, logger: &Logger, node_id: PublicKey,
+	address: SocketAddress,
+) {
+	if let Err(e) = connection_manager.connect_peer_if_necessary(node_id, address).await {
+		log_debug!(logger, "Failed to connect to LSP {} for protocol discovery: {}", node_id, e);
+		return;
+	}
+	match liquidity_source.discover_lsp_protocols(&node_id).await {
+		Ok(protocols) => {
+			log_info!(logger, "Discovered protocols for LSP {}: {:?}", node_id, protocols)
+		},
+		Err(e) => log_debug!(logger, "Protocol discovery failed for LSP {}: {:?}", node_id, e),
+	}
 }
 
 #[cfg(test)]
--- a/src/liquidity/mod.rs
+++ b/src/liquidity/mod.rs
@@ -523,6 +523,15 @@ where
 			.collect()
 	}
 
+	pub(crate) fn get_lsp_trust_0conf(&self, node_id: &PublicKey) -> Option<bool> {
+		self.lsp_nodes
+			.read()
+			.expect("lock")
+			.iter()
+			.find(|n| &n.node_id == node_id)
+			.map(|n| n.trust_peer_0conf)
+	}
+
 	/// Flips the `discovery_done` watch to `true`.
 	///
 	/// Called once after the *initial* batch of LSPs configured at build time has been
--- a/src/liquidity/mod.rs
+++ b/src/liquidity/mod.rs
@@ -345,7 +345,7 @@ where
 	lsps1_client: Arc<LSPS1Client<L>>,
 	lsps2_client: Arc<LSPS2Client<L>>,
 	lsps2_service: Arc<LSPS2ServiceLiquiditySource<L>>,
-	pending_lsps0_discovery: Mutex<HashMap<PublicKey, oneshot::Sender<Vec<u16>>>>,
+	pending_lsps0_discovery: Mutex<HashMap<PublicKey, Vec<oneshot::Sender<Vec<u16>>>>>,
 	discovery_done_tx: tokio::sync::watch::Sender<bool>,
 	discovery_done_rx: tokio::sync::watch::Receiver<bool>,
 	liquidity_manager: Arc<LiquidityManager>,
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -748,6 +751,62 @@ impl Node {
 			}
 		});
 
+		// Background retry for LSPs that failed the initial protocol-discovery batch, so a
+		// transient startup failure doesn't leave a configured LSP permanently unusable.
+		let mut stop_retry = self.stop_sender.subscribe();
+		let retry_ls = Arc::clone(&self.liquidity_source);
+		let retry_logger = Arc::clone(&self.logger);
+		let retry_cm = Arc::clone(&self.connection_manager);
+		self.runtime.spawn_background_task(async move {
```
tnull: That likely should be a cancellable task?

src/liquidity/mod.rs
```diff
@@ -164,6 +164,40 @@ impl Liquidity {
...
 
+	/// Re-runs bLIP-50 / LSPS0 protocol discovery for an already-configured LSP.
+	///
+	/// Use this to recover an LSP that failed protocol discovery at startup or to
+	/// pick up newly supported protocols after the LSP has rolled out an upgrade.
+	///
+	/// The `node_id` must belong to an LSP configured at build time or added via
+	/// [`Liquidity::add_liquidity_source`]; otherwise [`Error::LiquiditySourceUnavailable`]
+	/// is returned.
+	pub fn retry_discovery(&self, node_id: PublicKey) -> Result<(), Error> {
```
tnull: Not a fan of exposing this publicly. Retrying should not be the concern of a user?

src/liquidity/mod.rs
```diff
@@ -432,6 +466,17 @@ where
 			.collect()
 	}
 
+	pub(crate) fn get_single_lsp_details(
```
tnull: Please avoid such ~single-call helpers (especially if they are only used in one place). Please rather inline them so they don't clutter up the file as much.

src/lib.rs
```diff
@@ -748,6 +751,62 @@ impl Node {
...
+			let mut backoff = LIQUIDITY_DISCOVERY_RETRY_INITIAL_DELAY_SECS;
+			loop {
+				let undiscovered_lsps = retry_ls.get_undiscovered_lsps();
+				if undiscovered_lsps.is_empty() {
+					tokio::select! {
```
tnull: Hmm? If there's nothing to do, why do need this select?

src/lib.rs
```diff
@@ -748,6 +751,62 @@ impl Node {
...
+						_ = stop_retry.changed() => return,
+						_ = tokio::time::sleep(Duration::from_secs(
+							LIQUIDITY_DISCOVERY_RETRY_MAX_DELAY_SECS,
+						)) => continue,
+					}
+				}
+
+				tokio::select! {
```
tnull: It seems an `tokio::time::interval` would be more suitable?

src/lib.rs
```diff
@@ -748,6 +751,62 @@ impl Node {
...
+					_ = stop_retry.changed() => return,
+					_ = tokio::time::sleep(Duration::from_secs(backoff)) => {},
+				}
+
+				for (node_id, address) in undiscovered_lsps {
+					if let Err(e) =
+						retry_cm.connect_peer_if_necessary(node_id, address.clone()).await
```
tnull: Hmm, this seems redundant to our general reconnection loop, or not?

src/liquidity/client/mod.rs
```diff
@@ -1,11 +1,11 @@
...
-// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
-// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
-// http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
-// accordance with one or both of these licenses.
-
-pub(crate) mod lsps1;
-pub(crate) mod lsps2;
-
-pub use lsps1::LSPS1OrderStatus;
+// This file is Copyright its original authors, visible in version control history.
```
tnull, line 1: This commit diff seems to have a lot of unnecessary whitespace/line ending changes. Please amend to avoid that, will re-review then.

tnull (commented):
Excuse the delay here. Second commit seems broken, please avoid the unrelated changes as otherwise it's not really reviewable. Also needs a minor rebase by now.

src/liquidity/mod.rs
Camillarhi: It was meant to let a user trigger re-discovery for an LSP they know has rolled out a new protocol, the background retry only covers LSPs that never completed discovery in the first place, so an already-discovered LSP adding a protocol wouldn't get picked up.
But thinking about it more, the user usually won't know when an LSP adds a protocol, so leaning on them to call this isn't great. Probably makes more sense to have the background task periodically re-run discovery for already-discovered LSPs too, so new protocols get picked up automatically and there's nothing to expose. I will extend the background check

src/lib.rs
Camillarhi: It is there to stop the loop from busy-spinning when there's nothing to discover. The idea was to keep the task alive in case an LSP became undiscovered later, but a runtime `add_liquidity_source` cleans up the node if its discovery fails, so nothing new ever lands in the undiscovered set after startup. So once it's empty, there's nothing to wait for, and it should just return. I'll clean it up as part of the re-discovery rework, for re-discovering already-discovered LSPs.

src/lib.rs
Camillarhi: Configured LSPs aren't added to the peer store at startup, so the reconnection loop doesn't keep them connected. This is what connects us to run discovery.

src/liquidity/client/mod.rs
Camillarhi: This is the CRLF→LF normalization, which was already flagged as a nit in the earlier review. I'll amend it to pull it out of this commit and into a separate one.

tnull: > This is the CRLF→LF normalization, which was already flagged as a nit in the earlier review. I'll amend it to pull it out of this commit and into a separate one.

I think it's been fixed by #965, so shouldn't be necessary if you rebase.

Camillarhi: Thanks, I’ll rebase

src/lib.rs
```diff
@@ -762,6 +740,65 @@ impl Node {
...
+				let mut discovery_set = tokio::task::JoinSet::new();
+				for (node_id, address) in rediscovery_ls.get_all_lsp_details() {
+					let cm = Arc::clone(&rediscovery_cm);
+					let ls = Arc::clone(&rediscovery_ls);
+					let logger = Arc::clone(&rediscovery_logger);
+					discovery_set.spawn(async move {
+						connect_and_discover_lsp(&cm, &ls, &logger, node_id, address).await;
+					});
+				}
+				discovery_set.join_all().await;
```
tnull, line 768: Codex:  

>   - [P2] Make rediscovery cancellation-safe — /home/tnull/worktrees/ldk-node/pr-956/src/lib.rs:798
> 
>     Node::stop() aborts this task while join_all() may be awaiting an LSPS0 request. Dropping discover_lsp_protocols bypasses its timeout cleanup, leaving the node ID permanently present in pending_lsps0_discovery. After restarting the same Node, every discovery attempt reports “already in
>     flight,” preventing protocol refresh unless the old response eventually arrives. Use drop-based cleanup for pending requests or allow in-flight discovery tasks to finish during shutdown.
>

src/lib.rs
```diff
@@ -762,6 +740,65 @@ impl Node {
...
+				backoff *= 2;
+			}
+
+			// periodically re-discover all configured LSPs to pick up newly
```
tnull: Are we positive we need this? I think when an LSP updates it's fine to only discover that on next reconnection. And wouldn't peers that never finish discovery also be retried above?

tnull (commented):
Two minor comments.

src/lib.rs
Camillarhi: > Are we positive we need this? I think when an LSP updates it's fine to only discover that on next reconnection.

Unused LSPs aren't persisted to the peer store, so the reconnection loop won't cover them, and discovery doesn't currently run on reconnection, but I could watch for when an LSP disconnects and re-run discovery once it reconnects. 

> And wouldn't peers that never finish discovery also be retried above?

The retry loop above this only covers never-discovered peers until the backoff caps out, after that the sweep is what keeps retrying them. If I swap the sweep for the rediscover-on-reconnect approach, I'd let the retry loop keep going at the cap instead, so those stay covered.

src/lib.rs
Camillarhi: Fixed with drop-based cleanup, a guard now removes the pending entry when the discovery future is dropped, so an aborted task can't leave a stale "in flight" entry behind.

src/lib.rs
```diff
@@ -762,6 +740,65 @@ impl Node {
...
+			// Fast retries for LSPs that failed the startup discovery batch, backing off
+			// until we reach the periodic re-discovery cadence.
+			let mut backoff = LIQUIDITY_DISCOVERY_RETRY_INITIAL_DELAY;
+			while backoff < LIQUIDITY_DISCOVERY_RETRY_MAX_DELAY {
+				tokio::select! {
+						_ = stop_rediscovery.changed() => return,
+						_ = tokio::time::sleep(backoff) => {},
+				}
+
+				let undiscovered_lsps = rediscovery_ls.get_undiscovered_lsps();
```
tnull: Codex:

>   - [P2] Background discovery can make add_liquidity_source fail spuriously. The /home/tnull/worktrees/ldk-node/pr-956-latest/src/lib.rs:760 can select a runtime-added LSP after it is inserted with supported_protocols: None but before /home/tnull/worktrees/ldk-node/pr-956-latest/src/liquidity/
>     mod.rs:156. If the background task claims the pending-discovery slot first, the public call receives LiquidityRequestFailed and removes the otherwise valid LSP. Same-node discovery should be coalesced or initializing entries excluded from background sweeps.

src/liquidity/mod.rs
```diff
@@ -352,6 +352,21 @@ where
...
+// Removes the pending LSPS0 discovery entry for a node when dropped, ensuring cleanup happens
+// even if the discovery future is cancelled mid-await (e.g., when `Node::stop` aborts the
+// background task). Otherwise a stale entry would reject any subsequent discovery attempt for
+// this node as already being in flight.
+struct PendingDiscoveryGuard<'a> {
+	pending: &'a Mutex<HashMap<PublicKey, oneshot::Sender<Vec<u16>>>>,
+	node_id: PublicKey,
+}
+
+impl Drop for PendingDiscoveryGuard<'_> {
```
tnull: Codex
>  - [P2] An old discovery guard can delete a newer request. The response handler removes the pending sender before waking its receiver, while /home/tnull/worktrees/ldk-node/pr-956-latest/src/liquidity/mod.rs:364 later removes whichever entry currently exists for that node. A second discovery
>     can insert during that gap, only to have the first guard delete its sender. The guard should be disarmed after receiving its response or verify request identity before removal.
>

src/lib.rs
tnull: I'm confused: we now have two APIs, right? Users can either add LSPs at build time, or at runtime. Both are not persisted, but the build-time one is likely 'persisted' in the user code. Why do we need to 'rediscover' these configured LSPs during the runtime - do we really expect our nodes to never restart and we hence require live protocol upgrades? I'm not sure we need to / should.

tnull (commented):
Some comments that Codex had, mostly.

The cancellation safety part is now also addressed in #997.

src/lib.rs
Camillarhi: This is correct, a restart will cover this. I'll drop the rediscovery for already-discovered LSPs and keep only the retry for the ones where the initial discovery failed, running at the capped backoff until they come up, and stopping once there's nothing left undiscovered

Camillarhi:
> Some comments that Codex had, mostly.
> 
> The cancellation safety part is now also addressed in #997.

Thanks, I reverted the cancellation-safety guard since #997 covers it. If #997 lands first, I'll rebase on top of it.

tnull (approved):

Issue #269, opened 2024-03-05
Add a paginated version of `list_payments`


tnull:
We should eventually add a paginated version of `list_payments` to avoid overloading the caller in the face of many payments. This will be epecially crucial for a LDK Node 'server' deployment which we intend to add more support for generally in the future.

VanshulB:
Hey, I would like to work on this issue, is this up for grabs?

johncantrell97:
> > Hey, I would like to work on this issue, is this up for grabs?
> 
> Sure, however, I think we should do #57 first. I.e., base the work on #270 and add a `timestamp`/`last_updated` field by which we can sort `list_payments`. Then, we can add a paginated version in a much more reasonable manner. What do you think?

Would be nice if the KVStore key used to store the payments included the timestamp to avoid having to read _all_ payments in order to sort and page them.  If there was timestamp in key then can list the namespace, sort + page keys, and then only read payments needed for that page.

johncantrell97:
Do you think the PaymentStore should be changed to not keep all payments in memory? 

I was thinking if you were going to support pagination you'd change the store to just read from the store anytime it needed a particular payment and then the paginated list payments would only need to load the keys into memory and then only read payments for the specific page being requested.

With that model, in order to support pagination by time/recent payments you'd also need a separate index in the kvstore that stores a mapping with timestamp as key to the payment id.

johncantrell97:
> > Do you think the PaymentStore should be changed to not keep all payments in memory?
> 
> Especially for the use cases we're mainly focussing right now (self-custodial wallets, (mobile) apps, light server use) the sent/received payments are not a lot of data, so for the time being it's probably fine to keep in memory. We will likely add optional invoice/offer storage however (cf. #51), as invoice/offer data might be a bit heavier (even though still not _that much_ data). If/when we move the focus to heavier server use we might want to consider not keeping everything in memory, but we'll see when we get there (i.e., we have a user that really requires it).
> 

Makes sense, does seem like the data is pretty small even for thousands of payments.  Can always do it later as you say if needed.

> > I was thinking if you were going to support pagination you'd change the store to just read from the store anytime it needed a particular payment and then the paginated list payments would only need to load the keys into memory and then only read payments for the specific page being requested.
> > With that model, in order to support pagination by time/recent payments you'd also need a separate index in the kvstore that stores a mapping with timestamp as key to the payment id.
> 
[...]

PR #959, opened 2026-07-02
Add paginated payment listing API

benthecarman:
Add paginated payment listing API

Return payment history in reverse creation order one page at a time.
Keep ordering metadata in memory, support opaque tokens in language
bindings, and deprecate the unpaginated filtering API.

This commit was created with assistance from OpenAI Codex.

### DIFF
--- a/bindings/ldk_node.udl
+++ b/bindings/ldk_node.udl
@@ -151,7 +151,8 @@ interface Node {
 	[Throws=NodeError]
 	void remove_payment([ByRef]PaymentId payment_id);
 	BalanceDetails list_balances();
-	sequence<PaymentDetails> list_payments();
+	[Throws=NodeError]
+	PaymentDetailsPage list_payments(PageToken? page_token);
 	sequence<PeerDetails> list_peers();
 	sequence<ChannelDetails> list_channels();
 	NetworkGraph network_graph();
@@ -235,6 +236,7 @@ enum NodeError {
 	"InvalidDateTime",
 	"InvalidFeeRate",
 	"InvalidScriptPubKey",
+	"InvalidPageToken",
 	"DuplicatePayment",
 	"UnsupportedCurrency",
 	"InsufficientFunds",
@@ -276,6 +278,10 @@ enum PaymentFailureReason {
 };
 
 typedef dictionary PaymentDetails;
+
+typedef dictionary PaymentDetailsPage;
+
+typedef interface PageToken;
 
 [Remote]
 dictionary RouteParametersConfig {
--- a/src/data_store.rs
+++ b/src/data_store.rs
@@ -5,11 +5,11 @@
 // http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
 // accordance with one or both of these licenses.
 
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 use std::ops::Deref;
 use std::sync::{Arc, Mutex};
 
-use lightning::util::persist::KVStore;
+use lightning::util::persist::{KVStore, PageToken};
 use lightning::util::ser::{Readable, Writeable};
 
 use crate::logger::{log_error, LdkLogger};
@@ -25,7 +25,7 @@ pub(crate) trait StorableObject: Clone + Readable + Writeable {
 	fn to_update(&self) -> Self::Update;
 }
 
-pub(crate) trait StorableObjectId: std::hash::Hash + PartialEq + Eq {
+pub(crate) trait StorableObjectId: Clone + std::hash::Hash + Eq {
 	fn encode_to_hex_str(&self) -> String;
 }
 
@@ -40,11 +40,18 @@ pub(crate) enum DataStoreUpdateResult {
 	NotFound,
 }
 
+struct InMemoryObjects<SO: StorableObject> {
+	objects: HashMap<SO::Id, SO>,
+	creation_order: BTreeMap<u64, SO::Id>,
+	creation_order_by_key: HashMap<String, u64>,
+	next_creation_order: u64,
+}
+
 pub(crate) struct DataStore<SO: StorableObject, L: Deref>
 where
 	L::Target: LdkLogger,
 {
-	objects: Mutex<HashMap<SO::Id, SO>>,
+	objects: Mutex<InMemoryObjects<SO>>,
 	mutation_lock: tokio::sync::Mutex<()>,
 	primary_namespace: String,
 	secondary_namespace: String,
@@ -60,8 +67,23 @@ where
 		objects: Vec<SO>, primary_namespace: String, secondary_namespace: String,
 		kv_store: Arc<DynStore>, logger: L,
 	) -> Self {
-		let objects =
-			Mutex::new(HashMap::from_iter(objects.into_iter().map(|obj| (obj.id(), obj))));
+		let object_count = objects.len();
+		let mut creation_order = BTreeMap::new();
+		let mut creation_order_by_key = HashMap::with_capacity(object_count);
+		let mut objects_by_id = HashMap::with_capacity(objects.len());
+		for (creation_index, object) in objects.into_iter().rev().enumerate() {
+			let id = object.id();
+			let order = creation_index as u64;
+			creation_order.insert(order, id.clone());
+			creation_order_by_key.insert(id.encode_to_hex_str(), order);
+			objects_by_id.insert(id, object);
+		}
+		let objects = Mutex::new(InMemoryObjects {
+			objects: objects_by_id,
+			creation_order,
+			creation_order_by_key,
+			next_creation_order: object_count as u64,
+		});
 		Self {
 			objects,
 			mutation_lock: tokio::sync::Mutex::new(()),
@@ -77,7 +99,15 @@ where
 
 		self.persist(&object).await?;
 		let mut locked_objects = self.objects.lock().expect("lock");
-		let updated = locked_objects.insert(object.id(), object).is_some();
+		let id = object.id();
+		let updated = locked_objects.objects.insert(id.clone(), object).is_some();
+		if !updated {
+			let creation_order = locked_objects.next_creation_order;
+			locked_objects.next_creation_order =
+				creation_order.checked_add(1).expect("creation order overflow");
+			locked_objects.creation_order.insert(creation_order, id.clone());
+			locked_objects.creation_order_by_key.insert(id.encode_to_hex_str(), creation_order);
+		}
 		Ok(updated)
 	}
 
@@ -87,7 +117,7 @@ where
 		let id = object.id();
 		let data_to_persist = {
 			let locked_objects = self.objects.lock().expect("lock");
-			if let Some(existing_object) = locked_objects.get(&id) {
+			if let Some(existing_object) = locked_objects.objects.get(&id) {
 				let mut updated_object = existing_object.clone();
 				let updated = updated_object.update(object.to_update());
 				if updated {
@@ -104,7 +134,16 @@ where
 			Some(updated_object) => {
 				self.persist(&updated_object).await?;
 				let mut locked_objects = self.objects.lock().expect("lock");
-				locked_objects.insert(id, updated_object);
+				let is_new = locked_objects.objects.insert(id.clone(), updated_object).is_none();
+				if is_new {
+					let creation_order = locked_objects.next_creation_order;
+					locked_objects.next_creation_order =
+						creation_order.checked_add(1).expect("creation order overflow");
+					locked_objects.creation_order.insert(creation_order, id.clone());
+					locked_objects
+						.creation_order_by_key
+						.insert(id.encode_to_hex_str(), creation_order);
+				}
 				Ok(true)
 			},
 			None => Ok(false),
@@ -113,7 +152,7 @@ where
 
 	pub(crate) async fn remove(&self, id: &SO::Id) -> Result<(), Error> {
 		let _guard = self.mutation_lock.lock().await;
-		let should_remove = { self.objects.lock().expect("lock").contains_key(id) };
+		let should_remove = { self.objects.lock().expect("lock").objects.contains_key(id) };
 		if should_remove {
 			let store_key = id.encode_to_hex_str();
 			KVStore::remove(
@@ -135,7 +174,11 @@ where
 				);
 				Error::PersistenceFailed
 			})?;
-			self.objects.lock().expect("lock").remove(id);
+			let mut locked_objects = self.objects.lock().expect("lock");
+			locked_objects.objects.remove(id);
+			if let Some(creation_order) = locked_objects.creation_order_by_key.remove(&store_key) {
+				locked_objects.creation_order.remove(&creation_order);
+			}
 		}
 		Ok(())
 	}
@@ -146,7 +189,7 @@ where
 	/// Until store reads are async, callers may temporarily see in-memory state that has not yet
 	/// caught up to a write in progress.
 	pub(crate) fn get(&self, id: &SO::Id) -> Option<SO> {
-		self.objects.lock().expect("lock").get(id).cloned()
+		self.objects.lock().expect("lock").objects.get(id).cloned()
 	}
 
 	pub(crate) async fn update(&self, update: SO::Update) -> Result<DataStoreUpdateResult, Error> {
@@ -154,7 +197,7 @@ where
 		let id = update.id();
 		let updated_object = {
 			let locked_objects = self.objects.lock().expect("lock");
-			let Some(object) = locked_objects.get(&id) else {
+			let Some(object) = locked_objects.objects.get(&id) else {
 				return Ok(DataStoreUpdateResult::NotFound);
 			};
 			let mut updated_object = object.clone();
@@ -166,7 +209,7 @@ where
 
 		self.persist(&updated_object).await?;
 		let mut locked_objects = self.objects.lock().expect("lock");
-		locked_objects.insert(id, updated_object);
+		locked_objects.objects.insert(id, updated_object);
 		Ok(DataStoreUpdateResult::Updated)
 	}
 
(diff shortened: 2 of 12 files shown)


src/ffi/types.rs
```diff
@@ -893,6 +893,16 @@ uniffi::custom_type!(OfferId, String, {
 	},
 });
 
+uniffi::custom_type!(PageToken, String, {
```
tnull: Can we use the remote type rather than adding a custom type (as we try to get rid of the latter over time)?

src/lib.rs
```diff
@@ -2135,6 +2146,20 @@ impl Node {
 		self.payment_store.list_filter(|_| true)
 	}
 
+	/// Retrieves a page of payments from the underlying paginated store, ordered from most
+	/// recently created to least recently created.
+	///
+	/// Pass `None` to start listing from the most recently created payment. If the returned
+	/// [`PaymentDetailsPage::next_page_token`] is `Some`, pass it to a subsequent call to
+	/// retrieve the next page.
+	pub fn list_payments_paginated(
```
tnull: Hmm, do we really want an additional method? At some point (hopefully for v0.9) we'll stop keeping all entries in memory and at that point at the latest the current shape of `list_payments` isn't feasible anymore. So should this just replace `list_payments`? Or not yet?

src/data_store.rs
```diff
@@ -179,6 +180,104 @@ where
 		self.objects.lock().expect("lock").values().filter(f).cloned().collect::<Vec<SO>>()
 	}
 
+	/// Returns a page of objects read from the underlying store, ordered from most recently
+	/// created to least recently created, together with a token that can be passed to a
+	/// subsequent call to retrieve the next page.
+	pub(crate) async fn list_page(
+		&self, page_token: Option<PageToken>,
+	) -> Result<(Vec<SO>, Option<PageToken>), Error> {
+		let response = PaginatedKVStore::list_paginated(
```
tnull: Currently, we're still leaning on all `DataStore` entries being kept in memory everywhere else (which we should change soon as mentioned above). So right now it's a bit odd to have the paginated version being the only one calling through to `KVStore`, and not even using any cache (so it's likely pretty slow).

So, should this just use the in-memory entries? If not, we could consider already making the jump to *not* keep all entries in memory, but then we'd need some (presumably LRU?) caching logic for `DataStore` first, before we add pagination support/switch it to use pagination?

tnull (commented):
Few questions on the general approach / general direction going forward.

src/data_store.rs
```diff
@@ -179,6 +180,104 @@ where
...
+		})?;
+
+		// Read the page's objects through a sliding window: keep up to
+		// `DATA_STORE_READ_CONCURRENCY_LIMIT` reads in flight and await them in key order to
+		// preserve the page's newest-first ordering. The page size is implementation-defined,
+		// so we can't rely on pages being small.
+		let mut spawn_iter = response.keys.iter();
+		let mut read_handles = VecDeque::with_capacity(DATA_STORE_READ_CONCURRENCY_LIMIT);
+		for key in spawn_iter.by_ref().take(DATA_STORE_READ_CONCURRENCY_LIMIT) {
+			read_handles.push_back(tokio::spawn(KVStore::read(
```
joostjager: Drive-by question: are there plans for something like `list_paginated_values`, so SQL backends can fetch a page of serialized values in a single query instead of doing `list_paginated` plus one `read` per key?

tnull: Seems like a follow-up optimization that could eventually make sense (though, maybe even worth benchmarking if necessary for the concrete use cases). Mind opening an issue for it so we don't forget to revisit?

joostjager: https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/issues/4779

src/lib.rs
benthecarman: Yeah didn't originally because of the second point but changed here to now replace the normal `list_payments`

src/data_store.rs
benthecarman: Changed to just use in memory entries. Changing to not keep everything in memory would be a separate PR and maybe should wait for 0.9 as we have a lot of in flight storage changes right now.

benthecarman:
Changed to continue using the in memory entires in the `DataStore`. One note on this, codex suggested to make `list_payments_with_filter` deprecated because now this doesn't really work with pagination. To really support this in the future we'd need to keep everything in memory or make our KV Store api more SQL based.

tnull (commented):
Seems tests are failing and this also needs another rebase.

benthecarman:
> Seems tests are failing and this also needs another rebase.

fixed but seems like we're still getting random ci failures from out of storage

joostjager:
The discussion above mentions that we eventually need to stop keeping all payments in memory. Since #269 only tracks pagination, would it be worth opening a separate issue for the remaining scalability work?

src/data_store.rs
```diff
@@ -179,6 +179,47 @@ where
...
 
+	/// Returns a page of objects, ordered from most recently created to least recently created,
+	/// together with a token that can be passed to a subsequent call to retrieve the next page.
+	///
+	/// The underlying store is only queried for the page's key order, which the in-memory map
+	/// doesn't track; the objects themselves are served from memory.
+	pub(crate) async fn list_page(
+		&self, page_token: Option<PageToken>,
+	) -> Result<(Vec<SO>, Option<PageToken>), Error> {
+		let response = PaginatedKVStore::list_paginated(
```
joostjager: Mixing disk reads and memory reads seems risky. Is there locking that prevents discrepancies between the two?

If we go with in-memory values for now, it might be better to store the pagination information in memory too.

src/data_store.rs
```diff
@@ -179,6 +179,47 @@ where
...
+				"Listing object data under {}/{} failed due to: {}",
+				&self.primary_namespace,
+				&self.secondary_namespace,
+				e
+			);
+			Error::PersistenceFailed
+		})?;
+
+		let locked_objects = self.objects.lock().expect("lock");
+		let objects_by_store_key: HashMap<String, &SO> =
```
joostjager: It looks inefficient to build an all-values map for each page?

src/lib.rs
```diff
@@ -2199,15 +2200,44 @@ impl Node {
...
 		&self, f: F,
 	) -> Vec<PaymentDetails> {
 		self.payment_store.list_filter(f)
 	}
 
-	/// Retrieves all payments.
-	pub fn list_payments(&self) -> Vec<PaymentDetails> {
-		self.payment_store.list_filter(|_| true)
+	/// Retrieves a page of payments from the underlying paginated store, ordered from most
+	/// recently created to least recently created.
```
joostjager: >P2, backend migration loses payment chronology
The migration helper copies keys in arbitrary list_all_keys order. Destination stores assign new creation ordering from that write sequence, so payment order after migration can differ from the documented newest-payment-first order. The migration test hides this by sorting both results by payment ID.
Preserve ordering metadata or migrate entries in an order that reconstructs the original chronology. The test should compare pagination order without sorting.

joostjager:
> The discussion above mentions that we eventually need to stop keeping all payments in memory. Since #269 only tracks pagination, would it be worth opening a separate issue for the remaining scalability work?

Went ahead and created https://github.com/lightningdevkit/ldk-node/issues/998

src/data_store.rs
benthecarman: Added the lock here to prevent potential races

src/data_store.rs
benthecarman: Changed the in-memory map to use the encoded key instead so we can directly look up the entries

src/lib.rs
benthecarman: fixed

joostjager:
Looks like the force-push also rebased. That makes it harder to review. I'd suggest to do a separate push for the rebase next time, so reviewers can have a clean diff of the new PR changes.

src/data_store.rs
```diff
@@ -44,7 +44,7 @@ pub(crate) struct DataStore<SO: StorableObject, L: Deref>
 where
 	L::Target: LdkLogger,
 {
-	objects: Mutex<HashMap<SO::Id, SO>>,
+	objects: Mutex<HashMap<String, SO>>,
```
joostjager: Seems expensive to do heap allocated strings here?

src/data_store.rs
```diff
@@ -60,8 +60,9 @@ where
 		objects: Vec<SO>, primary_namespace: String, secondary_namespace: String,
 		kv_store: Arc<DynStore>, logger: L,
 	) -> Self {
-		let objects =
-			Mutex::new(HashMap::from_iter(objects.into_iter().map(|obj| (obj.id(), obj))));
+		let objects = Mutex::new(HashMap::from_iter(
```
joostjager: Looks like this refactor is independent and could go into a separate commit?

src/data_store.rs
joostjager: But since `DataStore` is currently an in-memory write-through store, it still feels odd for pagination ordering and cursor state to come from persistence while the values come from memory? The lock prevents mutations through this `DataStore` from racing, but each page is still assembled from two different sources and missing in-memory entries are silently skipped.

If we are keeping all entries in memory for now, could the pagination ordering information live in memory as well?

src/lib.rs
joostjager: It seems that the migration helper `migrate_kv_store_data_inner` is still using `list_all_keys` for which the contract says "may return the items in arbitrary order"?

src/data_store.rs
tnull: Yes, I agree that everything needs to live in memory for now, and for 0.9 we can explore moving to `KVStore`+in-memory cache architecture. Right now it's odd to mix the two, and if we keep everything in memory anyways we at least want to also benefit from the faster latency, not get the worst of both worlds (as we currently might).

tests/common/mod.rs
```diff
@@ -68,6 +70,48 @@ pub(crate) use in_memory_store::InMemoryStore;
 /// Shared timeout (in seconds) for waiting on LDK events and external node operations.
 pub(crate) const INTEROP_TIMEOUT_SECS: u64 = 60;
 
+pub(crate) trait NodePaymentExt {
```
tnull, line 73: Why do we need this extension trait exactly?

tnull (commented):
Also some comments from claude:

>   Issues and concerns
> 
>   1. Backend migration scrambles pagination order (open thread, confirmed real). I verified at the pinned LDK rev that migrate_kv_store_data_inner iterates MigratableKVStore::list_all_keys(), whose contract allows arbitrary order, and target stores assign creation order by write sequence (e.g.
>   SqliteStore's creation column, InMemoryStore's counter-on-first-write). After a store migration, the documented "most recently created first" order of list_payments is no longer true. The migrate_node_across_all_backends test masks this by sorting both sides by payment ID. This is
>   joostjager's still-unanswered 2026-07-23 comment; it likely needs an LDK-side fix (ordered list_all_keys or migration via list_paginated), but at minimum the ordering caveat should be documented on list_payments / in the changelog, or tracked in a follow-up issue.
>   2. mutation_lock is held across a full KV round trip per page. For remote backends (VSS), each page blocks all payment writes for one network call. A user paginating a large history in a loop serializes with concurrent payment persistence on the hot path (event handling persists payment
>   updates). The lock is needed for the consistency argument above, so I don't think it's wrong — but it's a throughput trade-off worth an explicit code comment, and a reason to prioritize #998.
>   3. FFI users can't persist a PageToken. [Remote] interface PageToken {} exposes no constructor and no accessor, so bindings users can only resume pagination within a single process lifetime — a mobile app can't store the cursor and continue after restart. Rust users get PageToken::new/as_str
>   via the re-export. If cross-session resumption matters for bindings consumers, a follow-up exposing the token's string form would fix it.
>   4. CHANGELOG: the entry doesn't mention the list_payments_with_filter deprecation and carries no (#959) reference (neighboring entries do reference PRs). Also, since the old list_payments() signature is gone entirely, consider stating explicitly that this is an API-breaking change for both
>   Rust and bindings.
>   5. Commit structure (already raised in review): the HashMap<SO::Id, SO> → HashMap<String, SO> re-keying is an independent refactor that would review more cleanly as its own commit, with the pagination feature on top.
> 
>   Minor / nits
> 
>   - StorableObjectId's Hash + PartialEq + Eq bounds appear to be dead now that the map is keyed by String — could be dropped in the refactor commit.
>   - insert computes store_key = object.id().encode_to_hex_str() and then persist re-encodes the same key inside encode_object — trivially redundant.
>   - Every get/contains_key now allocates a 64-char hex String per call (joostjager already flagged this). Negligible next to I/O, but it is on the payment-event hot path.
>   - Tests migrated to list_first_page_payments* implicitly assume payment counts stay under the smallest page size (50 for InMemoryStore/VSS). True today with wide margin; the helpers' naming makes the assumption visible, which is good. Tests that genuinely need everything
>   (payment_persistence_after_restart with 200 payments, migration, closing-tx checks) correctly use list_all_payments*.

tests/common/mod.rs
benthecarman: Guess it doesn't need to be a trait, could be functions that take in a `&Node`. A lot of the tests assume we are still fetching all the payments instead of just the first page so added this to have easy functions

src/data_store.rs
benthecarman: Okay put the pagination ordering in memory too

src/data_store.rs
benthecarman: Now that the creation ordering is in memory was able to revert this

src/data_store.rs
benthecarman: Now that the creation ordering is in memory was able to revert this

src/lib.rs
benthecarman: sorry yeah i had something originally but undid it, this would require deeper in ldk so just documented it in the changelog

joostjager:
@tnull is this indeed ready for re-review including addressing of the claude comments?

tnull:
> @tnull is this indeed ready for re-review including addressing of the claude comments?

You'd need to ask Ben for the details, but would be good to keep this moving, esp. given it also blocks #772.

src/lib.rs
joostjager: Do you want to open an issue for that?

src/data_store.rs
```diff
@@ -60,8 +66,19 @@ where
...
 		kv_store: Arc<DynStore>, logger: L,
 	) -> Self {
-		let objects =
-			Mutex::new(HashMap::from_iter(objects.into_iter().map(|obj| (obj.id(), obj))));
+		let next_creation_order = objects.len() as u64;
+		let mut creation_order = VecDeque::with_capacity(objects.len());
+		let mut objects_by_id = HashMap::with_capacity(objects.len());
+		for (index, object) in objects.into_iter().enumerate() {
+			let id = object.id();
+			creation_order.push_back((next_creation_order - index as u64, id.clone()));
```
joostjager: AI review (P2): These creation-order values are regenerated densely from the current object count on every startup, while the returned page token contains only the numeric order. This makes a token unstable across restarts if an unseen older payment is deleted. I reproduced this with 101 payments: after fetching the first page with token `52`, deleting the oldest unseen payment, and reloading the store, resuming with `52` returns the final payment from the first page again because its reconstructed order changed from `52` to `51`. The public API says the token can be passed to a subsequent call but does not limit it to the current process lifetime.

src/data_store.rs
```diff
@@ -176,7 +206,55 @@ where
...
+		&self, page_token: Option<PageToken>,
+	) -> Result<(Vec<SO>, Option<PageToken>), Error> {
+		const PAGE_SIZE: usize = 50;
+
+		let locked_objects = self.objects.lock().expect("lock");
+		let token_creation_order = page_token
+			.map(|token| {
+				token.as_str().parse::<u64>().map_err(|_| {
+					log_error!(self.logger, "Invalid object page token: {}", token);
+					Error::PersistenceFailed
```
joostjager: AI review (P3): A malformed page token is reported as `Error::PersistenceFailed` here. A numeric token greater than `next_creation_order` is reported as the same error below. Neither case involves a failed persistence operation, so callers receive a misleading public error. I reproduced both paths with focused tests.

joostjager (commented):
Just left two AI comments. Let me know when this PR is ready for final review.

src/data_store.rs
benthecarman: fixed

src/data_store.rs
benthecarman: fixed

benthecarman:
Should be good for review again

bindings/ldk_node.udl
```diff
@@ -277,6 +279,14 @@ enum PaymentFailureReason {
 
 typedef dictionary PaymentDetails;
 
+typedef dictionary PaymentDetailsPage;
+
+[Remote]
+interface PageToken {
+	constructor(string token);
+	string to_string();
```
joostjager: AI review (blocker): Exposing this remote method as `to_string` causes UniFFI to generate a Kotlin `toString` method without the required `override` modifier. The current `check-kotlin` job fails with `'toString' hides member of supertype 'Any' and needs 'override' modifier`, so the Kotlin bindings do not compile.

src/data_store.rs
```diff
@@ -176,7 +196,48 @@ where
...
+		const PAGE_SIZE: usize = 50;
+
+		let locked_objects = self.objects.lock().expect("lock");
+		let start_index = if let Some(token) = page_token {
+			locked_objects
+				.creation_order
+				.iter()
+				.position(|id| id.encode_to_hex_str() == token.as_str())
+				.map(|index| index + 1)
+				.ok_or_else(|| {
```
joostjager, line 240: AI review (P2): The continuation token contains the ID of the previous page's final payment, but `remove` also deletes that ID from `creation_order`. I reproduced this with 60 payments: after fetching page one and removing its final payment, requesting page two returns `InvalidPageToken`. This is a valid mutation between calls, and the upstream `PageToken` contract says pagination should resume from the next valid position when the referenced entry has been deleted.

src/data_store.rs
```diff
@@ -176,7 +196,48 @@ where
...
+		&self, page_token: Option<PageToken>,
+	) -> Result<(Vec<SO>, Option<PageToken>), Error> {
+		const PAGE_SIZE: usize = 50;
+
+		let locked_objects = self.objects.lock().expect("lock");
+		let start_index = if let Some(token) = page_token {
+			locked_objects
+				.creation_order
+				.iter()
+				.position(|id| id.encode_to_hex_str() == token.as_str())
```
joostjager: AI review (P2): Every continuation scans `creation_order` from the newest payment to the token boundary, and every comparison calls `encode_to_hex_str`, allocating a new string. A complete traversal is therefore quadratic: page two revisits 50 entries, page three revisits 100, and so on. Listing 100,000 payments in pages of 50 performs roughly 100 million comparisons and allocations while holding the store mutex.

joostjager (commented):
PR description needs update

bindings/ldk_node.udl
benthecarman: fixed

src/data_store.rs
benthecarman: Fixed. The token now includes the cursor’s creation order, so pagination can resume from that position after the cursor payment is removed

src/data_store.rs
benthecarman: Fixed. `creation_order` now uses a `BTreeMap` now

src/data_store.rs
```diff
@@ -176,7 +219,63 @@ where
...
+				.split_once(':')
+				.and_then(|(order, key)| order.parse::<u64>().ok().map(|order| (order, key)))
+				.ok_or_else(|| {
+					log_error!(self.logger, "Object page token not found: {}", token);
+					Error::InvalidPageToken
+				})?;
+			if let Some(current_order) = locked_objects.creation_order_by_key.get(key) {
+				// Reloading after an older removal shifts the current order down, while removing and
+				// reinserting the cursor shifts it up. In either case, resume from the older position.
+				encoded_order.min(*current_order)
```
joostjager, line 247: AI: This can repeat already-returned payments when the cursor and older payments are deleted before a restart. `DataStore::new` densely renumbers the surviving payments, but this branch interprets the token's old order in that new numbering. For example, after page one returns IDs 59 through 10, deleting IDs 0 through 10 and restarting renumbers IDs 11 through 59 to orders 0 through 48. Continuing from encoded order 10 then returns IDs 20 through 11 again. Please preserve stable ordering across reloads, or otherwise make missing-cursor continuation independent of dense renumbering, and add a regression test combining cursor deletion, deletion of older payments, and restart.

src/io/utils.rs
```diff
@@ -233,56 +234,67 @@ where
...
-	// Fill JoinSet with tasks if possible
-	while set.len() < BATCH_SIZE && !stored_keys.is_empty() {
-		if let Some(next_key) = stored_keys.pop() {
-			let fut = KVStore::read(kv_store, primary_namespace, secondary_namespace, &next_key);
-			set.spawn(fut);
-			debug_assert!(set.len() <= BATCH_SIZE);
+	let mut page_token = None;
+
+	loop {
+		let page = PaginatedKVStore::list_paginated(
```
joostjager, line 241: Perhaps this ordering fix can be an independent commit?

joostjager (commented):
I'd rather not have this 600-line change for a intermediate in-memory solution that also isn't all that simple. Wonder if it really isn't better to jump to a scalable on-disk solution right away.

joostjager:
Following up on this, isn't [rust-lightning #4807](https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/pulls/4807) the only thing needed to make this PR fully disk-based?

TheBlueMatt:
That should only be an optimization, in any case, anything you can do with it you can do by listing and reading separately, which is slower but for a local db shouldn't be insane.

joostjager:
I do think that slowness was the original reason for using the in-memory datastore though (https://github.com/lightningdevkit/ldk-node/pull/959#discussion_r3511391006).

tnull:
> I'd rather not have this 600-line change for a intermediate in-memory solution that also isn't all that simple. Wonder if it really isn't better to jump to a scalable on-disk solution right away.

Okay, attempted to go for the real deal and put up a draft in #1024 . Please check it out and give it some conceptual review.

tnull:
Pulling this off the milestone as superseded by #1024.

PR #960, opened 2026-07-02
Read stored objects page-by-page at startup

benthecarman:
`read_all_objects` previously fetched every key of a namespace in a single `KVStore::list` call, which can be expensive for large namespaces on remote backends like VSS. Now that all node stores implement `PaginatedKVStore`, list keys via `list_paginated` and read each page's objects with the existing bounded-concurrency batching. This covers the payment and pending payment stores read while building the node, and lays groundwork for reading stores lazily rather than loading them fully into memory.

### DIFF
--- a/src/io/utils.rs
+++ b/src/io/utils.rs
@@ -26,8 +26,8 @@ use lightning::routing::scoring::{
 	ChannelLiquidities, ProbabilisticScorer, ProbabilisticScoringDecayParameters,
 };
 use lightning::util::persist::{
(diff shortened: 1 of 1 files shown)


tnull (commented):
I'll hold off on this one until we decided the more general questions w.r.t. pagination etc in #959.

joostjager:
If we stick with loading all payments at startup for now, https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/issues/4779 could avoid the per-key reads by fetching each page’s keys and values together.

benthecarman:
this was done by #1024

PR #964, opened 2026-07-03
BOLT12 LSPS2 Support via persisted BOLT12 state

TheBlueMatt:
This is an alternative to #817 which uses the persisted LSPS2 parameters from https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/pulls/4781. It moves away from having the receive methods specify LSPS parameters in favor of just always using them. This is somewhat required for BOLT 12 as we may have an offer from 6 months ago that we still want to receive payments to, but we've since swapped the LSPS parameters we use and we don't want to keep reusing an old LSP just because we configured it as the LSP once. Instead, the new `lightning-liquidity` persists the latest state for our LSPs services and, on startup, ldk-node deletes the negotiated parameters for LSPs we're no longer configured to use.

This is all just claude output, I only skimmed the code to make sure its doing what I intended.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -183,16 +183,16 @@ harness = false
 #vss-client-ng = { path = "../vss-client" }
 #vss-client-ng = { git = "https://github.com/lightningdevkit/vss-client", branch = "main" }
 #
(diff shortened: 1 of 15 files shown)


tnull:
Superseded by #999.

PR #971, opened 2026-07-08
Bump LDK dependency

tnull:
Update LDK to revision 506cb91f2e0fb87906188b79777bcf42595d3623 from the 0.3 branch, and adapt onion-message interception to the new event and messenger APIs so the crate compiles.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -14,6 +14,7 @@ use std::sync::{Arc, Mutex};
 use bitcoin::blockdata::locktime::absolute::LockTime;
 use bitcoin::secp256k1::PublicKey;
 use bitcoin::{Amount, OutPoint};
+use lightning::blinded_path::message::NextMessageHop;
 use lightning::events::bump_transaction::BumpTransactionEvent;
 #[cfg(not(feature = "uniffi"))]
 use lightning::events::PaidBolt12Invoice;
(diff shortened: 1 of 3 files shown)


src/event.rs
```diff
@@ -1725,14 +1726,18 @@ where
...
+							self.logger,
+							"Onion message intercepted, but no onion message mailbox available"
+						);
+					}
 				} else {
-					log_trace!(
-						self.logger,
-						"Onion message intercepted, but no onion message mailbox available"
-					);
+					log_trace!(self.logger, "Onion message intercepted for unknown SCID");
```
joostjager: Shouldn't this be an error because the onion messenger is now created with `intercept_for_unknown_scids` set to false?

tnull: Hmm, not sure, that seems overly strict? I.e., seems asserting that only intercepted events are emitted should be part part of LDK's test coverage, not part of LDK Node's production code?

joostjager: I mean just logging it as error, not necessarily returning an error.

tnull: Okay, fine by me if you prefer. Amend the commit.

joostjager: Was just asking because it seemed that this branch is never expected to be hit.

joostjager (approved):

PR #972, opened 2026-07-08
Fix flaky reorg tests

tnull:
Previously, the reorg tests could be somewhat flaky. Here we attempt to stabilize them.

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -753,7 +753,7 @@ pub(crate) async fn generate_blocks_and_wait<E: ElectrumApi>(
 	let address = bitcoind.new_address().expect("failed to get new address");
 	// TODO: expect this Result once the WouldBlock issue is resolved upstream.
 	let _block_hashes_res = bitcoind.generate_to_address(num, &address);
-	wait_for_block(electrs, cur_height as usize + num).await;
+	wait_for_block(bitcoind, electrs, cur_height as usize + num).await;
 	print!(" Done!");
 	println!("\n");
 }
@@ -773,27 +773,25 @@ pub(crate) fn invalidate_blocks(bitcoind: &BitcoindClient, num_blocks: usize) {
 	assert!(new_cur_height + num_blocks == cur_height);
 }
 
-pub(crate) async fn wait_for_block<E: ElectrumApi>(electrs: &E, min_height: usize) {
-	let mut header = match electrs.block_headers_subscribe() {
-		Ok(header) => header,
-		Err(_) => {
-			// While subscribing should succeed the first time around, we ran into some cases where
-			// it didn't. Since we can't proceed without subscribing, we try again after a delay
-			// and panic if it still fails.
-			tokio::time::sleep(Duration::from_secs(3)).await;
-			electrs.block_headers_subscribe().expect("failed to subscribe to block headers")
-		},
-	};
-	loop {
-		if header.height >= min_height {
-			break;
+pub(crate) async fn wait_for_block<E: ElectrumApi>(
+	bitcoind: &BitcoindClient, electrs: &E, min_height: usize,
+) {
+	let expected_block_hash = exponential_backoff_poll(|| {
+		let bitcoind_height =
+			bitcoind.get_blockchain_info().expect("failed to get blockchain info").blocks as usize;
+		if bitcoind_height < min_height {
+			return None;
 		}
-		header = exponential_backoff_poll(|| {
-			electrs.ping().expect("failed to ping electrs");
-			electrs.block_headers_pop().expect("failed to pop block header")
-		})
-		.await;
-	}
+		bitcoind.get_block_hash(min_height as u64).ok()?.block_hash().ok()
+	})
+	.await;
+	// A height-only wait can return the old header during a same-height reorg. Require the
+	// replacement hash so callers cannot sync against the stale chain by mistake.
+	exponential_backoff_poll(|| {
+		let header = electrs.block_header(min_height).ok()?;
+		(header.block_hash() == expected_block_hash).then_some(())
+	})
+	.await;
 }
 
 pub(crate) async fn wait_for_tx<E: ElectrumApi>(electrs: &E, txid: Txid) {
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -810,15 +810,14 @@ pub(crate) async fn wait_for_outpoint_spend<E: ElectrumApi>(electrs: &E, outpoin
 	let tx = electrs.transaction_get(&outpoint.txid).unwrap();
 	let txout_script = tx.output.get(outpoint.vout as usize).unwrap().clone().script_pubkey;
 
-	let is_spent = !electrs.script_get_history(&txout_script).unwrap().is_empty();
-	if is_spent {
-		return;
-	}
-
+	// Script history already contains the funding transaction itself, so wait until the exact
+	// funding outpoint leaves the unspent set instead of treating any history as a spend.
 	exponential_backoff_poll(|| {
 		electrs.ping().unwrap();
 
-		let is_spent = !electrs.script_get_history(&txout_script).unwrap().is_empty();
+		let is_spent = !electrs.script_list_unspent(&txout_script).unwrap().iter().any(|output| {
+			output.tx_hash == outpoint.txid && output.tx_pos == outpoint.vout as usize
+		});
 		is_spent.then_some(())
 	})
 	.await;
--- a/tests/reorg_test.rs
+++ b/tests/reorg_test.rs
@@ -76,7 +76,9 @@ proptest! {
 				nodes_funding_tx.insert(node.node_id(), funding_txo);
 			}
 
-			generate_blocks_and_wait(bitcoind, electrs, 6).await;
+			// Keep funding confirmed across the deepest reorg. rust-lightning PR #4231 exempts
+			// only trusted zero-conf channels; regular channels still force-close at zero confirmations.
+			generate_blocks_and_wait(bitcoind, electrs, 7).await;
 			sync_wallets!();
 
 			reorg!(reorg_depth);
--- a/tests/reorg_test.rs
+++ b/tests/reorg_test.rs
@@ -8,10 +8,26 @@ use proptest::prelude::prop;
 use proptest::proptest;
 
 use crate::common::{
-	expect_event, generate_blocks_and_wait, invalidate_blocks, open_channel,
-	premine_and_distribute_funds, random_chain_source, random_config, setup_bitcoind_and_electrsd,
-	setup_node, wait_for_outpoint_spend,
+	expect_event, exponential_backoff_poll, generate_blocks_and_wait, invalidate_blocks,
+	open_channel, premine_and_distribute_funds, random_chain_source, random_config,
+	setup_bitcoind_and_electrsd, setup_node, wait_for_outpoint_spend, wait_for_tx,
 };
+
+async fn wait_for_pending_sweep_balance<F>(
+	node: &ldk_node::Node, mut matches_balance: F,
+) -> PendingSweepBalance
+where
+	F: FnMut(&PendingSweepBalance) -> bool,
+{
+	exponential_backoff_poll(|| {
+		node.sync_wallets().unwrap();
+		node.list_balances()
+			.pending_balances_from_channel_closures
+			.into_iter()
+			.find(|balance| matches_balance(balance))
+	})
+	.await
+}
 
 proptest! {
 	#![proptest_config(proptest::test_runner::Config::with_cases(5))]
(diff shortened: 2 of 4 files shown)


joostjager:
>Here we attempt to stabilize them

Were you able to reliably repro the flakes and see that this PR fixed them?

tests/common/mod.rs
```diff
@@ -358,13 +358,14 @@ pub(crate) fn random_storage_path() -> PathBuf {
 	temp_path
 }
 
-static NEXT_PORT: AtomicU16 = AtomicU16::new(20000);
-
 pub(crate) fn generate_listening_addresses() -> Vec<SocketAddress> {
-	let port = NEXT_PORT.fetch_add(2, Ordering::Relaxed);
+	let listener_a = TcpListener::bind(("127.0.0.1", 0)).expect("available listener port");
```
joostjager: The PR doesn't give more info on the nature of the flakes. Was it port collisions again?

tests/common/mod.rs
```diff
@@ -358,13 +358,14 @@ pub(crate) fn random_storage_path() -> PathBuf {
...
+	let listener_b = TcpListener::bind(("127.0.0.1", 0)).expect("available listener port");
+	let port_a = listener_a.local_addr().expect("listener address").port();
+	let port_b = listener_b.local_addr().expect("listener address").port();
```
joostjager: The problem here is also again that the listener is dropped before the node binds, leaving a window where it can be re-allocated in another thread.

tests/reorg_test.rs
```diff
@@ -143,43 +143,56 @@ proptest! {
...
-						return;
-					}
-					match node.list_balances().lightning_balances[0] {
-						LightningBalance::ClaimableAwaitingConfirmations {
-							confirmation_height,
-							..
-						} => {
-							let cur_height = node.status().current_best_block.height;
-							let blocks_to_go = confirmation_height - cur_height;
+				if force_close {
```
joostjager: This commit seems to combine two unrelated fixes.

tests/reorg_test.rs
```diff
@@ -143,43 +143,56 @@ proptest! {
...
+								} => Some(*confirmation_height),
+								_ => None,
+							}
+						});
+						let Some(confirmation_height) = confirmation_height else {
+							continue;
+						};
+						found_claimable_balance = true;
+
+						let cur_height = node.status().current_best_block.height;
```
joostjager: Seems quite a few things are changing in this diff hunk. Is it really minimal without unnecessary restructuring?

tests/reorg_test.rs
```diff
@@ -236,6 +236,8 @@ proptest! {
 
 				assert_eq!(node.next_event(), None);
 			});
+
+			stop_nodes_concurrently(nodes).await;
```
joostjager: [P2] tests/reorg_test.rs:240 only stops nodes on the success path. All assertions and waits above it can still panic or time out, and then nodes is dropped during unwinding before stop_nodes_concurrently(nodes).await runs. Since Node::drop calls stop(), failing proptest cases can still hit the exact shutdown-on-runtime path this commit is trying to avoid. That matters especially for proptest, because failures need to unwind cleanly for shrinking and diagnostics.

tests/common/mod.rs
```diff
@@ -730,6 +730,23 @@ where
 	}
 }
 
+pub(crate) async fn stop_nodes_concurrently(nodes: Vec<TestNode>) {
+	let stop_receivers = nodes
+		.into_iter()
+		.map(|node| {
+			let (stop_sender, stop_receiver) = tokio::sync::oneshot::channel();
+			std::thread::spawn(move || {
```
joostjager: Add comment why we need to do it concurrently? And how it would otherwise block the tokio worker?

tests/common/mod.rs
```diff
@@ -656,26 +660,65 @@ pub(crate) fn invalidate_blocks(bitcoind: &BitcoindClient, num_blocks: usize) {
...
-			tokio::time::sleep(Duration::from_secs(3)).await;
-			electrs.block_headers_subscribe().expect("failed to subscribe to block headers")
-		},
-	};
+	let mut delay = Duration::from_millis(64);
+	let mut tries = 0;
 	loop {
-		if header.height >= min_height {
-			break;
+		if electrs.block_header(min_height).is_ok() {
```
joostjager: [P2] tests/common/mod.rs:709-714 makes wait_for_block return on a stale reorged-out header. The two remaining callers use it after invalidate_blocks and manual generateblock calls, for example tests/integration_tests_rust.rs:746-756 and tests/integration_tests_rust.rs:1995-2005. At original_height + 1, electrs may already have the old block header from before the invalidation, so block_header(min_height).is_ok() can return immediately before electrs has switched to the replacement chain. Then the following wallet sync can still observe the pre-reorg chain, preserving the flake this helper is supposed to avoid. These callers need to wait for the replacement block hash, not just any header at that height.

tests/common/mod.rs
```diff
@@ -656,26 +660,65 @@ pub(crate) fn invalidate_blocks(bitcoind: &BitcoindClient, num_blocks: usize) {
...
+			return;
+		}
+		assert!(tries < 120, "electrs did not serve block header {} within 60 seconds", min_height);
+		tries += 1;
+		tokio::time::sleep(delay).await;
+		if delay.as_millis() < 512 {
```
joostjager: why not the exponential backoff anymore?

tests/common/mod.rs
```diff
@@ -656,26 +660,65 @@ pub(crate) fn invalidate_blocks(bitcoind: &BitcoindClient, num_blocks: usize) {
...
-		Ok(header) => header,
-		Err(_) => {
-			// While subscribing should succeed the first time around, we ran into some cases where
-			// it didn't. Since we can't proceed without subscribing, we try again after a delay
-			// and panic if it still fails.
-			tokio::time::sleep(Duration::from_secs(3)).await;
-			electrs.block_headers_subscribe().expect("failed to subscribe to block headers")
-		},
-	};
+	let mut delay = Duration::from_millis(64);
```
joostjager: What is the reason this method is touched?

tests/common/mod.rs
```diff
@@ -656,26 +660,65 @@ pub(crate) fn invalidate_blocks(bitcoind: &BitcoindClient, num_blocks: usize) {
 	assert!(new_cur_height + num_blocks == cur_height);
 }
 
+async fn wait_for_bitcoind_block_in_electrs<E: ElectrumApi>(
+	bitcoind: &BitcoindClient, electrs: &E, min_height: usize,
+	mut expected_block_hash: Option<BlockHash>,
```
joostjager: `min_height` and `expected_block_hash` and their interaction might benefit from a comment explaining how to use it.

tnull:
Alright, excuse the delay here. I now revisited the failure cases and narrowed the changes down to what's absolutely necessary, while improving the commit messages and comments. Also found another root cause for a flake, as we'd force-close if we entirely unconfirm a funding for a non-0conf channel.

tests/common/mod.rs
tnull: Removed.

tests/reorg_test.rs
tnull: Split out the fixes into 4 clearer-cut commits. Let me know what you think.

tests/reorg_test.rs
tnull: Should be better now.

tests/common/mod.rs
tnull: We no longer do this.

tests/common/mod.rs
tnull: No longer applies.

tests/common/mod.rs
tnull: No longer applies.

tests/common/mod.rs
tnull: Done.

tests/reorg_test.rs
tnull: No longer applies.

tests/common/mod.rs
tnull: Fixed.

tests/reorg_test.rs
```diff
@@ -76,7 +76,9 @@ proptest! {
 				nodes_funding_tx.insert(node.node_id(), funding_txo);
 			}
 
-			generate_blocks_and_wait(bitcoind, electrs, 6).await;
+			// Keep funding confirmed across the deepest reorg. rust-lightning PR #4231 exempts
+			// only trusted zero-conf channels; regular channels still force-close at zero confirmations.
+			generate_blocks_and_wait(bitcoind, electrs, 7).await;
```
joostjager, line 97: Curious how this could have caused a flake?

tests/common/mod.rs
```diff
@@ -753,7 +753,8 @@ pub(crate) async fn generate_blocks_and_wait<E: ElectrumApi>(
 	let address = bitcoind.new_address().expect("failed to get new address");
 	// TODO: expect this Result once the WouldBlock issue is resolved upstream.
 	let _block_hashes_res = bitcoind.generate_to_address(num, &address);
-	wait_for_block(electrs, cur_height as usize + num).await;
+	let min_height = cur_height as usize + num;
+	wait_for_block(bitcoind, electrs, min_height).await;
```
joostjager: Unnecessary var extraction, not consistent with calls below in the integration test?

tests/common/mod.rs
```diff
@@ -811,15 +811,22 @@ pub(crate) async fn wait_for_outpoint_spend<E: ElectrumApi>(electrs: &E, outpoin
 	let tx = electrs.transaction_get(&outpoint.txid).unwrap();
 	let txout_script = tx.output.get(outpoint.vout as usize).unwrap().clone().script_pubkey;
 
-	let is_spent = !electrs.script_get_history(&txout_script).unwrap().is_empty();
+	// Script history already contains the funding transaction itself, so wait until the exact
+	// funding outpoint leaves the unspent set instead of treating any history as a spend.
+	let is_spent =
+		!electrs.script_list_unspent(&txout_script).unwrap().iter().any(|output| {
+			output.tx_hash == outpoint.txid && output.tx_pos == outpoint.vout as usize
+		});
```
joostjager: Initial poll looks like an optimization, but is it needed?

tests/reorg_test.rs
```diff
@@ -146,40 +162,59 @@ proptest! {
...
+					sync_wallets!();
+				}
 
-					assert!(node.list_balances().lightning_balances.len() < 2);
-					assert!(node.list_balances().pending_balances_from_channel_closures.len() > 0);
-					match node.list_balances().pending_balances_from_channel_closures[0] {
-						PendingSweepBalance::BroadcastAwaitingConfirmation { .. } => {},
-						_ => panic!("Unexpected balance state!"),
+				// Mining for one node advances the shared chain for every node. Mature all
+				// claimable outputs together and wait for every sweep to reach the mempool.
```
joostjager: Comment might not fully match the code below, because it's not just waiting for reaching the mempool, but also AwaitingThresholdConfirmations

joostjager (commented):
Looks good, clear improvements. Only minor questions.

tests/reorg_test.rs
tnull: LDK currently force-closes a non-0conf channel if the funding is entirely unconfirmed due to a reorg. In the reorg test we randomize reorg height to be 1..=6, and if it happen to roll 6 in this particular instance the channel would force-close and we'd get the flake. The fix here is to just generate one more block to assert that we never completely reorg out the funding transaction here.

tests/common/mod.rs
tnull: Will revert.

tests/reorg_test.rs
tnull: Amended.

tests/reorg_test.rs
joostjager: I understand the fix, but in my opinion non-deterministic tests are something to avoid. Obviously it's a cheap way to expand coverage, but the cost of that is felt down the line.

tnull: Oh, I disagree on that sentiment. Exactly for things like reorgs you want to introduce some cases that make use of randomization (essentially chaos testing) to test cases that you'd otherwise only ever hit in production.

> but the cost of that is felt down the line

I have no idea what 'the cost of that' is meant to be in this case.

Also, didn't you *just* expand/improve reorg and mempool coverage of LDK's fuzzer? How's that materially different from the proptests here?

tests/common/mod.rs
tnull: It was an optimization that is now gone as you seem to prefer it.

joostjager: If we need the optimization, it doesn't need to be removed. But if it doesn't add anything meaningful, it's just duplicated code that can get out of sync.

tests/reorg_test.rs
joostjager: Fair, I should have been clearer. I’m not against randomized/chaos testing, quite the opposite, that’s also why I expanded the reorg/mempool fuzzer coverage.

The cost I meant is mostly CI signal: before a counterexample is persisted, a deterministic boundary case can look intermittent because it only fails when proptest generates that input. The repro output helps, but there is still triage overhead.

My unease is mostly about the overlap between fuzzing and proptests, since both serve a discovery role. I’m trying to get clearer on which randomized coverage we want in regular CI versus in fuzz testing.

joostjager (approved):

PR #973, opened 2026-07-08
Require `ChannelClosed::counterparty_node_id`

tnull:
Quick follow-up to #895.

Require `ChannelClosed` events to carry the counterparty node id. This removes a legacy missing value that callers could no longer handle. Persisted v0.1 queues with missing ids now fail to read. Newer optional encodings with the id still read correctly.


benthecarman (approved):
test is kinda overkill but lgtm

tnull:
> test is kinda overkill but lgtm

Hmm, well, one less `unwrap` we leave to the downstream user. In particular doesn't make sense as we have an `expect` on the field since #895.

PR #974, opened 2026-07-08
Remove test-helper warnings

benthecarman:
fixes the annoying unused warnings in tests

The shared integration-test helpers are macros used by different test binaries in different combinations. Export them as macros instead of reexporting private macro definitions from common, so targets that use only a subset no longer emit unused helper warnings.


tnull (commented):
Sir, this is broken right now.

Also not a big fan of using `macro_export` here. It might make more sense to `cfg`-gate the respective methods so they are only exported when really needed.

PR #977, opened 2026-07-13
Cache LSPS2 get_info responses 

benthecarman:
Reuse opening fee offers while they remain valid and match the configured LSP token. This avoids repeated LSPS2 round trips across node restarts while retaining a fresh-response fallback.

Evict cached offers when a buy request fails so a retry fetches updated parameters.


tnull (commented):
Not the biggest fan of this as we *just* got rid of LSPS2 parameters persistence and switched to store them in the metadata. 

Also, there is a larger ongoing discussion on how to fit all of this into the BOLT12 flow, so if anything all of that would need to be considered if/when we'd persist anything.

tnull:
Superseded by #999.

PR #660, opened 2025-10-13
Zero-fee commitments support

tankyleo:

PR #979, opened 2026-07-14
Do not record unilateral closes as payments

tankyleo:
Unilateral closes are not onchain payments. The real on-chain credits to our onchain balance after a unilateral close happen on `Sweep` and `Claim` transactions.

Also, `UnilateralClose` payments currently never graduate from the `Pending` state, as no BDK events are emitted when these transactions confirm.

Co-Authored-By: HAL 9000

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1183,8 +1183,8 @@ impl Wallet {
 			LdkTransactionType::InteractiveFunding { candidates } => {
 				self.classify_interactive_funding(tx, candidates, tx_type.clone().into()).await
 			},
+			LdkTransactionType::UnilateralClose { .. } => Ok(()),
 			LdkTransactionType::CooperativeClose { .. }
-			| LdkTransactionType::UnilateralClose { .. }
 			| LdkTransactionType::AnchorBump { .. }
 			| LdkTransactionType::Claim { .. }
 			| LdkTransactionType::Sweep { .. } => {
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -438,7 +438,7 @@ fn assert_any_node_has_onchain_tx_type<F: Fn(&TransactionType) -> bool + Copy>(
 }
 
 fn assert_all_nodes_have_onchain_tx_type<F: Fn(&TransactionType) -> bool + Copy>(
-	nodes: &[(&str, &TestNode)], tx_type_name: &str, predicate: F,
+	nodes: &[(&str, &TestNode)], panic_msg: &str, tx_type_name: &str, predicate: F,
 ) {
 	if nodes.iter().all(|(_, node)| has_onchain_tx_type(node, predicate)) {
 		return;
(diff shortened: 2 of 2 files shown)


tankyleo:
Continued from https://github.com/lightningdevkit/ldk-node/pull/660

The original motivation for this change: commitment transactions in 0FC channels have zero-fees and zero-amount according to BDK (since they don't interact with the on-chain wallet), so these were not recorded in the payment store. I then failed the assertion in the `channel_full_cycle` test that asserts that `UnilateralClose` was present in the payment store.

> Hmm, these don't directly interact with the onchain wallet, but wouldn't it still make sense to record them as otherwise users would have a 'gap' between channel funding and the resulting Sweep transaction? If we keep them that could allow them to track the funds end-to-end, on each step?

Would have to think more about the overall purpose of payment store, as a user I think I would find it quite confusing for there to be a zero-amount inbound payment, with a non-zero fee. In the case of zero-fee commitment channels, these would be inbound payment for zero-amount, zero-fee (but the entry could still allow us to track the movement of funds).

Also note that currently `UnilteralClose` only gets recorded for outbound channels, as BDK can tracks the funding output and assigns a non-zero fee only in this case. This makes sense as for inbound channels, the fee of the commitment transaction comes from the counterparty's balance, not our balance.

tests/common/mod.rs
```diff
@@ -1707,7 +1726,7 @@ pub(crate) async fn do_channel_full_cycle<E: ElectrumApi>(
 			node_a.list_peers().iter().any(|p| p.node_id == node_b.node_id() && p.is_persisted),
 			"node_b should remain persisted in node_a peer store after locally-initiated force-close"
 		);
-		assert_any_node_has_onchain_tx_type(
+		assert_no_node_has_onchain_tx_type(
```
tnull: nit: Can we just reuse `assert_all_nodes_have_onchain_tx_type` and invert the predicate?

tnull (dismissed):
Yeah, makes sense to me, though I still think it somehow would be nice to show users the intermediate transaction state somehow. But you're probably right that the payment store is not the right place to do it.

tests/common/mod.rs
tankyleo: Thanks I amended the commit below

tankyleo:
Hmm @tnull let me know what you think of this "unknown splice funding txid" CI failure in Rust Tests. So far on my machine I've been unable to reproduce with 
```
RUSTFLAGS="--cfg cycle_tests --cfg tokio_unstable" cargo test channel_full_cycle_0conf_0reserve`
```

tnull:
> Hmm @tnull let me know what you think of this "unknown splice funding txid" CI failure in Rust Tests. So far on my machine I've been unable to reproduce with
> 
> ```
> RUSTFLAGS="--cfg cycle_tests --cfg tokio_unstable" cargo test channel_full_cycle_0conf_0reserve`
> ```

I think it's a flake that I have yet to investigate, so may not be related to this PR.

src/wallet/mod.rs
```diff
@@ -1183,8 +1183,15 @@ impl Wallet {
 			LdkTransactionType::InteractiveFunding { candidates } => {
 				self.classify_interactive_funding(tx, candidates, tx_type.clone().into()).await
 			},
+			LdkTransactionType::UnilateralClose { .. } => {
+				log_trace!(
+					self.logger,
+					"Not recording unilateral close broadcast {} as a payment",
```
tnull: Bit odd to log *not* to do something, especially since we don't do so for all the other variants. Maybe just drop the log?

tankyleo: Done

tnull (approved):

PR #981, opened 2026-07-14
Stabilize probing restart test

tnull:
Wait for monitor persistence and queued probe failures to settle before accepting an in-flight probe for the restart assertion. This prevents transient locked amounts from failing the test under CI load.

Co-Authored-By: HAL 9000


TheBlueMatt (approved):
makes sense to me

PR #982, opened 2026-07-14
wallet: persist connected blocks without holding the wallet lock

randomlogin:
`block_connected` applied the block and then called `block_on(persist_async)` while still holding the wallet mutex (`self.inner`) — the old `persist_async` was a method on the locked wallet, so the lock was structurally required.

`Runtime::block_on` parks the current worker via `block_in_place` while driving the persist future. Holding the blocking (std) wallet lock across that park means any other task touching the wallet blocks the thread for the whole persist I/O; on a single-worker runtime that can wedge the block-application pipeline, and it widens the window for the documented `block_on` I/O-starvation deadlock.

  Apply the block under the lock, clone the staged changeset, drop the lock, then   persist the clone via `AsyncWalletPersister::persist`, clearing the stage only   once the persist succeeds. We still block on the persist before returning (block   N durable before N+1 is applied). Clearing only on success mirrors BDK's   `persist_async`: a failed persist leaves the changeset staged so the next block  retries it and the on-disk state stays a consistent prefix. A reader observing   applied-but-unflushed state is safe because a crash reverts to the last   persisted checkpoint.

This was heavily AI-coded.
Originally encountered the deadlock problem in one of the tests during kyoto chain source implementation.

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -16,7 +16,9 @@ use bdk_wallet::descriptor::ExtendedDescriptor;
 use bdk_wallet::error::{BuildFeeBumpError, CreateTxError};
 #[allow(deprecated)]
 use bdk_wallet::SignOptions;
-use bdk_wallet::{Balance, KeychainKind, LocalOutput, PersistedWallet, Update, WalletEvent};
+use bdk_wallet::{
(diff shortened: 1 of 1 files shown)


tnull:
Thanks, though this was reported in #978 and more fundamental fix is up over at #980. Closing as duplicate.

PR #983, opened 2026-07-14
Remove force-close peers after reconnect

tnull:
In #895 we started keeping peer store entries if the last channel was force-closed to allow for recovery via the channel establish flow.

However, that would result in the peer being persisted indefinitely in our peer store. Here, we add a new task that explictly disconnects and reconnects once before dropping the peer.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -53,6 +53,10 @@ where
 		self.do_connect_peer(node_id, addr).await
 	}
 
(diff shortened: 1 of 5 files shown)


valentinewallace (approved):

valentinewallace:
The CI error looks unrelated but I'll hold off on landing just in case

tnull:
> The CI error looks unrelated but I'll hold off on landing just in case

Yes, see #984.

Issue #478, opened 2025-03-03
bLIP-52/LSPS2 service: Revisit service params


tnull:
In https://github.com/lightningdevkit/ldk-node/pull/420 we added initial support for acting as an bLIP-52/LSPS2 service.

While we introduced a first batch of config knobs, we should (after feedback from LSPs) consider providing further config options, including:

- min/max channel size
- max fee (?)

PR #985, opened 2026-07-14
Allow setting a ChannelConfig for LSPS2 JIT channels

### LINKED ISSUE #961: Allow configuring channel parameters based on whether the channel is an LSP client channel
For an LSP, parameters set for channels might be different between public channels and LSP channels. We should find a way to allow these to be configured separately.

IgnacioPorte:
Hey this PR closes #961.

## Context

When acting as an LSPS2 service, the LSP always reuse the node-wide configuration
for the JIT channels it opens towards clients, hard coding their forwarding fees
to zero. An LSP may however want different parameters on client channels than
on its public channels, e.g., different forwarding fees or a different
`cltv_expiry_delta`.

## Changes

Adds an optional `channel_config` field to `LSPS2ServiceConfig`:

- If set, the given `ChannelConfig` is applied as-is to newly-opened JIT
  channels. In particular, configured forwarding fees are charged in addition
  to the channel opening fee.
- If unset (`None`), behavior is unchanged: we reuse the node-wide config but
  override the forwarding fees to zero, as we're already compensated via the
  channel opening fee.

## Compatibility

This is a breaking change, existing `LSPS2ServiceConfig` initializers need to
provide the new field. I assumed this is acceptable based on the struct's
history.

## Testing

Extended `lsps2_client_service_integration` to pass a custom config and assert
the opened JIT channel carries it.

---

This PR was developed with the assistance of AI tooling (Claude Code).

### DIFF
--- a/src/liquidity/service/lsps2.rs
+++ b/src/liquidity/service/lsps2.rs
@@ -22,6 +22,7 @@ use lightning_liquidity::lsps2::msgs::LSPS2RawOpeningFeeParams;
 use lightning_liquidity::lsps2::service::LSPS2ServiceConfig as LdkLSPS2ServiceConfig;
 use lightning_types::payment::PaymentHash;
 
+use crate::config::MaxDustHTLCExposure;
 use crate::logger::{log_error, LdkLogger};
 use crate::types::{ChannelManager, KeysManager, LiquidityManager, PeerManager, Wallet};
 use crate::{total_anchor_channels_reserve_sats, Config};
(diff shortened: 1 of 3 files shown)


src/liquidity/service/lsps2.rs
```diff
@@ -106,6 +107,15 @@ pub struct LSPS2ServiceConfig {
 	///
 	/// [`Node::open_0reserve_channel`]: crate::Node::open_0reserve_channel
 	pub disable_client_reserve: bool,
+	/// The channel config we'll apply to the JIT channels we open to clients, allowing to
+	/// configure them separately from regular channels.
+	///
+	/// If set, the given config is used without modification. Note that this means any
+	/// forwarding fees configured here are charged in addition to the channel opening fee.
+	///
+	/// If set to `None`, we'll reuse the node-wide config, but override the forwarding fees
```
tnull: It's odd to reuse the object, but then honor only a few fields of it.

tnull (commented):
Thanks for looking into this. While I think it makes sense to let the LSP configure certain values, `ChannelConfig` also has a few fields that don't make sense / we don't want to override (notably forwarding fees, but also stuff like `accept_underpaying_htlcs`). So it's not the right type for this API. If anything we need to create an LSP-specific channel config variant.

src/liquidity/service/lsps2.rs
IgnacioPorte: Thanks for the review! Read through #478 and I see the problem with the forwarding fees now, the client builds the JIT invoice's route hints before the channel exists, so fees set only in the channel config would never be covered by the sender. Communicating them to the client would need protocol support that bLIP-52 doesn't have today.

I pushed a new version with a dedicated LSPS2ChannelConfig that only exposes cltv_expiry_delta, max_dust_htlc_exposure and force_close_avoidance_max_fee_satoshis. Fees stay at zero and accept_underpaying_htlcs is untouched. Let me know if that make sense.

PR #990, opened 2026-07-16
 Support BOLT12 LSPS2 JIT-channel receives

tnull:
This adds LSPS2 JIT-channel support for long-lived BOLT12 offers without tying offers to expiring intercept SCIDs. LDK Node persists pending offers and safely reusable payment leases, restores sufficiently long-lived leases at startup, negotiates and replenishes them on demand, and consumes each lease at most once when responding to an invoice request. The same lease machinery is shared with BOLT11, including multi-LSP selection, retry and failover handling, and opening-fee validation.
 
Unlike #817 negotiated parameters are not embedded into an offer’s blinded message paths. Unlike #964 the implementation does not apply one globally cached parameter set to every BOLT12 receive (which is non-spec compliant and doesn't support fixed amount offers); leases are selected by LSP and payment mode or amount, consumed once, and refreshed independently while offers remain node-ID-addressed.

Based on https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/pulls/4809


tnull:
Superseded by #999.

PR #992, opened 2026-07-17
Follow-ups to #660

### LINKED ISSUE #989: Zero-fee commitments follow-ups
Just tracking some minor follow-ups mentioned here:

https://github.com/lightningdevkit/ldk-node/pull/660#pullrequestreview-4711590576

tankyleo:
Fixes https://github.com/lightningdevkit/ldk-node/issues/989

### DIFF
--- a/.github/workflows/0fc-integration.yml
+++ b/.github/workflows/0fc-integration.yml
@@ -46,4 +46,4 @@ jobs:
           echo "ELECTRS_EXE=$( pwd )/bin/electrs-${{ runner.os }}-${{ runner.arch }}" >> "$GITHUB_ENV"
       - name: Test with 0FC enabled
         run: |
-          RUSTFLAGS="--cfg no_download --cfg cycle_tests --cfg tokio_unstable --cfg zero_fee_commitment_tests" cargo test -- --test-threads=1
+          RUSTFLAGS="--cfg no_download --cfg cycle_tests --cfg tokio_unstable --cfg zero_fee_commitment_tests" cargo test -- --test-threads=1 --no-capture
--- a/src/ffi/types.rs
+++ b/src/ffi/types.rs
@@ -44,7 +44,10 @@ pub use lightning_liquidity::lsps0::ser::LSPSDateTime;
 pub use lightning_liquidity::lsps1::msgs::{
 	LSPS1ChannelInfo, LSPS1OrderId, LSPS1OrderParams, LSPS1PaymentState,
 };
-use lightning_types::features::{InitFeatures as LdkInitFeatures, NodeFeatures as LdkNodeFeatures};
+use lightning_types::features::{
+	ChannelTypeFeatures as LdkChannelTypeFeatures, InitFeatures as LdkInitFeatures,
+	NodeFeatures as LdkNodeFeatures,
+};
 pub use lightning_types::payment::{PaymentHash, PaymentPreimage, PaymentSecret};
 pub use lightning_types::string::UntrustedString;
 use vss_client::headers::{
@@ -1819,6 +1822,102 @@ impl From<LdkNodeFeatures> for NodeFeatures {
 
 #[derive(Debug, Clone, PartialEq, Eq, uniffi::Object)]
 #[uniffi::export(Debug, Eq)]
+pub struct ChannelTypeFeatures {
+	pub(crate) inner: LdkChannelTypeFeatures,
+}
+
+#[uniffi::export]
+impl ChannelTypeFeatures {
+	/// Constructs channel type features from big-endian BOLT 9 encoded bytes.
+	#[uniffi::constructor]
+	pub fn from_bytes(bytes: &[u8]) -> Self {
+		Self { inner: LdkChannelTypeFeatures::from_be_bytes(bytes.to_vec()) }
+	}
+
+	/// Returns the BOLT 9 big-endian encoded representation of these features.
+	pub fn to_bytes(&self) -> Vec<u8> {
+		self.inner.encode()
+	}
+
+	/// Whether this channel type advertises support for `option_static_remotekey`.
+	pub fn supports_static_remote_key(&self) -> bool {
+		self.inner.supports_static_remote_key()
+	}
+
+	/// Whether this channel type requires `option_static_remotekey`.
+	pub fn requires_static_remote_key(&self) -> bool {
+		self.inner.requires_static_remote_key()
+	}
+
+	/// Whether this channel type advertises support for `option_anchors_zero_fee_htlc_tx`.
+	pub fn supports_anchors_zero_fee_htlc_tx(&self) -> bool {
+		self.inner.supports_anchors_zero_fee_htlc_tx()
+	}
+
+	/// Whether this channel type requires `option_anchors_zero_fee_htlc_tx`.
+	pub fn requires_anchors_zero_fee_htlc_tx(&self) -> bool {
+		self.inner.requires_anchors_zero_fee_htlc_tx()
+	}
+
+	/// Whether this channel type advertises support for `option_anchors_nonzero_fee_htlc_tx`.
+	pub fn supports_anchors_nonzero_fee_htlc_tx(&self) -> bool {
+		self.inner.supports_anchors_nonzero_fee_htlc_tx()
+	}
+
+	/// Whether this channel type requires `option_anchors_nonzero_fee_htlc_tx`.
+	pub fn requires_anchors_nonzero_fee_htlc_tx(&self) -> bool {
+		self.inner.requires_anchors_nonzero_fee_htlc_tx()
+	}
+
+	/// Whether this channel type advertises support for `option_taproot`.
+	pub fn supports_taproot(&self) -> bool {
+		self.inner.supports_taproot()
+	}
+
+	/// Whether this channel type requires `option_taproot`.
+	pub fn requires_taproot(&self) -> bool {
+		self.inner.requires_taproot()
+	}
+
+	/// Whether this channel type advertises support for `option_scid_alias`.
+	pub fn supports_scid_privacy(&self) -> bool {
+		self.inner.supports_scid_privacy()
+	}
+
+	/// Whether this channel type requires `option_scid_alias`.
+	pub fn requires_scid_privacy(&self) -> bool {
+		self.inner.requires_scid_privacy()
+	}
+
+	/// Whether this channel type advertises support for `option_zeroconf`.
+	pub fn supports_zero_conf(&self) -> bool {
+		self.inner.supports_zero_conf()
+	}
+
+	/// Whether this channel type requires `option_zeroconf`.
+	pub fn requires_zero_conf(&self) -> bool {
+		self.inner.requires_zero_conf()
+	}
+
+	/// Whether this channel type advertises support for `option_zero_fee_commitments`.
+	pub fn supports_anchor_zero_fee_commitments(&self) -> bool {
+		self.inner.supports_anchor_zero_fee_commitments()
+	}
+
+	/// Whether this channel type requires `option_zero_fee_commitments`.
+	pub fn requires_anchor_zero_fee_commitments(&self) -> bool {
+		self.inner.requires_anchor_zero_fee_commitments()
+	}
+}
+
+impl From<LdkChannelTypeFeatures> for ChannelTypeFeatures {
+	fn from(features: LdkChannelTypeFeatures) -> Self {
+		Self { inner: features }
+	}
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, uniffi::Object)]
+#[uniffi::export(Debug, Eq)]
 pub struct InitFeatures {
 	pub(crate) inner: LdkInitFeatures,
 }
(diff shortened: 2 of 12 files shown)


src/tx_broadcaster.rs
```diff
@@ -18,7 +18,7 @@ use crate::logger::{log_error, LdkLogger};
 use crate::types::Wallet;
 use crate::Error;
 
-const BCAST_PACKAGE_QUEUE_SIZE: usize = 50;
+const BCAST_PACKAGE_QUEUE_SIZE: usize = 256;
```
tankyleo, line 21: I went back and forth with codex on this one, for now I lean against drawing transactions at random order: we do not do this now, but in the future we might want to make sure we broadcast parents before children no ?

src/chain/esplora.rs
```diff
@@ -521,6 +521,23 @@ impl EsploraChainSource {
 	}
 }
 
+fn esplora_submitpackage_error_implies_unsupported(e: &esplora_client::Error) -> bool {
+	matches!(e, esplora_client::Error::HttpResponse { status: 400 | 404, .. })
```
tankyleo: We map 400 here to ChainSourceNotSupported because Bitcoin Core v26 returns a RPC error when submitting the dummy package, which maps to error code 400.

src/tx_broadcaster.rs
tnull: > but in the future we might want to make sure we broadcast parents before children no ?

Well, in short, I'm not sure we could even begin to guarantee this? Broadcast is inherently fallible, we never know when the network connection could drop, when the backend won't accept anything into the mempool, and when it will just decide to drop any transaction again. So without mempool introspection I'd always lean on treating broadcast as an entirely opaque operation: we submit and retry,  and only stop once we see what we expect confirmed in a block.

src/chain/esplora.rs
tnull: Grrr, this is all very brittle. I can't wait to drop all of this logic again once we can just assume `submitpackage` is available if we have any chain source at all.

src/lib.rs
```diff
@@ -335,14 +335,18 @@ impl Node {
...
 			tokio::try_join!(
 				chain_source.update_fee_rate_estimates(),
 				chain_source.validate_zero_fee_commitments_support_if_required(
 					zero_fee_commitments_support_required
 				)
 			)
-		})?;
+		});
+		if let Err(e) = startup_chain_check_res {
+			self.chain_source.stop();
```
tnull: Actually, it seems there are a bunch of (pre-existing) potential cases below where we'd error. To solve this once and for all, let's rename `start` to `start_inner` and add a new `pub fn start` that wraps that `start_inner` (probably also takes the running lock) and stops the chain source *for any* error returned.

.github/workflows/0fc-integration.yml
```diff
@@ -47,3 +47,54 @@ jobs:
       - name: Test with 0FC enabled
         run: |
           RUSTFLAGS="--cfg no_download --cfg cycle_tests --cfg tokio_unstable --cfg zero_fee_commitment_tests" cargo test -- --test-threads=1
+  eclair-interop-test:
```
tnull: Hmm, any chance we could make this an interop test for 0fc channels, not a 0fc test that does interop? I.e., can this be (a `cfg`-gated) part of the regular eclair tests, where we already have the docker setup etc.?

tankyleo: done

src/lib.rs
tankyleo: done

src/tx_broadcaster.rs
tankyleo: sounds good added the random draw below

src/lib.rs
```diff
@@ -291,6 +291,16 @@ impl Node {
 			return Err(Error::AlreadyRunning);
 		}
 
+		match self.start_inner(&mut is_running_lock) {
+			Ok(()) => Ok(()),
+			Err(e) => {
+				self.chain_source.stop();
```
tnull, line 298: Codex:

>  - [P2] Roll back background tasks after startup failure. /home/tnull/worktrees/ldk-node/pr-992-latest-20260721/src/lib.rs:294 now stops only the chain source when start_inner fails. However, wallet sync, RGS, and pathfinding tasks are spawned before listener resolution/binding can fail at
>     lines 427–473. The node remains “not running,” so stop() cannot clean them up, while another start() creates duplicate loops. Move fallible listener setup before task spawning or perform a complete task rollback.

tnull (commented):
This needs a rebase post #996.

Btw, it seems the 0fc integration test is flaky, i.e., timing out in CI very often. Would be good to investigate why and include a fix as part of this PR.

src/wallet/mod.rs
```diff
@@ -337,6 +338,10 @@ impl Wallet {
 								})
 								.collect()
 						};
+						for i in (1..txs_to_broadcast.len()).rev() {
```
tnull: Hmm, shouldn't this be a feature of the broadcaster queue in general rather than just randomizing this one callsite? Maybe we could just have it drop random entries when the queue is full?

src/chain/mod.rs
```diff
@@ -31,19 +31,20 @@ use crate::{Error, PersistedNodeMetrics};
 
 /// We use this parent-child TRUC package to make sure the configured chain source supports
 /// broadcasting packages via the `submitpackage` Bitcoin Core RPC.
-const PARENT_TXID: &str = "9a015f93fac6cb203c2b994e18b85176eb0354a22a468255516f3c6002d3f696";
+const DUMMY_PACKAGE_EXPECTED_ERROR: &str = "bad-txns-inputs-missingorspent";
```
tnull: Hmm, matching on the exact string seems not very robust. Seems some minor change in how Bitcoin Core handles errors could lead to LDK Node not starting anymore. 

As mentioned on https://github.com/lightningdevkit/ldk-node/pull/660#discussion_r3593631809, maybe we just need to accept that we currently don't have a good way of checking all backends support v29+? Should we drop this commit?

.github/workflows/0fc-integration.yml
```diff
@@ -46,4 +46,4 @@ jobs:
           echo "ELECTRS_EXE=$( pwd )/bin/electrs-${{ runner.os }}-${{ runner.arch }}" >> "$GITHUB_ENV"
       - name: Test with 0FC enabled
         run: |
-          RUSTFLAGS="--cfg no_download --cfg cycle_tests --cfg tokio_unstable --cfg zero_fee_commitment_tests" cargo test -- --test-threads=1
+          RUSTFLAGS="--cfg no_download --cfg cycle_tests --cfg tokio_unstable --cfg zero_fee_commitment_tests" cargo test
```
tnull: Seems that should be a fixup commit?

tnull (commented):
Thanks! Seems the 0fc integration tests are still timing out in CI here?

src/lib.rs
```diff
@@ -292,6 +292,16 @@ impl Node {
 			return Err(Error::AlreadyRunning);
 		}
 
+		match self.start_inner(&mut is_running_lock) {
```
tnull, line 295: Codex:

>   - [P1] Fully unwind late startup failures — /home/tnull/worktrees/ldk-node/pr-992-latest-20260724/src/lib.rs:295
> 
>     A listener resolution/bind failure can occur after wallet sync and other background tasks have already spawned. The new error handler only stops the chain source. Since is_running remains false, callers cannot invoke stop(), and retrying start() leaves duplicate tasks running. Fallible
>     listener setup should occur before spawning tasks, or the error path must fully unwind them.

Seems that might be worth an additional PR though. Let me know if you prefer I pick that up.

src/ffi/types.rs
```diff
@@ -1817,6 +1820,82 @@ impl From<LdkNodeFeatures> for NodeFeatures {
...
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, uniffi::Object)]
+#[uniffi::export(Debug, Eq)]
+pub struct ChannelTypeFeatures {
+	pub(crate) inner: LdkChannelTypeFeatures,
+}
+
+#[uniffi::export]
+impl ChannelTypeFeatures {
```
tnull, line 1830: Codex:
 

>  - [P2] Expose all channel-type feature flags to bindings — /home/tnull/worktrees/ldk-node/pr-992-latest-20260724/src/ffi/types.rs:1830
> 
>     ChannelTypeFeatures omits typed accessors for option_scid_alias and option_zeroconf, although both are valid ChannelTypeContext features supported by the underlying type. Rust callers retain those methods, but UniFFI users must manually decode to_bytes().

src/wallet/mod.rs
tankyleo: The broadcast queue itself doesn't currently allow us to drop random entries when it is full; this would require a new data structure inside the queue that allows for this random removal.

If we want this, I think this should be a separate PR no?

src/chain/mod.rs
tankyleo: I could totally see us dropping this commit. Here's my argument against dropping it:

- We'd only abort startup in case `enable_zero_fee_commitments` is `true`, and we default to `false` at the moment. So given this default, I think we can be more aggressive in how we check for 0FC support once the flag is turned on. If someone turns this flag on and fails to start, they can easily turn this flag back off.


- With codex, I've tested this validation across 26, 27, 28, 29, 30, and 31, and across mempool, blockstream, romanz, Fulcrum, and ElectrumX. With codex I think we can easily test this logic against upcoming releases as they come out, since Core only ships two releases a year, and we can adjust this as the reality on the ground changes.

.github/workflows/0fc-integration.yml
tankyleo: I was just testing things out to see if removing this flag helped. It doesn't look like it I am going to revisit this.

src/ffi/types.rs
tankyleo: Done below.

src/lib.rs
tankyleo: Let me know what you think of the commit below, we move the fallible operations before starting the tasks.

.github/workflows/0fc-integration.yml
```diff
@@ -8,7 +8,7 @@ concurrency:
 
 jobs:
   build-and-test:
-    timeout-minutes: 60
+    timeout-minutes: 75
```
tnull: Hmm, if this is the only issue, let's bump the timeout to to hours across the board to avoid other workflows hitting the same issue?

src/chain/mod.rs
tnull: > We'd only abort startup in case enable_zero_fee_commitments is true, and we default to false at the moment. So given this default, I think we can be more aggressive in how we check for 0FC support once the flag is turned on. If someone turns this flag on and fails to start, they can easily turn this flag back off.

Hmm? What happens if they enable it, then open 0fc channels, then the backend changes something subtly and they fail to startup. Also note that the user might not be an operator that is able to debug this, it might simply be that the backend changes something and this results in their mobile app crashing as we fail to startup.

src/lib.rs
tnull: I don't think we'd want to do this, as we deliberately allow inbound connections as one of the last steps after we're positive we're ready-to-go.

src/wallet/mod.rs
tnull: Hmm, fair. This PR seems to keep growing in scope. Let's just revert to only bumping the number of entries here and we can revisit once we think we have actual evidence it's an issue.

tnull (commented):
Feel free to squash fixups (or drop as discussed above).

src/lib.rs
tankyleo: Help me understand thank you, once we've passed all the can-fail calls, aren't we positive we are ready to go ?

src/chain/mod.rs
tankyleo: Thanks I previously did not give much weight to this possibility, but now I see how this could happen, I dropped the commit below.

.github/workflows/0fc-integration.yml
tankyleo: This was not the issue, I reverted this, and instead I now fetch feerates and validate zero-fee commitments support serially, see the commit below.

src/wallet/mod.rs
tankyleo: Done

tankyleo:
The third run just passed, I believe CI for 0FC is now stable, will launch another run to check things again.

.github/workflows/0fc-integration.yml
tnull: Hmm, if indeed the described (kinda well-known) Electrum issue might be the source, we should probably not just patch around this one instance in production code to make CI pass. Either we find a hotfix that is test-only for now or we need to look into if/how to tackle the electrum issue properly (in a separate PR though). As a first step, mind reverting the diff here and opening an issue that documents your findings?

src/lib.rs
tnull: Ah, I think I was misunderstanding what you're doing. Still, just moving code around so we don't start tasks before possibly emitting an error seems very brittle. We might easily add another error'ing case sometime soon, and then have the same issue again. We probably need to fix it 'properly' by signalling stop to the tasks etc., but doing this is probably out-of-scope for this PR as mentioned above.

.github/workflows/0fc-integration.yml
tankyleo: Gave it some thought, I have some concerns about reverting the diff here. Yes we've only hit this issue in CI so far, but this means we could hit this issue in prod on some user's machine at some point ?

I would revert the diff *after* we've found and shipped a proper fix for the Electrum issue.

Let me know what you think.

src/lib.rs
tankyleo: Sounds good I kept the commit below, and we can add the proper signalling in the next PR.

.github/workflows/0fc-integration.yml
tnull: > Gave it some thought, I have some concerns about reverting the diff here. Yes we've only hit this issue in CI so far, but this means we could hit this issue in prod on some user's machine at some point ?

Well, AFAIU the core issue is that we might end up blocking forever on shutdown, in particular as we have no way to abort any tasks created via `spawn_blocking`? This has been kinda known for quite a while and this specific location seem to be only a symptom of the more general issue, i.e., it's not a solution to work around it for *this one case* (let along apply the hotfix with all the latency drawbacks irrespective of chain source used).

> I would revert the diff _after_ we've found and shipped a proper fix for the Electrum issue.

Hmm, I'm not sure if we have an immediate solution for the larger issue, but I'm all ears.

tankyleo: Sounds good I reverted the commit, filed an issue regarding the abort of tasks created via `spawn_blocking`, and will see what I can do on the electrum client side, we'd want to fix this part:

> When concurrent requests share an Electrum client connection, a clean EOF can cause the active reader to return an error without waking other requests waiting on the connection.

src/lib.rs
tnull: No, I'd prefer to drop the extra diff there, as splitting the listener logic in two just makes it harder to follow going forward. We'll want to keep that code block together, but need to address the 'wind down spawned tasks on error otherwise'.

tankyleo: Ok I dropped the diff, and filed an issue.

tnull (approved):
Thanks!

PR #993, opened 2026-07-17
Add LSPS5 webhook notification support

### LINKED ISSUE #1017: LSPS5 webhook notification support
## Context

Integrate [LSPS5/bLIP-55](https://github.com/lightning/blips/blob/master/blip-0055.md) from lightning-liquidity to enable webhook-based push notifications for LDK Node clients, so clients can receive notifications from their LSP while offline.

## Todo

- [ ] Detect LSPS5 support via LSPS0's `list_protocols` on configured LSP nodes
- [ ] Add event system integration for LSPS5 events
- [ ] Add public API to register, query, and remove webhooks (`set_webhook`, `list_webhooks`, `remove_webhook`)
- [ ] Add bindings support
- [ ] Add integration tests

Camillarhi:
Integrates LSPS5 (bLIP-0055) from lightning-liquidity, enabling webhook-based push notifications so clients can be alerted to events while their app is offline. Built on the refactored multi-LSP liquidity module `(src/liquidity/{client,service})`.

- set_webhook - register/update a webhook.
- list_webhooks - query registered webhooks.
- remove_webhook - remove a registration.

When `node_id` is `None`, `set_webhook` and `remove_webhook` fan out to every LSPS5-capable LSP, so a webhook can be configured once across all configured LSPs.

Lets a node act as an LSPS5 server:

- Delivers outgoing webhook notifications over HTTP in response to `LSPS5ServiceEvent::SendWebhookNotification`.
- Auto-triggers an `onion_message_incoming` notification when an intercepted onion message targets a client that is currently offline (wired from `LdkEvent::OnionMessageIntercepted`, gated on peer connectivity via the peer manager).

Reopening after an accidental force-push pushed the branch to main's tip and auto-closed #729.

Fixes: #1017

### DIFF
--- a/src/liquidity/client/lsps5.rs
+++ b/src/liquidity/client/lsps5.rs
@@ -1,0 +1,659 @@
+// This file is Copyright its original authors, visible in version control history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
+// accordance with one or both of these licenses.
+
+use std::collections::HashMap;
+use std::ops::Deref;
+use std::sync::{Arc, Mutex, RwLock};
+use std::time::Duration;
+
+use bitcoin::secp256k1::PublicKey;
+use lightning::log_debug;
+use lightning_liquidity::lsps0::ser::LSPSRequestId;
+use lightning_liquidity::lsps5::event::LSPS5ClientEvent;
+use lightning_liquidity::lsps5::msgs::{
+	LSPS5Error, ListWebhooksResponse, RemoveWebhookResponse, SetWebhookResponse,
+};
+use tokio::sync::oneshot;
+
+use crate::connection::ConnectionManager;
+use crate::liquidity::{
+	select_all_lsps_for_protocol, select_lsps_for_protocol, LspConfig, LspNode,
+	LIQUIDITY_REQUEST_TIMEOUT_SECS, LSPS_DISCOVERY_WAIT_TIMEOUT_SECS,
+};
+use crate::logger::{log_error, log_info, LdkLogger, Logger};
+use crate::runtime::Runtime;
+use crate::types::LiquidityManager;
+use crate::Error;
+
+pub(crate) struct LSPS5Client<L: Deref>
+where
+	L::Target: LdkLogger,
+{
+	pub(crate) lsp_nodes: Arc<RwLock<Vec<LspNode>>>,
+	pub(crate) pending_set_webhook_requests:
+		Mutex<HashMap<LSPSRequestId, oneshot::Sender<Result<SetWebhookResponse, LSPS5Error>>>>,
+	pub(crate) pending_list_webhooks_requests:
+		Mutex<HashMap<LSPSRequestId, oneshot::Sender<Result<ListWebhooksResponse, LSPS5Error>>>>,
+	pub(crate) pending_remove_webhook_requests:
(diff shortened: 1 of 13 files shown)


src/liquidity/client/lsps5.rs
```diff
@@ -0,0 +1,559 @@
...
+					log_error!(
+						self.logger,
+						"Failed to send set webhook request to liquidity service: {:?}",
+						e
+					);
+					return Err(self.map_lsps5_error(&e, Error::LiquiditySetWebhookFailed));
+				},
+			};
+
+			pending_set_webhook_requests_lock.insert(request_id, sender);
```
f3r10: LSPS1/LSPS2 wrap the same kind of map in PendingRequest/PendingRequestGuard (see client/lsps2.rs), held across the .await and removing its own entry on drop.
Here nothing removes the entry when tokio::time::timeout(...) in lsps5_set_webhook/lsps5_list_webhooks/lsps5_remove_webhook gives up — the oneshot::Sender and its HashMap entry stay behind for the life of the node. On a node with a slow or flaky LSPS5 LSP this grows unbounded. Same pattern at L124 (pending_list_webhooks_requests) and L176 (pending_remove_webhook_requests)

Could we reuse PendingRequestGuard here, the way LSPS2 does? None of these three calls need the fan-out (followers) side of PendingRequest — only one caller ever awaits a given set_webhook/list_webhooks/remove_webhook — but the drop-cleanup is exactly what's missing.

f3r10 (commented):
Great work 👍

src/liquidity/mod.rs
```diff
@@ -285,18 +313,32 @@ where
...
-				lsps1_service_config: None,
-				lsps2_service_config,
-				lsps5_service_config,
-				advertise_service,
-			}
-		});
+		let lsps2_service_config =
+			self.lsps2_service.as_ref().map(|s| s.ldk_service_config.clone());
+		let lsps5_service_config = self.lsps5_service.clone();
+		let advertise_service = self
```
f3r10, line 325: advertise_service sets the shared LSPS feature bit for any configured service, not just LSPS2 — but this only reads it off lsps2_service. A node with `enable_liquidity_provider(None, Some(lsps5_cfg))` always gets advertise_service = false, with no way to turn it on.

Is that intentional, or should LSPS5-only providers be able to advertise too?

src/liquidity/client/lsps5.rs
Camillarhi: Thanks! This has been updated to use `PendingRequestGuard` just like LSPS2 does

src/liquidity/service/lsps5.rs
```diff
@@ -0,0 +1,284 @@
...
+					client_id
+				),
+				_ => log_error!(
+					self.logger,
+					"Failed to notify liquidity management request for client {}: {:?}",
+					client_id,
+					e
+				),
+			}
+			Error::LiquidityNotifyWebhookFailed
```
f3r10, line 96: SlowDownError (the notification cooldown) and every other failure both map to Error::LiquidityNotifyWebhookFailed. A caller can't tell "you're rate-limited, retry shortly" from "this genuinely failed" without parsing logs.

Worth a distinct Error::LiquidityNotifyRateLimited (or similar) so callers can branch on it?

Camillarhi: Thanks. This will be updated to return a distict error for slow down

src/liquidity/mod.rs
Camillarhi: Yeah, it's reachable. Though the flag sets the shared `LSPS0`, so it was never really an `LSPS2` thing. If I add it to the `LSPS5` config too, then `LSPS1` service lands, and that's three copies of the same flag. One node-level setting is probably where this should end up, so I'll take a look at that instead of duplicating it

PR #994, opened 2026-07-17
Add `flake.nix`

rustaceanrob:
NixOS cannot execute the generic Linux `electrs` binary that the `electrsd` crate downloads for the integration tests. Provides a devshell that supplies bitcoind and a source-built Blockstream electrs from nixpkgs, along with `BITCOIND_EXE` and `ELECTRS_EXE` environment variables. Encountered when reviewing #828

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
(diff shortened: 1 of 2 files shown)


tnull (commented):
Thanks, but not sure we want to go there. If you run nixOS that's great, but I'm not sure we want to start going down the road of add any larger OS-specific config files to this repo.

rustaceanrob:
Fair enough. Perhaps I framed this poorly. Anyone on a unix-like system that has nix installed can benefit from this environment, but if nix isn't a common development tool for the project I will close this. For anyone that uses nix following along, I had an LLM build a more comprehensive development environment: https://github.com/rustaceanrob/ldk-node-flake/tree/master

PR #997, opened 2026-07-20
Release completed cancellable tasks

tnull:
Avoid retaining completed Tokio task allocations for the node lifetime while preserving shutdown cancellation and restart semantics. This fixes a considerable memory leak as previously we'd retain the `JoinSet`ed task's allocations until `stop`/`join_next`, as tokio thankfully only notes in the https://docs.rs/tokio-util/latest/tokio_util/task/task_tracker/struct.TaskTracker.html#comparison-to-joinset.

Thanks to @elnosh for reporting this leak.

### DIFF
--- a/src/chain/bitcoind.rs
+++ b/src/chain/bitcoind.rs
@@ -31,7 +31,7 @@ use lightning_block_sync::{
 };
 use serde::Serialize;
 
-use super::WalletSyncStatus;
+use super::{WalletSyncGuard, WalletSyncStatus};
 use crate::config::{
 	BitcoindRestClientConfig, Config, DEFAULT_FEE_RATE_CACHE_UPDATE_TIMEOUT_SECS,
 	DEFAULT_TX_BROADCAST_TIMEOUT_SECS,
@@ -152,12 +152,14 @@ impl BitcoindChainSource {
 	) {
 		// First register for the wallet polling status to make sure `Node::sync_wallets` calls
 		// wait on the result before proceeding.
-		{
+		let initial_sync_guard = {
 			let mut status_lock = self.wallet_polling_status.lock().expect("lock");
 			if status_lock.register_or_subscribe_pending_sync().is_some() {
 				debug_assert!(false, "Sync already in progress. This should never happen.");
+				return;
 			}
-		}
+			WalletSyncGuard::new(&self.wallet_polling_status, Error::WalletOperationFailed)
+		};
 
 		log_info!(
 			self.logger,
@@ -285,7 +287,7 @@ impl BitcoindChainSource {
 		}
 
 		// Now propagate the initial result to unblock waiting subscribers.
-		self.wallet_polling_status.lock().expect("lock").propagate_result_to_subscribers(Ok(()));
+		initial_sync_guard.complete(Ok(()));
 
 		let mut chain_polling_interval =
 			tokio::time::interval(Duration::from_secs(CHAIN_POLLING_INTERVAL_SECS));
@@ -396,6 +398,8 @@ impl BitcoindChainSource {
 				Error::WalletOperationFailed
 			})?;
 		}
+		let sync_guard =
+			WalletSyncGuard::new(&self.wallet_polling_status, Error::WalletOperationFailed);
 
 		let res = self
 			.poll_and_update_listeners_inner(
@@ -406,7 +410,7 @@ impl BitcoindChainSource {
 			)
 			.await;
 
-		self.wallet_polling_status.lock().expect("lock").propagate_result_to_subscribers(res);
+		sync_guard.complete(res);
 
 		res
 	}
--- a/src/chain/bitcoind.rs
+++ b/src/chain/bitcoind.rs
@@ -31,7 +31,7 @@ use lightning_block_sync::{
 };
 use serde::Serialize;
 
-use super::WalletSyncStatus;
+use super::{WalletSyncGuard, WalletSyncStatus};
 use crate::config::{
 	BitcoindRestClientConfig, Config, DEFAULT_FEE_RATE_CACHE_UPDATE_TIMEOUT_SECS,
 	DEFAULT_TX_BROADCAST_TIMEOUT_SECS,
@@ -51,6 +51,31 @@ const CHAIN_POLLING_TIMEOUT_SECS: u64 = 10;
 
 type BitcoindSpvClient =
 	SpvClient<ChainPoller<Arc<BitcoindClient>, BitcoindClient>, Arc<ChainListener>>;
+
+async fn acquire_initial_wallet_sync_guard<'a>(
+	wallet_polling_status: &'a Mutex<WalletSyncStatus>,
+	stop_sync_receiver: &mut tokio::sync::watch::Receiver<()>,
+) -> Option<WalletSyncGuard<'a>> {
+	loop {
+		let mut pending_sync = {
+			let mut status_lock = wallet_polling_status.lock().expect("lock");
+			match status_lock.register_or_subscribe_pending_sync() {
+				Some(pending_sync) => pending_sync,
+				None => {
+					return Some(WalletSyncGuard::new(
+						wallet_polling_status,
+						Error::WalletOperationFailed,
+					));
+				},
+			}
+		};
+		tokio::select! {
+			biased;
+			_ = stop_sync_receiver.changed() => return None,
+			_ = pending_sync.recv() => {},
+		}
+	}
+}
 
 pub(super) struct BitcoindChainSource {
 	api_client: Arc<BitcoindClient>,
@@ -160,12 +185,13 @@ impl BitcoindChainSource {
 	) {
 		// First register for the wallet polling status to make sure `Node::sync_wallets` calls
 		// wait on the result before proceeding.
-		{
-			let mut status_lock = self.wallet_polling_status.lock().expect("lock");
-			if status_lock.register_or_subscribe_pending_sync().is_some() {
-				debug_assert!(false, "Sync already in progress. This should never happen.");
-			}
-		}
+		let Some(initial_sync_guard) =
+			acquire_initial_wallet_sync_guard(&self.wallet_polling_status, &mut stop_sync_receiver)
+				.await
+		else {
+			log_trace!(self.logger, "Stopping initial chain sync.");
+			return;
+		};
 
 		log_info!(
 			self.logger,
@@ -302,7 +328,7 @@ impl BitcoindChainSource {
 		}
 
 		// Now propagate the initial result to unblock waiting subscribers.
-		self.wallet_polling_status.lock().expect("lock").propagate_result_to_subscribers(Ok(()));
+		initial_sync_guard.complete(Ok(()));
 
 		let mut chain_polling_interval =
 			tokio::time::interval(Duration::from_secs(CHAIN_POLLING_INTERVAL_SECS));
@@ -413,6 +439,8 @@ impl BitcoindChainSource {
 				Error::WalletOperationFailed
 			})?;
 		}
+		let sync_guard =
+			WalletSyncGuard::new(&self.wallet_polling_status, Error::WalletOperationFailed);
 
 		let res = self
 			.poll_and_update_listeners_inner(
@@ -423,7 +451,7 @@ impl BitcoindChainSource {
 			)
 			.await;
 
-		self.wallet_polling_status.lock().expect("lock").propagate_result_to_subscribers(res);
+		sync_guard.complete(res);
 
 		res
 	}
@@ -1588,6 +1616,9 @@ impl std::error::Error for BitcoindClientError {}
 
 #[cfg(test)]
 mod tests {
+	use std::sync::Mutex;
+	use std::time::Duration;
+
 	use bitcoin::hashes::Hash;
 	use bitcoin::{FeeRate, OutPoint, ScriptBuf, Transaction, TxIn, TxOut, Txid, Witness};
 	use lightning_block_sync::http::JsonResponse;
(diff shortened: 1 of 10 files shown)


tnull:
Rebased after #956 landed.

src/connection.rs
elnosh: similar to fb5fbc868e16347b01e544b905f378e703a3c4f2 this `debug_assert` could be removed as receiver from this pending connection attempt might have been dropped.

elnosh (commented):
ACK fix in 08d3b5d02ef651aec0a5f18aa3ff313f9271c67a

I'm less familiar with the code in some of the other commits, specially in `liquidity`. Some of the changes implementing the different `Guard`s took me quite a bit to understand so I would've appreciated some comments but they look correct. Specifically: 

- reason for `PendingRequestGuard` needing the tokens for the pending requests in `liquidity/mod.rs`
- `PendingRequest` in `liquidity` separating the owner `Sender` from the other `followers` `Vec<oneshot::Sender<T>>`. In `connection.rs` there's just one vec for all the senders.

These make sense now but not particularly obvious so adding comments is just a suggestion, I'm fine with it as-is

src/runtime.rs
```diff
@@ -28,13 +30,18 @@ pub(crate) struct Runtime {
 }
 
 struct CancellableBackgroundTasks {
-	tasks: JoinSet<()>,
+	tasks: TaskTracker,
+	cancellation_token: CancellationToken,
 	accepting_tasks: bool,
```
tankyleo: We can delete this field here, and track `is_closed` on `TaskTracker` instead

src/runtime.rs
```diff
@@ -122,11 +128,32 @@ impl Runtime {
...
+			runtime_handle,
+		);
 	}
 
 	pub fn allow_cancellable_background_task_spawns(&self) {
-		self.cancellable_background_tasks.lock().expect("lock").accepting_tasks = true;
+		let mut cancellable_background_tasks =
+			self.cancellable_background_tasks.lock().expect("lock");
+		if cancellable_background_tasks.cancellation_token.is_cancelled() {
+			debug_assert!(
```
tankyleo: I believe this `debug_assert` is reachable here: further below we cancel the token, then drop the lock, then wait. So here the token could be canceled, but the tasks not yet empty.

src/runtime.rs
```diff
@@ -164,15 +191,15 @@ impl Runtime {
...
 			cancellable_background_tasks.accepting_tasks = false;
-			core::mem::take(&mut cancellable_background_tasks.tasks)
+			cancellable_background_tasks.tasks.close();
+			cancellable_background_tasks.cancellation_token.cancel();
+			cancellable_background_tasks.tasks.clone()
 		};
-		debug_assert!(tasks.len() > 0, "Expected some cancellable background_tasks");
-		tasks.abort_all();
-		self.block_on(async { while let Some(_) = tasks.join_next().await {} })
+		self.block_on(tasks.wait())
```
tankyleo, line 199: IIRC while we are waiting on `tasks.wait`, we could call `allow_cancellable_background_task_spawns`, reopen the `TaskTracker`, and thus `tasks.wait()` would never finish.

tankyleo (commented):
So far just looked at the first commit, will continue tomorrow

src/connection.rs
```diff
@@ -18,12 +18,44 @@ use crate::logger::{log_debug, log_error, log_info, LdkLogger};
...
+	node_id: PublicKey,
+	active: bool,
+}
+
+impl<'a> PendingConnectionGuard<'a> {
+	fn new(pending_connections: &'a PendingConnections, node_id: PublicKey) -> Self {
+		Self { pending_connections, node_id, active: true }
+	}
+
+	fn disarm(&mut self) {
```
tankyleo, line 35: nit: Is this really needed ? On drop, we'd briefly lock the pending connections, see that we have no subscribers for the node_id, and do nothing.

src/connection.rs
```diff
@@ -277,3 +313,26 @@ where
...
+			"0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798".parse().unwrap();
+		let pending_connections = Mutex::new(HashMap::new());
+		let (sender, mut receiver) = tokio::sync::oneshot::channel();
+		pending_connections.lock().expect("lock").insert(node_id, vec![sender]);
+
+		let connection_guard = PendingConnectionGuard::new(&pending_connections, node_id);
+		drop(connection_guard);
+
+		assert!(
+			pending_connections.lock().expect("lock").is_empty(),
```
tankyleo, line 328: nit: Should we also make sure that key-values for a *different* node_id are not dropped ?

src/liquidity/mod.rs
```diff
@@ -540,3 +576,83 @@ where
...
+			.get_mut("request")
+			.expect("request should be pending")
+			.followers
+			.push(follower_sender);
+
+		drop(receiver);
+		drop(request_guard);
+
+		assert!(
+			pending_requests.lock().expect("lock").is_empty(),
```
tankyleo, line 610: Similar here perhaps some coverage to make sure we don't drop requests with a different request key.

src/runtime.rs
```diff
@@ -122,11 +128,32 @@ impl Runtime {
 		// Since it seems to make a difference to `tokio` (see
 		// https://docs.rs/tokio/latest/tokio/time/fn.timeout.html#panics) we make sure the futures
 		// are always put in an `async` / `.await` closure.
-		cancellable_background_tasks.tasks.spawn_on(async { future.await }, runtime_handle);
+		let cancellation_token = cancellable_background_tasks.cancellation_token.clone();
+		// Detach the handle while the tracker continues tracking the task.
+		let _ = cancellable_background_tasks.tasks.spawn_on(
```
tankyleo, line 135: Codex: 

  - [P2] Preserve abort-on-drop for detached cancellable tasks _ /home/ubuntu/ldk-node/src/runtime.rs:133-133
    Dropping the JoinHandle here changes the old JoinSet drop behavior: TaskTracker explicitly does not abort tracked tasks when it is dropped. With an externally supplied Tokio runtime, if Node::start() returns an error after spawning cancellable tasks but before is_running is set, Drop won't call
    stop(), and these detached tasks can keep running and holding node Arcs after the node is dropped instead of being aborted as before. Please retain abort-on-drop handles or cancel/abort tracked tasks from Runtime's drop path.


This might be related to the PR we've had in mind about proper clean up after `fn start`

src/liquidity/mod.rs
```diff
@@ -42,6 +42,45 @@ use crate::{Config, Error};
...
+}
+
+impl<K: Clone + Eq + Hash, T> Drop for PendingRequestGuard<'_, K, T> {
+	fn drop(&mut self) {
+		let mut pending_requests = self.pending_requests.lock().expect("lock");
+		if pending_requests
+			.get(&self.request_key)
+			.is_some_and(|request| Arc::ptr_eq(&request.token, &self.token))
+		{
+			pending_requests.remove(&self.request_key);
```
tankyleo, line 79: Codex:

  - [P2] Avoid panicking on late liquidity responses _ /home/ubuntu/ldk-node/src/liquidity/mod.rs:79-79
    When a caller times out or is cancelled, this guard removes the request from the pending map. For LSPS1/LSPS2 a slow LSP can still deliver the matching response later; after this removal the event handlers take their existing unknown request branch, which contains debug_assert!(false), so debug/
    test builds panic and the liquidity event loop dies for any response arriving after the 5s timeout. Please either keep enough tombstone state/treat abandoned IDs as expected, or remove those assertions before deleting entries this way.

src/chain/bitcoind.rs
```diff
@@ -152,12 +152,14 @@ impl BitcoindChainSource {
 	) {
 		// First register for the wallet polling status to make sure `Node::sync_wallets` calls
 		// wait on the result before proceeding.
-		{
+		let initial_sync_guard = {
 			let mut status_lock = self.wallet_polling_status.lock().expect("lock");
 			if status_lock.register_or_subscribe_pending_sync().is_some() {
 				debug_assert!(false, "Sync already in progress. This should never happen.");
+				return;
```
tankyleo: Codex:

  - [P2] Keep the bitcoind sync loop alive _ /home/ubuntu/ldk-node/src/chain/bitcoind.rs:157-159
    If a manual bitcoind Node::sync_wallets() call owns wallet_polling_status when this background task starts (for example immediately after start() before the spawned task has registered), this branch now returns from continuously_sync_wallets and permanently stops background chain polling/fee
    updates for that run. The task should wait for or otherwise handle the in-progress sync instead of exiting.

src/runtime.rs
tnull: Hmm, it's def. not reachable from the current callsites as they are serialized.

src/connection.rs
tnull: Dropped.

src/runtime.rs
tnull: Hmm, yes, if we only look at `Runtime` alone, but note that we are always acquiring the `is_running` lock in `start`/`stop`. Let me know if you think it's crucial to make `Runtime` more robust by itself.

src/runtime.rs
tankyleo: Agreed the current callsites are serialized. Might be good to not rely on this external lock for serialization, I vibed with codex on this issue, here's the patch it came up with below. Your call if you want to incorporate this or leave the PR as is :)


```diff
    Isolate cancellable task generations

    Replace cancelled cancellable-task state as a unit so a restart cannot reopen the tracker being drained by an in-flight abort.

    AI-Assisted-By: OpenAI Codex

diff --git a/src/runtime.rs b/src/runtime.rs
index d4ea7ce..3fe04d1 100644
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -146,14 +146,10 @@ impl Runtime {
                let mut cancellable_background_tasks =
                        self.cancellable_background_tasks.lock().expect("lock");
                if cancellable_background_tasks.cancellation_token.is_cancelled() {
-                       debug_assert!(
-                               cancellable_background_tasks.tasks.is_empty(),
-                               "Expected all cancellable background tasks to be stopped"
-                       );
-                       cancellable_background_tasks.cancellation_token = CancellationToken::new();
+                       // An abort may still be waiting on a clone of the previous task tracker. Start a new
+                       // generation instead of reopening that tracker underneath the in-flight wait.
+                       *cancellable_background_tasks = CancellableBackgroundTasks::new();
                }
-               cancellable_background_tasks.tasks.reopen();
-               cancellable_background_tasks.accepting_tasks = true;
        }

        pub fn spawn_background_processor_task<F>(&self, future: F)
```

src/runtime.rs
tankyleo: let me know what you think of the patch above, it would address this here too.

src/runtime.rs
tnull: Done.

src/connection.rs
tnull: Yes. Although result propagation normally removes the entry, another thread could register a replacement after the mutex is released but before this guard drops. Disarming prevents the old guard from removing that new attempt.

src/connection.rs
tnull: I’d prefer to leave this out since removal is explicitly scoped by node_id, and the existing test covers the relevant cancellation behavior.

src/liquidity/mod.rs
tnull: See above, would also prefer to leave it out for now.

src/runtime.rs
tnull: Done.

src/liquidity/mod.rs
tnull: Done.

src/chain/bitcoind.rs
tnull: Done.

tnull:
Now updated to address pending feedback. This is now great quite a bit in scope, which I'm not super happy with. But let me know what you think.

Jolah1:
Reviewed runtime.rs, connection.rs, chain/mod.rs, liquidity/mod.rs and both LSPS clients - looks correct to me.

 One thing worth a second look: PendingRequest<T> backs six maps, but only pending_lsps0_discovery ever touches followers (pushed at liquidity/mod.rs:477, drained at :432). The identifier doesn't appear in lsps1.rs or lsps2.rs at
  all — their five handlers do request.sender.send(..) and drop request.followers. Harmless today, since each LSPS1/LSPS2 request has a unique LSPSRequestId and nothing pushes followers there. But if someone later dedups 

e.g.check_order_status, those followers would never be notified and it'd surface as an LSP timeout rather than a local bug. Maybe drain them in those handlers too, or split the type.

 On the beta failure: it died at tests/common/mod.rs:865 in premine_blocks, where exponential_backoff_poll allows only ~9.2s (20 tries, 512ms cap) for bitcoind to generate and electrs to index 101 blocks. build (self-hosted, 1.85.0) passed on the same runner, so it's probably worth a re-run before digging.

src/chain/mod.rs
```diff
@@ -65,6 +65,34 @@ pub(crate) enum WalletSyncStatus {
...
+	}
+}
+
+impl Drop for WalletSyncGuard<'_> {
+	fn drop(&mut self) {
+		if self.active {
+			self.status
+				.lock()
+				.expect("lock")
+				.propagate_result_to_subscribers(Err(self.cancellation_error));
```
tankyleo, line 91: Codex brought this up:
> - [P1] Keep Electrum sync busy until blocking work exits — /home/ubuntu/ldk-node/src/chain/mod.rs:87-91
    With the Electrum backend, if shutdown aborts an active Lightning sync after the 5-second background-task timeout, ElectrumRuntimeClient::sync_confirmables's spawn_blocking closure cannot be cancelled and can keep invoking Confirm, yet this drop resets the status to Completed. An immediate restart can then run another sync concurrently, allowing the older worker to deliver a stale tip after the newer sync; track and wait for that worker before reopening sync.

I asked codex to double check, and this issue seems to have been introduced by this PR in commit 7d60fd5b

tankyleo (commented):
Thank you feel free to squash the existing fixups

tankyleo:
This is what codex is happy with on my end, please feel free to fold into your own commits, no need for attribution.

src/chain/mod.rs
tankyleo: Asked codex to fix this, see below for what it came up with "Keep Electrum sync busy through blocking work"

tnull: Yeah, not sure we should go out of our way to work around this in this PR if the much more important issue is that Electrum might still mutate confirmables during/after shutdown (cf. https://github.com/lightningdevkit/ldk-node/issues/1007). Dropping your commit for this for now.

tnull:
Squashed fixups generally, also folded in the version bump to `0.7.10` (though note it's somewhat unnecessary due to cargo's version resolution rules, i.e., "0.7" does *not* mean "0.7.0", it means 'newest 0.7.X').

tankyleo:
> Squashed fixups generally, also folded in the version bump to `0.7.10` (though note it's somewhat unnecessary due to cargo's version resolution rules, i.e., "0.7" does _not_ mean "0.7.0", it means 'newest 0.7.X').

Yes we agree on newest 0.7.X, Codex suggested that commit because of this paragraph in the recommendations from cargo:

> Try to keep the dependency versions up-to-date with the actual minimum versions that your library requires. For example, if you have a requirement of bar="1.0.12", and then in a future release you start using new features added in the 1.1.0 release of “bar”, update your dependency requirement to bar="1.1.0".
> If you fail to do this, it may not be immediately obvious because Cargo can opportunistically choose the newest version when you run a blanket cargo update. However, if another user depends on your library, and runs cargo update your-library, it will not automatically update “bar” if it is locked in their Cargo.lock. It will only update “bar” in that situation if the dependency declaration is also updated. Failure to do so can cause confusing build errors for the user using cargo update your-library.

https://doc.rust-lang.org/cargo/reference/resolver.html#recommendations

tankyleo (approved):

tnull:
> Yes we agree on newest 0.7.X, Codex suggested that commit because of this paragraph in the recommendations from cargo:

Sure, technically preferable, just usually not worth the noise changing.

PR #999, opened 2026-07-22
 Add BOLT12 LSPS2 JIT receive support

tnull:
Based on https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/pulls/4819.

This adds BOLT12 receive support over LSPS2 JIT channels using a node-local `OffersMessageHandler` and `OffersMessageFlow`. `Offer`s retain ordinary blinded message paths and never contain an intercept SCID, allowing them to outlive individual `PaymentLeases`.

Single-use `PaymentLeases` are persisted through `DataStore`, while a bounded LRU persists fixed-amount and variable `LeaseCacheTargets`. Leases are restored and prefilled at startup, replenished after consumption, and periodically pruned when stale or close to expiry. Cache misses negotiate on demand with serialized requests, retries, and multi-LSP failover. BOLT11 uses the same cache.

Every BOLT12 offer is eligible for JIT fallback when LSPS2 is configured. We first try ordinary payment paths and omit the JIT path entirely when existing inbound liquidity is sufficient. Otherwise, we consume or negotiate a lease and include a JIT path constrained to the full payment amount. Fixed invoices advertise MPP, variable invoices disable it, and `PaymentMetadata` retains the fee policy needed to validate the LSP skim.

## Potential follow-ups:

- [ ] Currently we by-default include JIT paths whenever an LSPS2-enabled LSP is configured.  Likewise, we're automatically auto-renegotiating leases. We should consider whether/how (read: best API) to give users more control here.


src/payment/bolt12/offers.rs
```diff
@@ -0,0 +1,537 @@
...
+				payment_metadata.clone()
+			},
+			_ => None,
+		};
+		let invoice_request = match self.flow.verify_invoice_request(invoice_request, context) {
+			Ok(InvreqResponseInstructions::SendInvoice(invoice_request)) => invoice_request,
+			Ok(InvreqResponseInstructions::SendStaticInvoice { .. }) | Err(()) => return None,
+		};
+
+		let allow_mpp = invoice_request.amount().is_some();
```
TheBlueMatt: Shouldn't this always be `true` for the first (non-LSP) attempt here?

src/payment/bolt12/offers.rs
```diff
@@ -0,0 +1,537 @@
...
+					OffersMessage::InvoiceError(InvoiceError::from(error)),
+					responder.respond(),
+				));
+			},
+		};
+		let pending_request = match self.pending_jit_invoice_requests.try_acquire() {
+			Some(pending_request) => pending_request,
+			None => {
+				return Some((
+					OffersMessage::InvoiceError(InvoiceError::from_string(
```
TheBlueMatt: Should we not consider duplicatively returning an existing LSPS2 cached negotiation here? In cases where the LSP over-allocated liquidity and we have enough for a second payment, or where we're getting DoS'd and most of the invreqs will not result in a payment, returning something might actually get us the payment.

FWIW if we do this it might be worth considering a marginally different router design, which might be cleaner - rather than deciding on which lease to use and then adding it to the metadata and using that metadata to tell the router what to inject, give the router a reference to the LSPS2 lease store and have it simply ask for the latest (potentially-unused) lease for a payment, then inject the metadata it wants.

tnull: Hmm, in my mind both of these fall in the category 'further optimizations', i.e., I'd lean towards taking a new look at them in a follow-up rather than trying to land all of it in this initial implementation. 

In particular I had considered doing something along the line of what you describe in the second paragraph to avoid transferring all of the parameters via the payment metadata, but for simplicity's sake (though nothing is really simple here) for now opted to keep it in the metadata for now as that makes it at least easier to reason about compared to introducing yet-another side channel that could race if we're not super careful.

src/payment/bolt12/offers.rs
tnull: Ah, yes, it seems for the first `build_invoice` call we'll want always `true`. Will add a fixup.

src/payment/bolt12/offers.rs
TheBlueMatt: That's fine with me, I only raised it because it would likely imply a refactor of the router, so if we want to do that refactor ideally we should do it now rather than rewriting the code later. The feature itself would be nice but could be a followup.

src/liquidity/client/lsps2/mod.rs
```diff
@@ -0,0 +1,1153 @@
...
+		if let Some((lease, total_fee_msat, lsp)) =
+			self.take_cached_fixed_lease(amount_msat).await?
+		{
+			self.schedule_fixed_lease_refill(amount_msat, connection_manager);
+			return Ok((lease, total_fee_msat, lsp, false));
+		}
+
+		let (negotiated_lease, min_total_fee_msat, cheapest_lsp) =
+			self.negotiate_fixed_lease(amount_msat, connection_manager).await?;
+		let lease = self.consume_lease(&negotiated_lease.id).await?;
```
jkczyz: Won't this re-fetch through valid() and reject the lease we just bought if its validity is under 24h? Likewise in `acquire_variable_lease`.

src/liquidity/client/lsps2/mod.rs
```diff
@@ -0,0 +1,1153 @@
...
+		self.consume_selected_lease(lease).await
+	}
+
+	async fn consume_selected_lease(&self, lease: PaymentLease) -> Result<PaymentLease, Error> {
+		let lease_store = Arc::clone(&self.lease_store);
+		consume_after_persisted_removal(
+			lease,
+			move |lease| async move { lease_store.remove(&lease.id).await },
+			|lease| {
+				self.lease_state.lock().expect("lock").remove(&lease.id);
```
jkczyz: Can't two concurrent requests both select the same lease? Both callers would get the same single-use intercept SCID.

https://github.com/jkczyz/ldk-node/commit/70e350e74e72fa2455a526f079ce7880f33570ae

src/liquidity/client/lsps2/router.rs
```diff
@@ -0,0 +1,392 @@
...
+				node_id: params.lsp_node_id,
+				htlc_maximum_msat: htlc_amount_msat,
+			};
+			if let Ok(path) = BlindedPaymentPath::new(
+				&[forward_node],
+				recipient,
+				local_node_receive_key,
+				tlvs.clone(),
+				htlc_amount_msat,
+				MIN_FINAL_CLTV_EXPIRY_DELTA,
```
jkczyz: Should we use `MIN_FINAL_CLTV_EXPIRY_DELTA + 2` as is done for BOLT11?

src/event.rs
jkczyz, line 670: Claude says this now can cause a debug assertion to be reached.

https://github.com/jkczyz/ldk-node/commit/919a8c3642ad889b7683f636a1b04b203f1b46c5

src/payment/bolt12/offers.rs
```diff
@@ -0,0 +1,540 @@
...
+						Ok((invoice, context)) => (
+							OffersMessage::Invoice(invoice),
+							responder.respond_with_reply_path(context),
+						),
+						Err(error) => (
+							OffersMessage::InvoiceError(error.into_invoice_error()),
+							responder.respond(),
+						),
+					}
+				},
```
jkczyz: A payer may send many invoice requests (for robustness) but only pay one. Won't this cause us to acquire too many leases / is a DoS vector?

src/payment/bolt12/offers.rs
```diff
@@ -0,0 +1,540 @@
...
+			"JIT invoice handling must only be initialized once"
+		);
+	}
+
+	pub(crate) fn current_best_block(&self) -> BlockLocator {
+		*self.best_block.read().expect("lock")
+	}
+}
+
+fn build_invoice(
```
jkczyz, line 236: Ideally, LDK provides some abstraction so we don't need to repeat much of this logic.

jkczyz:
I had Claude review this and it came up with quite a few issues. I already left comments for ones that I was able to verify. But there some others here: https://gist.github.com/jkczyz/3324c2914601a2728605a49294f25bc9. Failing tests in this branch: https://github.com/jkczyz/ldk-node/tree/claude/pr-999-edge-cases-725f70

src/liquidity/client/lsps2/mod.rs
tnull: Hmm, good catch. It seems this is a regression as originally we had a slightly different design there that would have made this impossible. Will fix.

src/payment/bolt12/offers.rs
tnull: Well, discussed this quite a bit with Matt before. Yes, it's not ideal, but we'll have to make sure to not reuse leases/intercept SCIDs as otherwise subsequent payments will not trigger channel opens. This is unfortunately a limitation of the bLIP-52/LSPS2 spec that we can't really work around. One follow-up we could consider is to add rate-limiting and more user control, e.g., via an event-based flow in LDK Node (also see the TODO mentioned in OP).

src/payment/bolt12/offers.rs
tnull: Yeah, possibly, but at least for v0.3 that ship has sailed I'm afraid.

src/liquidity/client/lsps2/router.rs
tnull: Hmm, done.

src/liquidity/client/lsps2/mod.rs
tnull: Done in https://github.com/lightningdevkit/ldk-node/pull/999/changes/cf217700321bf720b5471785eba3cb904a5e0810

src/event.rs
tnull: Ah, yeah, good catch. Can't fully recall why we added this `debug_assert` in the first place, but it's reachable now. Dropping in fixup.

tnull:
> I had Claude review this and it came up with quite a few issues. I already left comments for ones that I was able to verify. But there some others here: https://gist.github.com/jkczyz/3324c2914601a2728605a49294f25bc9. Failing tests in this branch: https://github.com/jkczyz/ldk-node/tree/claude/pr-999-edge-cases-725f70

Still have to go through the list fully. Some seem valid, some bogus.

tnull:
> > I had Claude review this and it came up with quite a few issues. I already left comments for ones that I was able to verify. But there some others here: https://gist.github.com/jkczyz/3324c2914601a2728605a49294f25bc9. Failing tests in this branch: https://github.com/jkczyz/ldk-node/tree/claude/pr-999-edge-cases-725f70
> 
> Still have to go through the list fully. Some seem valid, some bogus.

Addressed most of the points now. @jkczyz please let me know what you think.

tnull:
Rebased to address minor conflicts.

tnull:
Rebased to resolve minor conflicts.

tnull:
Rebased to resolve minor conflicts.

src/liquidity/client/lsps2/router.rs
```diff
@@ -100,6 +101,23 @@ impl<R: Router, ES: EntropySource> Router for LSPS2Router<R, ES> {
...
+					.fold(0u64, u64::saturating_add)
+					>= amount_msats
+			} else {
+				first_hops.iter().any(|channel| {
+					channel.inbound_capacity_msat >= amount_msats
+						&& channel.inbound_htlc_minimum_msat.unwrap_or(0) <= amount_msats
+						&& channel.inbound_htlc_maximum_msat.unwrap_or(u64::MAX) >= amount_msats
+				})
+			}
+		});
```
jkczyz, line 135: :robot: The sufficiency check sums `inbound_capacity_msat` over all first hops, but LDK's router won't route through a counterparty that is unannounced or has no forwarding info for us. A node whose inbound capacity sits on such channels passes this check, gets a direct-path invoice, and the payer can't pay it — with JIT never triggering, which is the case this fixup fixes. Should the sum apply the same usability filter the router does? Related test gap: both new tests pass empty `first_hops`, so neither the sufficient-capacity accept path nor the non-MPP per-channel branch (`router.rs:114-117`) is covered.

src/liquidity/client/lsps2/mod.rs
```diff
@@ -0,0 +1,1254 @@
...
+	}
+
+	pub(crate) async fn prepare_invoice_response(
+		self: Arc<Self>, request: JitInvoiceRequest, connection_manager: Arc<ConnectionManager<L>>,
+	) -> Result<JitInvoiceResponse, Error> {
+		// Cache targets are only an optimization. Recording an older, evicted offer again makes it
+		// eligible for startup pre-caching, but a persistence failure must not make the current
+		// invoice request unusable.
+		let (target_id, absolute_expiry) = request.cache_target();
+		if let Err(error) = self.register_cache_target(target_id, absolute_expiry).await {
```
jkczyz, line 216: :robot: Registration happens before lease negotiation or any serviceability check, keyed by the payer-resolved amount. The dedup fixup and the 100-slot cap stop identical replays, but anyone holding the offer can still send distinct amounts serially, churn the 100-entry cache-target LRU, evict the legitimate targets (e.g. the variable-amount one), and seed up to 100 attacker-chosen warm buys on every restart. Won't registering only after an invoice was actually produced avoid this?

src/liquidity/client/lsps2/state.rs
```diff
@@ -0,0 +1,899 @@
...
+	pub(crate) fn prune(&mut self) {
+		self.leases.retain(|_, lease| is_lease_cacheable(lease));
+	}
+}
+
+pub(crate) fn is_lease_usable(lease: &PaymentLease) -> bool {
+	checked_cltv_expiry_delta(lease.cltv_expiry_delta).is_some() && lease.valid_until > now_secs()
+}
+
+pub(crate) fn checked_cltv_expiry_delta(cltv_expiry_delta: u32) -> Option<u16> {
```
jkczyz, line 496: :robot: `checked_cltv_expiry_delta` only enforces the `u16` bound, but the BOLT12 router adds `MIN_FINAL_CLTV_EXPIRY_DELTA + 2` (= 44) on top and LDK aggregates the path CLTV in `u16` with `checked_add`. So a delta in `65492..=65535` passes validation at buy and at selection, then path building fails *after* the lease was consumed — the same silent consume-then-destroy loop 34fac3fa prevents for `65536`, just 44 values lower. An LSP can hit it deliberately by returning `65535`. Should the accepted bound be `u16::MAX - (MIN_FINAL_CLTV_EXPIRY_DELTA + 2)`?

src/payment/bolt12/offers.rs
```diff
@@ -0,0 +1,575 @@
...
+				return Some((
+					OffersMessage::InvoiceError(InvoiceError::from(error)),
+					responder.respond(),
+				));
+			},
+		};
+		let pending_request =
+			match self.pending_jit_invoice_requests.try_acquire(invoice_request_id) {
+				Ok(pending_request) => pending_request,
+				Err(PendingJitInvoiceRequestError::AlreadyPending) => return None,
```
jkczyz: :robot: Two questions on the dedup scope (28d55cce). First, it only covers copies in flight at the same time: once the first response is enqueued, an identical copy starts a fresh flow — another single-use lease, invoice, and refill buy. With a warm cache the in-flight window is only two KV writes, while a payer's up-to-10 copies arrive staggered by onion latency, so the serial case looks like the common one (and a deliberate replayer just waits for each response). Second, duplicates get `return None`, so only the first copy's reply path is ever used — payers send copies over several reply paths precisely so one broken path doesn't fail the payment, and pre-fix each copy got its own response. Would caching the response for a short window and re-sending it for a matching hash address both?

src/payment/store.rs
```diff
@@ -181,11 +181,6 @@ impl StorableObject for PaymentDetails {
 		if let Some(hash_opt) = update.hash {
 			match self.kind {
 				PaymentKind::Bolt12Offer { ref mut hash, .. } => {
-					debug_assert_eq!(
-						self.direction,
-						PaymentDirection::Outbound,
-						"We should only ever override payment hash for outbound BOLT 12 payments"
-					);
```
jkczyz, line 188: :robot: Removing this assert has no test pinning the new behavior, so re-tightening it would go unnoticed until a debug build fails a payment. Happy to contribute the test from the first review round (`inbound_bolt12_payments_can_be_failed`) — it passes on this head.

src/payment/store.rs
jkczyz: :robot: Should the relaxation in ca0f79c3 also apply to the `Bolt12Refund` arm? It keeps the identical `debug_assert_eq!(self.direction, Outbound)`, inbound refund records are created with their hash already set (`src/event.rs:1217`), and refunds carry no LSPS2 metadata, so a skimmed refund delivery takes the `fail_claimable_payment` branch and applies the same hash-bearing update — panicking in debug builds. I have a failing test demonstrating this.

src/liquidity/client/lsps2/state.rs
```diff
@@ -0,0 +1,899 @@
...
+			.filter(|lease| is_lease_cacheable(lease))
+			.filter_map(|lease| {
+				compute_opening_fee(
+					amount_msat,
+					lease.params.min_fee_msat,
+					lease.params.proportional as u64,
+				)
+			})
+			.any(|fee_msat| max_fee_msat.map_or(true, |max| fee_msat <= max))
+	}
```
jkczyz: :robot: `fixed_amount`/`variable_amount` filter by `available_lsps`, but these `has_*` variants (from c114927d) don't. A cached lease from an LSP that is no longer discovered/configured makes the refill check (`mod.rs:685`, `:708`) report the cache as full while consumers can never select it — so every receive pays on-demand negotiation until that lease ages out. Should the `has_*` variants take the same filter?

tests/integration_tests_rust.rs
```diff
@@ -4911,6 +5207,30 @@ async fn do_lsps2_multi_lsp_picks_cheapest(reverse_order: bool) {
 	let route_hint_src = first_hop.expect("route hint should have at least one hop").src_node_id;
 	assert_eq!(route_hint_src, cheap_id, "expected cheaper LSP to be selected.");
 
+	// Consuming the BOLT11 lease schedules a replacement. The BOLT12 flow shares that cache and
+	// must retain the same cheapest-LSP selection regardless of registration order.
```
jkczyz, line 5216: :robot: This section (from 71709d4a) is unreachable in the fee-limit variant: when `max_total_lsp_fee_limit_msat` is set, the test returns early above (`:5196`), so this only runs in the no-limit variant and the BOLT12 fee-rejection path is untested.

src/event.rs
```diff
@@ -1126,6 +1166,8 @@ where
 								hash: Some(payment_hash),
 								preimage: payment_preimage,
 								secret: Some(payment_secret),
+								counterparty_skimmed_fee_msat: (counterparty_skimmed_fee_msat > 0)
+									.then_some(counterparty_skimmed_fee_msat),
```
jkczyz, line 1175: :robot: The BOLT11 claimable insert records the gross amount (`invoice_amount_msat`, i.e. `amount_msat + counterparty_skimmed_fee_msat`), while this BOLT12 insert (f4c3c997) keeps the net `amount_msat`. Both kinds carry `counterparty_skimmed_fee_msat` separately, so `amount_msat` means different things across the two. Intentional?

src/liquidity/client/lsps2/mod.rs
```diff
@@ -0,0 +1,1254 @@
...
+		let all_offers = self.gather_lsps2_offers(connection_manager).await?;
+		let mut candidates = all_offers
+			.into_iter()
+			.flat_map(|(lsp, resp)| {
+				resp.opening_fee_params_menu
+					.into_iter()
+					.map(move |params| (lsp.clone(), params))
+			})
+			.filter_map(|(lsp, params)| {
+				if !opening_fee_params_valid_for(&params, min_remaining_secs) {
```
jkczyz, line 412: :robot: The candidate filter (from 3fba5a1f) and the post-buy check use the same `min_remaining_secs`, so a candidate whose remaining validity barely exceeds it passes the filter, spends a buy round-trip, and then fails the re-check — the LSP-side buy succeeded and the SCID is discarded. Should the pre-buy filter require a small extra margin to cover the round-trip?

jkczyz (commented):
Apologies if any of these are dups. Any tests mentioned are on https://github.com/jkczyz/ldk-node/commits/claude/pr-999-v2-edge-cases/.

:robot:

> Thanks for the thorough rework — I re-verified the earlier findings against the new head and four of the five are properly fixed (the fresh-lease validity split, the atomic single-use consumption, the BOLT12 CLTV buffer, and the `Bolt12Offer` fail path), and the mid-series compile break and fork pin are resolved too. The inline comments cover three remaining bugs (each reproduced with a failing unit test) and a few follow-ups on the new fix code.
>
> Before squashing, a few structural notes: 3fba5a1f sits on 60e22706 but mostly re-parameterizes validity policy from 541f2bb6/b24e9104/271a129b, so it's worth splitting across those or promoting to a standalone commit; edd6d4ab fixes a race introduced in 271a129b, 13 commits before where it currently lands; 34fac3fa bundles a pre-existing `as u16` fix that deserves its own commit; and the three "Update ... test payment IDs" fixups plus bbbac12e look ready to fold. There's still no CHANGELOG entry despite the breaking API changes (removed per-call fee-limit params, new `Config` field, new `PaymentKind` field).
>
> Docs nits, in passing: the "may instead arrive over pre-existing channels" sentence (`bolt11.rs:582`, `:601`, `:645`, `:664`) is only true for BOLT12 — the BOLT11 invoice's sole route hint is the intercept SCID; `receive_via_jit_channel` now fails when `expiry_secs` exceeds every LSP's advertised validity, which callers passing long expiries will notice and the docs don't mention; "thousands of a satoshi" (`store.rs:540`, `:560`) should be "thousandths"; and `Bolt12Payment::receive` doesn't mention that the LSP fee is deducted from the received amount.
>
> From the first round, I still think the startup abort on unreadable cache state (`src/liquidity/mod.rs:273`), the flat fee cap silently overriding the negotiated ppm for BOLT11 variable-amount invoices, and the worst-case blocking latency of sync JIT calls are worth reconsidering, but I won't re-litigate them here.

src/liquidity/client/lsps2/state.rs
tnull: Fixed.

src/payment/store.rs
tnull: Done.

src/payment/store.rs
tnull: Done.

src/liquidity/client/lsps2/state.rs
tnull: Done.

tests/integration_tests_rust.rs
tnull: Done.

src/liquidity/client/lsps2/router.rs
tnull: Now added a freestanding commit for this as I see it as an additioanal optimization. It's also a bit annoying as we now pull in a graph reference to the router, and have to replicate the LDK selection logic here. I previously omitted it.

src/liquidity/client/lsps2/mod.rs
tnull: Done in fixup.

src/payment/bolt12/offers.rs
tnull: Hmm, previously avoided that to not add yet another dataset that could become a DoS vector, though it seems with appropriate limits it's not materially worse. Will now add a fixup.

src/event.rs
tnull: Hmm, this is a good point. I think we switched from gross to net reporting in #948 to keep the previous behavior. But it seems thereby we introduced a mismatch between `PaymentClaimable` and `PaymentClaimed` *and* we have a mismatch between BOLT11 and BOLT12 now. Now switched all to net accounting. Seems this might also be more intuitive? WDYT @jkczyz ?

tnull:
Rebased on latest main to resolve some conflicts.

src/event.rs
jkczyz: Yeah, using net (i.e., how much was actually received) makes sense to me.

Issue #932, opened 2026-06-11
Add locking for safe node failover


benthecarman:
We currently support postgres and VSS as remote persistence backends, which makes it possible to run LDK Node in environments where another process could restart the same node identity using the same persisted state.

The problem is that there is currently no ownership mechanism or locking for accessing the persisted state and we can't the 2 copies of the same node running at once. We need a way to ensure that exactly one process owns and mutates a node’s persisted state at a time, while allowing another instance to wait and take over quickly if the active instance crashes.

We should add a locking mechanism in our remote storage to allow for nodes to fail over to another instance if one crashes. Ideally this would also work for mobile use cases where if one mobile wallet is running, it stops another from spinning up until the other has closed.

joostjager:
What is the ldk-node-specific part of this? Once remote persistence exists, doesn’t every serious LDK integration need an ownership or lease primitive to ensure only one process mutates a given node identity and persisted state at a time?

joostjager:
I looked into locking in more detail. Nearly all of the lease logic belongs inside the store implementations. Since the PostgreSQL and VSS stores both live in ldk-node, this also seems like the sensible place to land these changes.

For rust-lightning specifically, we need to ensure that every critical external effect is preceded by the required persistence. It could also optionally preserve and propagate a specific lease-lost error, but that is not required for safety if the store freezes the node when it detects lease loss.

joostjager:
I am thinking of the following scope for an initial minimal implementation:

A. Scope the first implementation to ldk-node’s built-in PostgreSQL store. VSS and generic user-provided KVStore implementations remain out of scope.

B. Use an ordinary PostgreSQL lease table rather than advisory locks. Each named KV table gets a companion lease table containing one singleton lease row. Multiple node identities can share a database when they use distinct KV table names.

C. Acquire the lease before deserializing any node state. A fallback that cannot acquire it does not construct a standby Node. Its supervisor retries later by constructing a fresh node.

D. Treat PostgreSQL time and the lease row as authoritative. The lease has a random per-process owner ID, is renewed periodically, and ownership loss is terminal for that in-memory Node.

E. Fence every PostgreSQL write and removal in one transaction. The transaction locks the lease row, validates ownership and expiry, renews the lease, performs the KV mutation, and commits. Reads remain unfenced.

F. Do not treat API-entry lease checks as a safety boundary. Leadership can be lost immediately after such a check. Only the fenced PostgreSQL mutation establishes authority for a critical operation.

[...]

PR #1000, opened 2026-07-22
Add PostgreSQL node failover leases

joostjager:
Addresses #932 by adding minimal active/passive failover support for nodes using the built-in PostgreSQL store.

`PostgresStore` acquires an exclusive lease for its KV table, renews it in the background, and fences every mutation by validating and renewing the lease in the same transaction. Lease loss is terminal for the node and is exposed to the application (ldk-server) so the process can exit and restart from persisted state.

This initial implementation is intentionally limited to PostgreSQL. Generic stores and VSS are out of scope. The API, timing configuration, and backend support can be extended later if operational experience requires it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -110,6 +110,7 @@ mod util;
 mod wallet;
 
 use std::default::Default;
+use std::future::Future;
 use std::sync::{Arc, Mutex, RwLock};
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 #[cfg(cycle_tests)]
@@ -145,6 +146,7 @@ use fee_estimator::{
 use ffi::*;
 use gossip::GossipSource;
 use graph::NetworkGraph;
+use io::node_lease::NodeLease;
 use io::utils::update_and_persist_node_metrics;
 pub use lightning;
 use lightning::chain::BlockLocator;
@@ -267,6 +269,7 @@ pub struct Node {
 	payment_store: Arc<PaymentStore>,
 	lnurl_auth: Arc<LnurlAuth>,
 	is_running: Arc<RwLock<bool>>,
+	node_lease: Option<Arc<NodeLease>>,
 	node_metrics: Arc<PersistedNodeMetrics>,
 	om_mailbox: Option<Arc<OnionMessageMailbox>>,
 	async_payments_role: Option<AsyncPaymentsRole>,
@@ -277,6 +280,40 @@ pub struct Node {
 }
 
 impl Node {
+	fn node_lease_is_lost_or_expired(&self) -> bool {
+		self.node_lease.as_ref().is_some_and(|node_lease| {
+			if node_lease.renewal_deadline_elapsed() {
+				node_lease.mark_lost();
+			}
+			node_lease.is_lost()
+		})
+	}
+
+	#[cfg(feature = "postgres")]
+	pub(crate) fn install_node_lease(&mut self, node_lease: Arc<NodeLease>) -> bool {
+		debug_assert!(self.node_lease.is_none());
+		let lease_stop_sender = self.stop_sender.clone();
+		let lease_runtime = Arc::downgrade(&self.runtime);
+		let lease_peer_manager = Arc::downgrade(&self.peer_manager);
+		let lease_logger = Arc::downgrade(&self.logger);
+		// Store fencing is the safety boundary; this only limits activity until process exit.
+		node_lease.set_loss_handler(Box::new(move || {
+			if let Some(logger) = lease_logger.upgrade() {
+				log_error!(logger, "PostgreSQL node lease lost, stopping background processing");
+			}
+			lease_stop_sender.send_replace(());
+			if let Some(runtime) = lease_runtime.upgrade() {
+				runtime.abort_background_processor_task();
+			}
+			if let Some(peer_manager) = lease_peer_manager.upgrade() {
+				peer_manager.disconnect_all_peers();
+			}
+		}));
+
+		self.node_lease = Some(node_lease);
+		!self.node_lease_is_lost_or_expired()
+	}
+
 	/// Starts the necessary background tasks, such as handling events coming from user input,
 	/// LDK/BDK, and the peer-to-peer network.
 	///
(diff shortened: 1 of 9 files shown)


joostjager:
I’m primarily looking for a concept ACK on the overall direction at this stage. The code has already gone through several iterations, and it will probably need another pass once we agree that the lease, fencing, and process-exit approach is the right path.

`ldk-server` branch I've been testing with: https://github.com/lightningdevkit/ldk-server/compare/main...joostjager:ldk-server:postgres-node-failover

benthecarman:
Would be nice if this is abstracted out to also work with VSS. That way multi-device with a VSS backend can work

joostjager:
The safety-critical part seems mostly backend-specific to me, since lease validation and the KV mutation must be atomic within the backend. What layer of this do you envision abstracting?

If generalizing this later would not require much rewriting, I would prefer to leave any necessary code movement to a future VSS lease PR.

joostjager:
Friendly ping @tnull. I would like to agree on the overall approach before doing further implementation cleanup. If you do not currently have the bandwidth to take a look, please let me know too.

tnull:
> Friendly ping @tnull. I would like to agree on the overall approach before doing further implementation cleanup. If you do not currently have the bandwidth to take a look, please let me know too.

I'll try to get back to a first look some time this week, but review on PRs for v0.8 has priority right now.

I think to begin with my first question is whether we don't need a more general version of this anyways, i.e., something a `Node` wrapper that allows to safely shutdown/restart the node when connectivity drops and we need to safely restart the node on persistence failures.

joostjager:
>whether we don't need a more general version of this anyway

For an initial version, I think rebuilding can remain the responsibility of the infrastructure, for example Kubernetes. ldk-node reports the lease lost, ldk-server exits, and the infrastructure restarts the process.

The wrapper you propose seems quite a bit broader, and I do not think it is needed to meet the initial objective of failover? As mentioned in https://github.com/lightningdevkit/ldk-node/pull/1000#issuecomment-5055146554, making lease+mutation atomic is backend-specific, so a general wrapper would not replace that work.

joostjager:
Pushed a few AI fixes in the mean time. Once we agree on general direction and scope, I can go in and make this PR merge ready.

tnull:
Moved this to the v0.9 milestone. It seems we need more time here for general architectural discussions, but even out side of these "big picture" questions (whether to find a solution that also incorporates VSS or not etc) it seems that this PR still has a lot of edge cases uncovered. E.g: 

>   - External-effect window: a process can pause after a fenced persistence commit, lose its lease, then resume and send the already-authorized peer message or broadcast. PostgreSQL cannot fence Lightning peers or Bitcoin broadcasts.
>   - Non-PostgreSQL stores: VSS and custom KVStore implementations have no lease/fencing support.
>   - Non-participating processes: older LDK Node versions, direct SQL writers, or other implementations can ignore the lease.
>   - Copied state: the same node state duplicated into another table, database, or backend gets a separate lease and can still produce duplicate active nodes.
>   - Application containment: safe process termination depends on the application observing wait_for_lease_loss() and exiting correctly.
>   - Generic builder path: constructing PostgresStore manually and passing it through build_with_store retains write fencing but does not install the node-level shutdown/notification handler.
>   - Unusual schema resolution: normal qualified/unqualified names are substantially improved, but physical table identity and unusual search_path configurations still deserve explicit integration tests.
>   - Database-level data loss: promotion of an asynchronously replicated PostgreSQL standby can lose both recent lease and node-state commits; application-level leasing cannot compensate for that.
>   - Planned handoff ergonomics: Node::stop() deliberately retains the lease; takeover requires dropping the node/process.
>   - Stale reads/API calls: reads are unfenced, and application threads may continue calling APIs after loss, although mutations should fail.
>   - Current implementation readiness: the present PR head also has a non-Postgres build failure and needs cleanup plus broader failure-injection testing before merging.

joostjager:
> Moved this to the v0.9 milestone. It seems we need more time here for general architectural discussions, but even out side of these "big picture" questions (whether to find a solution that also incorporates VSS or not etc) it seems that this PR still has a lot of edge cases uncovered. E.g:

My review request was explicitly for feedback on the general direction and proposed scope. Many of the items in this list restate scope boundaries and threat-model assumptions that were already called out, rather than identifying newly uncovered implementation edge cases.

Within the supported configuration, I do not currently see a concrete safety violation in the list. Mutations are atomically fenced by the lease. Some other scenarios may cause stale reads, duplicate network activity, reduced availability, or unsupported behavior, but it is not clear which one results in state corruption or funds risk.

Upgrades were not previously a compatibility concern because PostgreSQL support had not shipped. If #1012 ships first, that changes: its advisory lock and this PR's lease row do not coordinate.

tnull:
> Upgrades were not previously a compatibility concern because PostgreSQL support had not shipped. If #1012 ships first, that changes: its advisory lock and this PR's lease row do not coordinate.

Hmm? Why can't we revert #1012 in v0.9, worst case with a migration step?

joostjager:
> Hmm? Why can't we revert #1012 in v0.9, worst case with a migration step?

The migration step indeed. How to migrate from one type of lock to another.

joostjager:
I’ve rebased this on the latest main, which includes the temporary advisory lock from #1012 and a migration path to safer locking.

This PR is still blocked on a concept ack. I outlined the intended scope in https://github.com/lightningdevkit/ldk-node/pull/1000#issuecomment-5091325955. @tnull, could you elaborate on how you would like to deviate from or extend that scope (for this PR only).

src/lib.rs
```diff
@@ -900,6 +952,19 @@ impl Node {
...
 
+	/// Returns a future which completes when this node loses its PostgreSQL lease.
+	///
+	/// Returns [`None`] when the node was not built with the built-in PostgreSQL store.
+	/// Lease loss is terminal for this [`Node`]. The caller must immediately terminate the process
+	/// without calling [`Node::stop`] or dropping the node. A new process may reconstruct the node
+	/// from persistence. Store fencing cannot revoke an external effect from a process that pauses
+	/// after a fenced mutation, so every critical external effect must immediately follow one and
+	/// finish within that mutation's renewed lease interval.
+	pub fn wait_for_lease_loss(&self) -> Option<impl Future<Output = ()> + Send + 'static> {
```
tnull, line 963: I still don't think we'd want something like this in the public LDK Node API. To properly deal with lease losses, we'll need to refactor `Node` and handle the loss recovery/restarts internally. FWIW, this would also help with cascading persistence failures. 

Currently we can end up in a zombie state where we hit a `PersistenceFailed` but then continue the node operation, which leads to us having to add recovery/replay/retry logic all over the place. I think architecturally we should rather consider towards an STM pattern: some operation in `Node` hits `PersistenceFailed` or we lose the lease -> clean shutdown/stop -> try restart.

joostjager: How about if we first replace #1012's temporary mechanism with the leases and atomic fencing in this PR, but retain the panic when we lose the lease (no `wait_for_lease_loss`)? And then follow up with internal recovery, which is a bigger refactor.

tnull: Yes, converging now on a lease design that we know we want to have going forward seems like a good middle ground, to avoid running into any migration issues. Mind updating the PR to only switch the approach (also needs a rebase)?

joostjager: I'll work it out in a new one so that we can keep the relevant discussion about how to proceed with the lifecycle here.

https://github.com/lightningdevkit/ldk-node/pull/1111

PR #1004, opened 2026-07-24
Probing backround service follow-up.

### LINKED ISSUE #975: Probing service follow-ups
In #815 we pointed out some minor nits/comments before merging: https://github.com/lightningdevkit/ldk-node/pull/815#pullrequestreview-4662391033

We should address them in a follow-up.

(cc @randomlogin)

randomlogin:
Closes: https://github.com/lightningdevkit/ldk-node/issues/975

1. Move probing code in builder before the memory leak checker.

2. Prevent potential overflow due to public channel info fee values.

3. Account for aggregated fee potentially being bigger than constructed probing path's max htlc bound (doing second pass once we constructed path with fees).

### DIFF
--- a/src/probing.rs
+++ b/src/probing.rs
@@ -595,7 +595,7 @@ impl RandomWalkStrategy {
 			// Retrieve the direction-specific update via the public ChannelInfo fields.
 			// as_directed_from already checked both directions are Some, but we break
 			// defensively rather than unwrap.
-			let update = match if directed.source() == &next_channel.node_one {
+			let channel_update_info = match if directed.source() == &next_channel.node_one {
 				next_channel.one_to_two.as_ref()
 			} else {
 				next_channel.two_to_one.as_ref()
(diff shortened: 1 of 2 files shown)


src/probing.rs
```diff
@@ -682,25 +682,36 @@ impl RandomWalkStrategy {
...
 			};
-			let fee = update.fees.base_msat as u64
-				+ (forwarded * update.fees.proportional_millionths as u64 / 1_000_000);
-			forwarded += fee;
+			// Overflow prevention
+			let proportional_fee = forwarded
+				.checked_mul(channel_update_info.fees.proportional_millionths as u64)
+				.map(|v| v / 1_000_000)?;
+			let fee = (channel_update_info.fees.base_msat as u64).checked_add(proportional_fee)?;
+			forwarded = forwarded.checked_add(fee)?;
```
tnull, line 707: Codex:

> - [P2] Check the HTLC bound before adding the current hop’s fee — /home/tnull/worktrees/ldk-node/pr-1004-latest-20260724/src/probing.rs:698
> 
>     forwarded initially represents the amount sent through next_channel, including only downstream fees. The code adds the current node’s fee and then compares against that channel’s bounds. That fee is collected on the preceding channel and must not count toward
>     next_channel.htlc_maximum_msat.
> 
>     For example, if B→C has a 1,000,000-msat maximum and B charges 1,000 msat, delivering exactly 1,000,000 msat over B→C is valid. This code checks 1,001,000 and rejects the path.
> 
>     BOLT 7 defines the maximum as what the origin node will send through that channel, while its fee is calculated from amount_to_forward. BOLT 7 channel updates and HTLC fees (https://github.com/lightning/bolts/blob/master/07-routing-gossip.md#htlc-fees)
> 
>     The bound check should occur before calculating and adding fee.

tnull (commented):
One comment otherwise LGTM

src/probing.rs
randomlogin: fixed

tnull (approved):

PR #1006, opened 2026-07-27
bitcoind: fix raw TX fetching when using REST

jharveyb:
I noticed this issue while testing switching my nodes from the Electrum ChainSource to bitcoind.

An alternate fix here would also be to fetch the TX as binary instead of JSON - I can put up a PR for that if that's preferred.


tnull (commented):
Thanks, good catch! 

While you're here, mind adding a second commit that changes the range in `random_chain_source` to also cover the REST arm (i.e. should use 0..4 I believe), which would give us improved test coverage for the REST chain source?

tnull (approved):
Thanks! 

The CI change showed that we have another bug in the REST client implementation. Landing this and working on a fix for it.

PR #1008, opened 2026-07-30
Bump LDK to 0.3.0-beta1

tnull:
Track the first LDK 0.3 beta and its matching payment-instructions revision. Preserve Esplora configuration across the client versions and expose splice negotiation only to local contributors, as defined by the beta API.

Co-Authored-By: HAL 9000

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -196,9 +196,7 @@ pub enum Event {
 		/// transaction.
 		claim_from_onchain_tx: bool,
 		/// The final amount forwarded, in milli-satoshis, after the fee is deducted.
-		///
-		/// The caveat described above the `total_fee_earned_msat` field applies here as well.
-		outbound_amount_forwarded_msat: Option<u64>,
+		outbound_amount_forwarded_msat: u64,
 	},
 	/// A payment for a previously-registered payment hash has been received.
 	///
@@ -271,8 +269,10 @@ pub enum Event {
 		/// This will be `None` for events serialized by LDK Node v0.2.1 and prior.
 		reason: Option<ClosureReason>,
 	},
-	/// A channel splice has been negotiated and the funding transaction is pending
-	/// confirmation on-chain.
+	/// A channel splice with local inputs or outputs has been negotiated and the funding
+	/// transaction is pending confirmation on-chain.
+	///
+	/// This event is not emitted when only the counterparty contributes to a splice.
 	SpliceNegotiated {
 		/// The `channel_id` of the channel.
 		channel_id: ChannelId,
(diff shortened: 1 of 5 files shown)


src/event.rs
```diff
@@ -271,8 +271,10 @@ pub enum Event {
 		/// This will be `None` for events serialized by LDK Node v0.2.1 and prior.
 		reason: Option<ClosureReason>,
 	},
-	/// A channel splice has been negotiated and the funding transaction is pending
-	/// confirmation on-chain.
+	/// A channel splice with local inputs or outputs has been negotiated and the funding
+	/// transaction is pending confirmation on-chain.
+	///
+	/// This event is not emitted when only the counterparty contributes to a splice.
```
jkczyz, line 275: Same for `SpliceNegotiationFailed`?

src/event.rs
```diff
@@ -1545,7 +1547,7 @@ where
 					total_fee_earned_msat,
 					skimmed_fee_msat,
 					claim_from_onchain_tx,
-					outbound_amount_forwarded_msat,
+					outbound_amount_forwarded_msat: Some(outbound_amount_forwarded_msat),
```
jkczyz: Should we remove the `Option` from the LDK Node event, too?

src/event.rs
tnull: Done.

src/event.rs

jkczyz (approved):

src/event.rs
```diff
@@ -196,9 +196,7 @@ pub enum Event {
 		/// transaction.
 		claim_from_onchain_tx: bool,
 		/// The final amount forwarded, in milli-satoshis, after the fee is deducted.
-		///
-		/// The caveat described above the `total_fee_earned_msat` field applies here as well.
-		outbound_amount_forwarded_msat: Option<u64>,
+		outbound_amount_forwarded_msat: u64,
```
joostjager, line 199: AI: This still exposes only the aggregate forwarded amount. LDK 0.3 adds `amount_msat` to each `HTLCLocator` so callers can attribute multi-HTLC and trampoline forwards to individual channels, but our `From<LdkHtlcLocator>` conversion drops it. Could we add `amount_msat: Option<u64>` to the local locator, copy it during conversion, and persist it under TLV 7? Making it optional preserves compatibility with existing serialized events.

joostjager: Just a drive-by, can be done any time in a later PR

tnull: > Just a drive-by, can be done any time in a later PR

Said follow-up: https://github.com/lightningdevkit/ldk-node/pull/1021

PR #1012, opened 2026-08-03
Lock Postgres stores on initialization

benthecarman:
If we aren't going to get #1000 in before the release, we can add a simple lock on the postgres using its native `pg_try_advisory_lock`. 

Prevent multiple nodes from opening the same PostgreSQL database table at once while allowing separate database and table pairs to coexist. Retain the session-scoped advisory lock for the store lifetime.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -696,6 +696,14 @@ impl NodeBuilder {
 	/// The given `kv_table_name` will be used or default to
 	/// [`DEFAULT_KV_TABLE_NAME`](io::postgres_store::DEFAULT_KV_TABLE_NAME).
 	///
+	/// # Warning
+	///
+	/// Do not point multiple [`Node`] instances at the same database and table. Concurrent access is
+	/// unsafe and can corrupt node state. You must make sure that only one node accesses each
+	/// database and table. The store uses a PostgreSQL advisory lock to reduce this risk. This lock
+	/// is only a temporary safeguard and does not make concurrent access safe.
+	/// Nodes using a different database or table on the same server may coexist.
+	///
 	/// If `certificate_pem` is `Some`, TLS will be used for database connections and the
 	/// provided PEM-encoded CA certificate will be added to the system's default root
 	/// certificates (it does not replace them). If `certificate_pem` is `None`, connections
@@ -1229,6 +1237,14 @@ impl ArcedNodeBuilder {
 	///
 	/// The given `kv_table_name` will be used or default to
 	/// [`DEFAULT_KV_TABLE_NAME`](io::postgres_store::DEFAULT_KV_TABLE_NAME).
+	///
+	/// # Warning
+	///
+	/// Do not point multiple [`Node`] instances at the same database and table. Concurrent access is
+	/// unsafe and can corrupt node state. You must make sure that only one node accesses each
+	/// database and table. The store uses a PostgreSQL advisory lock to reduce this risk. This lock
+	/// is only a temporary safeguard and does not make concurrent access safe.
+	/// Nodes using a different database or table on the same server may coexist.
 	///
 	/// If `certificate_pem` is `Some`, TLS will be used for database connections and the
 	/// provided PEM-encoded CA certificate will be added to the system's default root
--- a/src/io/postgres_store/mod.rs
+++ b/src/io/postgres_store/mod.rs
@@ -11,6 +11,7 @@ use std::future::Future;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{Arc, Mutex};
 
+use bitcoin::hashes::{sha256, Hash, HashEngine};
 use lightning::io;
 use lightning::util::persist::{
 	KVStore, MigratableKVStore, PageToken, PaginatedKVStore, PaginatedListResponse,
@@ -44,6 +45,18 @@ const PAGE_SIZE: usize = 50;
 // Keep this small while still allowing progress if one runtime worker blocks on sync store access.
 const INTERNAL_RUNTIME_WORKERS: usize = 2;
 
+fn advisory_lock_id(db_name: &str, kv_table_name: &str) -> i64 {
+	let mut engine = sha256::Hash::engine();
+	engine.input(b"ldk-node:postgres-store");
+	for component in [db_name, kv_table_name] {
+		engine.input(&(component.len() as u64).to_be_bytes());
+		engine.input(component.as_bytes());
+	}
+
+	let hash = sha256::Hash::from_engine(engine).to_byte_array();
+	i64::from_be_bytes(hash[..8].try_into().expect("SHA-256 prefix has the expected length"))
+}
+
 fn sql_identifier(identifier: &str) -> io::Result<String> {
 	if identifier.is_empty() || identifier.contains('\0') {
 		return Err(io::Error::new(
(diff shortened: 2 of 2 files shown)


src/io/postgres_store/mod.rs
```diff
@@ -373,6 +388,9 @@ impl MigratableKVStore for PostgresStore {
 
 struct PostgresStoreInner {
 	pool: SmallPool,
+	// PostgreSQL advisory locks are session-scoped, so keep the connection that acquired our lock
+	// alive for the lifetime of the store.
+	_lock_client: ClientConnection,
```
joostjager: **[P1] Fail closed when the lock session disconnects**

This client is retained but never monitored. If its PostgreSQL session ends, the advisory lock is released while the independent pool can reconnect and continue serving operations. A second store can then acquire the lock while this store resumes writing. Please treat lock-session loss as terminal before any further operation, or otherwise reacquire and validate ownership without allowing stale writes. A regression test should terminate this backend, start a replacement store, and verify that the original store cannot operate.

src/io/postgres_store/mod.rs
```diff
@@ -426,6 +444,23 @@ impl PostgresStoreInner {
 		Self::create_database_if_not_exists(&config, &tls, logger.as_deref()).await?;
 
 		let client = make_config_connection(&config, &tls).await?;
+		let lock_id = advisory_lock_id(&db_name, &kv_table_name);
```
joostjager, line 447: **[P1] Derive the lock from the canonical schema and table identity**

This hashes the configured table string, so `ldk_data` and `public.ldk_data` normally produce different lock IDs despite resolving to the same physical table. Both stores can consequently initialize and write concurrently.

Please parse the table into optional schema and table components, resolve an omitted schema using `current_schema()`, and derive the lock from the actual database OID, schema OID, and table component. This identity is available before the table exists, allowing the lock to remain ahead of table creation and persisted-state reads. Please also reject identifier components exceeding PostgreSQL's `max_identifier_length`, since PostgreSQL otherwise truncates them and can make distinct strings resolve to the same relation.

Add an integration test opening the same table through qualified and unqualified names and verify that the second store receives `AlreadyExists`.

--
Not sure if we need to go this far... it is easy to oversee though when code changes

benthecarman: yeah this seems like total overkill

tnull:
@joostjager Mind sharing again why you think this PR is not sufficient for a first stopgap measure for v0.8/LDK Server v0.1?

From a first impression it *looks* like the kind of minimally invasive thing we want to land right now, before we can then take time to discuss the right approach / review #1000 etc.

joostjager:
The reason it is not sufficient is above in https://github.com/lightningdevkit/ldk-node/pull/1012#discussion_r3702709588. Maybe there are other edge cases that can happen with a connection-level lock.

If it is not possible to get the safer fix in and postgres must be included in the release, the advisory lock is of course better than no lock.

tnull:
Let's move forward with this minimalistic approach for v0.8, and do the more thorough #1000 in the v0.9 release cycle.

src/builder.rs
```diff
@@ -695,6 +695,8 @@ impl NodeBuilder {
 	///
 	/// The given `kv_table_name` will be used or default to
 	/// [`DEFAULT_KV_TABLE_NAME`](io::postgres_store::DEFAULT_KV_TABLE_NAME).
+	/// Building fails if another PostgreSQL-backed node using the same database and table is still
```
tnull: Let's make these docs a more explicit warning that it's generally discouraged and considered unsafe to point multiple instances towards the same backend. It should be clear that it's the users responsibility to avoid this for now, while we do have some stop gaps in place.

src/io/postgres_store/mod.rs
joostjager: If we go ahead with advisory locks for now, maybe some form of monitoring can be added?

benthecarman: Yeah if we're going with this, will add it so we attempt to reacquire the lock if it drops

src/builder.rs
benthecarman: done

benthecarman:
Added it so it'll reacquire the lock if we lose it.

joostjager:
Do we already need to think how we can upgrade from this intermediary solution later?

src/io/postgres_store/mod.rs
```diff
@@ -589,9 +639,61 @@ impl PostgresStoreInner {
...
+
+	async fn revalidate_store_lock(&self) -> io::Result<()> {
+		let mut lock_client = self.lock_client.lock().await;
+		if !lock_client.is_closed() && lock_client.simple_query("SELECT 1").await.is_ok() {
+			return Ok(());
+		}
+		self.reacquire_store_lock(&mut lock_client).await
+	}
+
+	async fn reacquire_store_lock(&self, lock_client: &mut ClientConnection) -> io::Result<()> {
```
joostjager: The whole concept of reacquire sounds unsafe. If the node continued running with it's in-memory state, and another instance had the lock in the mean time and wrote data, we have a problem? #1000 hard-exits in this case, and a process restart can reload the latest state then.

benthecarman: Changed so we panic on loss of the lock

src/io/postgres_store/mod.rs
```diff
@@ -128,6 +159,13 @@ impl PostgresStore {
 	/// the default `postgres` database to create it.
 	///
 	/// The given `kv_table_name` will be used or default to [`DEFAULT_KV_TABLE_NAME`].
+	/// # Warning
```
tnull, line 163: nit: This needs a whitespace line above.

src/io/postgres_store/mod.rs
```diff
@@ -999,6 +1081,48 @@ mod tests {
...
+		let locking_key = store.build_locking_key("test_ns", "test_sub", "key");
+		let inner_lock_ref = store.inner.get_inner_lock_ref(locking_key);
+		let inner_lock = inner_lock_ref.lock().await;
+
+		let write_task =
+			tokio::spawn(KVStore::write(&store, "test_ns", "test_sub", "key", vec![1u8]));
+		// Ensure the public method's initial check has passed and the internal write is queued on
+		// the per-key lock before closing the store lock connection.
+		tokio::time::sleep(std::time::Duration::from_millis(10)).await;
+		assert!(!write_task.is_finished());
```
tnull: Codex:

>   2. [P2] Queued-write test does not prove the recheck occurred.
>      /home/tnull/worktrees/ldk-node/pr-1012-review-20260810/src/io/postgres_store/mod.rs:1105 sleeps for 10 ms and treats !is_finished() as evidence that the task passed the initial assertion. An unpolled task also satisfies that condition, then panics at the initial check after the connection
>      is killed. Thus the test can pass without the intended inner recheck. It needs deterministic synchronization after the initial check.

tnull (commented):
Mostly looks good, just 1-2 nits.

src/io/postgres_store/mod.rs
benthecarman: added

src/io/postgres_store/mod.rs
benthecarman: fixed

tnull (approved):

PR #1021, opened 2026-08-05
Expose per-HTLC forwarded amounts

tnull:
Preserve LDK's per-HTLC amount when forwarding events cross the LDK Node API boundary. This lets callers attribute multi-HTLC forwards to individual channels.

Store the amount in an optional TLV so previously serialized events continue to decode without it.

Co-Authored-By: HAL 9000

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -68,6 +68,10 @@ use crate::{
 pub struct HTLCLocator {
 	/// The channel that the HTLC was sent or received on.
 	pub channel_id: ChannelId,
+	/// The amount, in milli-satoshis, of the HTLC that was sent or received, if known.
+	///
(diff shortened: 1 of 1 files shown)


src/event.rs
```diff
@@ -2027,6 +2035,56 @@ mod tests {
...
+
+		assert_eq!(
+			HTLCLocator::from(ldk_locator).encode(),
+			expected_encoding,
+			"converted HTLC locator should preserve its amount"
+		);
+	}
+
+	#[test]
+	fn htlc_locator_defaults_amount_for_legacy_encoding() {
```
joostjager: These tests look like tlv macro testing and not needed?

tnull: Dropped.

joostjager (approved):

PR #1022, opened 2026-08-05
Follow reorgs with bitcoind REST

tnull:
Bitcoin Core's REST headers endpoint cannot serve headers from a stale branch. This prevented REST-backed nodes from finding a common ancestor and following the replacement chain after a reorg.

In #1006 we enabled bitcoind REST chain source support in CI, which exposed this failure.

Co-Authored-By: HAL 9000

### DIFF
--- a/src/chain/bitcoind.rs
+++ b/src/chain/bitcoind.rs
@@ -1280,8 +1280,13 @@ impl BlockSource for BitcoindClient {
 				BitcoindClient::Rpc { rpc_client, .. } => {
 					rpc_client.get_header(header_hash, height_hint).await
 				},
-				BitcoindClient::Rest { rest_client, .. } => {
-					rest_client.get_header(header_hash, height_hint).await
+				BitcoindClient::Rest { rest_client, rpc_client, .. } => {
+					match rest_client.get_header(header_hash, height_hint).await {
+						Err(e) if e.kind() == BlockSourceErrorKind::Persistent => {
+							rpc_client.get_header(header_hash, height_hint).await
+						},
+						result => result,
+					}
 				},
 			}
 		}
--- a/src/chain/bitcoind.rs
+++ b/src/chain/bitcoind.rs
@@ -49,8 +49,12 @@ use crate::{Error, PersistedNodeMetrics};
 const CHAIN_POLLING_INTERVAL_SECS: u64 = 2;
 const CHAIN_POLLING_TIMEOUT_SECS: u64 = 10;
 
+type BitcoindSpvClient =
+	SpvClient<ChainPoller<Arc<BitcoindClient>, BitcoindClient>, Arc<ChainListener>>;
+
 pub(super) struct BitcoindChainSource {
 	api_client: Arc<BitcoindClient>,
+	spv_client: tokio::sync::Mutex<Option<BitcoindSpvClient>>,
 	latest_chain_tip: RwLock<Option<ValidatedBlockHeader>>,
 	wallet_polling_status: Mutex<WalletSyncStatus>,
 	fee_estimator: Arc<OnchainFeeEstimator>,
@@ -74,9 +78,11 @@ impl BitcoindChainSource {
 		));
 
 		let latest_chain_tip = RwLock::new(None);
+		let spv_client = tokio::sync::Mutex::new(None);
 		let wallet_polling_status = Mutex::new(WalletSyncStatus::Completed);
 		Self {
 			api_client,
+			spv_client,
 			latest_chain_tip,
 			wallet_polling_status,
 			fee_estimator,
@@ -103,10 +109,12 @@ impl BitcoindChainSource {
 		));
 
 		let latest_chain_tip = RwLock::new(None);
+		let spv_client = tokio::sync::Mutex::new(None);
 		let wallet_polling_status = Mutex::new(WalletSyncStatus::Completed);
 
 		Self {
 			api_client,
+			spv_client,
 			latest_chain_tip,
 			wallet_polling_status,
 			fee_estimator,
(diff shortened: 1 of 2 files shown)


TheBlueMatt:
Hmm, that's somewhat surprising. Can you file an issue on https://github.com/bitcoin/bitcoin as well?

tnull:
> Hmm, that's somewhat surprising. Can you file an issue on https://github.com/bitcoin/bitcoin as well?

Will do.

src/chain/bitcoind.rs
```diff
@@ -1280,8 +1280,13 @@ impl BlockSource for BitcoindClient {
 				BitcoindClient::Rpc { rpc_client, .. } => {
 					rpc_client.get_header(header_hash, height_hint).await
 				},
-				BitcoindClient::Rest { rest_client, .. } => {
-					rest_client.get_header(header_hash, height_hint).await
+				BitcoindClient::Rest { rest_client, rpc_client, .. } => {
+					match rest_client.get_header(header_hash, height_hint).await {
+						Err(e) if e.kind() == BlockSourceErrorKind::Persistent => {
+							rpc_client.get_header(header_hash, height_hint).await
```
joostjager: It looks a bit weird that we have to fallback to the other interface. From history it seems we need rest because rpc is too slow for large sync ops?

joostjager (commented):
I wonder why we need a chain backend to find a common ancestor. What happens if someone switches to a different backend that doesn't know about the stale branch at all?

src/chain/bitcoind.rs
tnull: REST was added both for its unauthenticated read-only interface and for more efficient/cacheable binary block retrieval; it isn’t simply that RPC is unusably slow.

tnull:
> I wonder why we need a chain backend to find a common ancestor. What happens if someone switches to a different backend that doesn't know about the stale branch at all?

 Good question. We mostly need the backend to resolve the stale branch here because we accidentally stopped retaining the header cache in #874 (though note after restart that cache is also empty, so the fallback still makes sense anyways). Before that change / https://github.com/lightningdevkit/rust-lightning/pull/4266, `BitcoindChainSource` kept one cache across polls. When `SpvClient` changed to own its `HeaderCache`, we started constructing it with a fresh cache every time and discarded the cache returned by initial synchronization.

I’ll see to restore cache reuse by retaining the `SpvClient` across polls.

 Switching backends happens across a restart. In that case, initial synchronization uses the recent hashes stored in `BlockLocator` and tries them until the new backend recognizes a common ancestor, so the new backend need not know the stale tip. This is bounded by the stored history, currently 12 blocks. Also cf. #963.

tnull:
Now updated to preserve the `SpvClient` across polls.

src/chain/bitcoind.rs
joostjager: Is the unauth interface an advantage if you also may need rpc?

tnull: It is to enable easier caching, but see your point.

src/chain/bitcoind.rs
```diff
@@ -1280,13 +1318,8 @@ impl BlockSource for BitcoindClient {
...
 				},
-				BitcoindClient::Rest { rest_client, rpc_client, .. } => {
-					match rest_client.get_header(header_hash, height_hint).await {
-						Err(e) if e.kind() == BlockSourceErrorKind::Persistent => {
-							rpc_client.get_header(header_hash, height_hint).await
-						},
-						result => result,
-					}
+				BitcoindClient::Rest { rest_client, .. } => {
+					rest_client.get_header(header_hash, height_hint).await
```
joostjager: In your comment, you said "though note after restart that cache is also empty, so the fallback still makes sense anyways", but you remove it anyway?

Or is the shared and initialized spv client all that is needed?

src/chain/bitcoind.rs
```diff
@@ -415,19 +432,24 @@ impl BitcoindChainSource {
...
+		if spv_client_lock.is_none() {
+			let latest_chain_tip_opt = self.latest_chain_tip.read().expect("lock").clone();
+			let chain_tip = if let Some(tip) = latest_chain_tip_opt {
+				tip
+			} else {
+				self.poll_chain_tip().await?
+			};
+			*spv_client_lock = Some(self.new_spv_client(
+				chain_tip,
+				HeaderCache::new(),
```
joostjager, line 445: Is it possible to hit this? And how is a stale branch walked backwards in that case?

src/chain/bitcoind.rs
tnull: I think the shared SPV client is the core fix, but there are edge cases where the fallback helps (belt-and-suspenders I guess):

>   There is a concrete race:
> 
>   1. REST chaininfo returns tip A.
>   2. Bitcoin Core reorganizes to tip B.
>   3. REST headers/A returns 404, classified as persistent.
>   4. During initial synchronization, that causes a 300-second backoff
> 
>   RPC can still resolve stale header A, allowing synchronization to complete; the retained SpvClient then handles B on the next poll. The fallback also helps beyond the 12-block restart locator window or when locator history is incomplete.

joostjager: Yes, but I mean your second commit removes the fallback again.

tnull: > Yes, but I mean your second commit removes the fallback again.

Ah, whoops, amended.

src/chain/bitcoind.rs
tnull: It is possible to hit this in a narrow startup race: `start` spawns the initial-sync future, but `sync_wallets` may run before that future is first polled and registers the pending sync. A persistent REST miss falls back to RPC, so stale headers remain available.

Fixing that narrow scheduling race would require coordinating registration with task spawning. Given that background synchronization corrects the state and the fallback remains available, I’d prefer to handle that some time separately rather than expand this reorg fix.

joostjager: Yes, fair

joostjager (approved):

PR #1024, opened 2026-08-06
Stop keeping all `DataStore` entries in-memory, add pagination

### LINKED ISSUE #998: Stop keeping the full payment history in memory
#269 tracks paginating `Node::list_payments`, and #959 adds that API. However, `PaymentStore` is still initialized with `read_all_objects`, and `DataStore` retains every `PaymentDetails` in a `HashMap` for the lifetime of the node. #960 makes the startup reads paginated, which avoids one unbounded list response, but it still reads and decodes the full payment namespace before the node is usable.

This means startup work and steady-state memory usage continue to grow with the complete payment history, even though callers retrieve that history page by page. This will become increasingly relevant for long-lived server deployments.

Related work:

- #269 tracks the caller-facing pagination requirement.
- #959 adds the paginated `Node::list_payments` API while continuing to serve values from memory.
- #960 reads persisted objects page by page at startup, but still materializes the full namespace in memory.
- [rust-lightning #4779](https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/issues/4779) tracks paginated key-and-value reads, but does not address retaining the complete payment history in memory.
- [ldk-server #240](https://github.com/lightningdevkit/ldk-server/pull/240) contains earlier related payment-listing work and was closed without merging.

tnull:
Alternative/prefactor to #959.

Fixes #998.

Draft for now until concept ack.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Pending
 
 ## Compatibility Notes
+- Migrating between storage backends does not preserve the relative creation order of
+  pre-existing payments, as the generic KV store migration copies entries in an unspecified
+  order. Expect the order in which `Node::list_payments` returns pre-existing payments to
+  change once after such a migration. Payment contents and completeness are unaffected.
 - Pending JIT-channel payments created before upgrading may fail after upgrade because the
   prior LSPS2 fee-limit state stored in `PaymentKind::Bolt11Jit` is not migrated.
 - Upgrading from LDK Node v0.1 is no longer supported if the event queue still contains
@@ -9,6 +13,13 @@
   `v0.1.0-alpha.0` before upgrading LDK Node.
 
 ## Feature and API updates
+- `Node::list_payments` is now paginated: it takes an optional `PageToken` and returns a
+  `PaymentDetailsPage` holding one page of payments, ordered from most recently created to
+  least recently created, plus the token for the next page. Ordering and page tokens come
+  from the configured storage backend, so a token stays valid across restarts. This replaces
+  the previous unpaginated `Node::list_payments`, and `Node::list_payments_with_filter` has
+  been removed; filter the returned pages instead.
+- `Node::payment` now returns a `Result`, as retrieving a payment may fail.
 - The Bitcoin Core RPC and REST chain-source builder methods now accept an optional
   `wallet_rescan_from_height` argument. Passing a height lets fresh wallets rescan from a known
   birthday block instead of checkpointing at the current tip, which is useful when restoring a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Pending
 
 ## Compatibility Notes
+- Migrating between storage backends does not preserve the relative creation order of
+  pre-existing payments, as the generic KV store migration copies entries in an unspecified
+  order. Expect the order in which `Node::list_payments` returns pre-existing payments to
+  change once after such a migration. Payment contents and completeness are unaffected.
 - Pending JIT-channel payments created before upgrading may fail after upgrade because the
   prior LSPS2 fee-limit state stored in `PaymentKind::Bolt11Jit` is not migrated.
 - Upgrading from LDK Node v0.1 is no longer supported if the event queue still contains
@@ -19,6 +23,13 @@
   `Event::PaymentClaimable`.
 
 ## Feature and API updates
+- `Node::list_payments` is now paginated: it takes an optional `PageToken` and returns a
+  `PaymentDetailsPage` holding one page of payments, ordered from most recently created to
+  least recently created, plus the token for the next page. Ordering and page tokens come
+  from the configured storage backend, and token lifetime follows that backend's guarantees.
+  This replaces the previous unpaginated `Node::list_payments`, and
+  `Node::list_payments_with_filter` has been removed; filter the returned pages instead.
+- `Node::payment` now returns a `Result`, as retrieving a payment may fail.
 - The Bitcoin Core RPC and REST chain-source builder methods now accept an optional
   `wallet_rescan_from_height` argument. Passing a height lets fresh wallets rescan from a known
   birthday block instead of checkpointing at the current tip, which is useful when restoring a
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -51,6 +51,7 @@ use crate::config::{
 	BitcoindRestClientConfig, Config, ElectrumSyncConfig, EsploraSyncConfig, HRNResolverConfig,
 	TorConfig, DEFAULT_ESPLORA_SERVER_URL, DEFAULT_LOG_FILENAME, DEFAULT_LOG_LEVEL,
 	DEFAULT_MAX_PROBE_AMOUNT_MSAT, DEFAULT_MIN_PROBE_AMOUNT_MSAT, PAYMENT_CACHE_CAPACITY,
+	PAYMENT_CACHE_WARMUP_COUNT,
 };
 use crate::connection::ConnectionManager;
 use crate::data_store::{KeepAllEntries, KeepLeastRecentlyUsed};
@@ -61,8 +62,8 @@ use crate::gossip::GossipSource;
 use crate::io::sqlite_store::SqliteStore;
 use crate::io::utils::{
 	open_or_migrate_fs_store, read_all_objects, read_event_queue,
-	read_external_pathfinding_scores_from_cache, read_network_graph, read_node_metrics,
-	read_output_sweeper, read_peer_info, read_scorer,
+	read_external_pathfinding_scores_from_cache, read_n_objects, read_network_graph,
+	read_node_metrics, read_output_sweeper, read_peer_info, read_scorer,
 };
 use crate::io::vss_store::VssStoreBuilder;
 use crate::io::{
@@ -1456,9 +1457,16 @@ fn build_with_store_internal(
 
 	let kv_store_ref = Arc::clone(&kv_store);
 	let logger_ref = Arc::clone(&logger);
-	let (node_metris_res, pending_payment_store_res, address_pool_res) =
-		runtime.block_on(async move {
+	let (payment_store_res, node_metris_res, pending_payment_store_res, address_pool_res) = runtime
+		.block_on(async move {
 			tokio::join!(
+				read_n_objects(
+					&*kv_store_ref,
+					PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
+					PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
+					PAYMENT_CACHE_WARMUP_COUNT,
+					Arc::clone(&logger_ref),
+				),
 				read_node_metrics(&*kv_store_ref, Arc::clone(&logger_ref)),
 				read_all_objects(
 					&*kv_store_ref,
@@ -1483,17 +1491,23 @@ fn build_with_store_internal(
 		},
 	};
 
-	// The payment store caches a bounded number of payments and reads the rest back on demand, so
-	// we start it empty rather than paying to read a node's entire payment history at startup only
-	// to immediately drop all but the most recent entries.
-	let payment_store = Arc::new(PaymentStore::new(
-		Vec::new(),
-		KeepLeastRecentlyUsed::new(PAYMENT_CACHE_CAPACITY),
-		PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE.to_string(),
-		PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE.to_string(),
-		Arc::clone(&kv_store),
-		Arc::clone(&logger),
-	));
+	let payment_store = match payment_store_res {
+		Ok(payments) => Arc::new(PaymentStore::new(
+			// The read hands us the newest payments first, while the cache treats the objects it
+			// is seeded with as increasingly recently used. Reverse them, so that the newest
+			// payment is the last one to be evicted rather than the first.
+			payments.into_iter().rev().collect(),
+			KeepLeastRecentlyUsed::new(PAYMENT_CACHE_CAPACITY),
+			PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE.to_string(),
+			PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE.to_string(),
+			Arc::clone(&kv_store),
+			Arc::clone(&logger),
+		)),
+		Err(e) => {
+			log_error!(logger, "Failed to read payment data from store: {}", e);
+			return Err(BuildError::ReadFailed);
+		},
+	};
 
 	let (chain_source, chain_tip_opt) = match chain_data_source_config {
 		Some(ChainDataSourceConfig::Esplora { server_url, headers, sync_config }) => {
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -51,6 +51,7 @@ use crate::config::{
 	BitcoindRestClientConfig, Config, ElectrumSyncConfig, EsploraSyncConfig, HRNResolverConfig,
 	TorConfig, DEFAULT_ESPLORA_SERVER_URL, DEFAULT_LOG_FILENAME, DEFAULT_LOG_LEVEL,
 	DEFAULT_MAX_PROBE_AMOUNT_MSAT, DEFAULT_MIN_PROBE_AMOUNT_MSAT, PAYMENT_CACHE_CAPACITY,
+	PAYMENT_CACHE_WARMUP_COUNT,
 };
 use crate::connection::ConnectionManager;
 use crate::data_store::{KeepAllEntries, KeepLeastRecentlyUsed};
@@ -61,8 +62,8 @@ use crate::gossip::GossipSource;
 use crate::io::sqlite_store::SqliteStore;
 use crate::io::utils::{
 	open_or_migrate_fs_store, read_all_objects, read_event_queue,
-	read_external_pathfinding_scores_from_cache, read_network_graph, read_node_metrics,
-	read_output_sweeper, read_peer_info, read_scorer,
+	read_external_pathfinding_scores_from_cache, read_n_objects, read_network_graph,
+	read_node_metrics, read_output_sweeper, read_peer_info, read_scorer,
 };
 use crate::io::vss_store::VssStoreBuilder;
 use crate::io::{
@@ -1440,17 +1441,25 @@ fn build_with_store_internal(
 
 	let kv_store_ref = Arc::clone(&kv_store);
 	let logger_ref = Arc::clone(&logger);
-	let (node_metris_res, pending_payment_store_res) = runtime.block_on(async move {
-		tokio::join!(
-			read_node_metrics(&*kv_store_ref, Arc::clone(&logger_ref)),
-			read_all_objects(
-				&*kv_store_ref,
-				PENDING_PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
-				PENDING_PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
-				Arc::clone(&logger_ref),
+	let (payment_store_res, node_metris_res, pending_payment_store_res) =
+		runtime.block_on(async move {
+			tokio::join!(
+				read_n_objects(
+					&*kv_store_ref,
+					PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
+					PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
+					PAYMENT_CACHE_WARMUP_COUNT,
+					Arc::clone(&logger_ref),
+				),
+				read_node_metrics(&*kv_store_ref, Arc::clone(&logger_ref)),
+				read_all_objects(
+					&*kv_store_ref,
+					PENDING_PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE,
+					PENDING_PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE,
+					Arc::clone(&logger_ref),
+				)
 			)
-		)
-	});
+		});
 
 	// Initialize the status fields.
 	let node_metrics = match node_metris_res {
@@ -1465,17 +1474,23 @@ fn build_with_store_internal(
 		},
 	};
 
-	// The payment store caches a bounded number of payments and reads the rest back on demand, so
-	// we start it empty rather than paying to read a node's entire payment history at startup only
-	// to immediately drop all but the most recent entries.
-	let payment_store = Arc::new(PaymentStore::new(
-		Vec::new(),
-		KeepLeastRecentlyUsed::new(PAYMENT_CACHE_CAPACITY),
-		PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE.to_string(),
-		PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE.to_string(),
-		Arc::clone(&kv_store),
-		Arc::clone(&logger),
-	));
+	let payment_store = match payment_store_res {
+		Ok(payments) => Arc::new(PaymentStore::new(
+			// The read hands us the newest payments first, while the cache treats the objects it
+			// is seeded with as increasingly recently used. Reverse them, so that the newest
+			// payment is the last one to be evicted rather than the first.
+			payments.into_iter().rev().collect(),
+			KeepLeastRecentlyUsed::new(PAYMENT_CACHE_CAPACITY),
+			PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE.to_string(),
+			PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE.to_string(),
+			Arc::clone(&kv_store),
+			Arc::clone(&logger),
+		)),
+		Err(e) => {
+			log_error!(logger, "Failed to read payment data from store: {}", e);
+			return Err(BuildError::ReadFailed);
+		},
+	};
 
 	let (chain_source, chain_tip_opt) = match chain_data_source_config {
 		Some(ChainDataSourceConfig::Esplora { server_url, headers, sync_config }) => {
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -53,6 +53,7 @@ use crate::config::{
 	DEFAULT_MAX_PROBE_AMOUNT_MSAT, DEFAULT_MIN_PROBE_AMOUNT_MSAT,
 };
 use crate::connection::ConnectionManager;
+use crate::data_store::KeepAllEntries;
 use crate::entropy::NodeEntropy;
 use crate::event::EventQueue;
 use crate::fee_estimator::OnchainFeeEstimator;
@@ -1491,6 +1492,7 @@ fn build_with_store_internal(
 	let payment_store = match payment_store_res {
 		Ok(payments) => Arc::new(PaymentStore::new(
 			payments,
+			KeepAllEntries,
 			PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE.to_string(),
 			PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE.to_string(),
 			Arc::clone(&kv_store),
@@ -1745,8 +1747,12 @@ fn build_with_store_internal(
 	};
 
 	let pending_payment_store = match pending_payment_store_res {
+		// NOTE: This store must keep all its entries in memory: the wallet scans it in full on
+		// every chain tip change and to resolve replaced transactions. It stays bounded anyway,
+		// as entries are removed once a payment is no longer pending.
 		Ok(pending_payments) => Arc::new(PendingPaymentStore::new(
 			pending_payments,
+			KeepAllEntries,
 			PENDING_PAYMENT_INFO_PERSISTENCE_PRIMARY_NAMESPACE.to_string(),
 			PENDING_PAYMENT_INFO_PERSISTENCE_SECONDARY_NAMESPACE.to_string(),
 			Arc::clone(&kv_store),
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -53,6 +53,7 @@ use crate::config::{
 	DEFAULT_MAX_PROBE_AMOUNT_MSAT, DEFAULT_MIN_PROBE_AMOUNT_MSAT,
 };
 use crate::connection::ConnectionManager;
+use crate::data_store::KeepAllEntries;
 use crate::entropy::NodeEntropy;
 use crate::event::EventQueue;
 use crate::fee_estimator::OnchainFeeEstimator;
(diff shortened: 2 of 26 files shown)


benthecarman (commented):
concept ack, this approach looks good to me

CHANGELOG.md
```diff
@@ -9,6 +13,13 @@
   `v0.1.0-alpha.0` before upgrading LDK Node.
 
 ## Feature and API updates
+- `Node::list_payments` is now paginated: it takes an optional `PageToken` and returns a
+  `PaymentDetailsPage` holding one page of payments, ordered from most recently created to
+  least recently created, plus the token for the next page. Ordering and page tokens come
+  from the configured storage backend, so a token stays valid across restarts. This replaces
+  the previous unpaginated `Node::list_payments`, and `Node::list_payments_with_filter` has
+  been removed; filter the returned pages instead.
+- `Node::payment` now returns a `Result`, as retrieving a payment may fail.
```
benthecarman, line 32: also async now

src/data_store.rs
```diff
@@ -40,143 +50,622 @@ pub(crate) enum DataStoreUpdateResult {
...
+	/// shift position and so be skipped or returned twice.
+	///
+	/// Note this deliberately does not hold the mutation lock across its reads: a listing must not
+	/// block every writer for the duration of a round trip to a remote backend. Objects created or
+	/// removed while paginating may or may not be observed.
+	///
+	/// Note also that a page may hold fewer objects than the backend's page size, because objects
+	/// removed between listing the keys and reading them are skipped. Iterate until
+	/// `next_page_token` is `None` rather than until a short page.
+	pub(crate) async fn list_page(
```
benthecarman, line 486: claude:

A single page can mix stale cache hits with fresh store reads — src/data_store.rs list_page

 The read guard is dropped after the cache peek, before read_missing runs. Objects served from the cache reflect the state at peek time; objects read from the store reflect a possibly-later state. The doc comment covers concurrent creation and removal but not this. One sentence would close it.

src/data_store.rs
```diff
@@ -396,6 +420,163 @@ where
 		self.contains(id).await
 	}
 
+	/// Returns a page of objects, ordered from most recently created to least recently created.
+	///
+	/// Pass `None` to start at the most recently created object, and the returned
+	/// [`DataStorePage::next_page_token`] to continue from where the previous call left off.
+	///
+	/// The ordering and the tokens are the storage backend's own: we hand its opaque token back to
+	/// it unchanged and never derive an order of our own. That keeps tokens valid across restarts
```
joostjager: Delegating ordering to the backend sounds risky. I also believe we already had an issue with list_all where the order wasn't maintained across migrations. Wondering what the reason is we don't have an ldk-node level creation id that is used for ordering?

src/data_store.rs
```diff
@@ -396,6 +420,163 @@ where
...
+	/// and across changes to our caching, and it is why an object updated mid-pagination cannot
+	/// shift position and so be skipped or returned twice.
+	///
+	/// Note this deliberately does not hold the mutation lock across its reads: a listing must not
+	/// block every writer for the duration of a round trip to a remote backend. Objects created or
+	/// removed while paginating may or may not be observed.
```
joostjager: Is this a premature optimization? Especially with pagination and the upcoming more efficient postgres list with values, this may not be necessary. If we can avoid "may or may not be observed" I think that would be good.

src/builder.rs
```diff
@@ -1472,20 +1465,17 @@ fn build_with_store_internal(
...
-			log_error!(logger, "Failed to read payment data from store: {}", e);
-			return Err(BuildError::ReadFailed);
-		},
-	};
+	// The payment store caches a bounded number of payments and reads the rest back on demand, so
+	// we start it empty rather than paying to read a node's entire payment history at startup only
+	// to immediately drop all but the most recent entries.
+	let payment_store = Arc::new(PaymentStore::new(
+		Vec::new(),
+		KeepLeastRecentlyUsed::new(PAYMENT_CACHE_CAPACITY),
```
joostjager: Does this PR already need to include caching of payments that are no longer pending, or is it something to consider later when we see that disk reads are unacceptable?

src/data_store.rs
```diff
@@ -40,143 +50,622 @@ pub(crate) enum DataStoreUpdateResult {
...
+	///
+	/// Note also that a page may hold fewer objects than the backend's page size, because objects
+	/// removed between listing the keys and reading them are skipped. Iterate until
+	/// `next_page_token` is `None` rather than until a short page.
+	pub(crate) async fn list_page(
+		&self, page_token: Option<PageToken>,
+	) -> Result<DataStorePage<SO>, Error> {
+		let response = PaginatedKVStore::list_paginated(
```
joostjager, line 489: AI: The bounded cache does not make pagination scalable for the default filesystem backend. At the pinned rust-lightning revision, `FilesystemStoreV2::list_paginated_impl` collects every key in the namespace and sorts the complete vector on every page request: https://github.com/lightningdevkit/rust-lightning/blob/9174965af9437196c527a9aa0df36bbcf050c8bb/lightning-persister/src/fs_store/v2.rs#L147-L216

Consequently, one page still requires O(N) temporary key memory and O(N log N) work, while traversing the complete history repeats that work for every page. Could this be fixed upstream, or should the scalability claim be explicitly limited for the filesystem backend?

joostjager: Me says: drop filesystem store in ldk-node

src/lib.rs
```diff
@@ -2194,13 +2195,26 @@ impl Node {
 		}
 	}
 
-	/// Retrieves all payments that match the given predicate.
+	/// Retrieves a page of payments, ordered from most recently created to least recently created.
+	///
+	/// Pass `None` to start at the most recently created payment, and the
+	/// [`PaymentDetailsPage::next_page_token`] returned by the previous call to continue from
+	/// where it left off. Ordering and pagination are backed by the configured storage backend, so
+	/// a token stays valid across restarts of the node.
```
joostjager: AI: This guarantee is stronger than the `PaginatedKVStore` contract. That trait only requires tokens to remain valid across calls for a reasonable timeframe; it does not require them to survive backend reconstruction or restart. `build_with_store` accepts arbitrary implementations, so forwarding an opaque token unchanged is not enough to guarantee this.

Could we either make restart-stable tokens an explicit custom-store requirement or qualify this statement for backends that provide that guarantee?

src/data_store.rs
```diff
@@ -40,143 +50,622 @@ pub(crate) enum DataStoreUpdateResult {
...
+						log_error!(
+							self.logger,
+							"Failed to deserialize object for key {}/{}/{}: {}",
+							&self.primary_namespace,
+							&self.secondary_namespace,
+							key,
+							e
+						);
+						set.abort_all();
+						return Err(Error::PersistenceFailed);
```
joostjager, line 588: AI: Because startup now validates only the newest 50 payments, an unreadable older payment can first surface here during pagination. Returning an error discards `response.next_page_token`, leaving the caller unable to identify or skip the bad entry and therefore unable to reach any older pages.

Should pagination expose the continuation token alongside entry-level failures, provide another recovery mechanism, or deliberately retain fail-at-startup behavior?

joostjager (commented):
Concept ack, but I'd split this PR in two parts. First do the disk read, and only in a follow up add caching if we are absolutely sure we need it.

tnull:
> Concept ack, but I'd split this PR in two parts. First do the disk read, and only in a follow up add caching if we are absolutely sure we need it.

Thanks, but I think the caching is an important part here, e.g. for VSS-based mobile wallets (but also server) not immediately hitting huge latency spikes when listing recent payments. I however did refrain from further optimizations (e.g. lazy look-ahead caching etc).

joostjager:
> Thanks, but I think the caching is an important part here, e.g. for VSS-based mobile wallets (but also server) not immediately hitting huge latency spikes when listing recent payments. I however did refrain from further optimizations (e.g. lazy look-ahead caching etc).

For server deployments, it seems unlikely to me that reading one page from disk is expensive enough to warrant caching.

For VSS, the individual object reads are already performed concurrently. Does that still cause unacceptable latency in practice?

Either way, I think this is easier to review and reason about as two changes: first land disk-backed pagination, then add caching separately.

tnull:
> Either way, I think this is easier to review and reason about as two changes: first land disk-backed pagination, then add caching separately.

I mean, if you add caching later, you'll have to re-review the same code twice as the whole design is now rearranged to make that fit?

joostjager:
> I mean, if you add caching later, you'll have to re-review the same code twice as the whole design is now rearranged to make that fit?

You'd have to re-arrange the commit stack, but still I think there is enough that doesn't need to be reviewed twice. If adding caching later requires a whole rewrite, I'd think there is something not quite right in the design. Generally caching should be something that can be added incrementally.

If reviewers are fine with a PR of this size, then it matters less. Although even then I still think it is safer to take small steps.

CHANGELOG.md
tnull: Still sync? Or maybe misunderstanding?

src/data_store.rs
tnull: Added a comment in fixup.

src/data_store.rs
tnull: > Delegating ordering to the backend sounds risky.

This is the pre-existing API contract of `PaginatedKVStore`, no?

src/data_store.rs
tnull: AI:

> Valid tradeoff, not a correctness defect. Holding the lock would block payment mutations during remote reads and still could not provide consistency across multiple pages or external writers.

src/builder.rs
tnull: Yes, as discussed below, I think it's one of the main features of the refactoring.

src/data_store.rs
tnull: Yeah, that seems a valid concern but nothing we can do about it here. Now opened https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/issues/4870

> Me says: drop filesystem store in ldk-node

We could at best drop the constructor, but we'll still allow users to construct with any store that matches the traits.

Me says: drop filesystem store ~~in ldk-node~~

src/lib.rs
tnull: Added a fixup to clarify.

tnull:
> If reviewers are fine with a PR of this size, then it matters less.

Yup, checked with @benthecarman and he's fine with it and will do most of the review now.

tnull:
Rebased to resolve some conflicts. Ready for review now.

src/lib.rs
```diff
@@ -2136,9 +2136,10 @@ impl Node {
 
 	/// Retrieve the details of a specific payment with the given id.
 	///
-	/// Returns `Some` if the payment was known and `None` otherwise.
-	pub fn payment(&self, payment_id: &PaymentId) -> Option<PaymentDetails> {
-		self.payment_store.get(payment_id)
+	/// Returns `Ok(Some(..))` if the payment was known and `Ok(None)` otherwise. Returns an error
+	/// if the payment could not be retrieved from the store.
+	pub fn payment(&self, payment_id: &PaymentId) -> Result<Option<PaymentDetails>, Error> {
+		self.runtime.block_on(self.payment_store.get(payment_id))
```
benthecarman, line 2142: For here, `list_payments` and `Bolt11Payment::claim_for_hash()`, they changed from panic-free in-memory reads to `Runtime::block_on`, which uses `tokio::task::block_in_place`. If an app calls them from async code on a `current_thread` tokio runtime, they now panic. The `send` paths already used `block_on` before this PR, so only these three methods are a regression.

benthecarman: This seems more like we should just block `current_thread` and only allow multi threaded ones tho. We kinda already have this requirement implicitly

src/wallet/mod.rs
```diff
@@ -401,13 +405,13 @@ impl Wallet {
 					// The payment already exists in the store at this point: `bump_fee_rbf` updates
 					// the payment store with the replacement txid before the next sync cycle, so we
 					// can safely fetch it here.
+					let stored_payment = self.payment_store.get(&payment_id).await?;
```
benthecarman, line 546: **Store read failure can drop wallet sync events for good** — `src/wallet/mod.rs:408` (plausible) `payment_store.get()` in the `TxReplaced` arm now reads through to the KVStore, and a transient failure aborts the rest of the event batch after the in-memory BDK wallet was already mutated — the dropped events never regenerate. Caveat: this failure class predates the PR (the write path had it); the PR only widens the trigger surface.

src/event.rs
```diff
@@ -801,7 +801,13 @@ where
 				..
 			} => {
 				let payment_id = PaymentId(payment_hash.0);
-				let payment_info = self.payment_store.get(&payment_id);
+				let payment_info = match self.payment_store.get(&payment_id).await {
+					Ok(payment_info) => payment_info,
+					Err(e) => {
+						log_error!(self.logger, "Failed to access payment store: {}", e);
+						return Err(ReplayEvent());
```
benthecarman: **Remote store availability now gates inbound HTLC claims** — `src/event.rs:804` (confirmed, behavior change) The `PaymentClaimable` handler’s `get()` is now fallible and returns `ReplayEvent` on store failure, so an inbound HTLC cannot be claimed while the store is down. A sustained outage past the claim deadline fails the payment back to the sender. This is fail-safe by design, but it is a real behavior change worth documenting.

src/event.rs
```diff
@@ -1219,22 +1225,29 @@ where
...
+							} else {
+								"".to_string()
+							},
+							hex_utils::to_string(&payment_hash.0),
+						);
+					},
+					Ok(None) => {},
+					Err(e) => {
+						log_error!(self.logger, "Failed to access payment store: {}", e);
+						return Err(ReplayEvent());
```
benthecarman: **`PaymentSent` replays the whole event when a log-only read fails** — `src/event.rs:1246` (confirmed, low) A `get()` used only to log the sent amount returns `ReplayEvent` on error — even though a _missing_ payment on the line above is tolerated. The required update already succeeded, so the replay is wasteful but self-healing. Fix: log and continue.

src/data_store.rs
```diff
@@ -40,143 +50,625 @@ pub(crate) enum DataStoreUpdateResult {
...
+	///
+	/// Returns whether an object was previously stored under that id.
 	pub(crate) async fn insert(&self, object: SO) -> Result<bool, Error> {
-		let _guard = self.mutation_lock.lock().await;
+		let _guard = self.mutation_lock.write().await;
 
+		let id = object.id();
+		// Callers treat the return value as "this id was already known", so a cache miss is not
+		// enough to answer it under a bounded policy.
+		let replaced = self.contains(&id).await?;
```
benthecarman: **`insert()` pays a guaranteed extra read for an unused bool** — `src/data_store.rs:329` `insert()` computes its `replaced` return value via `contains()` → a full KVStore read that always misses for new payment ids. The bool is discarded at 12 of 14 call sites; the other two only log and `debug_assert`. Every insert on a remote backend pays an extra round trip under the write lock.

src/payment/bolt11.rs
```diff
@@ -276,7 +276,7 @@ impl Bolt11Payment {
 
 		let payment_hash = invoice.payment_hash();
 		let payment_id = PaymentId(invoice.payment_hash().0);
-		if let Some(payment) = self.payment_store.get(&payment_id) {
+		if let Some(payment) = self.runtime.block_on(self.payment_store.get(&payment_id))? {
```
benthecarman, line 211: **The send path issues 3 sequential store round trips** — `src/payment/bolt11.rs:279` Duplicate-check `get()` (miss) → `insert()`'s internal `contains()` re-reads the same key (miss) → the write. One write plus a cache check would do. Same shape in `spontaneous.rs`; bolt12 sends pay two.

src/data_store.rs
```diff
@@ -40,143 +50,625 @@ pub(crate) enum DataStoreUpdateResult {
...
+			cache_policy: PhantomData,
 		}
 	}
 
+	/// Stores `object`, overwriting any object previously stored under the same id.
+	///
+	/// Returns whether an object was previously stored under that id.
 	pub(crate) async fn insert(&self, object: SO) -> Result<bool, Error> {
-		let _guard = self.mutation_lock.lock().await;
+		let _guard = self.mutation_lock.write().await;
```
benthecarman, line 316: **The write lock is held across sequential KVStore round trips** — `src/data_store.rs:324` Mutations hold the store-wide `RwLock` write guard across read-then-write remote I/O, so unrelated payments serialize and all readers block for the full duration. Reader-blocking is a documented consistency choice, but the critical section is longer than it needs to be.

benthecarman: this kinda seems overkill

src/data_store.rs
```diff
@@ -40,143 +50,625 @@ pub(crate) enum DataStoreUpdateResult {
...
+	/// Removes the object stored under `id`, if any.
 	pub(crate) async fn remove(&self, id: &SO::Id) -> Result<(), Error> {
-		let _guard = self.mutation_lock.lock().await;
-		let should_remove = { self.objects.lock().expect("lock").contains_key(id) };
-		if should_remove {
-			let store_key = id.encode_to_hex_str();
-			KVStore::remove(
+		let _guard = self.mutation_lock.write().await;
+
+		if !self.contains(id).await? {
```
benthecarman: **`remove()` reads the full value to gate an idempotent remove** — `src/data_store.rs:366` `remove()` checks `contains()` first, which downloads the entire serialized payment on a cache miss — but the KVStore contract guarantees `remove` succeeds whether or not the key exists. Two round trips where one suffices.

src/data_store.rs
```diff
@@ -40,143 +50,625 @@ pub(crate) enum DataStoreUpdateResult {
...
+			next_page_token: response.next_page_token,
+		})
+	}
+
+	/// Reads the objects we couldn't serve from the cache into their slots in `objects`.
+	///
+	/// Reads run concurrently but are tracked by slot, as the order in which they finish says
+	/// nothing about the order of the page. Note the objects read here are deliberately *not*
+	/// cached, see [`Self::list_page`].
+	async fn read_missing(
```
benthecarman, line 544: **`read_missing` duplicates the batched-read machinery** — `src/data_store.rs:503` It reimplements the JoinSet fill/refill/abort machinery of `read_objects_internal` (`src/io/utils.rs`) near-verbatim. The drift risk is proven: main’s copy shipped a missing-`abort_all` bug that this PR fixed only incidentally while creating the second copy. A shared helper extracts cleanly.

src/data_store.rs
```diff
@@ -40,143 +50,625 @@ pub(crate) enum DataStoreUpdateResult {
 	NotFound,
 }
 
-pub(crate) struct DataStore<SO: StorableObject, L: Deref>
+/// How many of a namespace's objects a [`DataStore`] keeps in memory.
+#[derive(PartialEq, Eq, Debug, Clone, Copy)]
+pub(crate) enum CacheLimit {
```
benthecarman, line 57: **Cache policy encoded in four parallel layers** — `src/data_store.rs:55` (plausible, optional nit) Marker types + `CachePolicy` trait + `CacheLimit` enum + `ObjectCache` enum all encode one binary choice. The `CacheLimit` layer is collapsible while keeping the compile-time gate on `list_filter`.

benthecarman: also overkill imo

src/builder.rs
```diff
@@ -1489,7 +1492,11 @@ fn build_with_store_internal(
 
 	let payment_store = match payment_store_res {
 		Ok(payments) => Arc::new(PaymentStore::new(
-			payments,
+			// The read hands us the newest payments first, while the cache treats the objects it
+			// is seeded with as increasingly recently used. Reverse them, so that the newest
+			// payment is the last one to be evicted rather than the first.
+			payments.into_iter().rev().collect(),
```
benthecarman, line 1499: **LRU seeding order is guarded only by a comment** — `src/builder.rs:1495` `read_n_objects` returns newest-first, but `DataStore::new` treats the _last_ seed element as most-recently-used, so the builder must `.rev()`. Dropping that call would silently evict the newest payments first with no test to catch it.

benthecarman (commented):
Some Ai review first

src/lib.rs
tnull: Yes, we have that requirement, period. Not sure if there is a way to 'block' it per se.

src/event.rs
tnull: I mean, yes, if IO fails continuously we're screwed either way?

src/lib.rs
benthecarman: it suggested checking `runtime_flavor` in `Runtime`'s constructor, still doesn't fix at compile time but you fail faster potentially

tnull: Yeah, idk, we already mention this requirement in `Node::start` docs. Seems sufficient for now, and we'll need to revisit when going for WASM related changes next release anyways.

src/event.rs
tnull: Addressed in fixup.

src/data_store.rs
tnull: Addressed in fixup.

src/payment/bolt11.rs
tnull: Addressed in fixup.

src/data_store.rs
tnull: Addressed in fixup.

src/data_store.rs
tnull: Done by pulling out prefactor.

tnull: But also seems a bit overkill, @benthecarman  let me know if you prefer the earlier non-DRY approach.

src/data_store.rs
```diff
@@ -179,36 +424,280 @@ where
...
+		self.cache.lock().expect("lock").insert(new_object.id(), new_object.clone());
+		Ok(Some(new_object))
+	}
+
+	/// Like [`Self::mutate`], but allows the transformation to await fallible reads.
+	///
+	/// The mutation lock remains held while `f` runs. This is useful when the new state must be
+	/// decided from an async read of another store without letting a concurrent writer invalidate
+	/// that decision. Callers must keep cross-store lock ordering consistent to avoid deadlocks.
+	pub(crate) async fn mutate_async<F, Fut>(&self, id: &SO::Id, f: F) -> Result<Option<SO>, Error>
```
benthecarman, line 452: this is almost exact same as `mutate` can we DRY these

src/data_store.rs
```diff
@@ -40,135 +52,368 @@ pub(crate) enum DataStoreUpdateResult {
...
-			let Some(object) = locked_objects.get(&id) else {
-				return Ok(DataStoreUpdateResult::NotFound);
-			};
-			let mut updated_object = object.clone();
-			if !updated_object.update(update) {
-				return Ok(DataStoreUpdateResult::Unchanged);
-			}
-			updated_object
+		// As in `insert_or_update`, a cache miss is not evidence of absence: reporting `NotFound`
+		// for a merely evicted object would drop the update on the floor.
```
benthecarman: this comment is kinda confusing here, can you just delete it here, i think the code is more obvious of what it does without it

src/data_store.rs
```diff
@@ -40,135 +52,368 @@ pub(crate) enum DataStoreUpdateResult {
...
+				Self::BoundedLru(lru)
+			},
+		}
+	}
+
+	/// Whether the cache holds *every* object of the namespace.
+	///
+	/// If it does, a miss proves the object is absent from the store, and iterating the cache
+	/// yields a complete listing. If it doesn't, both require reading from the store.
+	fn is_authoritative(&self) -> bool {
```
benthecarman: Don't love this name, i think like `is_keep_all` would be more clear

src/data_store.rs
```diff
@@ -179,36 +424,280 @@ where
...
+			} else {
+				Error::PersistenceFailed
+			}
+		})?;
+
+		// Serve whatever we already hold, and note the rest to read below. We take the mutation
+		// lock only for this, so that we observe a consistent view of the cache without holding up
+		// writers while we read.
+		let mut objects: Vec<Option<SO>> = vec![None; response.keys.len()];
+		let mut missing = Vec::new();
```
benthecarman: should we vec with capacity here?

src/config.rs
```diff
@@ -48,6 +49,21 @@ pub(crate) const DEFAULT_FEE_RATE_CACHE_UPDATE_TIMEOUT_SECS: u64 = 10;
...
+// per cached payment, this bounds the payment store's share of memory at well under a megabyte,
+// while still covering the recent payments a node actually works with.
+pub(crate) const PAYMENT_CACHE_CAPACITY: NonZeroUsize = NonZeroUsize::new(1000).unwrap();
+
+// The number of payments we read into the cache when starting up.
+//
+// This matches the storage backends' page size, so warming the cache costs a single page listing
+// and one batch of reads, and the first page of `Node::list_payments` is served without going to
+// the store at all. The remaining capacity fills as payments are used.
+pub(crate) const PAYMENT_CACHE_WARMUP_COUNT: NonZeroUsize = NonZeroUsize::new(50).unwrap();
```
benthecarman, line 66: we have this `50` as a const somewhere, can we use that?

src/config.rs
```diff
@@ -48,6 +49,21 @@ pub(crate) const DEFAULT_FEE_RATE_CACHE_UPDATE_TIMEOUT_SECS: u64 = 10;
...
+// payments and read the rest back from the store as they are needed. At roughly 400 to 500 bytes
+// per cached payment, this bounds the payment store's share of memory at well under a megabyte,
+// while still covering the recent payments a node actually works with.
+pub(crate) const PAYMENT_CACHE_CAPACITY: NonZeroUsize = NonZeroUsize::new(1000).unwrap();
+
+// The number of payments we read into the cache when starting up.
+//
+// This matches the storage backends' page size, so warming the cache costs a single page listing
+// and one batch of reads, and the first page of `Node::list_payments` is served without going to
+// the store at all. The remaining capacity fills as payments are used.
```
benthecarman: > and the first page of `Node::list_payments` is served without going to the store at all

is only true if the LRU hasn't been touched but payments could effect this, no?

src/data_store.rs
```diff
@@ -40,135 +52,368 @@ pub(crate) enum DataStoreUpdateResult {
...
+		}
+		debug_assert_eq!(self.entries.len(), self.recency.len());
+	}
+}
+
+/// The in-memory part of a [`DataStore`].
+///
+/// Modelled as an enum rather than a map plus a policy field so that a [`KeepAllEntries`] store
+/// provably pays nothing for the bookkeeping a bounded one needs: its representation is just the
+/// map it always was.
```
benthecarman: this comment feels like codex restating your pompts

src/data_store.rs
```diff
@@ -243,27 +732,75 @@ where
...
 	use lightning::{impl_writeable_tlv_based, io};
+	use tokio::sync::Notify;
 
 	use super::*;
 	use crate::hex_utils;
-	use crate::io::test_utils::InMemoryStore;
+	use crate::io::test_utils::{InMemoryStore, IN_MEMORY_PAGE_SIZE};
 	use crate::types::DynStoreWrapper;
 
+	const TEST_PRIMARY_NAMESPACE: &str = "datastore_test_primary";
```
benthecarman, line 763: you added these but didn't add them to all the existing tests, can you do that

src/data_store.rs
tnull: Done in prefactor.

src/data_store.rs
tnull: Done in fixup.

src/data_store.rs
tnull: Done in fixup.

src/data_store.rs
tnull: Done in fixup.

src/config.rs
tnull: Hmm, we have some `const`s that happen to be 50, but they are somewhat orthogonal? Which one do you have in mind exactly?

src/config.rs
tnull: Done in fixup.

src/data_store.rs
tnull: Done in fixup.

src/data_store.rs
tnull: No, I don't think this was the case exactly, but dropped in fixup as still bogus/not how we should write comments.

benthecarman:
ci failing

tnull:
> ci failing

Not any longer. Note that the timing out jobs might be fixed by #997, we'll see.

src/error.rs
```diff
@@ -117,6 +117,8 @@ pub enum Error {
 	InvalidFeeRate,
 	/// The given script public key is invalid.
 	InvalidScriptPubKey,
+	/// The given page token is invalid.
+	InvalidPageToken,
```
benthecarman, line 121: claude is yelling that we never use this for vss

src/data_store.rs
benthecarman: eh yeah these about equal in my head

src/config.rs
benthecarman: well the comment highlights that its equal ot the page size. We have separate consts for posgres and sqlite page size. Should unify those and use that const here

src/data_store.rs
tnull: > eh yeah these about equal in my head

Okay, I'm not sure what to make of this though? Do you prefer any further changes or not?

src/config.rs
tnull: Hmm, not sure? It's basically a coincidence that it's a page size, though likely a multiple of page size makes sense? Not sure if we want to couple the concepts strongly here, would also be a layer violation somewhat, and one additional thing we'd need to disentangle when upstreaming VSS/Postgres stores?

src/error.rs
tnull: As this is pre-existing I now added it as an orthogonal fix commit.

tnull:
Rebased to resolve minor conflicts. Also, let me know if I can squash fixups.

src/wallet/mod.rs
tnull: Yeah, that's pre-existing and honestly I don't know how to improve here as part of this PR.

Generally these issues are tracked on https://github.com/lightningdevkit/ldk-node/issues/1044, now mentioned over at https://github.com/lightningdevkit/ldk-node/issues/1044#issuecomment-5316162678

src/data_store.rs
benthecarman: Let's go with the changes you made

src/config.rs
benthecarman: I mostly say this because the comment says 

> This matches the built-in storage backends' page size, so warming the cache costs a single page listing and one batch of reads

So it implies that they are meant to be the same. If its a coincidence / orthogonal i guess that's fine and yeah when we eventually upstream we would have to change anyways.

benthecarman (commented):
good to squash

tnull:
> good to squash

benthecarman (approved):

src/data_store.rs
joostjager: @tnull to resolve this, I think we just accept this behavior assuming there are no bad entries?

tnull: Yes, at least for now. We don't deal with corrupted data (other than erroring out) elsewhere, and this change only marginally increases the chances of hitting such a thing. Coming up with a strategy to deal with corrupted data in general seems it could be worth it at some point, but it's somewhat orthogonal to this specific PR in my eyes.

PR #1025, opened 2026-08-06
Retain Electrum Filterregistrations acrossstop/start``

### LINKED ISSUE #1005: Electrum chain source `stop()` discards registered `Filter` entries
Discovered in https://github.com/lightningdevkit/ldk-node/pull/992 (but pre-existing)

Claude:

> The new failure-path chain_source.stop() silently discards Electrum Filter registrations, breaking a failed-start → retry cycle for nodes with existing channels.
> 
>   During Builder::build, channel monitors are loaded (src/builder.rs:1987) and register their watched txs/outputs with the chain source's Filter impl. For Electrum, these are queued in ElectrumRuntimeStatus::Stopped { pending_registered_txs, pending_registered_outputs } until start() applies
>   them. But ElectrumRuntimeStatus::stop() does *self = Self::new() (src/chain/electrum.rs:456), wiping both the running client and any pending registrations, and nothing re-registers them afterward — the outer ChainSource::registered_txids set survives but is never re-applied (and doesn't
>   carry scripts/outputs).
> 
>   Concrete scenario: a node with existing channels calls start(), the fee-rate update or 0FC validation fails transiently (network blip) → the new error path calls chain_source.stop() → the app retries start(), which now succeeds — but the fresh ElectrumRuntimeClient's tx-sync watches nothing.
>   Confirmations/spends of existing funding outputs (e.g. a counterparty force-close) go undetected until a full process restart.
> 
>   To be fair, a normal stop() → start() cycle already has this property on main, so it may be a known limitation — but pre-PR the failed-start path left the registrations intact (at the cost of the leak this PR fixes), and a failed first start with a retry loop is a much more common pattern
>   than a deliberate stop/restart. I'd suggest having ElectrumRuntimeStatus::stop() preserve the registration set (transition back to Stopped carrying the known registrations) rather than resetting to empty — that would fix both this path and the pre-existing stop/start case.

The long-term fix would still be to get rid of the `start`/`stop` mechanism entirely (see https://github.com/bitcoindevkit/rust-electrum-client/issues/166), but as a hotfix we probably should make sure we trigger re-registration after restart (via [`load_outputs_to_watch`](https://docs.rs/lightning/latest/lightning/chain/chainmonitor/struct.LockedChannelMonitor.html#method.load_outputs_to_watch)

Jolah1:
Fixes #1005.

  Problem

 ElectrumRuntimeStatus::stop reset itself via *self = Self::new(), which dropped the ElectrumRuntimeClient and the pending-registration vectors. Since the client owns the ElectrumSyncClient, which in turn owns its own copy of the
  registered transactions and outputs, all Filter state was gone. start then drain(..)ed the pending vectors, so registrations didn't survive even a single cycle.

 This matters because ChannelMonitors only register their watched transactions and outputs while being loaded in Builder::build. A stop/start cycle that doesn't rebuild the node therefore leaves the chain source with no
  registrations at all, and the node stops learning about confirmations and spends of, e.g., its funding outputs.

 As discussed on the issue, replaying ChannelMonitor::load_outputs_to_watch on start wouldn't be sufficient either: the OutputSweeper also registers outputs it wants to see spent, and exposes no way to reload them.

  The Esplora chain source isn't affected, as its tx_sync is long-lived and never torn down — so this also restores parity between the two backends.

  Fix

ElectrumRuntimeStatus becomes a struct holding an Option<Arc<ElectrumRuntimeClient>> alongside a canonical, deduplicated registration inventory (HashMap<Txid, ScriptBuf> and HashSet<WatchedOutput>) that is maintained regardless of whether we're currently started:

  - register_tx/register_output always record the entry, and additionally forward it to the client if one is live.
  - start replays the whole inventory to the fresh client without consuming it.
  - stop merely drops the client.

 Note the inventory intentionally retains entries for already-confirmed transactions. That mirrors what already happens on every process restart via monitor load, and matches the existing behaviour of ChainSource::registered_txids, which is likewise never pruned.

 client() keeps its existing Option<Arc<ElectrumRuntimeClient>> signature, so all call sites are untouched.

  Testing

Adds electrum_registrations_survive_chain_source_restart: it opens a channel, leaves the funding transaction unconfirmed, then stops and starts the node repeatedly before confirming it. The node can only emit ChannelReady if the chain source replayed its registrations on every start — the repetition specifically covers the non-draining behaviour.

Verified the test fails on current main (node times out after 60s waiting for ChannelReady) and passes with this fix. The other Electrum-backed tests still pass.

This change was developed with the assistance of Claude Code. The design, review, and testing decisions are my own.

### DIFF
--- a/src/chain/electrum.rs
+++ b/src/chain/electrum.rs
@@ -5,7 +5,7 @@
 // http://opensource.org/licenses/MIT>, at your option. You may not use this file except in
 // accordance with one or both of these licenses.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex, RwLock};
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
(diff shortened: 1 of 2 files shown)


src/chain/electrum.rs
```diff
@@ -386,79 +386,78 @@ impl Filter for ElectrumChainSource {
...
+		self.client.as_ref().map(Arc::clone)
 	}
 
 	fn register_tx(&mut self, txid: &Txid, script_pubkey: &Script) {
-		match self {
-			Self::Started(client) => client.register_tx(txid, script_pubkey),
-			Self::Stopped { pending_registered_txs, .. } => {
-				pending_registered_txs.push((*txid, script_pubkey.to_owned()))
-			},
+		self.registered_txs.insert(*txid, script_pubkey.to_owned());
```
tnull, line 449: When we do this, we should deduplicate the entries to make sure we're not re-registering all the entries every time. 

Same below

tnull (commented):
Thanks for looking into this!

src/chain/electrum.rs
Jolah1: > When we do this, we should deduplicate the entries to make sure we're not re-registering all the entries every time.
> 
> Same below

These dedupe by construction,  the inventory is a HashMap<Txid, ScriptBuf> / HashSet<WatchedOutput> rather than the Vecs it replaces, so start replays each entry exactly once. (The hunk cuts off right at the insert, so the forwarding branch below it may not have been visible.)

 If you meant pruning entries that are no longer needed, I don't think we can from here: whether a tx is buried or an output spent lives in lightning-transaction-sync's private SyncState, which is dropped with the client on stop.
 So we do re-register confirmed entries on each start, the same thing ChannelMonitor::load_outputs_to_watch does on every process restart. ChainSource::registered_txids has the same property today.

Let me know if you meant something else.

tnull: > These dedupe by construction, the inventory is a HashMap<Txid, ScriptBuf> / HashSet rather than the Vecs it replaces, so start replays each entry exactly once. (The hunk cuts off right at the insert, so the forwarding branch below it may not have been visible.)

Ah, lol, duh. Sorry, brain wasn't working right there.

> If you meant pruning entries that are no longer needed, I don't think we can from here: whether a tx is buried or an output spent lives in lightning-transaction-sync's private SyncState, which is dropped with the client on stop.
> So we do re-register confirmed entries on each start, the same thing ChannelMonitor::load_outputs_to_watch does on every process restart. ChainSource::registered_txids has the same property today.

Yeah, real pruning would likely require https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/issues/1664 - long standing issue.

tnull (approved):
Thanks!

PR #1030, opened 2026-08-06
eltoo with bip-448

elnosh:
This builds on top of https://github.com/bip448/rust-lightning/pull/2 to implement a node that can create eltoo channels.
Integration tests need to point a bitcoind binary with support for it which I was using https://github.com/bip448/bitcoin/pull/1

Comes with a repl to drive the demo node creating a channel, can broadcast an old state and rebind to broadcast newest update and then the settlement tx.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -141,6 +141,23 @@ check-cfg = [
 name = "payments"
 harness = false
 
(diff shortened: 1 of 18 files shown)


elnosh:
woops wrong repo

PR #1032, opened 2026-08-07
Fail closed on manager read errors

### LINKED ISSUE #1026: `ChannelManager` read errors start a fresh node
## TLDR

ldk-node treats any `ChannelManager` KV-store read error as a missing manager and creates a fresh, empty node. A timeout, connection failure, authentication error, or exhausted VSS 5xx can therefore return a successful node instead of `BuildError::ReadFailed`.

## Vulnerable code

[`src/builder.rs:1935`](https://github.com/lightningdevkit/ldk-node/blob/5e7250a7873e1c695672a0e86c4d148a008f5522/src/builder.rs#L1935-L1980):

```rust
if let Ok(reader) = channel_manager_bytes_res {
    // Restore ChannelManager.
} else {
    // Create a fresh ChannelManager.
}
```

The `else` branch does not distinguish `NotFound` from other storage failures.

## Impact

An established node can start with zero channels after a manager read failure. If the fresh manager is later persisted, it can overwrite the valid manager and permanently destroy the live off-chain channel state.

Loaded `ChannelMonitor`s remain active, so the expected manager-only outcome is unilateral channel closure, fees, and delayed access to funds—not loss of the full channel balance. Principal loss requires the monitors to also be missing or unsafe.

## Verification

The vulnerable behavior is confirmed by source inspection of upstream `main`.

The attached test [poc.txt](https://github.com/user-attachments/files/30803961/poc.txt) is an end-to-end regtest PoC using two real ldk-node instances and a real channel. It injects a manager-only `ErrorKind::Other` at the `KVStore` boundary and verifies that:

- a total store outage fails closed;
- a manager-only failure builds a node with zero channels;
- the channel is unilaterally closed on-chain; and
- graceful shutdown overwrites the valid manager with the smaller empty manager.

`ErrorKind::Other` is the error class produced by VSS for a 5xx after retry exhaustion. The PoC tests the resulting KV-store failure directly; it does not emulate the HTTP request or VSS retry layer itself.

## Proposed fix

```rust
match channel_manager_bytes_res {
    Ok(reader) => restore_manager(reader)?,
    Err(error) if error.kind() == std::io::ErrorKind::NotFound => create_fresh_manager(),
    Err(error) => {
        log_error!(logger, "Failed to read ChannelManager: {}", error);
        return Err(BuildError::ReadFailed);
    }
}
```

Additionally, a missing manager with existing monitors should require an explicit recovery path rather than appearing as a normal empty node.

tnull:
Only a missing ChannelManager should create a fresh node. Propagate other storage errors so transient failures cannot replace live channel state with an empty manager.

Fixes #1026

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1934,7 +1934,16 @@ fn build_with_store_internal(
 
 	// Initialize the ChannelManager
 	let channel_manager = {
-		if let Ok(reader) = channel_manager_bytes_res {
+		let channel_manager_bytes = match channel_manager_bytes_res {
+			Ok(reader) => Some(reader),
+			Err(e) if e.kind() == lightning::io::ErrorKind::NotFound => None,
+			Err(e) => {
+				log_error!(logger, "Failed to read channel manager from store: {}", e);
+				return Err(BuildError::ReadFailed);
+			},
+		};
+
+		if let Some(reader) = channel_manager_bytes {
 			let channel_monitor_references =
 				channel_monitors.iter().map(|(_, chanmon)| chanmon).collect();
 			let read_args = ChannelManagerReadArgs::new(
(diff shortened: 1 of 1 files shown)


src/builder.rs
```diff
@@ -1934,7 +1934,16 @@ fn build_with_store_internal(
 
 	// Initialize the ChannelManager
 	let channel_manager = {
-		if let Ok(reader) = channel_manager_bytes_res {
+		let channel_manager_bytes = match channel_manager_bytes_res {
+			Ok(reader) => Some(reader),
+			Err(e) if e.kind() == lightning::io::ErrorKind::NotFound => None,
```
TheBlueMatt, line 1939: Hmm, okay. I don't love reintroducing the assumption that the `ErrorKind` is "correct", but at worst we just refuse to start and the developer fixes their KVStore cause it should be obvious during development. Also I dunno how else to fix this so 🤷‍♂️

TheBlueMatt (approved):

src/builder.rs
tnull: Yeah, well, `NotFound` is at least part of the `KVStore` API contract.

TheBlueMatt: No we removed that when we fixed MonitorUpdatingPersister to no longer rely on it.

tnull: > No we removed that when we fixed MonitorUpdatingPersister to no longer rely on it.

Uh, but that's actually bad, as we do lean on it in LDK Node in several places. `NotFound` is always a special value for `KVStore`s, irrespective of whether `MonitorUpdatingPersister` now returns it or not. We should re-add it then.

PR #1037, opened 2026-08-10
Fix wallet UTXO reuse for funding transactions and onchain spends

### LINKED ISSUE #41: Don't double-spend ourselves on sequential channel opens
Unfortunately `bdk::TxBuilder` will reuse already used UTXOs, if no `sync()` call happens in-between.

We might therefore need to track UTXOs used in funding transactions and explicitly forbid reuse via `TxBuilder::unspendable`. Moreover we'll need to track current sync height and un-lock the UTXOs if they haven't been spent for some time to allow reuse.

tnull:
Fixes #41.

For the longest time BDK didn't offer any UTXO locking mechanisms and only considered transactions canonical once seen in the mempool during syncing. This always left a gap between the time of transaction signing/broadcast and the time of sync during which the wallet could double-spend itself. Since `bdk_wallet` v3.0 they finally offer UTXO locking APIs which we finally use here to close this gap for funding transactions and onchain spends.

**Note:** We intentionally leave splicing transactions out-of-scope of this PR because with #962 and #930 there are related PRs inflight. Depending on the order these land, this PR or they need to be updated to marry the two approaches. (cc @jkczyz)

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -538,6 +538,28 @@ impl Future for EventFuture {
 	}
 }
 
+fn discarded_funding_transaction(funding_info: FundingInfo) -> Option<bitcoin::Transaction> {
+	match funding_info {
+		FundingInfo::Tx { transaction } => Some(transaction),
+		FundingInfo::Contribution { inputs, outputs } => Some(bitcoin::Transaction {
+			version: bitcoin::transaction::Version::TWO,
+			lock_time: bitcoin::absolute::LockTime::ZERO,
+			input: inputs
+				.into_iter()
+				.map(|previous_output| bitcoin::TxIn {
+					previous_output,
+					..bitcoin::TxIn::default()
+				})
+				.collect(),
+			output: outputs
+				.into_iter()
+				.map(|script_pubkey| bitcoin::TxOut { value: bitcoin::Amount::ZERO, script_pubkey })
+				.collect(),
+		}),
+		FundingInfo::OutPoint { .. } => None,
+	}
+}
+
 pub(crate) struct EventHandler<L: Deref + Clone + Sync + Send + 'static>
 where
 	L::Target: LdkLogger,
@@ -1954,26 +1976,7 @@ where
 				}
 			},
 			LdkEvent::DiscardFunding { channel_id, funding_info } => {
-				let tx = match funding_info {
-					FundingInfo::Tx { transaction } => Some(transaction),
-					FundingInfo::Contribution { inputs: _, outputs } => {
-						Some(bitcoin::Transaction {
-							version: bitcoin::transaction::Version::TWO,
-							lock_time: bitcoin::absolute::LockTime::ZERO,
-							input: vec![],
-							output: outputs
-								.into_iter()
-								.map(|script_pubkey| bitcoin::TxOut {
-									value: bitcoin::Amount::ZERO,
-									script_pubkey,
-								})
-								.collect(),
-						})
-					},
-					FundingInfo::OutPoint { .. } => None,
-				};
-
-				if let Some(tx) = tx {
+				if let Some(tx) = discarded_funding_transaction(funding_info) {
 					log_info!(
 						self.logger,
 						"Reclaiming unused wallet state from channel {} funding",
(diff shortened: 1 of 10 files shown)


tnull:
Drafting this for now as it might make sense to wait for #962 to land first, and then rebase this.

tnull:
Rebased to resolve minor conflicts. Should be good for review now that #962 landed. 

@wpaulino @jkczyz Do you think we should also cover UTXO locking for splices in this PR, or would that be the concern of #930?

jkczyz:
> Rebased to resolve minor conflicts. Should be good for review now that #962 landed.
> 
> @wpaulino @jkczyz Do you think we should also cover UTXO locking for splices in this PR, or would that be the concern of #930?

I'd say either here or in a dedicated PR. It can be done independent of #930, which is already pretty big. That touches `SpliceNegotiationFailed` handling, but for unlocking we only need to touch `DiscardFunding` handling. The bigger change is using our own `CoinSelectionSource` for locking rather than relying on `LdkWallet`, IIUC.

tnull:
> > Rebased to resolve minor conflicts. Should be good for review now that #962 landed.
> > @wpaulino @jkczyz Do you think we should also cover UTXO locking for splices in this PR, or would that be the concern of #930?
> 
> I'd say either here or in a dedicated PR. It can be done independent of #930, which is already pretty big. That touches `SpliceNegotiationFailed` handling, but for unlocking we only need to touch `DiscardFunding` handling. The bigger change is using our own `CoinSelectionSource` for locking rather than relying on `LdkWallet`, IIUC.

Now added a commit here, let me know what you think

src/wallet/mod.rs
```diff
@@ -630,18 +653,21 @@ impl Wallet {
...
+				log_error!(self.logger, "Failed to extract transaction: {}", e);
+				e
+			})?;
+			for txin in &tx.input {
+				locked_wallet.lock_outpoint(txin.previous_output);
+			}
+
+			(tx, locked_wallet.take_staged().unwrap_or_default())
 		};
 		locked_persister.persist_changeset(change_set).await.map_err(|e| {
```
Jolah1, line 673: If this persistence fails, the inputs locked above are leaked. persist_changeset retains the staged changeset on failure, but this function returns without the transaction, so the caller cannot release the locks and LDK cannot later emit DiscardFunding. 

A subsequent successful wallet persistence makes the orphan locks durable. I reproduced this across a wallet reload with a fault-injecting store. The same issue occurs in select_confirmed_utxos below. Could we roll back both the in-memory and staged locks before returning the error, with regression tests for both paths?

Jolah1 (commented):
The overall approach looks good, but I found one persistence-failure issue that should be addressed before merging. Both channel-funding creation and splice coin selection can leave their selected inputs durably locked when persisting the locks fails. 
 I reproduced this with a fault-injecting store: the operation returned an error, the inputs remained locked, and after persistence recovered the locks survived a wallet reload. The same issue occurs in select_confirmed_utxos at line 1384.
I left the details inline.

tnull:
Rebased to resolve some conflicts.

src/event.rs
```diff
@@ -1941,27 +1954,33 @@ where
...
+							output: outputs
+								.into_iter()
+								.map(|script_pubkey| bitcoin::TxOut {
+									value: bitcoin::Amount::ZERO,
+									script_pubkey,
+								})
+								.collect(),
+						})
+					},
+					FundingInfo::OutPoint { .. } => None,
```
wpaulino: Is it not worth doing or are we not able to locate the transaction in the wallet via the txid to cancel it?

src/wallet/mod.rs
```diff
@@ -637,18 +660,21 @@ impl Wallet {
 				},
 			}
 
-			(psbt, locked_wallet.take_staged().unwrap_or_default())
+			let tx = psbt.extract_tx().map_err(|e| {
+				log_error!(self.logger, "Failed to extract transaction: {}", e);
+				e
+			})?;
+			for txin in &tx.input {
+				locked_wallet.lock_outpoint(txin.previous_output);
```
wpaulino: Will these ever get unlocked if we crash before calling `ChannelMananger::funding_transaction_generated`?

src/event.rs
wpaulino: Is ldk-node not expecting these at the moment? We could look up the transaction in the wallet via its txid. Otherwise a `debug_assert` that this is not reachable would be nice.

PR #1040, opened 2026-08-11
Fix CI: Keep Python README within package builds

tnull:
Load the repository README through Hatchling metadata so builds accept the Python package configuration. Include a regular copy in source distributions so archive extraction remains safe.

(As latest `hatchlings` release broke our CI)

### DIFF
--- a/bindings/python/pyproject.toml
+++ b/bindings/python/pyproject.toml
@@ -9,7 +9,7 @@ authors = [
   { name="Elias Rohrer", email="dev@tnull.de" },
 ]
 description = "A ready-to-go Lightning node library built using LDK and BDK."
-readme = "../../README.md"
+dynamic = ["readme"]
 requires-python = ">=3.8"
 classifiers = [
     "Topic :: Software Development :: Libraries",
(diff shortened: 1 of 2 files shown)


bindings/python/pyproject.toml
```diff
@@ -30,4 +30,9 @@ dev = ["requests"]
 [tool.hatch.build.targets.wheel]
 packages = ["src/ldk_node"]
 
+[tool.hatch.build.targets.sdist.force-include]
+"../../README.md" = "README.md"
```
tankyleo: For now I just had codex review this commit, it had this to say let me know if you want to fix this:

>   - [P2] Make the sdist README source self-contained _ /home/ubuntu/ldk-node/bindings/python/pyproject.toml:34-34
    When an extracted sdist is repackaged, this mapping still resolves ../../README.md outside the extracted package, where that file is absent. Consequently, running a standard build such as python -m build on the sdist raises FileNotFoundError: Forced include not found, even though the archive
    contains its own README.md; the sdist configuration should use that local copy when rebuilding.

tankyleo: Here's what codex came up with to fix this, feel free to grab it if it's useful: https://github.com/tankyleo/ldk-node/commit/d5bd4df428ecbdf92a43513721e471f0c1b5c9d1

tnull: Hmm, strange it didn't find it first. Now amended the commit.

tankyleo (approved):

PR #1041, opened 2026-08-11
Prepare Python wheel releases

tnull:
Build Python-ABI-independent wheels for supported Linux and macOS targets so users can install LDK Node directly from PyPI.

Keep build and publication available through local scripts. Require the complete artifact set on TestPyPI before production, and exercise the bindings on the oldest and newest supported CPython versions.

### DIFF
--- a/scripts/python_build_wheel.sh
+++ b/scripts/python_build_wheel.sh
@@ -138,7 +138,7 @@ trap cleanup EXIT
 echo "Building Python wheel from commit $(git rev-parse HEAD)"
 echo "Build host: $HOST_OS $HOST_ARCH"
 
-uv tool run --from "cibuildwheel[uv]==$CIBUILDWHEEL_VERSION" cibuildwheel \
+uv tool run --python 3.11 --from "cibuildwheel[uv]==$CIBUILDWHEEL_VERSION" cibuildwheel \
 	bindings/python \
 	--config-file bindings/python/pyproject.toml \
 	--output-dir "$OUTPUT_DIR"
--- a/scripts/python_build_wheel.sh
+++ b/scripts/python_build_wheel.sh
@@ -123,6 +123,7 @@ fi
 
 CARGO_TARGET_DIR="$(mktemp -d /tmp/cargo-target-ldk-node-python-wheels.XXXXXX)"
 export CARGO_TARGET_DIR
+BUILD_SOURCE_DIR="$(mktemp -d /tmp/ldk-node-python-wheels-source.XXXXXX)"
 SOURCE_DATE_EPOCH="$(git show -s --format=%ct HEAD)"
 export SOURCE_DATE_EPOCH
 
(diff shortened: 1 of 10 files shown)


valentinewallace:
This is pretty outside of my wheelhouse (ba dum tiss), and I was planning to sink a lot of my review bandwidth into https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/pulls/4872 in the coming days, so FYI my first impression is that it may take me a little while to get to this and then take a while to get up to speed on it.

tnull:
> This is pretty outside of my wheelhouse (ba dum tiss), and I was planning to sink a lot of my review bandwidth into https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/pulls/4872 in the coming days, so FYI my first impression is that it may take me a little while to get to this and then take a while to get up to speed on it.

Okay. Now reassigned @tankyleo as we'll need this for next week's release (worst case I just run the updated scripts locally without prior review). 

Rebased after #1040 landed.

scripts/python_build_wheel.sh
```diff
@@ -1,31 +1,154 @@
...
+			rm -rf -- "$CARGO_TARGET_DIR"
+			;;
+	esac
+}
+trap cleanup EXIT
+
+echo "Building Python wheel from commit $(git rev-parse HEAD)"
+echo "Target: $(uname -s) $(uname -m)"
+
+uv tool run --from "cibuildwheel[uv]==$CIBUILDWHEEL_VERSION" cibuildwheel \
```
tankyleo: Had codex do a full dress rehearsal, it came up with this, but seems low priority so feel free to dismiss:
```diff
commit a81dff8cd9864b68dc4506f086de18987adb9bc8
Author: Leo Nash <hello@leonash.net>
Date:   Fri Aug 14 00:57:22 2026 +0000

    Select Python 3.11 for cibuildwheel

    cibuildwheel 4.1 requires Python 3.11 or newer. When uv selected
    Python 3.10, the wheel build failed before Docker could start.

    Select Python 3.11 explicitly so uv installs a compatible interpreter
    and the release build works when Python 3.10 is the default.

    AI-Assisted-By: OpenAI Codex

diff --git a/scripts/python_build_wheel.sh b/scripts/python_build_wheel.sh
index 56ee4678..4b1337b6 100755
--- a/scripts/python_build_wheel.sh
+++ b/scripts/python_build_wheel.sh
@@ -127,7 +127,7 @@ trap cleanup EXIT
 echo "Building Python wheel from commit $(git rev-parse HEAD)"
 echo "Target: $(uname -s) $(uname -m)"

-uv tool run --from "cibuildwheel[uv]==$CIBUILDWHEEL_VERSION" cibuildwheel \
+uv tool run --python 3.11 --from "cibuildwheel[uv]==$CIBUILDWHEEL_VERSION" cibuildwheel \
        bindings/python \
        --config-file bindings/python/pyproject.toml \
        --output-dir "$OUTPUT_DIR"
```

tankyleo (dismissed):

scripts/python_build_wheel.sh
tnull: Thanks. Addressed this by amending.

tnull:
Now also amended to allow building macOS x86_64 wheels on aarch64 macs.

scripts/python_build_wheel.sh
tankyleo: Sounds like the `--python 3.11` flag is still missing from the `uv tool run` call ?

tankyleo (commented):
Pushed a couple of commits on top of your branch that came up after a few rounds with codex, feel free to drop and amend as needed.

tnull:
> Pushed a couple of commits on top of your branch that came up after a few rounds with codex, feel free to drop and amend as needed.

Yeah, def. disagree with adding `Cargo.lock` in this PR, and we also shouldn't already bump the version number, that will done as part of the release process.

scripts/python_build_wheel.sh
tnull: "worked for me", but now left your commit on the branch.

Jolah1:
I successfully built and tested the Linux x86-64 wheel from this head across Python 3.10–3.14. The checksum and packaged license files look correct, and the publication validator rejects the incomplete artifact set as expected. The Linux ARM64 and two macOS wheels still need confirmation on their respective build hosts.

tankyleo (approved):

PR #1042, opened 2026-08-11
Fall back to per-target defaults for sparse Esplora fee estimates

### LINKED ISSUE #1028: Sparse Esplora fee estimates may fall back to 1 sat/vB for urgent targets
## TLDR

On mainnet, the Esplora fee estimator rejects a completely empty estimate map but accepts a non-empty map that lacks estimates for important confirmation targets.

When conversion fails for one of those missing targets, the code silently substitutes 1 sat/vB. This applies even to urgent targets such as `UrgentOnChainSweep`, whose normal fallback is substantially higher.

## Affected code

[`src/chain/esplora.rs:368`](https://github.com/lightningdevkit/ldk-node/blob/5e7250a7873e1c695672a0e86c4d148a008f5522/src/chain/esplora.rs#L368-L396):

```rust
if estimates.is_empty() && self.config.network == Network::Bitcoin {
	return Err(Error::FeerateEstimationUpdateFailed);
}

for target in confirmation_targets {
	let num_blocks = get_num_block_defaults_for_target(target);

	let converted_estimate_sat_vb =
		esplora_client::convert_fee_rate(num_blocks, estimates.clone())
			.map_or(1.0, |converted| converted.max(1.0));

	let fee_rate =
		FeeRate::from_sat_per_kwu((converted_estimate_sat_vb * 250.0) as u64);

	let adjusted_fee_rate = apply_post_estimation_adjustments(target, fee_rate);
	new_fee_rate_cache.insert(target, adjusted_fee_rate);
}
```

Urgent targets include six-block confirmation estimates:

[`src/fee_estimator.rs:87`](https://github.com/lightningdevkit/ldk-node/blob/5e7250a7873e1c695672a0e86c4d148a008f5522/src/fee_estimator.rs#L87-L101):

```rust
match target {
	ConfirmationTarget::OnchainPayment => 6,
	ConfirmationTarget::ChannelFunding => 12,
	ConfirmationTarget::Lightning(ldk_target) => match ldk_target {
		LdkConfirmationTarget::MaximumFeeEstimate => 1,
		LdkConfirmationTarget::UrgentOnChainSweep => 6,
		// ...
	},
}
```

The existing target-specific fallback for `UrgentOnChainSweep` is 5,000 sat/kwu:

[`src/fee_estimator.rs:112`](https://github.com/lightningdevkit/ldk-node/blob/5e7250a7873e1c695672a0e86c4d148a008f5522/src/fee_estimator.rs#L112-L122):

```rust
LdkConfirmationTarget::UrgentOnChainSweep => 5000,
```

## Impact

A partial or malformed Esplora response can replace urgent fee estimates with the relay-floor value instead of the target-specific fallback.

During a high-fee period, this can delay on-chain payments or time-sensitive channel sweeps. LDK fee bumping and the available confirmation windows reduce the likelihood of immediate loss, but the fallback moves in the unsafe direction for urgent transactions.

An arbitrary Lightning counterparty cannot directly supply this response. Exploitation would require control of, or interference with, the configured Esplora source. The issue can also occur non-maliciously if the source returns a sparse estimate map.

[Issue #331](https://github.com/lightningdevkit/ldk-node/issues/331) and [PR #249](https://github.com/lightningdevkit/ldk-node/pull/249) concern fee estimation generally but do not appear to track this sparse-map fallback behavior.

## Proposed fix

On conversion failure, use the existing fallback for the requested target:

```rust
let fee_rate = esplora_client::convert_fee_rate(num_blocks, estimates.clone())
	.map(|estimate| {
		FeeRate::from_sat_per_kwu((estimate.max(1.0) * 250.0) as u64)
	})
	.unwrap_or_else(|| {
		FeeRate::from_sat_per_kwu(get_fallback_rate_for_target(target) as u64)
	});
```

For mainnet, also consider validating that the response provides sufficient coverage for all critical targets before replacing the current cache.

Jolah1:
Fixes #1028.

Esplora may return a non-empty estimate map that still has no usable entry for a given confirmation target, since convert_fee_rate only yields a value if the map holds a block count at or below the requested one. A map of {144: 
  2.0, 1008: 1.0} therefore leaves every target below 144 blocks — UrgentOnChainSweep, OnchainPayment, ChannelCloseMinimum, ... — without an estimate.

  Today we substitute 1 sat/vb in that case. That's not a missing fallback but an actively harmful one: OnchainFeeEstimator::estimate_fee_rate already falls back to get_fallback_rate_for_target whenever it misses the cache, so
  writing 1 sat/vb into the cache only serves to shadow the value we'd otherwise have used. For UrgentOnChainSweep that's 250 sats/kwu where the default is 5000 — a 20x underestimate on a target whose whole point is timeliness.

  So here we fall back to the target's own default instead. The post-estimation adjustments are deliberately skipped for it: there's no estimate to adjust, and skipping them lands us on exactly the rate the cache-miss path would
  have produced, which keeps the two paths consistent.

  One open question for reviewers: whether that's the behaviour you want, or whether the fallback should go through apply_post_estimation_adjustments anyway. It only matters for MaximumFeeEstimate (8000 unadjusted vs. 11300
  adjusted) and MinAllowedNonAnchorChannelRemoteFee (unchanged, as the subtraction clamps back to the floor). Happy to flip it — it's a one-line change plus one expected value in the test.

  To make this testable without an Esplora server, the cache construction moves out of update_fee_rate_estimates into a pure build_fee_rate_cache. The new tests cover sparse maps, empty maps on and off Mainnet, and assert the
  complete-map behaviour is unchanged.

  I confirmed the sparse test fails against the previous behaviour (250 vs. 5000 sats/kwu for UrgentOnChainSweep, 250 vs. 5000 for OnchainPayment).

### DIFF
--- a/src/chain/esplora.rs
+++ b/src/chain/esplora.rs
@@ -24,8 +24,8 @@ use crate::config::{
 	MAX_FULL_SCAN_STOP_GAP, MIN_FULL_SCAN_STOP_GAP,
 };
 use crate::fee_estimator::{
-	apply_post_estimation_adjustments, get_all_conf_targets, get_num_block_defaults_for_target,
-	OnchainFeeEstimator,
+	apply_post_estimation_adjustments, get_all_conf_targets, get_fallback_rate_for_target,
+	get_num_block_defaults_for_target, ConfirmationTarget, OnchainFeeEstimator,
 };
 use crate::io::utils::update_and_persist_node_metrics;
 use crate::logger::{log_bytes, log_debug, log_error, log_trace, log_warn, LdkLogger, Logger};
@@ -365,41 +365,20 @@ impl EsploraChainSource {
 			Error::FeerateEstimationUpdateFailed
 		})?;
 
-		if estimates.is_empty() && self.config.network == Network::Bitcoin {
-			// Ensure we fail if we didn't receive any estimates.
-			log_error!(
+		let new_fee_rate_cache =
+			build_fee_rate_cache(estimates, self.config.network).inspect_err(|_| {
+				log_error!(
 						self.logger,
 						"Failed to retrieve fee rate estimates: empty fee estimates are dissallowed on Mainnet.",
 					);
-			return Err(Error::FeerateEstimationUpdateFailed);
-		}
+			})?;
 
-		let confirmation_targets = get_all_conf_targets();
-
-		let mut new_fee_rate_cache = HashMap::with_capacity(10);
-		for target in confirmation_targets {
-			let num_blocks = get_num_block_defaults_for_target(target);
-
-			// Convert the retrieved fee rate and fall back to 1 sat/vb if we fail or it
-			// yields less than that. This is mostly necessary to continue on
-			// `signet`/`regtest` where we might not get estimates (or bogus values).
-			let converted_estimate_sat_vb =
-				esplora_client::convert_fee_rate(num_blocks, estimates.clone())
-					.map_or(1.0, |converted| converted.max(1.0));
-
-			let fee_rate = FeeRate::from_sat_per_kwu((converted_estimate_sat_vb * 250.0) as u64);
-
-			// LDK 0.0.118 introduced changes to the `ConfirmationTarget` semantics that
-			// require some post-estimation adjustments to the fee rates, which we do here.
-			let adjusted_fee_rate = apply_post_estimation_adjustments(target, fee_rate);
-
-			new_fee_rate_cache.insert(target, adjusted_fee_rate);
-
+		for (target, fee_rate) in new_fee_rate_cache.iter() {
 			log_trace!(
 				self.logger,
 				"Fee rate estimation updated for {:?}: {} sats/kwu",
 				target,
-				adjusted_fee_rate.to_sat_per_kwu(),
+				fee_rate.to_sat_per_kwu(),
 			);
 		}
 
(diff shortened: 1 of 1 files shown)


tnull (commented):
Please excuse the delay here!

Yes, we're currently considering Esplora/Electrum servers to be somewhat trusted (for fee rates, but also privacy-wise). It could make sense to apply additional checks to the returned fee rates (i.e., also somewhat bound them), but this is generally pretty delicate as we'd still need to make sure that we can handle legitimate fee rate spikes. 

As for falling back to default values - yes, that's somewhat legitimate also but a) different networks (e.g., regtest) probably need to be handled differently and b) if we do this, we'd probably at the very least implement the same behavior across all three chain sources?

Jolah1:
> Please excuse the delay here!
> 
> Yes, we're currently considering Esplora/Electrum servers to be somewhat trusted (for fee rates, but also privacy-wise). It could make sense to apply additional checks to the returned fee rates (i.e., also somewhat bound them), but this is generally pretty delicate as we'd still need to make sure that we can handle legitimate fee rate spikes.
> 
> As for falling back to default values - yes, that's somewhat legitimate also but a) different networks (e.g., regtest) probably need to be handled differently and b) if we do this, we'd probably at the very least implement the same behavior across all three chain sources?

Thanks for the review @tnull , and no worries about the delay!

 (a): turns out regtest doesn't serve estimates at all. I ran the electrs build
  the tests use (esplora_a33e97e1) against bitcoind 29.0:

      $ curl localhost:3002/fee-estimates
      {}

  estimatesmartfee has no data on regtest, so there's nothing for convert_fee_rate
  to pick up. That means the fallback decides every fee rate on regtest/signet, not
  just a rare sparse-map case, and as written this PR moves all of them onto
  mainnet-sized defaults: UrgentOnChainSweep 253 -> 5000 sat/kwu, ChannelFunding
  253 -> 1000. That seems clearly wrong for a chain with no fee market.

  What I'd do instead: per-target defaults on mainnet/testnet, keep the 1 sat/vb
  behavior on regtest/signet, i.e. the same split bitcoind.rs:612-621 already does.

(b): yes, agreed.

   On the apply_post_estimation_adjustments question from the description: I don't
  think the fallback should go through it. 8000 has been the no-estimate ceiling
  since 2024, and the adjustment was added later (02208ab) to relax the
  estimate-derived values, without touching the fallbacks. Sending the default
  through it would also mean the cache no longer matches what estimate_fee_rate
  returns on a cache miss. If 8000 is too low nowadays that feels like a separate
  change to get_fallback_rate_for_ldk_target.

PR #1043, opened 2026-08-11
Fix Python bindings build with recent `hatchling` releases

vincenzopalazzo:
## Summary

`check-python` is currently failing on `main` (and consequently on every open PR). Recent `hatchling` releases validate that `project.readme` resolves inside the project directory, so our `readme = "../../README.md"` aborts the build before anything else runs:

```
ValueError: Readme path must be within the project directory: ../../README.md
```

This affects `scripts/python_build_wheel.sh` too, not just CI.

Rather than duplicating the README into `bindings/python/`, this declares `readme` dynamic and reads the repository README from a custom metadata hook, which passes it along as literal text and hence skips the path validation.

## Test plan

- [x] `uv pip install -e bindings/python` succeeds again (fails on `main` today)
- [x] `uv build --wheel` succeeds, and the resulting `METADATA` has `Description-Content-Type: text/markdown` with a body byte-identical to the repository `README.md` — so the published PyPI description is unchanged

## Disclosure

This PR was prepared with AI assistance (Claude Code).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/bindings/python/hatch_build.py
+++ b/bindings/python/hatch_build.py
@@ -1,4 +1,16 @@
+import os
+
 from hatchling.builders.hooks.plugin.interface import BuildHookInterface
+from hatchling.metadata.plugin.interface import MetadataHookInterface
+
+
+class CustomMetadataHook(MetadataHookInterface):
+    # Hatchling refuses a `readme` path outside the project directory, so we read the
+    # repository README ourselves and hand it over as literal text instead.
+    def update(self, metadata):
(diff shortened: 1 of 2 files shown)


vincenzopalazzo:
Superseded by #1041, which fixes this with the same `dynamic = ["readme"]` + metadata-hook approach as part of the broader wheel-release work. I missed that it was already open — sorry for the noise.

PR #845, opened 2026-03-25
Add BOLT 12 payer proof support

vincenzopalazzo:
## Summary

Adds support for building [BOLT 12 payer proofs](https://github.com/lightningdevkit/rust-lightning/pull/4297) for previously succeeded outbound BOLT 12 payments:

- Persist the paid BOLT 12 invoice on `PaymentKind::Bolt12Offer` / `PaymentKind::Bolt12Refund`, set from `Event::PaymentSent` once the payment succeeds.
- Expose `Bolt12Payment::create_payer_proof`, which builds a payer proof from the persisted invoice and payment preimage. `PayerProofOptions` controls which optional invoice fields (offer description, issuer, amount, creation time, extra TLV types) are selectively disclosed, plus an optional proof note.
- Wire the new `PayerProof` type and error variants (`PayerProofCreationFailed`, `PayerProofUnavailable`, `InvalidPayerProof`) through the UniFFI surface.

Since the proof is built purely from data persisted in the payment store, payer proofs survive restarts without any additional storage: no new KV namespace and no in-memory side store. Payments that completed via a static invoice (async payments) do not support payer proofs and are rejected with `PayerProofUnavailable`.

The payer proof primitives are already present in the `rust-lightning` revision we pin, so this PR does not patch LDK to a fork — `Cargo.toml` is untouched.

## Verification

- `cargo fmt --all -- --check`
- `cargo check --lib --tests --benches`, `cargo check --lib --features uniffi`
- `cargo clippy --lib -- -A warnings -D clippy::unwrap_used -A clippy::tabs_in_doc_comments`
- `cargo test --lib`, `cargo test --test upgrade_downgrade_tests`
- `cargo test --test integration_tests_rust simple_bolt12_send_receive` — extended to build a payer proof from the persisted payment and assert the payment hash, disclosed amount, proof note, and that undisclosed fields stay absent
- Python UniFFI bindings generation (`scripts/uniffi_bindgen_generate_python.sh`)

## Disclosure

This PR was prepared with AI assistance (Claude Code).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1045, opened 2026-08-12
Add stateless BOLT 12 payer proof support

vincenzopalazzo:
Supersedes #845, reworked per the review feedback there: no payment-store persistence, and the API is stateless so we stay free to change or drop it once the verification side is worked out.

## Summary

Exposes `Bolt12Payment::create_payer_proof`, which builds a BOLT 12 payer proof for a payment this node made:

```rust
let proof = node.bolt12_payment().create_payer_proof(
    payment_id,        // from Event::PaymentSuccessful
    payment_preimage,  // from Event::PaymentSuccessful
    paid_invoice,      // Event::PaymentSuccessful::bolt12_invoice
    Some(options),
)?;
```

`PayerProofOptions` controls which optional invoice fields are selectively disclosed (offer description, offer issuer, invoice amount, invoice creation time, plus arbitrary extra TLV types) and lets the payer attach a note. A proof always commits to the payer id, payment hash, and issuer signing pubkey.

Nothing is persisted: `PaymentKind` and the payment store are untouched, so there's no serialization change and no compatibility note. Callers hold on to the invoice themselves if they want to build a proof later.

Payments settled via a static invoice, i.e., async payments, can't be proven this way and are rejected with `PayerProofUnavailable`.

## Why a node-side method at all

It was suggested that, since the invoice is already exposed on `Event::PaymentSuccessful`, we could skip the API and just document building a proof with LDK's methods directly. That doesn't work today — both LDK entry points need key material LDK Node deliberately doesn't expose:

- `PaidBolt12Invoice::prove_payer_derived(preimage, expanded_key, payment_id, secp)` takes an `&ExpandedKey`.
- `PaidBolt12Invoice::prove_payer(preimage)` avoids that, but returns an `UnsignedPayerProof` that must be signed with the *derived payer* key, which means `Bolt12Invoice::derive_payer_signing_keys(expanded_key, secp)` — `ExpandedKey` again.

In LDK Node `ExpandedKey` is only reachable via `KeysManager`, which is `pub(crate)`, and `Node::sign_message` is a node-key BIP137 message signer, so it's no help here. Hence one thin method that supplies the expanded key; everything else is caller-provided.

Happy to expose a lower-level signing hook instead if that's preferred.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo check --lib --tests`, `cargo check --lib --features uniffi`
- [x] `cargo clippy --lib -- -A warnings -D clippy::unwrap_used -A clippy::tabs_in_doc_comments`
- [x] `cargo test --lib`
- [x] `cargo test --test integration_tests_rust simple_bolt12_send_receive` — extended to build a proof from values captured off the event, asserting the preimage, disclosed amount and note, and that undisclosed fields stay absent
- [x] Python UniFFI bindings generate

## Disclosure

This PR was prepared with AI assistance (Claude Code).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/ffi/types.rs
+++ b/src/ffi/types.rs
@@ -23,7 +23,6 @@ use bitcoin::hashes::Hash;
 use bitcoin::secp256k1::PublicKey;
 pub use bitcoin::{Address, BlockHash, Network, OutPoint, ScriptBuf, Txid};
 pub use lightning::chain::channelmonitor::BalanceSource;
-use lightning::events::PaidBolt12Invoice as LdkPaidBolt12Invoice;
 pub use lightning::events::{ClosureReason, PaymentFailureReason};
 use lightning::ln::channel_state::{ChannelShutdownState, CounterpartyForwardingInfo};
 use lightning::ln::channelmanager::PaymentId;
@@ -32,6 +31,9 @@ pub use lightning::ln::types::ChannelId;
 use lightning::offers::invoice::Bolt12Invoice as LdkBolt12Invoice;
 pub use lightning::offers::offer::OfferId;
 use lightning::offers::offer::{Amount as LdkAmount, Offer as LdkOffer};
+use lightning::offers::payer_proof::{
+	PaidBolt12Invoice as LdkPaidBolt12Invoice, PayerProof as LdkPayerProof,
+};
 use lightning::offers::refund::Refund as LdkRefund;
 use lightning::offers::static_invoice::StaticInvoice as LdkStaticInvoice;
 use lightning::onion_message::dns_resolution::HumanReadableName as LdkHumanReadableName;
@@ -840,6 +842,17 @@ pub enum PaidBolt12Invoice {
 	Static(Arc<StaticInvoice>),
 }
 
+impl PaidBolt12Invoice {
+	/// Returns the [`Bolt12Invoice`] if the payment was for a standard BOLT 12 invoice, and
+	/// `None` for a static invoice, i.e., an async payment, which can't be proven.
+	pub fn bolt12_invoice(&self) -> Option<Arc<Bolt12Invoice>> {
+		match self {
+			PaidBolt12Invoice::Bolt12(invoice) => Some(Arc::clone(invoice)),
+			PaidBolt12Invoice::Static(_) => None,
+		}
+	}
+}
+
 impl From<LdkPaidBolt12Invoice> for PaidBolt12Invoice {
 	fn from(ldk: LdkPaidBolt12Invoice) -> Self {
 		match ldk {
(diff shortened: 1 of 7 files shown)


src/payment/bolt12.rs
```diff
@@ -389,6 +425,69 @@ impl Bolt12Payment {
...
+	/// store, so it's up to you to hold on to the invoice if you want to build a proof later on.
+	///
+	/// Note that payments settled via a static invoice, i.e., async payments, can't be proven this
+	/// way and will fail with [`Error::PayerProofUnavailable`].
+	///
+	/// [BOLT 12]: https://github.com/lightning/bolts/blob/master/12-offer-encoding.md
+	/// [`Event::PaymentSuccessful`]: crate::Event::PaymentSuccessful
+	pub fn create_payer_proof(
+		&self, payment_id: PaymentId, payment_preimage: PaymentPreimage,
+		paid_invoice: PaidBolt12Invoice, options: Option<PayerProofOptions>,
```
tnull: I do wonder if this should take `Bolt12Invoice` rather than `PaidBolt12Invoice`, which would avoid the entire `PayerProofError::IncompatibleInvoice` case AFAIU?

tnull (commented):
One question, otherwise LGTM

Feel free to undraft.

src/payment/bolt12.rs
vincenzopalazzo: **Fixed in aa9f3fb.** Good catch -- `create_payer_proof` now takes `&Bolt12Invoice` and wraps it into `PaidBolt12Invoice::Bolt12Invoice` internally, so `IncompatibleInvoice` is unreachable. That let me drop the `Error::PayerProofUnavailable` variant and the error-mapping helper entirely, and it pushes the "is this provable?" question to the call site, where `PaidBolt12Invoice::bolt12_invoice()` already returns `None` for static invoices.

src/ffi/types.rs
```diff
@@ -881,6 +883,93 @@ impl Readable for PaidBolt12Invoice {
...
+/// created; the signatures and Merkle root needed to actually verify a proof aren't surfaced
+/// yet, as the verification API is still being designed.
+#[derive(Debug, Clone, uniffi::Object)]
+#[uniffi::export(Debug, Display)]
+pub struct PayerProof {
+	pub(crate) inner: LdkPayerProof,
+}
+
+#[uniffi::export]
+impl PayerProof {
```
tnull, line 909: This omits `issuer_signing_pubkey` and `payer_signing_pubkey`?

src/payment/bolt12.rs
```diff
@@ -52,6 +56,33 @@ type HumanReadableName = lightning::onion_message::dns_resolution::HumanReadable
...
+///
+/// [BOLT 12]: https://github.com/lightning/bolts/blob/master/12-offer-encoding.md
+#[derive(Clone, Debug, PartialEq, Eq, Default)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct PayerProofOptions {
+	/// An optional note to attach to the payer proof itself.
+	pub note: Option<String>,
+	/// Whether to disclose the offer description.
+	pub include_offer_description: bool,
+	/// Whether to disclose the offer issuer.
```
tnull, line 79: >   1. [P2] Advertised arbitrary TLV disclosure is missing. The PR description (https://github.com/lightningdevkit/ldk-node/pull/1045) promises additional TLV selection, but /home/tnull/worktrees/ldk-node/pr-1045-review-20260812/src/payment/bolt12.rs:78 has no extra_tlv_types, and
>      create_payer_proof never calls LDK’s include_type.

src/ffi/types.rs
```diff
@@ -881,6 +883,93 @@ impl Readable for PaidBolt12Invoice {
...
+	pub fn proof_note(&self) -> Option<String> {
+		self.inner.proof_note().map(|value| value.to_string())
+	}
+
+	/// The raw TLV bytes of the proof.
+	pub fn bytes(&self) -> Vec<u8> {
+		self.inner.bytes().to_vec()
+	}
+
+	/// The bech32-encoded string form of the proof.
```
tnull, line 989: >   4. [P3] The string representation cannot round-trip through UniFFI. /home/tnull/worktrees/ldk-node/pr-1045-review-20260812/src/ffi/types.rs:947 is advertised for transporting proofs, but the wrapper only provides from_bytes; LDK supports FromStr, as do analogous wrappers in this file.

src/payment/bolt12.rs
```diff
@@ -52,6 +56,33 @@ type HumanReadableName = lightning::onion_message::dns_resolution::HumanReadable
...
+#[cfg(not(feature = "uniffi"))]
+type PayerProof = LdkPayerProof;
+#[cfg(feature = "uniffi")]
+type PayerProof = Arc<crate::ffi::PayerProof>;
+
+/// Options controlling which optional fields are disclosed in a [BOLT 12] payer proof.
+///
+/// A payer proof always commits to the payer id, the payment hash, and the issuer signing
+/// pubkey. Everything else is disclosed only if requested here, allowing to reveal just as much
+/// of the invoice as the verifier needs to see.
```
tnull: >   3. [P3] The selective-disclosure documentation is inaccurate. /home/tnull/worktrees/ldk-node/pr-1045-review-20260812/src/payment/bolt12.rs:69 says everything except three required fields is disclosed only when requested. Pinned LDK also automatically includes invoice features when present.

tests/integration_tests_rust.rs
```diff
@@ -2592,18 +2592,47 @@ async fn simple_bolt12_send_receive() {
...
+			ref payment_preimage,
+			..
+		} => {
 			println!("{} got event {:?}", node_a.node_id(), e);
 			assert_eq!(*evt_id, payment_id);
 			assert!(
 				bolt12_invoice.is_some(),
 				"bolt12_invoice should be present for BOLT12 payments"
 			);
+			let invoice = bolt12_invoice.as_ref().unwrap().bolt12_invoice().unwrap().clone();
```
tnull, line 2608: I think this fails to compile under `--features uniffi`

src/ffi/types.rs
vincenzopalazzo: **Fixed in d848ff5.** Restored `payer_signing_pubkey` and `issuer_signing_pubkey`, plus `invoice_signature`, `proof_signature` and `merkle_root` -- I had trimmed all five in an earlier "make it smaller" pass, reasoning they were only useful to a verifier and verification is still undesigned. That was the wrong call: it left bindings users without parity with LDK's `PayerProof`. Full accessor set is back.

src/payment/bolt12.rs
vincenzopalazzo: **Fixed in d848ff5.** `extra_tlv_types` is back on `PayerProofOptions`, and `create_payer_proof` calls `include_type` for each. Same earlier over-trim as the accessors above -- the PR description was the accurate one, the code was what drifted.

src/ffi/types.rs
vincenzopalazzo: **Fixed in d848ff5.** Added a `from_str` constructor and a `FromStr` impl on the FFI `PayerProof`, matching `Offer`/`Refund`/`Bolt12Invoice` in that file, so `as_string()` now round-trips.

src/payment/bolt12.rs
vincenzopalazzo: **Fixed in d848ff5.** The doc now says the proof always commits to payer id, payment hash and issuer signing pubkey, and additionally discloses the invoice features whenever the invoice carries any -- which is what `default_included_types` does in the pinned LDK.

tests/integration_tests_rust.rs
vincenzopalazzo: **Fixed in d848ff5.** You were right -- it did not compile. Under `uniffi` the event's `bolt12_invoice` is the FFI `PaidBolt12Invoice`, which had no `bolt12_invoice()` method. My fault for only ever running `cargo check --lib --features uniffi` and never `--tests`, so I never compiled that line against the FFI type.

Rather than cfg-gating the test, I gave the FFI enum the same accessor LDK's type has, returning `Option<Arc<Bolt12Invoice>>`, so the identical line compiles both ways -- and bindings users get a first-class way to pull the provable invoice out instead of hand-matching variants. Verified with `cargo check --tests --features uniffi` and by running `simple_bolt12_send_receive` under both feature sets.

tnull (approved):
LGTM.

Two non-blocking nits: commit description still mentions `PayerProofUnavailable` and the commit shows as 'Unverified' here on Github, might want to start signing commits.

PR #1047, opened 2026-08-12
Use default CI workflow timeouts

tnull:
Two-hour job limits continue to cancel otherwise healthy CI runs when integration tests are slow. Remove the overrides from every workflow so GitHub Actions applies its platform defaults.

### DIFF
--- a/.github/workflows/0fc-integration.yml
+++ b/.github/workflows/0fc-integration.yml
@@ -8,7 +8,6 @@ concurrency:
 
 jobs:
   build-and-test:
-    timeout-minutes: 120
     runs-on: self-hosted
     steps:
       - name: Checkout source code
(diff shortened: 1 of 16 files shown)


benthecarman:
When they finish they don't normally very long. If they are hitting the timeouts I assume it's because there is a hang in the test

tnull:
> When they finish they don't normally very long. If they are hitting the timeouts I assume it's because there is a hang in the test

Hmm, I think I saw some that we're just aborted mid-test. Let's wait until the other CI fixes that are pending have been merged to see if it's still necessary.

Jolah1:
for context, removing these entries changes the timeout from 120 minutes to GitHub’s 360-minute default. Since the observed issue appears limited to Rust test jobs, retaining explicit limits and increasing only the affected job’s timeout would seem safer than removing them from all workflows. 
Waiting for the pending CI fixes first, as suggested above, also makes sense.

PR #1048, opened 2026-08-12
fix: drop stale monitor lightning fork

ben-kaufman:
This PR drops the stale-monitor Lightning fork and pins upstream rust-lightning v0.2.2 by immutable revision.

### Description

Bitkit no longer recovers from stale channel monitors, so the Synonym `0.2.2-accept-stale-monitors` branch is unused. This removes `set_accept_stale_channel_monitors` and the healing path.

Lightning crates cannot all come from crates.io `0.2.2` because `lightning-block-sync` 0.2.2 is unpublished. They are patched to upstream `lightningdevkit/rust-lightning` at rev `0695da99` (v0.2.2), not a mutable Git branch.

Checked-in Swift/Kotlin/Python bindings are not regenerated here. The next rc bindgen must drop `set_accept_stale_channel_monitors`.

Related: [synonymdev/audit#55](https://github.com/synonymdev/audit/issues/55) (BITKIT-SEC-004).

#### Manual Tests
- [ ] **1.** Consumer app → start a healthy node: build and start succeed.
- [ ] **2.** `regression:` stale ChannelMonitor vs ChannelManager: build fails closed with `DangerousValue`.

#### Automated Checks
- `cargo check` passed after switching the patch pin and removing the stale-monitor API.

Made with [Cursor](https://cursor.com)


ben-kaufman:
Opened against the wrong repository. This change belongs on synonymdev/ldk-node.

PR #1052, opened 2026-08-16
Fix uniffi feature breaking postgres/migration integration tests

juliusjulyp:
## Context

`cargo test --all-features` fails to compile `tests/integration_tests_postgres.rs`
and `tests/integration_tests_migration.rs` with  type
mismatches (`Arc<NodeEntropy>` expected, `NodeEntropy` found; `Vec<u8>`
expected, `[u8; 64]` found; `Node` vs `Arc<Node>`).

## Cause

The `uniffi` feature changes the public shape of `NodeEntropy`, `Node`, and
related builder methods for FFI  — wrapping return types in `Arc<T>`
and switching `NodeEntropy::from_seed_bytes` to take `Vec<u8>` instead of a
fixed `[u8; 64]` array. 

## Fix

Both files now follow the same `#[cfg(feature = "uniffi")]` /
`#[cfg(not(feature = "uniffi"))]`  pattern similar
to `tests/integration_tests_rust.rs` and `tests/common/mod.rs`

## Testing

`cargo test --features postgres --test integration_tests_postgres` —
  3/3 passing, against  local Postgres instance

Open to feedback if a different approach is preferred — e.g. expanding CI
coverage to catch this combination going forward, in addition to or instead
of this fix

### DIFF
--- a/tests/integration_tests_migration.rs
+++ b/tests/integration_tests_migration.rs
@@ -88,10 +88,12 @@ macro_rules! with_opened_store {
 async fn build_migration_node(
 	instance: &BackendInstance, node_config: ldk_node::config::Config, node_entropy: NodeEntropy,
 	esplora_url: &str,
-) -> ldk_node::Node {
+) -> common::TestNode {
 	let mut builder = Builder::from_config(node_config);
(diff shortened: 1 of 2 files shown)


tnull (approved):
Thanks!

PR #1054, opened 2026-08-18
Prevent Electrum updates during shutdown

tnull:
Based on #997 for now, will rebase once that's merged.

```
A timed-out synchronous Electrum sync can outlive its driving task
and continue calling Confirm implementations while node shutdown is
draining persistence work.

Close a gate before task cancellation so no new callbacks can start.
Let callbacks already in progress finish atomically before shutdown
continues.
```

### DIFF
--- a/src/chain/electrum.rs
+++ b/src/chain/electrum.rs
@@ -7,7 +7,7 @@
 
 use std::collections::{HashMap, HashSet};
 use std::sync::atomic::{AtomicBool, Ordering};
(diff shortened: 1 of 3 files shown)


tankyleo:
Needs rebase

tnull:
> Needs rebase

Rebased.

tankyleo (approved):
CI failure looks unrelated, I pushed https://github.com/lightningdevkit/ldk-node/pull/1063 that should fix it.

PR #1058, opened 2026-08-19
Return an error from `NodeEntropy::from_bip39_mnemonic` in bindings

kaloudis:
The uniffi-exposed `from_bip39_mnemonic` constructor was infallible and took the `Mnemonic` custom type, which parses the string during argument lifting. An invalid mnemonic would fail the lift and surface as an unexpected-error call status. The generated Swift wrapper for an infallible function wraps the call in `try!`, so passing an invalid mnemonic (eg. a user typo during wallet restore) aborts the process with an uncatchable EXC_BREAKPOINT. Kotlin and Python raise their generic internal exceptions instead.

Following the `from_seed_bytes` precedent, give the constructor a uniffi-specific signature that takes the mnemonic as a plain string and returns `Result<NodeEntropy, EntropyError>`, parsing inside the function and reporting failures via a new
`EntropyError::InvalidMnemonic` variant. The non-uniffi Rust API is unchanged.

Discovered in an ZEUS iOS user crash report.

### DIFF
--- a/src/entropy.rs
+++ b/src/entropy.rs
@@ -19,6 +19,8 @@ use crate::io;
 #[derive(Debug, Clone, PartialEq)]
 #[cfg_attr(feature = "uniffi", derive(uniffi::Error))]
 pub enum EntropyError {
+	/// The given BIP 39 mnemonic is invalid.
+	InvalidMnemonic,
 	/// The given seed bytes are invalid, e.g., have invalid length.
 	InvalidSeedBytes,
 	/// The given seed file is invalid, e.g., has invalid length, or could not be read.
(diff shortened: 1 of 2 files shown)


tnull:
Ugh, sorry. We'll be working towards getting rid of all custom type to avoid these cases, seems we missed this so far :/

I think we should however do https://github.com/lightningdevkit/ldk-node/pull/1059, i.e., expose `Mnemonic` as a proper interface to bindings users, which also exposes more functionality and allows us to clean up some legacy code.

tnull:
Closing as superseded by #1059

PR #1013, opened 2026-08-04
Make mnemonic RNG selection explicit

benthecarman:
Mnemonic generation is infrequent, so source randomness directly from OS entropy instead of initializing a thread-local generator. Also given recent events, better have this more explicitly defined than in a dep.

PR #1059, opened 2026-08-19
Expose `Mnemonic` as interface to bindings users

tnull:
```
Invalid mnemonic strings currently fail during implicit custom-type
lifting, which leaves generated bindings without a catchable validation
error.

Parse mnemonic phrases through a fallible object constructor while
keeping bip39::Mnemonic in the native Rust API.
```

And, since we now expose `Mnemonic::generate` to bindings users directly, we can get rid of our global `generate_entropy_mnemonic` helper.

### DIFF
--- a/src/ffi/types.rs
+++ b/src/ffi/types.rs
@@ -151,7 +151,7 @@ impl VssClientHeaderProvider for VssHeaderProviderAdapter {
 
 use crate::builder::sanitize_alias;
 pub use crate::config::{default_config, ElectrumSyncConfig, EsploraSyncConfig, TorConfig};
-pub use crate::entropy::{generate_entropy_mnemonic, NodeEntropy, WordCount};
+pub use crate::entropy::NodeEntropy;
 use crate::error::Error;
 pub use crate::liquidity::LSPS1OrderStatus;
 pub use crate::logger::{LogLevel, LogRecord, LogWriter};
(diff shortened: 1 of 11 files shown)


src/ffi/types.rs
```diff
@@ -1151,15 +1151,107 @@ uniffi::custom_type!(BlockHash, String, {
...
+		Bip39Mnemonic::from_entropy(entropy).map(Self::from).map_err(|_| Error::InvalidMnemonic)
+	}
+
+	/// Generates a random English mnemonic with the specified word count.
+	///
+	/// Defaults to 24 words when no word count is specified.
+	#[uniffi::constructor]
+	pub fn generate(word_count: Option<WordCount>) -> Self {
+		let word_count = word_count.unwrap_or(WordCount::Words24).word_count();
+		let inner = Bip39Mnemonic::generate(word_count)
```
benthecarman: Can we do the same as #1013 here

src/ffi/types.rs
```diff
@@ -1151,15 +1151,107 @@ uniffi::custom_type!(BlockHash, String, {
...
+	}
+
+	/// Returns the indices of the mnemonic's words in the English BIP 39 word list.
+	pub fn word_indices(&self) -> Vec<u16> {
+		self.inner.word_indices().map(|index| index as u16).collect()
+	}
+
+	/// Returns the number of words in the mnemonic.
+	pub fn word_count(&self) -> u8 {
+		self.inner.word_count() as u8
```
benthecarman, line 1197: changing the type here is kinda weird

benthecarman (commented):
nits otherwise lgtm

src/ffi/types.rs
tnull: Hmm, no, I'd prefer to just call through to `bip39` here, though that should do it soon, see https://github.com/rust-bitcoin/rust-bip39/pull/105 (we just need to get a release out)

src/ffi/types.rs
tnull: Yeah, unfortunately no usize in uniffi. Would you prefer we just no expose the method?

tnull: Could also wait until after https://github.com/rust-bitcoin/rust-bip39/pull/98 lands and ships, when we'll re-introduce `WordCount` (just seemed weird to keep it for bindings only for now)

benthecarman: This is fine for now, we can move to that in the future. No reason we'd ever break the u8 here for any practical reason

benthecarman (dismissed):

tnull:
Amended with some fixes to make CI pass.

benthecarman (approved):

kaloudis:
Thanks for fixing!

PR #1065, opened 2026-08-20
Ignore late failures for successful payments

thesimplekid:
rust-lightning documents that PaymentFailed can arrive after PaymentSent in rare cases. In that ordering, the failure must be ignored and the payment must be treated as successful:

https://github.com/lightningdevkit/rust-lightning/blob/9174965af9437196c527a9aa0df36bbcf050c8bb/lightning/src/events/mod.rs#L1230-L1233

Keep succeeded outbound Lightning records monotonic and suppress the contradictory user-facing PaymentFailed event. Cover both BOLT11 and BOLT12 on the persistence-backed store path.

Developed with assistance from OpenAI Codex.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -1457,13 +1457,6 @@ where
 				};
 			},
 			LdkEvent::PaymentFailed { payment_id, payment_hash, reason, .. } => {
-				log_info!(
-					self.logger,
-					"Failed to send payment with ID {} due to {:?}.",
-					payment_id,
-					reason
-				);
-
 				let update = PaymentDetailsUpdate {
 					hash: Some(payment_hash),
 					status: Some(PaymentStatus::Failed),
@@ -1476,6 +1469,32 @@ where
 						return Err(ReplayEvent());
 					},
 				};
+
+				// LDK may emit `PaymentFailed` after `PaymentSent` in exceedingly rare cases.
+				// The payment-store update above preserves success in that case; re-read the
+				// resulting state so we also avoid surfacing a contradictory public event.
+				match self.payment_store.get(&payment_id).await {
+					Ok(Some(payment)) if payment.status == PaymentStatus::Succeeded => {
+						log_info!(
+							self.logger,
+							"Ignoring late payment failure for already-succeeded payment with ID {}.",
+							payment_id
+						);
+						return Ok(());
+					},
+					Ok(_) => {},
+					Err(e) => {
+						log_error!(self.logger, "Failed to access payment store: {}", e);
+						return Err(ReplayEvent());
+					},
+				}
+
+				log_info!(
+					self.logger,
+					"Failed to send payment with ID {} due to {:?}.",
+					payment_id,
+					reason
+				);
 
 				let event = Event::PaymentFailed { payment_id, payment_hash, reason };
 				match self.event_queue.add_event(event).await {
--- a/src/payment/store.rs
+++ b/src/payment/store.rs
@@ -317,7 +317,20 @@ impl StorableObject for PaymentDetails {
 		}
 
 		if let Some(status) = update.status {
-			update_if_necessary!(self.status, status);
+			// LDK may, in exceedingly rare cases, emit `PaymentFailed` after
+			// `PaymentSent` for the same outbound payment. In that case the
+			// failure must be ignored and the payment must remain succeeded.
+			//
+			// Keep this invariant scoped to outbound Lightning payments as
+			// on-chain payment state may legitimately be revised after a reorg.
+			let is_outbound_lightning_payment = self.direction == PaymentDirection::Outbound
+				&& !matches!(self.kind, PaymentKind::Onchain { .. });
+			let is_late_failure = is_outbound_lightning_payment
+				&& self.status == PaymentStatus::Succeeded
+				&& status == PaymentStatus::Failed;
+			if !is_late_failure {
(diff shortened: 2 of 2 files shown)


src/event.rs
```diff
@@ -1477,6 +1470,32 @@ where
...
+				log_info!(
+					self.logger,
+					"Failed to send payment with ID {} due to {:?}.",
+					payment_id,
+					reason
+				);
+
 				let event = Event::PaymentFailed { payment_id, payment_hash, reason };
 				match self.event_queue.add_event(event).await {
 					Ok(_) => return Ok(()),
```
joostjager, line 1501: This is again related to the split-brain situation with persistence, but can the initial payment store write fail after LDK accepts the payment? `PaymentSent` then accepts Ok(NotFound) and emits success, so a replayed late failure could escape this check after restart.

src/payment/store.rs
```diff
@@ -317,7 +317,20 @@ impl StorableObject for PaymentDetails {
 		}
 
 		if let Some(status) = update.status {
-			update_if_necessary!(self.status, status);
+			// LDK may, in exceedingly rare cases, emit `PaymentFailed` after
```
tnull: I think we should avoid the special case logic here.

src/event.rs
```diff
@@ -1477,6 +1470,32 @@ where
 					},
 				};
 
+				// LDK may emit `PaymentFailed` after `PaymentSent` in exceedingly rare cases.
+				// The payment-store update above preserves success in that case; re-read the
+				// resulting state so we also avoid surfacing a contradictory public event.
+				match self.payment_store.get(&payment_id).await {
```
tnull: Let's avoid the second store lookup. Rather, we can change above call from `update` to use `mutate`, and then only override the full payment store entry in case we hadn't seen a succeeded Lightning payment.

tnull (commented):
Excuse the delay here, good catch.

src/event.rs
thesimplekid: Addressed in https://github.com/lightningdevkit/ldk-node/pull/1065/commits/eab5e3639704897c63772950621006729dbf2504

src/payment/store.rs
thesimplekid: Addressed in https://github.com/lightningdevkit/ldk-node/commit/eab5e3639704897c63772950621006729dbf2504

src/event.rs
```diff
@@ -1472,34 +1523,15 @@ where
 				};
 			},
 			LdkEvent::PaymentFailed { payment_id, payment_hash, reason, .. } => {
-				log_info!(
-					self.logger,
-					"Failed to send payment with ID {} due to {:?}.",
+				return Self::handle_payment_failed(
```
tnull: No, please keep this inline for now. Arguably we might eventually split the event handler into individual handler methods, but if we go there we should make it consistent. For now it's fine to leave it inlined, also fine to drop/amend the test that depends on it.

src/payment/store.rs
```diff
@@ -1617,6 +1617,10 @@ mod bounded_cache_tests {
 		let data_store = new_bounded_payment_store(1);
 
 		let mut stored = bolt11_payment(1);
+		stored.status = PaymentStatus::Pending;
```
tnull: It seems this is an unnecessary leftover change?

src/event.rs
thesimplekid: Had a feeling you might say that. Addressed in https://github.com/lightningdevkit/ldk-node/pull/1065/commits/1accb18802bcd24657a37d81db4171b29ffde793

src/payment/store.rs
thesimplekid: Yes removed, thanks.

src/event.rs
```diff
@@ -1472,26 +1472,50 @@ where
...
 			},
 			LdkEvent::PaymentFailed { payment_id, payment_hash, reason, .. } => {
+				let mut ignore_late_failure = false;
+				self.payment_store
+					.mutate(&payment_id, |current| {
+						let current = current?;
+						// LDK may emit PaymentFailed after PaymentSent. Ignore the entire late
+						// failure so it cannot alter payment metadata or emit a contradictory event.
+						if current.direction == PaymentDirection::Outbound
+							&& !matches!(current.kind, PaymentKind::Onchain { .. })
```
tnull: Hmm, going forward we might add additional variants to `PaymentKind`, so matching by negation like this is a footgun. Let's add a prefactor commit that adds `PaymentKind::is_lightning_payment` or similar that matches on the Lightning variants. Then use it here.

tnull (commented):
Please squash, one minor comment though.

src/event.rs
thesimplekid: https://github.com/lightningdevkit/ldk-node/pull/1065/commits/e0e6dda86eed808c736186753f7d6f94f0d02342

tnull (commented):
LGTM, please squash.

thesimplekid:
Sorry, squashed into 1 commit now when you said prefactor commit I thought you wanted that one on its own.

tnull:
> Sorry, squashed into 1 commit now when you said prefactor commit I thought you wanted that one on its own.

Ah, indeed. Just skimmed the history and saw the `fix` so I assumed it was a fixup. Sorry for the confusion, but doesnt' matter all too much now.

tnull (approved):

PR #1066, opened 2026-08-20
Stabilize the connection restart test

tankyleo:
```
    Stabilize the connection restart test

    Give a few seconds for the connection to occur before asserting on it.
    Also verify the inbound peer's connected flag instead of rechecking the
    outbound peer.

    Co-Authored-By: HAL 9000
```


tankyleo:
This should fix the CI failure we hit here https://github.com/lightningdevkit/ldk-node/actions/runs/32375903151/job/96447097507?pr=1064

tnull (approved):

PR #1069, opened 2026-08-26
Rescan the `wallet_rescan_from_height` block

tankyleo:
```
    Rescan the `wallet_rescan_from_height` block

    It is surprising to have BDK rescan from one block above the given
    `wallet_rescan_from_height` block. So we tweak things to have the block
    height that the user gives us included in the rescan.

    Co-Authored-By: HAL 9000
```


benthecarman (dismissed):
Lgtm!

tankyleo:
Sorry for the churn ben, making sure the test actually keeps testing the previous case, which is to pick up transactions confirmed a few blocks after the rescan height.

benthecarman (approved):

tnull (commented):
Post-merge ACK

PR #1071, opened 2026-08-27
Require a multithreaded runtime when setting up the node

elnosh:
We already note on `Node::start` that we require the outer runtime to be of the multithreaded flavor, but we never check it. 
Given that this [block_in_place](https://github.com/lightningdevkit/ldk-node/blob/6480bff115bda0895009ca2cdf73b7f5957d639c/src/runtime.rs#L188) panics if called from a current-thread runtime, it seems we should error rather than just silently allow it.

https://docs.rs/tokio/latest/tokio/task/fn.block_in_place.html#panics

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -367,10 +367,18 @@ impl NodeBuilder {
 	///
 	/// If not provided, the node will spawn its own runtime or reuse any outer runtime context it
 	/// can detect.
+	///
+	/// Note we require the given runtime to be of the `multithreaded` flavor.
 	#[cfg_attr(feature = "uniffi", allow(dead_code))]
-	pub fn set_runtime(&mut self, runtime_handle: tokio::runtime::Handle) -> &mut Self {
+	pub fn set_runtime(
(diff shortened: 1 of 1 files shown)


src/builder.rs
```diff
@@ -367,6 +367,8 @@ impl NodeBuilder {
 	///
 	/// If not provided, the node will spawn its own runtime or reuse any outer runtime context it
 	/// can detect.
+	///
+	/// Note we require the given runtime to be of the `multithreaded` flavor.
 	#[cfg_attr(feature = "uniffi", allow(dead_code))]
 	pub fn set_runtime(&mut self, runtime_handle: tokio::runtime::Handle) -> &mut Self {
```
tnull: Can we return a new `BuildError` if the given handle does not belong to a multithreaded runtime?

tnull (commented):
Yeah, I guess it's better to be explicit about it. One comment though.

src/builder.rs
elnosh: done

tnull (approved):

PR #1086, opened 2026-09-03
Wind down spawned tasks when startup fails

### LINKED ISSUE #1009: Properly wind down all spawned tasks on any startup error
At the moment if we fail to bind to any of our listening addresses, we return a startup error, but do not properly wind down the wallet sync, RGS, and pathfinding scores tasks, which are started before we attempt to bind our listeners.

Jolah1:
Fixes #1009.

`Node::start` spawns background tasks — wallet sync, RGS gossip, pathfinding scores — before it can still fail, most visibly when resolving or binding the configured listening addresses. The error arm only called `chain_source.stop()`, so those tasks kept running behind a node that never came up.

This extracts the wind-down sequence out of `Node::stop` into a private `Node::shutdown` and runs it on any `start_inner` error.

**Reviewer note — relaxed assertions.** Because the wind-down now also runs after a *partial* startup, it can no longer assume every task exists:

- Both shutdown `watch::Sender::send` calls are allowed to find no receivers (previously `log_error!` + `debug_assert!(false)`).
- `Runtime::wait_on_background_tasks`' `debug_assert!(tasks.len() > 0)` and `Runtime::wait_on_background_processor_task`' `debug_assert!(false, "Expected a background processing task")` are dropped, with doc comments in their place explaining that the empty case is now reachable by design.

I considered keeping the assertions by threading an `expect_running: bool` through `shutdown()`, but that trades a real invariant for a parameter the caller can get wrong, and the assertions only ever restated "we got here via `stop`". Happy to go the other way if you'd rather keep them.

### Testing

- `tests/integration_tests_rust.rs::failed_start_winds_down_background_tasks` — squats one of the node's listening addresses so `bind` fails after the wallet-sync, RGS and pathfinding-scores tasks have been spawned, asserts via a `CollectingLogWriter` that the full shutdown sequence ran, then frees the address and confirms the node starts and stops cleanly.
- `src/runtime.rs::winding_down_without_spawned_tasks_is_a_noop` — the runtime wind-down calls tolerate finding nothing to wait on.

This PR was written with the assistance of Claude Code.

### DIFF
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -199,9 +199,12 @@ impl Runtime {
 		self.block_on(tasks.wait())
 	}
 
+	/// Waits for all non-cancellable background tasks to finish.
+	///
+	/// Note this may find no tasks at all, as it's also reached when winding down a startup that
+	/// failed before spawning any.
 	pub fn wait_on_background_tasks(&self) {
 		let mut tasks = core::mem::take(&mut *self.background_tasks.lock().expect("lock"));
-		debug_assert!(tasks.len() > 0, "Expected some background_tasks");
 		self.block_on(async {
 			loop {
 				let timeout_fut = tokio::time::timeout(
@@ -231,6 +234,10 @@ impl Runtime {
 		})
 	}
 
+	/// Waits for the background processor task to finish.
+	///
+	/// Note this may find no task at all, as it's also reached when winding down a startup that
+	/// failed before spawning it.
 	pub fn wait_on_background_processor_task(&self) {
 		if let Some(background_processor_task) =
 			self.background_processor_task.lock().expect("lock").take()
@@ -265,9 +272,7 @@ impl Runtime {
 					log_error!(self.logger, "Stopping event handling timed out: {}", e);
 				},
 			}
-		} else {
-			debug_assert!(false, "Expected a background processing task");
-		};
+		}
 	}
 
 	#[cfg(tokio_unstable)]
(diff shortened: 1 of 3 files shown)


elnafateh:
The test covers only wallet sync and pathfinding scores but never  RGS. Worth adding, or at least noting why it's left out, Overall good work though.

Jolah1:
> The test covers only wallet sync and pathfinding scores but never RGS. Worth adding, or at least noting why it's left out, Overall good work though.

Good catch, thanks! Looking into it turned up a second gap: the test had background sync disabled, so the wallet-sync task exited immediately and there was nothing left for the wind-down to stop. Fixed now: background sync  stays enabled, and the test asserts that the wallet-sync task logged its stop line.

src/runtime.rs
```diff
@@ -199,9 +199,12 @@ impl Runtime {
 		self.block_on(tasks.wait())
 	}
```
elnafateh, line 200: can we add a `log_debug!` here ?
// Stopped all cancellable background tasks.

tests/integration_tests_rust.rs
```diff
@@ -918,6 +918,51 @@ async fn start_stop_with_pathfinding_scores_sync() {
...
+	let node = builder.build(config.node_entropy.into()).unwrap();
+
+	assert_eq!(node.start(), Err(NodeError::InvalidSocketAddress));
+	assert!(!node.status().is_running);
+	assert_eq!(node.stop(), Err(NodeError::NotRunning));
+
+	// The failed startup ran the full shutdown sequence, rather than leaving the tasks it had
+	// already spawned running behind a node that never came up.
+	assert!(log_writer.contains("Stopped all background tasks"));
+	assert!(log_writer.contains("Disconnected all network peers."));
```
elnafateh, line 954: That new line is what the test would assert instead of trying to catch RGS's own internal log line
// assert!(log_writer.contains("Stopped all cancellable background tasks."));

elnafateh (commented):
Other than those two, the overall approach here looks good

PR #1088, opened 2026-09-03
Create private node directories and create private SQLite database files

tankyleo:


src/logger.rs
```diff
@@ -261,7 +263,7 @@ impl Logger {
 	/// are the path to the log file, and the log level.
 	pub fn new_fs_writer(file_path: String, max_log_level: LogLevel) -> Result<Self, ()> {
 		if let Some(parent_dir) = Path::new(&file_path).parent() {
-			fs::create_dir_all(parent_dir)
+			create_dir_all_private(parent_dir)
 				.map_err(|e| eprintln!("ERROR: Failed to create log parent directory: {}", e))?;
```
benthecarman, line 275: **Log file stays world-readable while its directory becomes private.** `src/logger.rs:270` and `src/logger.rs:213` still open the log file with default options, so it is created `0644`. The directory is now `0700`, so protection depends entirely on nobody copying or rotating the file out. Logs contain peer IDs, channel IDs, payment hashes, and amounts. Either give the file `0600` with the same pattern used for the database, or leave the log directory alone. Right now the PR hardens the wrong half of the pair.

benthecarman (commented):
small comment otherwise lgtm

src/logger.rs
tankyleo: Thanks I added a commit below to address this

benthecarman (commented):
lgtm on squash

tankyleo:
thanks squashed with no changes

benthecarman (approved):

tnull:
Post-merge ACK.

Fine by me, though some of these tests are so trivial that we might want to consider removing them at some point to avoid too much clutter/slow down.

PR #1089, opened 2026-09-05
Expose the probe amount range on ProbingConfigBuilder

joschisan:
The amount each probe is drawn from was fixed to the crate defaults of 1k - 10k sats. Nodes that mostly send larger payments want larger probes, since knowing a route passes 10k sats says little about whether it passes 500k. This adds `amount_range_msat` to `ProbingConfigBuilder` (and `set_amount_range_msat` on the uniffi variant) and threads it into both built-in strategies; the defaults are unchanged.

I am using this in prod already for a custom fedimint gateway.

### DIFF
--- a/src/probing.rs
+++ b/src/probing.rs
@@ -79,8 +79,8 @@ use lightning_invoice::DEFAULT_MIN_FINAL_CLTV_EXPIRY_DELTA;
 use lightning_types::features::{ChannelFeatures, NodeFeatures};
 
 use crate::config::{
-	DEFAULT_MAX_PROBE_LOCKED_MSAT, DEFAULT_PROBED_NODE_COOLDOWN_SECS,
-	DEFAULT_PROBING_INTERVAL_SECS, MIN_PROBING_INTERVAL,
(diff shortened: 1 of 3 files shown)


randomlogin:
Nice find! 

Perhaps we could also add the same setters for the `ArcedProbingConfigBuilder`?  

UPD: it's already there. LGTM

src/probing.rs
```diff
@@ -256,6 +262,21 @@ impl ProbingConfigBuilder {
...
+	/// Overrides the bounds each probe's amount is uniformly drawn from.
+	///
+	/// Only applies to the built-in strategies; custom strategies choose their
+	/// own probe amounts when building paths. Larger amounts teach the scorer
+	/// about liquidity in the range of larger payments, at the cost of locking
+	/// more liquidity per in-flight probe — size `max_locked_msat` accordingly.
+	///
+	/// Defaults to 1 000 000 - 10 000 000 msat (1k - 10k sats).
+	pub fn amount_range_msat(&mut self, min_msat: u64, max_msat: u64) -> &mut Self {
+		debug_assert!(min_msat <= max_msat, "min_msat must not exceed max_msat");
```
benthecarman, line 274: would be better to return a Result/Error here imo

tnull: @joschisan Would be good to address this before we can merge.

PR #1092, opened 2026-09-09
Pin `encoding_rs` for MSRV builds

tnull:
Keep the Rust 1.85 CI job resolving `encoding_rs` 0.8.35 because newer releases require Rust 1.88.


benthecarman (approved):
trivial

PR #1093, opened 2026-09-10
Replace funding_redeem_script with output script

benthecarman:
Expose the funding output's `scriptPubKey` on `ChannelDetails` as `funding_output_script` instead of the witness script. An output script stays meaningful for future taproot channels, where there is no witness script to expose, while a redeem script does not.

The field was added after the last release, so no compatibility note is needed.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -457,7 +457,7 @@ pub struct ChannelDetails {
 	/// state until the splice transaction reaches sufficient confirmations to be locked (and we
 	/// exchange `splice_locked` messages with our peer).
 	pub funding_txo: Option<OutPoint>,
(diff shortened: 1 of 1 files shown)


tankyleo (approved):
LGTM tnull let us know what you think. I plan to update the API in rust-lightning for 0.4.

tnull:
> LGTM tnull let us know what you think. I plan to update the API in rust-lightning for 0.4.

Hmm, is there a rush? I.e., shouldn't LDK Node just follow the API change once it happened in LDK?

benthecarman:
@tankyleo was asking for this mainly, discussed offline and fine to skip.

PR #1101, opened 2026-09-16
Drop rate-limited static invoice requests

tnull:
Drop static invoice requests when the rate limit is reached instead of retrying them and delaying later events. Keep retrying storage errors and only acknowledge invoices after saving them.

 Fixes an issue reported by @ajaysehwal.


ajaysehwal (commented):
@tnull , one gap in thorough test suite: there’s no test covering a sender that gets rate-limited once and then successfully retries after the bucket refills.

Not required for correctness — the existing tests already show that the rate-limited request is dropped cleanly without affecting other requests. But a small test using a short refill interval or deterministic bucket advancement would make the intended contract explicit: the request is dropped, and the sender is free to retry later rather than being blocked permanently.

Purely a documentation-via-test nice-to-have.

src/payment/asynchronous/static_invoice_store.rs
```diff
@@ -33,6 +33,12 @@ impl_writeable_tlv_based!(PersistedStaticInvoice, {
 	(2, request_path, required)
 });
 
+#[derive(Debug)]
+pub(crate) enum StaticInvoiceStoreError {
```
ajaysehwal, line 37: One small style nit: the two other `pub(crate)` error enums in the codebase (`AnnounceError` in `config.rs` and `BitcoindClientError` in `chain/bitcoind.rs`) both implement `Display` (`BitcoindClientError` also implements `std::error::Error`), while `StaticInvoiceStoreError` only derives `Debug`.

It doesn't cause any issues today since the log call destructures the inner `lightning::io::Error` directly, but for consistency with the existing pattern — and in case this type ever gets logged or propagated as a whole later — it might be worth adding:

```rust
impl std::fmt::Display for StaticInvoiceStoreError {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self {
            Self::RateLimited => write!(f, "rate limit exceeded"),
            Self::Io(e) => write!(f, "{e}"),
        }
    }
}
```
Purely optional — happy to see this merged without it.

tnull: Hmm, well, given we intentionally omit logging the errors, I don't think we need to add `Display`. Generally, no need to add implementations for internal types if we currently don't use them.

ajaysehwal (approved):
LGTM

src/event.rs
```diff
@@ -2115,7 +2117,11 @@ where
 						Ok(_) => {
 							self.channel_manager.static_invoice_persisted(invoice_persisted_path);
 						},
-						Err(e) => {
+						Err(StaticInvoiceStoreError::RateLimited) => {
```
joostjager, line 2120: Don't we want to skip the rate limiter once we are retrying persistence, so that we don't exit the retry loop because new requests came in?

src/payment/asynchronous/static_invoice_store.rs
```diff
@@ -154,16 +155,214 @@ mod tests {
...
+		let messages =
+			AsyncPaymentsMessageHandler::release_pending_messages(&*node.channel_manager);
+		assert_eq!(messages.len(), 1);
+		assert!(matches!(messages[0].0, AsyncPaymentsMessage::StaticInvoicePersisted(_)));
+		let store = StaticInvoiceStore::new(kv_store);
+		assert!(store.handle_static_invoice_requested(&[1, 1, 1], 1).await.unwrap().is_some());
+	}
+
+	#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+	async fn static_invoice_read_error_is_replayed() {
```
joostjager, line 350: Quite a bit of test code. Can't AI come up with something more compact with still reasonable coverage?

PR #1103, opened 2026-09-17
Expose pending HTLCs and channel dust exposure

### LINKED ISSUE #1102: Expose pending HTLCs and dust exposure in ChannelDetails
While working on ldk-server I ran into ldk-server#251, which tried to expose more channel data over its API but had to leave out pending HTLCs and dust exposure, because ldk-node's `ChannelDetails` doesn't include them.

Looking at `ChannelDetails::from_ldk` in `src/types.rs`, LDK's `ChannelDetails` already has these fields, but we don't copy them over:
  - `pending_inbound_htlcs: Vec<InboundHTLCDetails>`
  - `pending_outbound_htlcs: Vec<OutboundHTLCDetails>`
  - `current_dust_exposure_msat: Option<u64>`
  - `next_splice_out_maximum_sat: u64`

The HTLC details include the amount, CLTV expiry, payment hash, state and whether the HTLC is dust (plus `skimmed_fee_msat` for outbound ones).

I think these would be useful for:
  - debugging stuck payments, e.g. seeing which HTLCs are still pending on a channel and when
    they expire
  - checking how close a channel is to its force-close deadline
  - comparing `current_dust_exposure_msat` against the channel's `max_dust_htlc_exposure` config
  - showing how much can be spliced out before calling `splice_out`

LND and CLN both show pending HTLCs when listing channels, so node operators are probably used to having this. tnull also mentioned in #163 that most `ChannelDetails` fields would make sense to expose, and #993 already reads `pending_outbound_htlcs` internally for LSPS5 notifications, so this data seems useful outside LDK too.

For the bindings, I guess the HTLC detail structs and the `InboundHTLCStateDetails` / `OutboundHTLCStateDetails` enums would need uniffi wrappers, similar to what we do for `ChannelShutdownState` in `src/ffi/types.rs`.

I left `splice_details` out on purpose since there's ongoing work on splice tracking (#1080, #1079, #1057) that already uses it internally, but it could be added later if that makes sense.

  Does this sound reasonable? I'd be happy to work on it. @tnull @TheBlueMatt

tnull:
Make pending payment state and dust exposure available to callers so they can inspect stalled payments and assess force-close risk. Also expose the maximum splice-out amount to help plan withdrawals.

Expose the same information through Rust and the language bindings.

Rust and UniFFI checks, unit tests, the new integration test in both configurations, doctests, and binding generation pass. The full integration run had two failures in existing tests. Both passed in isolation, but channel_full_cycle_force_close_trusted_no_reserve failed again with Bitcoin Core RPC at tests/common/mod.rs:632. Its clean-main run passed, so the cause remains unresolved. UniFFI rustdoc also fails on three broken links reproduced on main.

Fixes #1102


src/types.rs
```diff
@@ -616,6 +619,21 @@ pub struct ChannelDetails {
...
+	/// The current total dust exposure on this channel, in millisatoshis.
+	///
+	/// This is the maximum of the dust exposure on the holder and counterparty commitment
+	/// transactions. It includes pending HTLCs below the dust threshold and the portion of
+	/// commitment transaction fees that contributes to dust exposure.
+	///
+	/// This is compared against [`ChannelConfig::max_dust_htlc_exposure`] when determining whether
+	/// new HTLCs can be accepted or offered on this channel.
+	///
+	/// Will be `None` for objects serialized with LDK versions prior to 0.3.
```
f3r10, line 635: This line is copied from LDK's docs, but it doesn't apply here: ldk-node only builds ChannelDetails from live channels in list_channels, and LDK always sets Some(..) for those (ChannelDetails::from_channel in channel_state.rs). Maybe drop it, or say it's always Some for channels returned by Node::list_channels? Keeping the Option seems right in case upstream changes. It's just the doc that's misleading.

f3r10 (commented):
Thanks, this looks good to me overall, and the test covering a dust and a non-dust HTLC at the same time is nice.

And about that comment on the commit: I ran `channel_full_cycle_force_close_trusted_no_reserve` 5× on this branch (4 pass, 1 fail [panic: …]) and 5× on the parent commit (5 pass). That's too few runs to tell whether the test is flaky or something changed here, but since this diff only adds read-only fields to ChannelDetails and nothing in the force-close path touches them, I'd expect a pre-existing flake.

src/types.rs
tnull: Ah, good catch, updated!

tnull:
> And about that comment on the commit: I ran `channel_full_cycle_force_close_trusted_no_reserve` 5× on this branch (4 pass, 1 fail [panic: …]) and 5× on the parent commit (5 pass). That's too few runs to tell whether the test is flaky or something changed here, but since this diff only adds read-only fields to ChannelDetails and nothing in the force-close path touches them, I'd expect a pre-existing flake.

Yeah, it's flaky currently.

f3r10 (approved):

PR #1107, opened 2026-09-18
Reclaim forwarded-payment replay markers instead of leaking them

ajaysehwal:
## Summary

`ForwardingStore::record_forward` writes a permanent `ForwardedPaymentReplayMarker` for every forwarded HTLC, used to guard against LDK replaying an event whose side effects we already recorded. Nothing ever removed these markers -- the existing aggregation pass reclaims detail records but never touches the marker store, so it grows without bound for the life of the node.

This is worse in the default `Stats` tracking mode: no detail record is ever written there, so aggregation has nothing to key cleanup off. Worse still, `run_forwarded_payment_aggregation`'s background loop exits for good the first time it observes an empty details store under `Stats` mode -- which is immediately, since that mode never populates one -- so the reclamation task stops running entirely after its first pass on a routing node's very first startup, in the default config.

Any peer with a channel to the node can trigger growth for the cost of their own routing fee (circular routing refunds it to the attacker). The write also happens synchronously inside the LDK event handler, so on a remote KV backend the accumulating read cost adds to the same hot path already shown to head-of-line block the event queue.

## Fix

- Add a `forwarded_at_timestamp` field to the marker (TLV-optional, defaults to `0` on read, so pre-existing leaked markers are swept on the first pass after upgrading).
- Add `prune_expired_replay_markers`, run every aggregation cycle independent of `retention_secs`, so `Stats` mode reclaims markers too. A marker is only removed once it's past a fixed one-bucket-wide age cutoff **and** has no corresponding detail record left --- age alone isn't sufficient: the aggregation pass defers an *entire*  bucket (every sibling detail in it, not just the record missing a marker) whenever any one detail in that bucket is still missing its own marker. An age-only version of this fix passed every existing test but silently corrupted exactly that scenario; caught it with a dedicated regression test before landing this version.
- Skip the per-marker detail lookup when the details store is empty (one read instead of one per marker) -- the common case in `Stats.` mode, where a detail can never exist to check for.
- Fix the background loop's early-exit check to also require the marker store to be empty, not just the details store, so it doesn't stop reclaiming markers while details happen to be empty.

## Testing

Six new tests, covering: markers leaking across both tracking modes, the exact bucket-width age cutoff, the sibling-bucket corruption case above, and the background loop no longer exiting while markers remain.

For both substantive changes (the sibling-bucket guard, and the details-empty fast path), I also manually reverted just that piece, confirmed the corresponding test fails with the predicted symptom, and then restored it and confirmed green -- not just written, verified to actually catch the regression it's meant to catch.

`cargo fmt --all -- --check`, `cargo clippy --lib -- -D
clippy::unwrap_used` (the repo's CI lint), and the full `cargo test
--lib` suite (200/200) all pass.


src/payment/forwarding_store.rs
```diff
@@ -549,12 +556,17 @@ async fn aggregate_forwarded_payments_and_log(
...
+}
+
 pub(crate) async fn run_forwarded_payment_aggregation(
 	mut stop_receiver: tokio::sync::watch::Receiver<()>, forwarding_store: Arc<ForwardingStore>,
 	retention_secs: u64,
 ) {
 	if retention_secs == 0 {
-		match forwarding_store.details.is_empty().await {
+		match has_nothing_left_to_reclaim(&forwarding_store).await {
 			Ok(true) => return,
```
juliusjulyp: Nice fix. One gap I noticed (not blocking): in `Stats` mode (`retention_secs == 0`) the loop still returns early when both stores are empty, e.g. on a fresh node before its first forward. The early exit predates this PR, but it seems to leave part of the problem this PR describes. Since `record_forward` writes a marker for every tracked forward in `Stats` mode, and I don't see anything that re-spawns the task (`start()` spawns it once, and `record_forward`/`event.rs` don't signal it), markers written after that point would only be reclaimed on the next restart.

I wrote a small test to check this: it spawns the loop on empty `Stats` stores and asserts it's still running after 200ms. It fails on this branch.

<details>
<summary>Test and output</summary>

```rust
#[tokio::test]
async fn loop_stays_alive_in_stats_mode_when_stores_start_empty() {
	let kv_store: Arc<DynStore> = Arc::new(DynStoreWrapper(InMemoryStore::new()));
	let logger = Arc::new(Logger::new_log_facade());
	let forwarding_store = Arc::new(ForwardingStore::new(
		Vec::new(),
		ForwardedPaymentTrackingMode::Stats,
		kv_store,
		logger,
	));
	let (_stop_sender, stop_receiver) = tokio::sync::watch::channel(());
	let handle = tokio::spawn(run_forwarded_payment_aggregation(
		stop_receiver,
		Arc::clone(&forwarding_store),
		0,
	));
	tokio::time::sleep(Duration::from_millis(200)).await;
	assert!(
		!handle.is_finished(),
		"loop exited on startup; markers written later will never be reclaimed"
	);
}
```

```
thread '...loop_stays_alive_in_stats_mode_when_stores_start_empty' panicked at src/payment/forwarding_store.rs:1875:5:
loop exited on startup; markers written later will never be reclaimed
```

</details>

The existing zero-retention test asserts the loop returns once drained, which made sense when `Stats` wrote nothing that needed later cleanup. Would it make sense to never exit early in `Stats` mode now, or to have `record_forward` wake or respawn the task? Happy to be corrected if the exit is intentional and I'm missing what covers later markers.

ajaysehwal: Good catch, thank you for the thorough look (and for the test — that made this trivial to verify and fix).

You're right on both counts: the early exit predates this PR, and it does undercut the whole point of it. A fresh node (or any node whose stores happen to be empty at the moment this task starts) returns before ever reaching the periodic loop, and nothing respawns it, so markers written afterward leak until the next restart -- exactly the failure mode this PR set out to fix, just moved to a different trigger condition.

Went with your first suggestion: removed the early-exit special-casing entirely, so the loop now runs for the life of the node in both tracking modes, matching how `Detailed` mode already behaved unconditionally. The idle cost is a couple of cheap, mostly-empty store reads once an hour, which is negligible next to silently leaking markers again. Also let me simplify the code a bit in the process (one less helper, three fewer duplicated exit checks).

Added your test essentially as you wrote it (`loop_stays_alive_in_stats_mode_when_stores_start_empty`), plus rewrote the existing `background_aggregation_runs_immediately` test, which previously relied on the function returning once drained -- now it spawns the task, polls for the first pass to land, asserts the loop is still running, then exercises clean shutdown via the stop signal.

Confirmed the fix the same way you found the bug: reverted just this change, reran your test, watched it fail with your exact panic message, then restored it and confirmed green along with the full suite.

juliusjulyp: Thanks, checked the updated branch and the tests pass locally. Looks good to me.

benthecarman:
Thanks! Approach looks right, two structural things claude and I discussed on how to do this best: 

1. Don't prune on the startup. The aggregation task runs before the background processor replays any events that were handled but not persisted before a crash, so after a long outage this deletes the markers right before they're needed. Waiting one bucket width of uptime before the first prune is enough.

2. In Detailed mode this does a `contains_key` on the details store for every marker older than an hour, every hour, which is nearly all of them for the whole retention period. Remove the marker in the aggregation pass right after it removes the detail instead. The sweep then only has to handle Stats mode and crash orphans, and the per-marker lookup can stay as is since it rarely fires.

Nits: the doc comment on `run_forwarded_payment_aggregation` describes a version that never existed on main. Forwarding tracking hasn't shipped yet, so `forwarded_at_timestamp` can just be `required` rather than defaulting to 0 for old markers.

ajaysehwal:
> Thanks! Approach looks right, two structural things claude and I discussed on how to do this best:
> 
> 1. Don't prune on the startup. The aggregation task runs before the background processor replays any events that were handled but not persisted before a crash, so after a long outage this deletes the markers right before they're needed. Waiting one bucket width of uptime before the first prune is enough.
> 2. In Detailed mode this does a `contains_key` on the details store for every marker older than an hour, every hour, which is nearly all of them for the whole retention period. Remove the marker in the aggregation pass right after it removes the detail instead. The sweep then only has to handle Stats mode and crash orphans, and the per-marker lookup can stay as is since it rarely fires.
> 
> Nits: the doc comment on `run_forwarded_payment_aggregation` describes a version that never existed on main. Forwarding tracking hasn't shipped yet, so `forwarded_at_timestamp` can just be `required` rather than defaulting to 0 for old markers.

Thanks for the review — both fixed in the latest commit: [eb251c4](https://github.com/lightningdevkit/ldk-node/pull/1107/commits/eb251c46c75b86ec918132ab71e93e388ae8082e)

- Don't prune on startup: the sweep now waits one bucket width of uptime before running, so it can't delete a marker a crash-replay still needs. Aggregation still runs immediately, since it only removes a marker alongside a detail it just confirmed is gone.
- Pair marker removal with detail removal: aggregation now removes the marker right after its detail in the same pass (detail first, so a failure between the two leaves a harmless orphan the sweep can catch, not a dangerous one).

Also fixed the stale doc comment and made forwarded_at_timestamp required.

Separately, juliusjulyp caught that the old early-exit logic could stop the loop forever on an idle/fresh node. Removed that logic entirely — the loop now just runs for the node's lifetime, which is negligible cost when idle.

Covered by updated/new tests. Let me know if anything still looks off.

src/payment/forwarding_store.rs
```diff
@@ -1181,6 +1298,91 @@ mod forwarding_stats_tests {
...
+	}
+
+	/// Regression test: the immediate startup pass must not sweep an old marker with no detail
+	/// behind it -- after a crash, that's indistinguishable from an event the background
+	/// processor still needs to replay, and `record_forward` depends on the marker being there to
+	/// recognize the replay. `prune_expired_replay_markers_respects_the_bucket_cutoff` above
+	/// already covers the sweep itself eventually reclaiming a marker this old; this test is
+	/// narrowly about the startup pass not being the one to do it.
+	#[tokio::test]
+	async fn startup_pass_does_not_sweep_an_old_marker() {
```
benthecarman, line 1507: Adding a `Detailed` version of what I described above would catch the issue.

src/payment/forwarding_store.rs
```diff
@@ -767,6 +852,15 @@ async fn aggregate_expired_forwarded_payments_at(
 			log_error!(logger, "Failed to remove forwarded payment {:?}: {}", payment_id, e);
 			e
 		})?;
+		// The marker's job ends with its detail: this bucket was only aggregated because the
+		// marker confirmed the detail was durably recorded, so nothing needs it anymore. Removing
+		// it here, rather than leaving it to the separate age-based sweep, means that sweep rarely
+		// has to fall back to its per-marker existence check -- almost every Detailed-mode marker
+		// is gone by the time it would otherwise become a candidate.
```
benthecarman: this is way too much docs for this line, how about:

```
The bucket is committed and the detail is gone, so the marker has nothing left to guard. Remove it here so the sweep only has to deal with markers that have no detail. If this fails the marker is orphaned and the sweep reclaims it.
```

src/payment/forwarding_store.rs
```diff
@@ -376,6 +383,24 @@ impl ForwardingStore {
 		)
 		.await
 	}
+
+	/// Sweeps markers the aggregation pass above didn't already remove alongside their detail --
+	/// `Stats` mode, which writes no detail to pair a removal with, and crash orphans. See
+	/// [`prune_expired_replay_markers`] for why age alone isn't a safe criterion here.
+	pub(crate) async fn prune_stale_replay_markers(&self) -> Result<u64, Error> {
```
benthecarman: nit: this doesn't need to be `pub(crate)`

benthecarman (changes requested):
Thanks, close now. 

One major thing from claude, `run_forwarded_payment_aggregation` needs to cover the paired removal too. In `Detailed` mode, the startup aggregation pass after a long outage deletes both the detail and its marker, and the background processor then replays the unpersisted events against an empty marker store. The doc comment says this path carries no such risk, but it does. Skipping marker removal on the startup pass (leave them for the sweep) would fix it.

ajaysehwal:
@benthecarman  All three are fixed now:

- Paired removal on startup — the startup aggregation pass now runs with marker removal skipped entirely; it still aggregates and removes details normally, but leaves every marker for the delayed, age-gated sweep to reclaim once it's actually safe. Added a Detailed-mode regression test (startup_pass_does_not_remove_a_paired_marker_in_detailed_mode) that reproduces the exact scenario you described — a detail+marker pair old enough to aggregate on the very first startup pass — and confirmed it fails without the fix and passes with it.
- Shortened the comment — used your wording verbatim.
- Dropped the pub(crate) — you're right, prune_stale_replay_markers is only ever called from prune_stale_replay_markers_and_log in the same file, so it's just a private method now.


While I was in there, I also swapped the remove_markers: bool parameter threaded through the aggregation functions for a small MarkerRemoval::{Retain, Remove} enum, since the bare true/false at call sites (especially in tests) wasn't self-explanatory.

src/payment/forwarding_store.rs
```diff
@@ -767,6 +871,15 @@ async fn aggregate_expired_forwarded_payments_at(
...
 			e
 		})?;
+		if remove_markers == MarkerRemoval::Remove {
+			// The bucket is committed and the detail is gone, so the marker has nothing left to
+			// guard. Remove it here so the sweep only has to deal with markers that have no
+			// detail. If this fails the marker is orphaned and the sweep reclaims it.
+			replay_marker_store.remove(&payment_id).await.map_err(|e| {
+				log_error!(logger, "Failed to remove replay marker {:?}: {}", payment_id, e);
+				e
+			})?;
```
benthecarman: we shouldn't do `?` on error here, otherwise we'll stop in the middle of the pass through, shold just log and continue on

benthecarman (commented):
also can you squash your commits

ajaysehwal:
> also can you squash your commits

Squashed all commits into a single commit now

src/payment/forwarding_store.rs
ajaysehwal: Good catch, you're right — fixed. Marker removal no longer uses ?; it now logs and continues, matching what the comment already said would happen ("the marker is orphaned either way, and the sweep reclaims it later"), but the code wasn't actually doing.

Added a regression test (marker_removal_failure_does_not_abort_the_rest_of_the_pass) with a small KV store wrapper that fails removal for one specific marker while everything else succeeds. It confirms the pass still aggregates both payments, the other payment's marker is removed normally, and only the one whose removal genuinely failed survives (orphaned, for the sweep to pick up later) — rather than the old behavior where that single failure would've aborted the whole batch, leaving every later payment's marker untouched too. Verified by reverting to the old behavior and confirming the test fails with PersistenceFailed as expected, then restoring the fix.

benthecarman:
Claude found some docs to fix: **Stale comment** (`forwarding_store.rs:325`): `record_forward` still says the marker is kept after the event is handled because LDK can replay an old event and gives no durability callback. Markers are now deleted, so this should say what they guard and until when: until the ChannelManager has been persisted after the event was handled.

otherwise, can you make your commit message have a description, then this should be good to go

ajaysehwal:
> Claude found some docs to fix: **Stale comment** (`forwarding_store.rs:325`): `record_forward` still says the marker is kept after the event is handled because LDK can replay an old event and gives no durability callback. Markers are now deleted, so this should say what they guard and until when: until the ChannelManager has been persisted after the event was handled.
> 
> otherwise, can you make your commit message have a description, then this should be good to go

Fixed both:


- The comment at forwarding_store.rs:325 now says what the marker guards and until when:
 This marker guards against LDK replaying this event: it must survive until the ChannelManager has been persisted after the event was handled, since only that persist durably drops the event from the replay queue. It's removed once enough time has passed that persistence is presumed complete -- either paired with its aggregated detail, or by the separate age-based sweep.
- The commit message now has a full description covering the original bug, each fix, and the testing approach, instead of just a title.

PR #1111, opened 2026-09-21
Replace PostgreSQL advisory locking with fenced leases

joostjager:
Acquire a table-scoped lease before loading persisted state and validate and renew it atomically with each mutation. Renew idle leases in the background and permanently invalidate the store on detected lease loss, preserving panic behavior and the existing lock-contention error kind.

Coordinate the one-way schema-v1 upgrade with the legacy advisory lock and commit the schema-v2 marker with the initial lease. Older releases then reject the upgraded store instead of bypassing lease fencing.

Keep lease handling inside PostgresStore, with node lifecycle and recovery unchanged. Cover contention, migration, queued-write fencing, lease release, and panic propagation in the store tests.


tnull (commented):
This needs a rebase for CI to run. Let me know when fully ready for review.

joostjager:
If we indeed add this on the 0.8 milestone, I'll remove the migration because the temp. locking solution won't be in a release.

tnull:
> If we indeed add this on the 0.8 milestone, I'll remove the migration because the temp. locking solution won't be in a release.

Do it.

joostjager:
I’ve removed the migration. This now directly replaces advisory locks with the lease table.

I’m particularly happy to see all the scattered checks before and after operations to confirm we still hold the lock disappear. PostgreSQL now checks lease validity within the transaction.

I had a good back-and-forth with AI to minimize the diff. I think it’s looking pretty good now.

src/io/postgres_store/mod.rs
```diff
@@ -630,18 +737,80 @@ impl PostgresStoreInner {
...
+		let update_sql = format!(
+			"UPDATE {lease_table}
+			SET expires_at = clock_timestamp() + ($2::bigint * interval '1 second')
+			WHERE id = 1 AND owner_id = $1 AND expires_at > clock_timestamp()"
 		);
+		let updated = client
+			.execute(&update_sql, &[&self.lease_owner_id.as_slice(), &lease_duration_secs])
+			.await?;
+		if updated != 1 {
+			panic!("PostgreSQL node lease was lost; continuing may corrupt node state");
```
tnull: This probably should be `std::process:abort` as panicking the tokio task *won't* abort the whole process, but just have the task return a `JoinError` in the end.

tnull (commented):
Thanks, yet to do a very detailed review. Also tagging @benthecarman as a secondary reviewer as he did the first approach.

src/io/postgres_store/mod.rs
joostjager: The previous advisory-lock implementation used `assert!` rather than `abort`, and LDK Server sets `panic = "abort"` for both dev and release builds. But definitely seems safer to use `std::process:abort`, will change.

Note that not panicking wouldn't be a data consistency issue, because the consistency is guarded in each transaction.

joostjager: Although, it seems in other places in ldk-node, it's not an explicit process abort?

PR #1114, opened 2026-09-22
Allow routing nodes to forward over unannounced channels

tankyleo:
```
    Allow routing nodes to forward over unannounced channels

    rust-lightning defaults to refusing private-channel forwards to protect
    nodes that are not reliably online from HTLC-forwarding risk. The flag
    expresses willingness to forward; negotiated SCID privacy independently
    rejects forwards using the real channel identifier.

    LDK-Node already treats a configured alias and non-empty listening
    addresses as indications of a forwarding node. Enabling private-channel
    forwards for these nodes aligns with that distinction and lets ordinary
    routers deliver payments to private recipients without requiring a
    liquidity-provider or async-server role.

    Co-Authored-By: HAL 9000
```


benthecarman (approved):
should get @tnull blessing as this is changing default behavior

vincenzopalazzo (approved):
LGTM, this will fix by default my publick node!

tnull:
Hmm, I want to note that besides the HTLC forwarding risk there is also an associated privacy impact as channels that don't enable `negotiate_scid_privacy` can be probed. 

Funnily enough I just discovered that LDK still defaults to `false` and LDK Node does not override that default setting. I think we should change that, and maybe even beyond that require it by now in the spec. So, can you pull in https://github.com/tnull/ldk-node/commit/36087a64a2c6afa9184486dfbfbe436b91c717c7 to this PR?

I also opened https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/pulls/5028 and https://github.com/lightning/bolts/pull/1362.

tankyleo:
Thanks good catch for some reason I was confident SCID alias was turned on in LDK ! I should have checked thank you

tnull (approved):

